PR #78, opened 2015-07-10
Show the correct number of paid items

Unicorn-rzl:
In the order overview the total of "paid orders" is the same as the total number of "ordered items". By using the right variable, we can easily fix that ;)


raphaelm:
Good catch, thanks! 

(Will be deployed to mrmcd.net in a couple of minutes :))

PR #101, opened 2015-10-02
improved top-to-bottom-adhoc-readability

cherti:
use long names in list comprehensions, so that we have not just one letter, but a more expressive word indicating where we pull the information from.


raphaelm:
Great, thanks!

PR #136, opened 2016-05-08
Added basic Django password validations

### LINKED ISSUE #113: Use Django 1.9's password validation feature
As we now only have admin users, I'd like to restrict them to somewhat-safe passwords. I'd like to use Django 1.9's new validation feature together with https://pypi.python.org/pypi/zxcvbn for this task.

JRodDynamite:
Partially fixes #113. Do let me know if there are any issues with the code.

### DIFF
--- a/src/pretix/base/forms/auth.py
+++ b/src/pretix/base/forms/auth.py
@@ -1,5 +1,8 @@
 from django import forms
 from django.contrib.auth import authenticate
+from django.contrib.auth.password_validation import (
+    password_validators_help_texts, validate_password,
+)
 from django.utils.translation import ugettext_lazy as _
 
 from pretix.base.models import User
@@ -84,7 +87,7 @@ class RegistrationForm(forms.Form):
     )
 
     def clean(self):
-        password1 = self.cleaned_data.get('password')
+        password1 = self.cleaned_data.get('password', '')
         password2 = self.cleaned_data.get('password_repeat')
 
         if password1 and password1 != password2:
@@ -93,6 +96,12 @@ class RegistrationForm(forms.Form):
                 code='pw_mismatch'
             )
 
+        user = User(email=self.cleaned_data.get('email'))
+        if validate_password(password1, user=user) is not None:
+            raise forms.ValidationError(
+                _(password_validators_help_texts()),
+                code='pw_invalid'
+            )
         return self.cleaned_data
 
     def clean_email(self):
@@ -107,7 +116,7 @@ class RegistrationForm(forms.Form):
 
 class PasswordRecoverForm(forms.Form):
     error_messages = {
-        'pw_mismatch': _("Please enter the same password twice")
+        'pw_mismatch': _("Please enter the same password twice"),
     }
     password = forms.CharField(
         label=_('Password'),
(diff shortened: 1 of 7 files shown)


raphaelm:
Thanks a lot! I will leave some comments in the code.

src/pretix/base/forms/auth.py
```diff
@@ -3,6 +3,9 @@
 from django.utils.translation import ugettext_lazy as _
 
 from pretix.base.models import User
+from django.contrib.auth.password_validation import (
+    validate_password, password_validators_help_texts
+)
```
raphaelm: We use ``isort`` to keep imports deterministically ordered. Just run ``isort -rc pretix`` on the codebase and those will be ordered alphabetically.

src/pretix/base/forms/auth.py
```diff
@@ -68,6 +71,7 @@ class RegistrationForm(forms.Form):
     error_messages = {
         'duplicate_email': _("You already registered with that e-mail address, please use the login form."),
         'pw_mismatch': _("Please enter the same password twice"),
+        'pw_invalid': _("Please enter a valid password")
```
raphaelm: I think this line is unnecessary as we never access it because ``password_validators_help_texts()`` already gives a human-readable message.

src/pretix/base/forms/auth.py
```diff
@@ -107,7 +117,8 @@ def clean_email(self):
 
 class PasswordRecoverForm(forms.Form):
     error_messages = {
-        'pw_mismatch': _("Please enter the same password twice")
+        'pw_mismatch': _("Please enter the same password twice"),
+        'pw_invalid': _("Please enter a valid password")
```
raphaelm: Ssame as above

raphaelm:
I left a few small comments, otherwise this is totally perfect. Thanks especially for the extensive tests. Less tests would probably have sufficed since a lot of this is tested in Django itself, but more tests are always good.

The flake8 style checker yields ``./pretix/control/views/auth.py:164:102: W291 trailing whitespace``, the line seems to contain a space character at the end.

The test suite timed out on travis (this sadly happens sometimes), but locally it fails only due to something that is not your fault and that I just fixed in 749e44f on the master branch. So its only the stylistic issues mentioned above and in the comments that are left to be fixed :)

raphaelm:
There is something else I missed on first sight: You changed RegistrationForm and PasswordRecoverForm but there is also a ``UserSettingsForm``. Could you change that as well?

JRodDynamite:
Thanks for the quick response and detailed. :+1:  Will fix them and and push it up. :)

JRodDynamite:
Hey @raphaelm, are there no test cases written for `UserSettingsForm`? Can't find them.

raphaelm:
There are a few in ``tests/control/test_user.py``. There surely is room for improvement in the organization of the test files.

JRodDynamite:
Hey @raphaelm, I've updated my code. Made the corrections as requested. Let me know if I've still missed out on anything. :)

raphaelm:
Looks near perfect. I have only one very minor thing to say: You put the new validation logic in  ``UserSettingsForm.clean_new_pw_repeat``. This means, if the input is wrong, the "password repeat" field will be red and has the error attached. From an UI point of view, think it would be better to attach the error to the password field itself, e.g. ``clean_new_pw``. Dou you agree?

JRodDynamite:
Okay. Understood. Will make the necessary changes.

JRodDynamite:
Hey @raphaelm I'm not sure why the that one test case failed in the build. Could you let me know why exactly that happened?

raphaelm:
It's probably a false negative, as it is totally unrelated. It is one of the Selenium interface tests and sadly, Selenium is very unstable and flaky which gets even worse on Travis.

I'll try locally later today and then merge this PR.

raphaelm:
Everything looks fine to me now. Thank you very very much for this contribution, I really appreciate it!

PR #140, opened 2016-05-28
Change language tag of input fields to standard to enable spell checking

Unicorn-rzl:
Currently textareas/inputs have a "data-lang" tag. How ever, the correct tag is "lang" and should therefore be used. This will enable spell checking for different languages depending on the text area contents.

See https://www.w3.org/International/questions/qa-html-language-declarations for reference.


raphaelm:
Thank you very much!

PR #144, opened 2016-06-15
Python 3.5 required

MrGirlyMan:
`from typing import List` in base/plugins.py was introduced in python 3.5.


raphaelm:
Thanks! But can just use ``pip install -r requirements/py34.txt`` or ``pip install typing`` to make it work on Python 3.4 just fine. I'll make sure to update the documentation accordingly.

raphaelm:
I updated the documentation in e09f213. Thanks again!

PR #154, opened 2016-07-25
Move middleware functionality to utils.py

rixx:
This simplifies providing an `event_view` method which works as
decorator for presale/ views (e.g. for plugins), providing passing
requests with event and organizer context if possible, and re-routing or
failing appropriately otherwise.

### DIFF
--- a/src/pretix/presale/middleware.py
+++ b/src/pretix/presale/middleware.py
@@ -1,14 +1,6 @@
-from urllib.parse import urljoin
-
-from django.core.exceptions import PermissionDenied
 from django.core.urlresolvers import resolve
-from django.http import Http404
-from django.shortcuts import redirect
-from django.utils.translation import ugettext_lazy as _
 
-from pretix.base.middleware import LocaleMiddleware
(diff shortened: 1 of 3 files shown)


src/pretix/presale/middleware.py
```diff
@@ -19,54 +11,7 @@ def process_request(self, request):
...
-                    LocaleMiddleware().process_request(request)
-
-                    if not request.event.live:
-                        if not request.user.is_authenticated() or not EventPermission.objects.filter(
-                                event=request.event, user=request.user).exists():
-                            raise PermissionDenied(_('The selected ticket shop is currently not available.'))
-
-            except IndexError:
-                raise Http404(_('The selected event or organizer was not found.'))
+            _detect_event(request)
```
raphaelm: Don't you need the same logic here as in the decorator, i.e. returning a response if one is given back by ``_detect_event``

raphaelm:
Added documentation in https://github.com/pretix/pretix/commit/03a56f81ca007a1dd21dc625da73cd9d8a55a174

PR #158, opened 2016-08-04
Add missing nav_event signal to docs

rixx:


doc/development/api/general.rst
```diff
@@ -47,8 +47,13 @@ Backend
 -------
 
 .. automodule:: pretix.control.signals
+   :members: nav_event
+
+
+.. automodule:: pretix.control.signals
```
raphaelm: Is there a reason you did not pull this together in one ``automodule`` block?

Issue #51, opened 2015-04-06
Allow rich text in product descriptions


raphaelm:
Currently, product descriptions and category descriptions only allow plain text.

This issue suggests to add support for Markdown markup to allow links and other simple formatting.

FlaviaBastos:
Hi, @raphaelm 
I am working on this issue and I have a couple questions for you when you have time, please.
I used [django-markup](https://django-markup.readthedocs.io/en/latest/installation.html) to implement this. 
1) where should I put this in the requirements? I'm guessing I have to add line 'django-markup' to 'pretix/src/requirements/production.txt' but I wanted to check with you first

2) I ran the code checks and unit tests as [instructed here](https://docs.pretix.eu/en/latest/development/setup.html#code-checks-and-unit-tests) and there are a bunch of fails from py.test and also an output from flake8. As far as I can tell they are not related to the two files I changed but I'm not sure. What would you like me to do?
Thank you for your time.

raphaelm:
Hey @FlaviaBastos that is really awesome! I'm very happy to respond to your questions! Starting on Saturday I'll be on holiday, so do expect *late* replies, but you can certainly *expect* replies ;)

1) ``production.txt`` is totally fine as it will be required by all installations! You should probably also add it in ``setup.py``, although we are not really using it at that time.

2a) ``flake8 .`` gives a number of warnings that we need to fix, yes. Most of them are complexity issues and those are currently ignored by our CI installation. As long as you believe that they are not related to your changes, you can safely ignore them. If you "violate" a rule that we care about deeply, TravisCI will tell you when you submit your pull requests or I will tell you when reviewing. Don't worry about them too much at this time.

2b) ``py.test`` shouldn't give any failures, however, we have some frontend browser tests in there. By default they run with PhantomJS and that has grown unstable over the last months on some systems and I'm not really sure why. I plan on replacing them with other test methods alltogether but for now they are there. You can try whether ``TEST_BROWSER=Chrome py.test`` or ``TEST_BROWSER=Firefox py.test`` works better for you, Chrome is pretty stable on my machine, I just run it and all 524 tests passed.

Thanks for taking your time to work on this!

FlaviaBastos:
Thanks @raphaelm for answering my questions.
I tested the tests separately (ha!) to make sure I didn't break anything and I'm pretty confident I didn't. Please let me know if you think otherwise.
As you can probably see, I followed your advice for requirements and setup.py and added support to rich text on category descriptions also.
[...]

PR #159, opened 2016-08-04
Allow rich text in product and category descriptions

FlaviaBastos:
Using [django-markup](https://django-markup.readthedocs.io/en/latest/installation.html) to allow rich text in product and category descriptions.
Added django-markup to requirements and setup

#51


raphaelm:
That is awesome! Don't care about travis failing with "no build output received for 10 minutes", that is selenium talking, I *really* need to get rid of this. I'm testing this manually and it works great.

One caveat: Apparently, not only ``django-markup`` is required, but the ``markdown`` package also needs to be installed explicitly. Would you like to address this?

If you *want* to, you can also add markdown to the recently added ``frontpage_text`` (see 60c1048), but if not I can do this as well after this is merged.

FlaviaBastos:
Hi @raphaelm 
Oh, my bad... I forgot about 'markdown' :blush: It's fixed now. Sorry about that! And I added my name to authors - thanks! :)

I tried to add markdown to the frontpage_text but I got an error when manually testing the page (even before changing the code). I reported the issue (#160) so you can take a look and confirm (it could be something wrong with my setup too...). Anyway, I know what needs to be done to add markdown support to the frontpage but I didn't want to do it in the dark, without testing. 

Will send a push request shortly.

raphaelm:
Hey @FlaviaBastos, I'm just gonna merge this now so it is in here before my holiday. If you want to add the frontpage text thing when we got #160 resolved, you are invited to do so; if you don't, I'm gonna do it, it probably boils down to a one-line change in the template. Thanks a lot for taking your time to do this! :)

FlaviaBastos:
Hi, @raphaelm 
I will poke around issue #160 as you suggested and will update with my findings.

I will also try to find out what's going on with my setup (local tests failing, etc). Maybe when I get this running smoothly I can grab other issues. :)

Have a nice holiday!

Issue #164, opened 2016-08-07
Mail sending issue


rixx:
When sending mails in a production environment, the pretix `mail()` method returned non-success values while the mail server actually succeeded in sending the mail.

While I understand that a positive return value does not guarantee that a mail has actually been sent (let alone delivered), a negative (well, zero) value should definitely indicate an issue with the mail delivery, imo.

raphaelm:
I came to the conclusion that ``mail()`` shouldn't return anything at all and should raise an exception whenever it is clear that it errored. Returning a boolean to indicate success is neither pythonic nor good style.

rixx:
Would you want to include error handling on all pretix-core calls of `mail()`? There are only about five of them, but we''d need to decide on what to do with them. None of these checked the return value of `mail()` up until now.

raphaelm:
Fixed in #169

PR #169, opened 2016-08-09
Make mail() raise an exception on obvious failures

rixx:
As per #164 

Can be either committed as-is, or I could add error-handling to all `mail()` calls. Not sure how useful this would be, especially since mail errors will be logged regardless. I probably should at least do this for the sendmail plugin?


raphaelm:
I believe this is fine for now.

rixx:
Erm, "obvious" failures while sending mails would now lead to user-facing 500s in some places - at least a generic try/except would probably be best?

PR #171, opened 2016-08-10
Voucher related fixes

rixx:
Two fixes for voucher-related issues, as referenced.

### DIFF
--- a/src/pretix/base/models/vouchers.py
+++ b/src/pretix/base/models/vouchers.py
@@ -160,10 +160,12 @@ class Voucher(LoggedModel):
 
     def is_ordered(self) -> int:
         """
-        Returns whether an order position exists that uses this voucher.
+        Returns whether a non-canceled order position exists that uses this voucher.
         """
         return OrderPosition.objects.filter(
             voucher=self
+        ).exclude(
+            order__status=Order.STATUS_CANCELLED
         ).exists()
 
     def is_in_cart(self) -> int:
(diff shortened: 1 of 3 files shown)


raphaelm:
Thank you for working on this. I'm very sorry to say it this way, but unfortunately neither of the two solves the actual problem at hand.

On the cancellation issue: ``is_ordered()`` is a legacy method that is not used for quota calculation. ``is_ordered`` and its two usages should be removed altogether. The quota calculation only uses the ``redeemed`` value. Therefore, to solve this problem, we'd need to actually set ``redeemed = False`` in ``cancel_order`` on the appropriate vouchers.

On the availability issue: If it where possible to block quota with a voucher that allows for multiple variations, we could have the following scenario: Voucher V is applicable to any variation of item P, however variation A is already sold out. As ``block_quota`` skips further availability checks as long as the voucher is valid, this could allow a user to buy A although the quota is empty. Therefore it is forbidden in ``Voucher.clean()`` in ``base/models/vouchers.py:146`` to have ``block_quota`` in such a case and we don't need to check for this case in the form. There are two ways to resolve the exception:

* just hardcode ``avail = (Quota.AVAILABILITY_OK, 99999)`` as it will throw an error later on anyway
* change ``clean()`` in the form in a way that the availability check is only performed in the case where ``block_quota`` is on and it makes sense.

Either way, this shows that we desperately need tests for this form.

rixx:
As for the first one: please note that I set `position.voucher.redeemed = False` in my patch, which is exactly what you are proposing. I just changed `is_ordered` in addition for consistency.

Which way do you want to move forward on the second issue, that I looked at in 998d9c3? Because at the moment, you'll get an error (500, introduced about a week ago) when you try to create a voucher that applies to all variations of a product. Like, any voucher. This really needs to be fixed one way or the other.

Issue #177, opened 2016-08-12
Add comment field to orders


rixx:
Analogous to voucher comment fields - available internally for staff/support to denote arrangements etc. Would not be displayed to the user at any point.

Questions:
 - where in the order detail view should the comment be displayed (including a "Save" button)

raphaelm:
Fixed in #178. Thanks!

PR #178, opened 2016-08-12
Add comment field to Orders

rixx:
as per #177 

TODO: consider a good position for the comment field in the order view. At the moment it is kind of just attached at the bottom.


raphaelm:
Would you please include the number of the issue that is fixed by this PR in the commit message?

Regarding the position of the field, I believe it is fine for now. :)

Issue #170, opened 2016-08-10
Cannot create/update voucher for an item with variations


rixx:
pretix crashes when trying to create a voucher for an item with variations (if you choose the dropdown entry for "Item X - Any variation").

PR #184, opened 2016-08-13
[WIP] Fixes bug in voucher generation for items with variations

rixx:
Bug was for voucher generation to fail when an item with variations was
chosen.
In items with variations, vouchers will be generated if at least one
variation is still available and the sum of available items is larger
than the amount of vouchers being generated.

It would maybe be worth consideration to make this the implementation of
Item.check_quotas for items with variations.

Refs #170.


raphaelm:
As discussed elsewhere, I need to close this because I need to rewrite big parts of the ``clean`` logic, as much as I appreciate the logic you wrote here.

Issue #195, opened 2016-08-19
Tell users that their invoice won't change automatically


rixx:
When a user changes their invoice address, they might expect their invoice to change, too. Add a notification on-change (maybe even pre-change?) that they should message the support to get a new version of their invoice.

PR #196, opened 2016-08-19
Add message explaining invoice address updating process (#195)

rixx:
Adds messages both pre- and post-update explaining to the user that the
invoice itself will not be automatically regenerated and that support
interaction is required for this step. Might reduce user frustration. (re: #195)

What this doesn't do

 - it will display "please contact us at None" if no contact email is configured
 - it doesn't display the email as mailto: link

I wasn't sure how to do *both* of these things idiomatically while using translation. Any help appreciated.

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/order_modify.html
+++ b/src/pretix/presale/templates/pretixpresale/event/order_modify.html
@@ -12,6 +12,12 @@
         {% csrf_token %}
         <div class="panel-group" id="questions_accordion">
             {% if event.settings.invoice_address_asked %}
+                <div class="alert alert-info">
+                    {% blocktrans trimmed %}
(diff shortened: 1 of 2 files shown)


rixx:
More random session key errors in travis tests.

rixx:
More random session key errors in travis tests.

raphaelm:
I believe it is hard to solve those problems in a really idiomatic way, especially the second one.

* The "contact us at None" problem needs to be solved before merging this. The only clean ways I see is to (a) provide two different translation strings or to (b) make that field required. I could live with both options (as I really want organizers to fill in an address to stop people from contacting *me*).
* Links in translation are hard, I often include the ``<a href="{{url}}"`` in the translation string and pass in the URL from outside


<a href="{{link}}>

Issue #122, opened 2016-02-11
Resend link option for users


raphaelm:
E.g. "Send me an email with links to the download and detail pages of all orders associated with this email address"

- [ ] Spam protection

PR #197, opened 2016-08-19
Allow users to re-send order links (#122)

rixx:
If the provided mail address has not ordered anything, there will still
be a mail generated and sent to an invalid mail address, to avoid
obvious timing attacks to determine active users. (#122)

TODO:

 - [x] is `resend/` a good URL?
 - [ ] link in frontend
 - [x] invalid address as a feature
 - [x] tests
 - [x] spam protection as in https://github.com/pretix/pretix/blob/master/src/pretix/control/views/auth.py#L99

From review:
 - [x] use exception.logger
 - [x] use separate form
 - [x] move mails to format string/config variable instead of template

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/resend_link.html
+++ b/src/pretix/presale/templates/pretixpresale/event/resend_link.html
@@ -1,0 +1,31 @@
+{% extends "pretixpresale/event/base.html" %}
+{% load i18n %}
+{% load bootstrap3 %}
+{% block title %}{% trans "Resend order links" %}{% endblock %}
+{% block content %}
+    <h2>
+        {% trans "Resend order links" %}
+    </h2>
+    <div class="row">
+        <div class="panel-body">
+        {% blocktrans trimmed %}
+            If you lost the link to your order or orders, please enter the email address you
+            used for your order. We will send you an email with links to all orders you placed
+            using this email address.
+        {% endblocktrans %}
+    </div>
+    </div>
+    <div class="row">
+    <form class="form" method="post">
+        {% csrf_token %}
+        <div class="col-md-8 col-sm-6 col-xs-12">
+                {% bootstrap_form form layout="inline" %}
+        </div>
+        <div class="col-md-4 col-sm-6 col-xs-12">
+            <button class="btn btn-primary btn-block" type="submit">
+                {% trans "Send links" %}
+            </button>
+        </div>
+    </form>
+    </div>
+{% endblock %}
--- a/src/pretix/presale/views/user.py
+++ b/src/pretix/presale/views/user.py
@@ -1,0 +1,73 @@
+import logging
+
+from django.conf import settings
+from django.contrib import messages
+from django.shortcuts import redirect
+from django.utils.functional import cached_property
+from django.utils.translation import ugettext_lazy as _
+from django.views.generic import TemplateView
+
+from pretix.base.services.mail import INVALID_ADDRESS, SendMailException, mail
+from pretix.multidomain.urlreverse import build_absolute_uri, eventreverse
+from pretix.presale.forms.user import ResendLinkForm
+from pretix.presale.views import EventViewMixin
+
+
+class ResendLinkView(EventViewMixin, TemplateView):
+    template_name = 'pretixpresale/event/resend_link.html'
+
+    @cached_property
+    def link_form(self):
+        return ResendLinkForm(data=self.request.POST if self.request.method == 'POST' else None)
+
+    def post(self, request, *args, **kwargs):
+        if not self.link_form.is_valid():
+            messages.error(self.request, _('We had difficulties processing your input.'))
+            return self.get(request, *args, **kwargs)
+
+        user = self.link_form.cleaned_data.get('email')
+
+        if settings.HAS_REDIS:
+            from django_redis import get_redis_connection
+            rc = get_redis_connection("redis")
+            if rc.exists('pretix_resend_{}'.format(user)):
+                messages.error(request, _('We already sent you an email in the last 24 hours.'))
+                return redirect('presale:user.resend')
+            else:
+                rc.setex('pretix_resend_{}'.format(user), 3600 * 24, '1')
+
+        orders = self.request.event.orders.filter(email=user)
+        order_context = []
+
+        for order in orders:
+            url = build_absolute_uri(
+                self.request.event,
+                'presale:event.order',
+                kwargs={'order': order.code, 'secret': order.secret}
+            )
+            order_context.append(' - {} - {}'.format(order, url))
+
+        if not orders:
+            user = INVALID_ADDRESS
+
+        subject = _('Your orders for {}'.format(self.request.event))
+        template = self.request.event.settings.mail_text_resend_all_links
+        context = {
+            'orders': '\n'.join(order_context),
+            'event': self.request.event,
+        }
+        try:
+            mail(user, subject, template, context, event=self.request.event, locale=self.request.LANGUAGE_CODE)
+        except SendMailException:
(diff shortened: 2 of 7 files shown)


src/pretix/presale/views/user.py
```diff
@@ -0,0 +1,67 @@
+from django.contrib import messages
+from django.shortcuts import redirect
+from django.utils.functional import cached_property
+from django.utils.translation import ugettext_lazy as _
+from django.views.generic import TemplateView
+
+from pretix.base.services.mail import mail
+from pretix.multidomain.urlreverse import build_absolute_uri, eventreverse
+from pretix.presale.forms.checkout import ContactForm
```
raphaelm: I don't *really* fancy using that form as it is semantically a very different thing.

src/pretix/presale/views/user.py
```diff
@@ -0,0 +1,67 @@
...
+        subject = _('Your orders for {}'.format(self.request.event))
+        template = 'pretixpresale/event/resend_link_email.txt'
+        context = {
+            'orders': order_context,
+            'event': self.request.event,
+        }
+        try:
+            mail(user, subject, template, context)
+        except Exception as e:
+            print(e)
```
raphaelm: I believe this is can be replaced by ``logger.exception``

raphaelm:
Thoughts on this:

> is resend/ a good URL?

I  believe it is.

> better invalid mail address (if we like this feature)

Wow. I wouldn't have thought of this timing attack vector. Very very nice. However, sending actual emails to the ``invalid_user@localhost`` will spam the return address with bounces and is therefore not an option. In a production setup, however, ``mail()`` will not talk to the SMTP server directly but will trigger a celery task. It is therefore sufficient to either

* Catch this case inside ``mail_send`` and drop the sending request
* Don't call ``mail`` but call another celery task that performs a noop

> spam protection

I'm in favor of a very simple redis-only approach like in the password reset feature: [auth.py](https://github.com/pretix/pretix/blob/master/src/pretix/control/views/auth.py#L99)

One more thing: You generate the settings from a template. I moved all other event-related emails from templates to simpler format strings that are stored as a setting with a sensible default value. The reason is that the wording of emails to users is something that some organizers take very seriously and that I'd therefore like to have configurable.

rixx:
Hm, as of now, links to all orders are sent, including cancelled orders. Do we want that?

rixx:
I included tests, the only thing missing now is a link in the frontend. PR is basically done on my part.

raphaelm:
> Hm, as of now, links to all orders are sent, including cancelled orders. Do we want that?

Yes, I think we want that.

src/pretix/presale/views/user.py
```diff
@@ -0,0 +1,77 @@
...
+
+        subject = _('Your orders for {}'.format(self.request.event))
+        template = self.request.event.settings.mail_text_resend_all_links
+        context = {
+            'orders': '\n'.join(order_context),
+            'event': self.request.event,
+        }
+        try:
+            mail(user, subject, template, context)
+        except Exception:
```
raphaelm: As I merged the other PR, this can be replaced by SendMailException

src/pretix/presale/views/user.py
```diff
@@ -0,0 +1,77 @@
...
+        orders = self.request.event.orders.filter(email=user)
+        order_context = []
+
+        for order in orders:
+            url = build_absolute_uri(
+                self.request.event,
+                'presale:event.order',
+                kwargs={'order': order.code, 'secret': order.secret}
+            )
+            order_context.append(' - {} ({})'.format(order, url))
```
raphaelm: Please use `` - {} - {}`` instead. With your format, some mail clients might interpret ``)`` as part of the URL

src/pretix/presale/views/user.py
```diff
@@ -0,0 +1,77 @@
...
+        }
+        try:
+            mail(user, subject, template, context)
+        except Exception:
+            logger = logging.getLogger('pretix.presale.user')
+            logger.exception('A mail resending order links to {} could not be sent.'.format(user))
+            messages.error(self.request, _('We have trouble sending emails right now, please check back later.'))
+            return self.get(request, *args, **kwargs)
+
+        messages.success(self.request, _('If there were any orders by this user, they will receive an email with their order codes.'))
```
raphaelm: "they"?

src/tests/presale/test_event.py
```diff
@@ -245,3 +246,47 @@ def test_in_time(self):
...
+            total=0,
+        )
+        mail.outbox = []
+        url = '/{}/{}/resend/'.format(self.orga.slug, self.event.slug)
+        resp = self.client.post(url, data={'email': 'dummy@dummy.dummy'})
+
+        self.assertEqual(resp.status_code, 302)
+        self.assertEqual(len(mail.outbox), 1)
+        self.assertIn('DUMMY1', mail.outbox[0].body)
+        self.assertIn('DUMMY2', mail.outbox[0].body)
```
raphaelm: Couldy you, just that we are 120% sure for eternity, include a test where an order exists that is NOT by the requesting user?

src/pretix/presale/views/user.py
```diff
@@ -0,0 +1,77 @@
...
+            user = INVALID_ADDRESS
+
+        subject = _('Your orders for {}'.format(self.request.event))
+        template = self.request.event.settings.mail_text_resend_all_links
+        context = {
+            'orders': '\n'.join(order_context),
+            'event': self.request.event,
+        }
+        try:
+            mail(user, subject, template, context)
```
raphaelm: I think we might need to pass ``event=self.request.event`` and ``locale=self.request.LANGUAGE_CODE`` to ``mail()``

src/pretix/presale/views/user.py
```diff
@@ -0,0 +1,77 @@
...
+
+class ResendLinkView(EventViewMixin, TemplateView):
+    template_name = 'pretixpresale/event/resend_link.html'
+
+    @cached_property
+    def link_form(self):
+        return ResendLinkForm(data=self.request.POST if self.request.method == 'POST' else None)
+
+    def dispatch(self, request, *args, **kwargs):
+        self.request = request
```
raphaelm: ``self.request`` is already set at this point thanks to ``View`` :)

src/pretix/presale/templates/pretixpresale/event/resend_link.html
```diff
@@ -0,0 +1,31 @@
...
+            using this email address.
+        {% endblocktrans %}
+    </div>
+    </div>
+    <div class="row">
+    <form class="form-horizontal" method="post">
+        {% csrf_token %}
+        <div class="col-md-3 col-xs-12">
+            <div class="panel-body">
+                {% bootstrap_form form layout="inline" %}
```
raphaelm: The form doesn't yet look very good as it is very small in horizontal size. I'd prefer it to be in one line, e.g. in an analogy to the voucher redemption form.

raphaelm:
I gave this a more thorough look and test and found some minor issues. The overall concept design good (still loving the timing attack part)

src/pretix/presale/views/user.py
rixx: It's an established gender-neutral pronoun - what would you like to use instead?

rixx:
I worked through all your comments, thanks for the review!

src/pretix/presale/views/user.py
raphaelm: Uh. Just not used to it yet and read it as plural (and therefore thought it refered to the "orders"). I just wasn't *that* awake either, maybe? Another gender-neutral solution could be: "If there were any orders with this email address, YOU will receive an email with their order codes."

raphaelm:
Merged. Thanks! :)

PR #199, opened 2016-08-20
Export an item's tax_rate in the JSON export

rixx:
This might be required for various depending services; knowing a price without the included tax isn't useful in a lot of places.


rixx:
Another example of those random failing tests regarding session keys etc. :\

raphaelm:
Thanks a lot!

Issue #180, opened 2016-08-13
"Add to cart" is shown when there are no products listed


rixx:
![2016-08-13-121250_1289x371_scrot](https://cloud.githubusercontent.com/assets/2657547/17642681/4655263c-614f-11e6-8584-c7277a52ac11.png)

raphaelm:
If someone wants to tackle this as his or her first issue in the project, do not hesitate to ask me questions if you get stuck!

vishalsodani:
I would like to have a go at this. I have figured out that we have to check - 

1. if there are no items active or 
2. if no vouchers exist for any of the items 

then the 'Add to cart' button can be hidden.  Am I correct?

rixx:
Hi! :)

I think your first point is sufficient - voucher redemption works separately from the normal way items are added to a cart.

vishalsodani:
Hi,

I actually meant to say "if no available quantity in quotas exist for any of the items" for second point.

I  ran the unit tests but some of them are failing. Are they supposed to? I am using sqllite3 for tests.

rixx:
That sounds basically correct, but it might be a bit more simple: Pretix is already doing all of these checks in the view (if an item is active, if the item or its variations are still available according to their quotas). It then supplies the template with a variable containing all items that are active and available, `items_by_category`, which is then rendered by the template.

As long as this variable is filled, we need the "Add to cart" button, but if it is empty, the button is not necessary. Does that sound about right to you?

vishalsodani:
Yes. I will push my change.

vishalsodani:
Hi,

I have pushed my change. I see 2 tests are failing as they expect 'Add to cart' row.

rixx:
I think @raphaelm is away atm, so it could take until next week until he has a look at your pull request.

As for the tests, I think the best thing would be for you to adjust the tests so that they don't expect the 'Add to cart' button on empty item lists - instead they could assert that the button is **not** there :)

PR #200, opened 2016-08-21
Hide add to cart button if no products active or no quantity left(#180)

vishalsodani:


raphaelm:
Hey @vishalsodani  just for your information, I'm on vacation right now and will get back to this as soon as possible, but maybe not before next week. Thank you for your work!

raphaelm:
Hi @vishalsodani thank you very much for this work!
However, I don't believe this behaves correctly for more than one item. Let's assume for a minute that there are two items of which the first one is available and the second one is not. In that case, in the first iteration of ``for item in items`` the ``display_add_to_cart`` flag will be set to ``True``, in the second iteration it will be overridden with ``False`` and is ``False`` in the end. Do I see this correctly? If so, a possible solution might be something like:

```
display_add_to_cart = display_add_to_cart or item.order_max > 0
```

vishalsodani:
Hi @raphaelm You are absolutely right! I will fix and commit as per your suggestion.

raphaelm:
Looks good now! Thank you very much for the contribution!

Issue #91, opened 2015-08-16
Make products/orders that are not cancelable


Unicorn-rzl:
May be there are some products which shouldn't be cancelable?

raphaelm:
Orders can only be cancelled before they are paid. So, introducing non-cancelable items only make sense if you're really going to forcefully collect the money for unpaid orders. 
I agree that this makes sense for some events and should be possible.

raphaelm:
To implement this, the following steps are required:

* Add a boolean property as ``allow_cancel`` to the ``Item`` model that defaults to ``True`` and create a migration
* Add this property to the item editing form
* Add an appropriate check to ``Order.can_user_cancel``

If someone wants to tackle this as his or her first issue in the project, do not hesitate to ask me questions if you get stuck!

PR #204, opened 2016-08-22
Make products/orders that are not cancelable

esaezgil:
As suggested in https://github.com/pretix/pretix/issues/91, make products/orders that are not cancelable
Update Unit Tests


src/pretix/base/models/orders.py
```diff
@@ -229,7 +229,14 @@ def can_modify_answers(self) -> bool:
 
     @property
     def can_user_cancel(self) -> bool:
-        return self.event.settings.cancel_allow_user
+        positions = self.positions.all().prefetch_related('item')
```
raphaelm: I'd prefer ``select_related`` over ``prefetch_related`` in this case as it will cause one query less. Would you mind changing this?

src/pretix/base/models/orders.py
```diff
@@ -229,7 +229,14 @@ def can_modify_answers(self) -> bool:
...
+        cancelable = True
+        cancelable_products = []
+        for i, op in enumerate(positions):
+            cancelable_products.append(op.item.allow_cancel)
+        if False in cancelable_products:
```
raphaelm: I believe this could be written in a more concise way that is still very readable, for example:

```
cancelable = not all([op.item.allow_cancel for op in positions])
```

raphaelm:
This looks very good, thank you very much! I left two minor comments on how I think this could be improved before merging.

As I just merged your other PR, this will cause a merge conflict. To resolve this, this branch would need to be rebased against the new master and then you would need to [merge migrations](https://docs.djangoproject.com/el/1.10/ref/django-admin/#cmdoption-makemigrations--merge). If you have problems with this, just ask here or if you do not want to dig into this, just say so and I'll fix this up myself before merging.

esaezgil:
Hi Raphael, thanks for reviewing the PR. Both comments are spot on, I'll try to update the PR over the next couple of days :)

src/pretix/base/models/orders.py
esaezgil: I think this should be `cancelable = all([op.item.allow_cancel for op in positions])` without the `not`

raphaelm:
Looks very good to me. Thanks a lot! :)

Issue #176, opened 2016-08-12
Provide localized "Bank account details" field in Bank transfer plugin


rixx:
Bank account details might include words like "Bank", "Address", "account holder" etc that should be localized.

raphaelm:
Yes, that shouldn't be too hard, but might require some insight into our settings localization toolchain, so I'll not mark this as easy fo rnow.

PR #205, opened 2016-08-22
Localize bankimport settings (#176)

rixx:
This is a non-destructive upgrade since LazyI18nString will return a normal string (which is content of the settings variable up until now) as-is. Tested on all provided sites and email.

Note: I had to wrap `I18nString` around the config value manually, which is kind of understandable, but I also had to set the event locales on the form field, since `SettingsForm` does its logic in `__init__`, but plugin fields get added afterwards. Do you think we could find a solution for localized forms where even add-on fields will get their locale automatically? Maybe just provide a helper method (`.add_field(self, field)`)?

Edit: Hm. Never mind the paragraph above, it would be fairly impractical to implement. Not worth the effort for the one line of code, I think.


src/pretix/plugins/banktransfer/payment.py
```diff
@@ -14,19 +15,22 @@ class BankTransfer(BasePaymentProvider):
...
+            list(super().settings_form_fields.items()) + [('bank_details', form_field)]
         )
 
     def payment_form_render(self, request) -> str:
         template = get_template('pretixplugins/banktransfer/checkout_payment_form.html')
-        ctx = {'request': request, 'event': self.event, 'settings': self.settings}
+        ctx = {
+            'request': request,
+            'event': self.event,
+            'details': LazyI18nString(self.settings.get('bank_details')),
```
raphaelm: You could use ``self.settings.get('bank_details', as_type=LazyI18nString)`` here :)

raphaelm:
The manual ``I18nString`` wrapping can be solved by ``as_type`` (which would make it more future-proof in theory, if settings serialization changes). Apart from that, looks good to me.

src/pretix/plugins/banktransfer/payment.py
rixx: Done.

raphaelm:
> Do you think we could find a solution for localized forms where even add-on fields will get their locale automatically?

Another thought regarding this: Could this be solved by a one-liner in ``ProviderForm.prepare_fields``? If that is the case, payment plugins would not need to care.

rixx:
Meh, as I said: I don't think it's worth the effort for one line of code.

raphaelm:
I believe it is worth to include one line in core to save plugin authors to think of one magic line every time. I merged your PR manually and changed it as I was working and testing with i18n forms anyway.

PR #207, opened 2016-08-22
Invoices

rixx:
This pull request adds

 - a footer field
 - a second text field, displayed above the invoice rows (for greetings etc)
 - a column displaying the net value of every invoice row

This pull request fixes a bug where line breaks were not rendered in the multi-line text field.


raphaelm:
Looks good to me, except this one question: Should we make all those fields ``LazyI18nString``s?

PR #210, opened 2016-08-23
Use get_random_string everywhere

rixx:
Django's get_random_string tries really hard to either use sysrandom or
be otherwise as unpredictable as possible.


raphaelm:
Awesome, thanks!

Issue #166, opened 2016-08-07
Don't display "resend link" in canceled order


rixx:
Canceled orders have no business being exposed again, and sending the link of a canceled order to a user would probably only lead to confusion.

raphaelm:
Agreed. A small template change in the order detail view should be sufficient.

If someone wants to tackle this as his or her first issue in the project, do not hesitate to ask me questions if you get stuck!

PR #214, opened 2016-08-24
Do not resend links to cancelled orders (#166)

rixx:


rixx:
Weird test failures that I can't reproduce locally. No idea what's up with that.

raphaelm:
Thanks!

PR #215, opened 2016-08-24
Catch and display mail sending errors

rixx:


src/pretix/plugins/sendmail/views.py
```diff
@@ -36,11 +36,18 @@ def form_valid(self, form):
...
+            try:
+                mail(o.email, form.cleaned_data['subject'], form.cleaned_data['message'],
+                     None, self.request.event, locale=o.locale, order=o)
+            except Exception:
+                failures.append(o.email)
+
+        if failures:
+            messages.error(self.request, _('Failed to send mails to the following users: {}'.format(' '.join(failures))))
+        else:
+            messages.success(self.request, _('Your message has been sent to the selected users.'))
```
raphaelm: No, in a production setup, it has only been queued for celery to send, therefore this change (and its translation) is wrong.

raphaelm:
I'm pretty strongly against ``catch Exception`` because in theory, totally mail-unrelated exceptions could completely bypass a system like Sentry this way. I'm okay with this if you add a statement like:

```
logger.exception('Error sending an email')
```

to every of the ``catch`` blocks. Python's ``logging`` module will automatically add the traceback to the logfile.

rixx:
Fixed the string to use "queued to be sent".

As for catching `Exception` I agree - let's introduce a `SendMailException` instead? We can still catch it and `logger.exception()` the thing, and we don't catch `Exception`, then.

raphaelm:
> As for catching `Exception` I agree - let's introduce a `SendMailException` instead? We can still catch it and `logger.exception()` the thing, and we don't catch `Exception`, then.

Even better.

rixx:
:ballot_box_with_check:

PR #219, opened 2016-08-24
Add a count of elements to pagination elements

rixx:
Also, add a second, upper pagination element to orders and attendees
list sites as those tend to get long.


raphaelm:
Thanks!

Issue #220, opened 2016-08-25
Refactor user input handling/validation in orders and cart views


rixx:
At the moment, cart and presale data are rendered by using hand-made html forms and "manual" validation of POST parameters. Even though Django forms might seem clumsy for relatively dynamic content, being able to just call `form.is_valid()` might provide more security and stability than a custom validation logic of raw POST params.

raphaelm:
This might be a solution, but I believe there are other, maybe easier ways to achive this than to squeeze this into Django forms. Let's discuss this when we refactored the services.orders into a state machine and use that structure to refactor services.cart.

rixx:
In that case, let's keep this issue as a general "unify user input validation/handling in orders and cart" issue.

raphaelm:
I think CartManager does a decent job here

PR #221, opened 2016-08-25
Assert that the item count is a positive integer

rixx:
Thanks to David Gullasch for pointing this one out. - This is an example of why refactoring user input validation as per #220 might be a good idea.

### DIFF
--- a/src/pretix/presale/views/cart.py
+++ b/src/pretix/presale/views/cart.py
@@ -56,13 +56,23 @@ class CartActionMixin:
                 parts = parts[:-1]
             else:
                 voucher = None
+
+            try:
+                amount = int(value)
+            except ValueError:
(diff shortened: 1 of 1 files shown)


src/pretix/presale/views/cart.py
```diff
@@ -56,13 +56,24 @@ def _items_from_post_data(self):
                 parts = parts[:-1]
             else:
                 voucher = None
+
+            try:
+                amount = int(value)
+                assert amount > 0
```
raphaelm: Could we use a simple ``if`` instead of an ``assert``? ``assert`` is not run if ``__debug__`` is ``False`` which might happen in production (``python -O``).

rixx:
Removed the `assert` statement.

Issue #152, opened 2016-07-19
Render linebreaks in multiline answers to questions


raphaelm:
If a product has a question attached that allows for multiline answers, the answer is currently rendered within one line. This can be easily fixed by adding appropriate template filters.

FlaviaBastos:
Hey, @raphaelm!
I was looking into this issue but I need some directions, please.
I understand that you can duplicate this issue when you add the item that has a question linked to it to the cart and checkout, right? 
I'm stuck on my 'Checkout' page where I have to choose payment method but it has 'go back' and 'continue' only. I'm wondering if I'm missing any step to run the dev server that will display payment methods so I can pick one and continue. OR could it be because my shop is not on live mode yet? :confused: 
Thanks!

raphaelm:
Hi :) It is partly related to live mode -- if you try to switch it to live mode, it will tell you what the problem is: No payment method is active. The easiest way is to first go to Settings->Plugins and activate Bank transfer, then go to Settings->Payment and activate it there as well :)

esaezgil:
@raphaelm maybe that explanation could be included in the docs?

raphaelm:
@esaezgil I don't think that should go into the developer docs, I rather think we need to improve the UI so that the solution becomes obvious as soon as the problem occurs. This will not only be encountered by developers.

PR #224, opened 2016-08-31
Linebreak for multiline answers - Issue 152

FlaviaBastos:
Multiline answers are now rendered appropriately. 

Issue #152


raphaelm:
Hi :) I just tested this locally and ``|linebreaks`` renders a ``<p>`` tag that adds a margin of a few pixels below the answer (which looks inconsistent with other question types). I believe the easiest fix would be to use ``|linebreaksbr`` instead.

FlaviaBastos:
ops... :flushed: Thanks for pointing that out. I just changed to `linebraksbr` instead. :)

raphaelm:
Well, it wasn't really easy to spot this :) Thanks for the PR!

PR #225, opened 2016-09-06
Invoice string from payment provider

rixx:
One commit for functionality, one for less code duplication. Please review.

Tested manually, and only a bit since I didn't have more than one testable payment provider set up.


src/pretix/base/models/vouchers.py
```diff
@@ -163,7 +163,7 @@ def delete(self, using=None, keep_parents=False):
         super().delete(using, keep_parents)
         self.event.get_cache().delete('vouchers_exist')
 
-    def is_ordered(self) -> int:
+    def is_ordered(self) -> bool:
```
raphaelm: Could you remove that change? :) This leads to a merge conflict, because as discussed I finally got rid of this method now. (Didn't push yesterday, my fault.)

raphaelm:
Looks good to me except the merge conflict that you could not know about yet (see comment) and one minor issue:
Directly accessing ``payment_provider.settings.get('_invoice_text', as_type=LazyI18nString)`` seems unclean to me, as this is an internal of the payment provider. I believe it would be cleaner to add a new ``render_invoice_text`` method to ``BasePaymentProvider`` that does this and may be overridden with custom behaviour.

src/pretix/base/payment.py
```diff
@@ -145,6 +153,15 @@ def settings_content_render(self, request: HttpRequest) -> str:
         pass
 
     @property
+    def invoice_text(self) -> str:
```
raphaelm: Wouldn't it be nice to give the provider even more flexbility by not using a property and passing the ``Order`` object in?

src/pretix/base/services/invoices.py
```diff
@@ -71,16 +81,11 @@ def regenerate_invoice(invoice: Invoice):
             invoice.invoice_to = ""
 
         invoice.file = None
+        invoice.lines.all().delete()
```
raphaelm: I haven't tested yet, but in the case of a new invoice this is called before the invoice even is created and has a primary key… does that actually work?

raphaelm:
Okay, I found a few last issues:

* If the new setting was never set, ``render_invoice_text`` returns ``None`` which is printed verbatim on the invoice. I beliefe a ``default=`` argument to the ``settings.get`` call should fix this
* If you cancel an invoice, a duplicate key error occurs. This is because you save the invoice in ``generate_cancellation`` *before* you reset ``invoice_no`` in ``build_cancellation``. You need to reset this value earlier. Also, you need to set ``invoice.is_cancellation = True`` there.
* Could include document the new method in ``doc/development/api/payment.rst``?

raphaelm:
Thanks! \o/

Issue #233, opened 2016-09-13
Filter orderlist by payment_provider


pc-coholic:
Oder overview offers a links to a filtered order-list (like payment pending, cancelled, refunded, etc).

There is even a breakdown on the "Payment method fees" per payment_provider. However clicking for example on the "Payment Pending"-link in the "Stripe"-row, won't just return the pending orders that are associated with this payment-method. Instead, all payments are returned.

PR #234, opened 2016-09-13
Make orderlist and overview filterable by payment_provider (fixing #233)

pc-coholic:


pc-coholic:
Also für diesen gefailten build fühle ich mich ja nicht so richtig verantwortlich ;-)

raphaelm:
Ja :/ Wir haben zwei nicht-deterministische Test-Failures die auf Travis ständig auftreten aber bei mir lokal nie. Super geil zum Debuggen.

src/pretix/control/views/orders.py
```diff
@@ -66,12 +66,29 @@ def get_queryset(self):
...
+            p = self.request.GET.get("provider", "")
+            qs = qs.filter(payment_provider=p)
         return qs.distinct()
 
+    def get_payment_providers(self):
+        providers = []
+        responses = register_payment_providers.send(self.request.event)
+        for receiver, response in responses:
+            provider = response(self.request.event)
+            if not provider.is_enabled or not provider.is_allowed(self.request):
```
raphaelm: You can remove the ``is_allowed`` constraint here, it barely matters whether the currently logged-in user can use this provider to *buy* something :)

raphaelm: One could probably argue that we should ignore ``is_enabled`` here as well, because even if a provider is (temporarily) disabled, there might be orders using that provider.

Providers from disabled plugin won't show up in the responses in the first place.

PR #242, opened 2016-09-16
Change signature-delimiter for e-mails

cherti:
`"-- "` on a single line as a signature-delimiter is noticed by a lot of mail programs out there as an actual signature separator and the text below may be marked that way, e.g. is greyed out a little in Mozilla Thunderbird.

This change should leverage that behavior for e-mails sent by pretix.


raphaelm:
Good catch, thanks a lot! :)

Issue #187, opened 2016-08-14
Unify usages of canceLed vs canceLLed


raphaelm:
currently is mixed througout the sourcecode and texts... LL leads in usage, L is probalby what we want.

FlaviaBastos:
Hi, @raphaelm 
Both 'cancelled' and 'canceled' [are correct](http://www.dictionary.com/browse/cancelled). It's a matter of preference (American x British), I think. 
I can work on this, if that's ok.
I'm assuming you will want all to be 'canceLed', correct?

raphaelm:
Hi @FlaviaBastos!

I prefer the L version, because American English is more common in the project and in programming in general, however, LL is already more widespread in the project. (153x LL and 26x L). But I guess, we should still go for L, as consistency wins.

Replacing the strings is straightforward, replacing variable names requires some caution. Especially renaming things like ``Order.STATUS_CANCELLED`` is changing a public API that plugins could depend on. However, there are not yet any relevant plugins out there so if there is a time to change this, it is now.

So, yeah, if you want to, go ahead :)

FlaviaBastos:
Got it! Thanks!
I will start working on this and reach out if needed.

PR #243, opened 2016-09-20
Unify spelling of 'canceled' - #187

FlaviaBastos:
Switching 'cancelled' (double 'L') to 'canceled' (single 'L') across all files.

#187

I skipped all `src/pretix/base/migrations/*` files (I understand that these files are overwritten when new migrations are done, but I'm not sure about this :confused: ). Please let me know if they need to be included.

### DIFF
--- a/doc/development/implementation/logging.rst
+++ b/doc/development/implementation/logging.rst
@@ -16,7 +16,7 @@ We recommend all relevant models to inherit from ``LoggedModel`` as it simplifie
 
 To actually log an action, you can just call the ``log_action`` method on your object::
 
-   order.log_action('pretix.event.order.cancelled', user=user, data={})
+   order.log_action('pretix.event.order.canceled', user=user, data={})
 
 The positional ``action`` argument should represent the type of action and should be globally unique, we
 recomment do prefix it with your packagename, e.g. ``paypal.payment.rejected``. The ``user`` argument is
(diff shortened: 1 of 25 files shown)


raphaelm:
Hey :) Looks good to me so far! Only one suggestion: Could you revert the changes in ``src/static/typeahead/typeahead.bundle.js``? That code is not by us but part of [typeahead.js](https://twitter.github.io/typeahead.js/) and this will lead to confusion if we ever update that library.

FlaviaBastos:
Sure! Just... hm... :flushed: any recommendation on how I can revert this file back - 'git speaking'? :grin: I've been reading about `git revert` but that seems to affect the whole commit, not just the one file. Is this right? 
Or should I just manually track the changes I made and undo them?

FlaviaBastos:
I found something that seems to fit my case ([from here](http://stackoverflow.com/questions/2733873/reverting-a-single-file-to-a-previous-version-in-git)):

> you'll be able to find the SHA1 (hash) of the commit with the version of the file you want. Now, all you have to do is this:
> 
> # get the version of the file from the given commit
> `git checkout <commit> path/to/file`
> # and commit this modification
> `git commit`

I guess I will try this.

raphaelm:
Yes, that is exactly what I just wanted to suggest :) If you do ``git commit --amend`` and then push with ``-f``, it will even squashed into the same commit so that there is just one single commit in the end.

FlaviaBastos:
Did it work? :open_mouth: What is the emoji for sweaty-shaky hands?? :laughing: 
I was stuck with a typo.... :rage:

raphaelm:
Yes it worked :) Thank you very much!

PR #251, opened 2016-09-26
Fix typo on Stripe plugin configuration alert.

FlaviaBastos:
Small typo fix on Stripe plugin configuration alert displayed in payment settings.


raphaelm:
Good catch, thansk!

FlaviaBastos:
:laughing: :+1:

Issue #259, opened 2016-10-04
Cannot login in control: TypeError at /control/ process() takes 2 positional arguments but 3 were given


FlaviaBastos:
This is on a fresh clone/install on mac os x el capitan (v 10.11.6), running python 3.4, following all the steps outlined in docs (https://docs.pretix.eu/en/latest/development/setup.html)

- run `python manage.py runserver`
- access: http://localhost:8000/control/ (same result with http://localhost:8000/mrmcd/2015/)
- login

After login page, I get:
`TypeError at /control/
process() takes 2 positional arguments but 3 were given
Request Method:	GET
Request URL:	http://localhost:8000/control/
Django Version:	1.10.2
Exception Type:	TypeError
Exception Value:	
process() takes 2 positional arguments but 3 were given
Exception Location:	/Users/iMac/Progs/test_pretix/pretix/env/lib/python3.4/site-packages/sqlparse/engine/__init__.py in run, line 44
Python Executable:	/Users/iMac/Progs/test_pretix/pretix/env/bin/python
Python Version:	3.4.4
Python Path:	
['/Users/iMac/Progs/test_pretix/pretix/src',
 '/Library/Frameworks/Python.framework/Versions/3.4/lib/python34.zip',
 '/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4',
 '/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/plat-darwin',
 '/Library/Frameworks/Python.framework/Versions/3.4/lib/python3.4/lib-dynload',
 '/Users/iMac/Progs/test_pretix/pretix/env/lib/python3.4/site-packages',
 '/Users/iMac/Progs/test_pretix/pretix/src']
Server time:	Tue, 4 Oct 2016 17:26:53 +0000
`

I can provide more info if required.

rixx:
I cannot reproduce this with python3.5 or python3.4 - after running `python manage.py migrate`, a normal login page shows. http://localhost:8000/control/ redirects to http://localhost:8000/control/login?next=/control/

FlaviaBastos:
Hi @rixx 👋 
Hmm... I did run migrate also... 😕 
Wait, did you try to login? Because I also get the login page fine. The issue is when it tries to authenticate user. 
If you can login fine, I guess I will try on python3.5

rixx:
Ohh, I missed that!

The problem exists on clean installs of both python3.4 and python3.5.

rixx:
[...]

PR #260, opened 2016-10-04
Fix incompatibility with sqlparse and debug-toolbar

rixx:
Thanks to @flaviabastos for reporting this in #259.
I think the issue is solved by higher versions of sqlparse in general,
but without further testing, version pinning is probably the safest
option.

Tested both with python 3.4 and 3.5, fixes the issue successfully.


raphaelm:
Thanks for debugging. I'm really curious why coveralls reports a decrease in coverage, floating point errors maybe? xD

Issue #155, opened 2016-07-28
CSP-compliant error pages


raphaelm:
Our error pages (404, 403, ...) contain links to e.g. ``javascript:history.back()`` that won't work due to our [Content Security Policy](https://de.wikipedia.org/wiki/Content_Security_Policy) headers that forbid javascript: URLs or ``<script>`` tags. The solution would be to move those JavaScript lines into a seperate ``.js`` file that is loaded by the error page and attaches those handlers to the links.

PR #261, opened 2016-10-04
remove javascript URIs for CSP

FlaviaBastos:
This refers to issue #155 
I removed the inline javascript on the error pages and replaced them with external files but I'm not sure I placed them in the right place.
Also, I know that there is some meta tag changes that need to be done ([according to this](https://developers.google.com/web/fundamentals/security/csp/)) but was not sure how you wanted and if I should do it.
So I'm submitting this to get some feedback. :wink:


raphaelm:
Hi! This looks good and is the correct way to do it. One minor thing: I think we can combine the JS files into a errors.js. There is no need.to have the browser load two files.

We don't need the metatags, they are just a different way to express what we already do with headers.

FlaviaBastos:
Argh... I think I messed up the commit... :unamused:

raphaelm:
Doesn't look that bad, it just has all three JS files in it :) Remove the two old ones and we're good to go :)

FlaviaBastos:
What a mess.... so sorry about that... :disappointed: 
It looks like the two files are removed but hey, who knows! :laughing: 
@raphaelm you are very patient! Thanks!

raphaelm:
Now all is perfect, thanks a lot! :)

PR #264, opened 2016-10-08
Add signal to display html in quota detail view

rixx:


raphaelm:
Could you also reference the signal in ``api/general.rst`` in the documentation? They are not automatically inserted there, sadly.

raphaelm:
Thanks! :)

Issue #148, opened 2016-06-27
Add more information to PDF tickets


raphaelm:
e.g. the order code

FlaviaBastos:
Hi, @raphaelm 
I added the order code to the PDF - please check file attached.
Is this ok? Is there anything else you think should be added also?
[1d73f4f9-8ae4-464d-becd-a0accfed05d0..pdf](https://github.com/pretix/pretix/files/514727/1d73f4f9-8ae4-464d-becd-a0accfed05d0.pdf)

raphaelm:
Yes, this is totally fine as a default :) Just like the other strings in the PDF, there should be settings for their position and size such that one day we can build a fancy drag-and-drop PDF editor :)
From the top of my head, I currently cannot think of other useful information to include there.

PR #265, opened 2016-10-08
Add order code to pdf and to ticket settings #148

FlaviaBastos:
Add order code to tickets generated by PDF
Add order code to ticket settings in control

#148

### DIFF
--- a/src/pretix/helpers/formats/en/formats.py
+++ b/src/pretix/helpers/formats/en/formats.py
@@ -1,2 +1,3 @@
+# Date according to https://docs.djangoproject.com/en/dev/ref/templates/builtins/#date
 SHORT_DATE_FORMAT = 'Y-m-d'
-SHORT_DATETIME_FORMAT = 'Y-d-m P'
+SHORT_DATETIME_FORMAT = 'Y-m-d H:i'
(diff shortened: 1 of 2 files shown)


raphaelm:
Hi! Thanks a lot! One direct comment, before I've actually tested it:
Could you replace ``"%s %s" % ('Order code:', str(order.code))`` with ``_('Order code: {code}').format(code=order.code)``? This way, the string "Order code" can later be translated to other languages.

FlaviaBastos:
Sure! Just to confirm: the underscore before the opening parenthesis on `_('Order code: {code}').format(code=order.code)` is expected to be there, right? I'm not familiar with that...( I think)

raphaelm:
Exactly! If you look at the imports at the top of the file, the function ``ugettext_lazy`` is imported with the name ``_``. ``ugettext_lazy`` is basically a magic function that takes a string and returns a translated version if one exists. Because one needs that function all the time and it is tedious to type it out, it is convention to rename it to a single underscore.

If you are interested, the matching docs are here: https://docs.djangoproject.com/en/1.10/topics/i18n/translation/
Be warned, this is a rather complicated matter as it follows the standard of GNU gettext, a nearly ancient translation tool ;)

FlaviaBastos:
Thank you so much for the explanation!

raphaelm:
Works like a charm, thanks a lot! :)

PR #270, opened 2016-10-11
Simple change to leak less information to make recon harder.

cptfoo:
Leaking less data is always a good idea


raphaelm:
Good idea, thanks! :)

PR #271, opened 2016-10-12
add basic instrumentation possibilities to pretix

cherti:
This PR adds the necessary datatypes to provide prometheus-like instrumentation natively to pretix.

Currently supported by this PR are Counters and Gauges.

### DIFF
--- a/src/pretix/base/metrics.py
+++ b/src/pretix/base/metrics.py
@@ -20,14 +20,11 @@ class Metric(object):
         self.help = helpstring
         self.labelnames = labelnames
 
-
     def __repr__(self):
         return self.name + "{" + ",".join(self.labelnames) + "}"
 
-
     def __str__(self):
         return self.__repr__()
-
 
     def _check_label_consistency(self, labelset):
         """
@@ -43,7 +40,6 @@ class Metric(object):
         if len(labelset) != len(self.labelnames):
             raise MetricsError("Unknown label used.")
 
-
     def _construct_metric_identifier(self, metricname, labelset=None):
         """
         Constructs the scrapable metricname usable in the output format.
@@ -57,14 +53,12 @@ class Metric(object):
 
             return metricname[:-1] + "}"
 
-
     def _inc_in_redis(self, key, amount):
         """
         Increments given key in Redis.
         """
         rkey = ns_prefix + key  # effective key for redis
         r.incrby(rkey, amount)
-
 
     def _set_in_redis(self, key, value):
         """
@@ -109,7 +103,6 @@ class Gauge(Metric):
         fullmetric = self._construct_metric_identifier(self.name, kwargs)
         self._set_in_redis(fullmetric, value)
 
-
     def inc(self, amount=1, **kwargs):
         """
         Increments Gauge by given amount for the labelset specified in kwargs.
@@ -122,7 +115,6 @@ class Gauge(Metric):
         fullmetric = self._construct_metric_identifier(self.name, kwargs)
         self._inc_in_redis(fullmetric, amount)
 
-
     def dec(self, amount=1, **kwargs):
         """
         Decrements Gauge by given amount for the labelset specified in kwargs.
@@ -132,9 +124,8 @@ class Gauge(Metric):
 
         self._check_label_consistency(self.labelnames, kwargs)
 
-        fullmetric = self._construct_metric_identifier(name, kwargs)
-        self._inc_in_redis(fullmetric, -1*amount)
-
+        fullmetric = self._construct_metric_identifier(self.name, kwargs)
+        self._inc_in_redis(fullmetric, amount * -1)
 
 
 def metrics_scrapable_textformat():
@@ -151,8 +142,6 @@ def metrics_scrapable_textformat():
         output.append(output_key + " " + str(value))
 
     return "\n".join(output)
-
-
 
 """
 Provided metrics
--- a/src/pretix/base/metrics.py
+++ b/src/pretix/base/metrics.py
@@ -124,23 +124,23 @@ class Gauge(Metric):
         self._inc_in_redis(fullmetric, amount * -1)
 
 
-def metrics_textformat():
+def metric_values():
     """
     Produces the scrapable textformat to be presented to the monitoring system
     """
     if not settings.HAS_REDIS:
         return ""
 
-    output = []
+    metrics = {}
 
     for key in redis.scan_iter(match=REDIS_KEY_PREFIX + "*"):
         dkey = key.decode("utf-8")
         _, _, output_key = dkey.split("_", 2)
         value = float(redis.get(dkey).decode("utf-8"))
 
-        output.append(output_key + " " + str(value))
+        metrics[output_key] = value
 
-    return "\n".join(output)
+    return metrics
 
 
 """
--- a/src/pretix/base/metrics.py
+++ b/src/pretix/base/metrics.py
@@ -6,151 +6,151 @@ r = redis.Redis(host=redishost, port=redisport)
 
 
 class MetricsError(Exception):
-	def __init__(self, msg):
-		self.msg = msg
+    def __init__(self, msg):
+        self.msg = msg
 
 
 class Metric(object):
-	"""
-	Base Metrics Object
-	"""
+    """
+    Base Metrics Object
+    """
 
-	def __init__(self, name, helpstring, labelnames=[]):
-		self.name = name
-		self.help = helpstring
-		self.labelnames = labelnames
-
-
-	def __repr__(self):
-		return self.name + "{" + ",".join(self.labelnames) + "}"
+    def __init__(self, name, helpstring, labelnames=[]):
+        self.name = name
+        self.help = helpstring
+        self.labelnames = labelnames
 
 
-	def __str__(self):
-		return self.__repr__()
+    def __repr__(self):
+        return self.name + "{" + ",".join(self.labelnames) + "}"
 
 
-	def _check_label_consistency(self, labelset):
-		"""
-		Checks if the given labelset provides exactly the labels that are required.
-		"""
-
-		# test if every required label is provided
-		for labelname in self.labelnames:
-			if labelname not in labelset:
-				raise MetricsError("Label {0} not specified.".format(labelname))
-
-		# now test if no further labels are required
-		if len(labelset) != len(self.labelnames):
-			raise MetricsError("Unknown label used.")
+    def __str__(self):
+        return self.__repr__()
 
 
-	def _construct_metric_identifier(self, metricname, labelset=None):
-		"""
-		Constructs the scrapable metricname usable in the output format.
-		"""
-		if not labelset:
-			return metricname
-		else:
-			metricname += "{"
-			for labelname in self.labelnames:
-				metricname += '{0}="{1}",'.format(labelname, labelset[labelname])
+    def _check_label_consistency(self, labelset):
+        """
+        Checks if the given labelset provides exactly the labels that are required.
+        """
 
-			return metricname[:-1] + "}"
+        # test if every required label is provided
+        for labelname in self.labelnames:
+            if labelname not in labelset:
+                raise MetricsError("Label {0} not specified.".format(labelname))
 
-
-	def _inc_in_redis(self, key, amount):
-		"""
-		Increments given key in Redis.
-		"""
-		rkey = ns_prefix + key  # effective key for redis
-		r.incrby(rkey, amount)
+        # now test if no further labels are required
+        if len(labelset) != len(self.labelnames):
+            raise MetricsError("Unknown label used.")
 
 
-	def _set_in_redis(self, key, value):
-		"""
-		Sets given key in Redis.
-		"""
-		rkey = ns_prefix + key  # effective key for redis
-		r.set(rkey, value)
+    def _construct_metric_identifier(self, metricname, labelset=None):
+        """
+        Constructs the scrapable metricname usable in the output format.
+        """
+        if not labelset:
+            return metricname
+        else:
+            metricname += "{"
+            for labelname in self.labelnames:
+                metricname += '{0}="{1}",'.format(labelname, labelset[labelname])
+
+            return metricname[:-1] + "}"
+
+
+    def _inc_in_redis(self, key, amount):
+        """
+        Increments given key in Redis.
+        """
+        rkey = ns_prefix + key  # effective key for redis
+        r.incrby(rkey, amount)
+
+
+    def _set_in_redis(self, key, value):
+        """
+        Sets given key in Redis.
+        """
+        rkey = ns_prefix + key  # effective key for redis
+        r.set(rkey, value)
 
 
 class Counter(Metric):
-	"""
-	Counter Metric Object
-	Counters can only be increased, they can neither be set to a specific value
-	nor decreased.
-	"""
+    """
+    Counter Metric Object
+    Counters can only be increased, they can neither be set to a specific value
+    nor decreased.
+    """
 
-	def inc(self, amount=1, **kwargs):
-		"""
-		Increments Counter by given amount for the labelset specified in kwargs.
-		"""
-		if amount < 0:
-			raise MetricsError("Counter cannot be decreased.")
+    def inc(self, amount=1, **kwargs):
+        """
+        Increments Counter by given amount for the labelset specified in kwargs.
+        """
+        if amount < 0:
+            raise MetricsError("Counter cannot be decreased.")
 
-		self._check_label_consistency(kwargs)
+        self._check_label_consistency(kwargs)
 
-		fullmetric = self._construct_metric_identifier(self.name, kwargs)
-		self._inc_in_redis(fullmetric, amount)
+        fullmetric = self._construct_metric_identifier(self.name, kwargs)
+        self._inc_in_redis(fullmetric, amount)
 
 
 class Gauge(Metric):
-	"""
-	Gauge Metric Object
-	Gauges can be set to a specific value, increased and decreased.
-	"""
+    """
+    Gauge Metric Object
+    Gauges can be set to a specific value, increased and decreased.
+    """
 
-	def set(self, value=1, **kwargs):
-		"""
-		Sets Gauge to a specific value for the labelset specified in kwargs.
-		"""
-		self._check_label_consistency(kwargs)
+    def set(self, value=1, **kwargs):
+        """
+        Sets Gauge to a specific value for the labelset specified in kwargs.
+        """
+        self._check_label_consistency(kwargs)
 
-		fullmetric = self._construct_metric_identifier(self.name, kwargs)
-		self._set_in_redis(fullmetric, value)
-
-
-	def inc(self, amount=1, **kwargs):
-		"""
-		Increments Gauge by given amount for the labelset specified in kwargs.
-		"""
-		if amount < 0:
-			raise MetricsError("Amount must be greater than zero. Otherwise use dec().")
-
-		self._check_label_consistency(kwargs)
-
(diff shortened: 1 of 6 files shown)


src/pretix/base/metrics.py
```diff
@@ -0,0 +1,161 @@
+import redis
+
+ns_prefix = "pretix_metrics_"
+
+r = redis.Redis(host=redishost, port=redisport)
```
raphaelm: We'll go for ``django_redis.get_redis_connection("redis")`` here. As far as I can see, django-redis does connection pooling, so we don't need it ourselves. Therefore, we can just use it here in global space.

src/pretix/base/metrics.py
```diff
@@ -0,0 +1,150 @@
...
+
+
+class MetricsError(Exception):
+    def __init__(self, msg):
+        self.msg = msg
```
raphaelm: That is the default constructor of ``Exception``, you can just omit it and do a ``class MetricsError(Exception):\n pass``

src/pretix/base/metrics.py
```diff
@@ -0,0 +1,150 @@
...
+    def __init__(self, name, helpstring, labelnames=[]):
+        self.name = name
+        self.help = helpstring
+        self.labelnames = labelnames
+
+    def __repr__(self):
+        return self.name + "{" + ",".join(self.labelnames) + "}"
+
+    def __str__(self):
+        return self.__repr__()
```
raphaelm: This is redundant, __str__ will call __repr__ if __str__ is not defined.

src/pretix/base/metrics.py
```diff
@@ -0,0 +1,150 @@
...
+        """
+
+        # test if every required label is provided
+        for labelname in self.labelnames:
+            if labelname not in labelset:
+                raise MetricsError("Label {0} not specified.".format(labelname))
+
+        # now test if no further labels are required
+        if len(labelset) != len(self.labelnames):
+            raise MetricsError("Unknown label used.")
```
raphaelm: The error message would be a lot more helpful if it said, which one is unknown. This should be easy if you just use actual ``set`` objects and use some primitive set operation.

src/pretix/base/metrics.py
```diff
@@ -0,0 +1,150 @@
...
+        Constructs the scrapable metricname usable in the output format.
+        """
+        if not labelset:
+            return metricname
+        else:
+            metricname += "{"
+            for labelname in self.labelnames:
+                metricname += '{0}="{1}",'.format(labelname, labelset[labelname])
+
+            return metricname[:-1] + "}"
```
raphaelm: Instead of slicing, you could use ``str.join`` to make the commas work:
```
nameparts = []
for labelname in self.labelnames:
    nameparts.append(...)
return "{" + ",".join(nameparts) + "}" 
```
but I believe that is a matter of taste

src/pretix/base/metrics.py
```diff
@@ -0,0 +1,150 @@
+import redis
+
+ns_prefix = "pretix_metrics_"
```
raphaelm: As this is a constant, I'd prefer an uppercase naming, ``NS_PREFIX`` or maybe to be more clear ``REDIS_KEY_PREFIX``.

cherti: how about REDIS_NAMESPACE? reflects more what the constant is used for.

src/pretix/base/metrics.py
cherti: this a little more difficult as labelset is a dict and labelnames is a list. I'd like to keep the list here to have guarantees of ordering, not sure if I have those in a set.
We could cast both to sets locally though, for this very purpose. How about that?

src/pretix/base/metrics.py
```diff
@@ -0,0 +1,150 @@
...
+    def __init__(self, msg):
+        self.msg = msg
+
+
+class Metric(object):
+    """
+    Base Metrics Object
+    """
+
+    def __init__(self, name, helpstring, labelnames=[]):
```
raphaelm: Something I missed: Never use a mutable object as a default argument! Better use ``labelnames=None`` and then ``self.labelnames = labelnames or []`` in the function body.

src/pretix/base/metrics.py
rixx: +1 for `','.join()`.

Also, using `'{0}, {1}'` in format strings is kind of redundant - consider either using unnamed placeholders `'{}, {}'` or named ones `'{name}, {value}'.format(name=labelname, value=labelset[labelname])`

cherti: I like that approach

src/pretix/base/metrics.py
```diff
@@ -0,0 +1,150 @@
...
+        self.name = name
+        self.help = helpstring
```
rixx: Is the helpstring used anywhere?

In any case, consider naming the object variable `helpstring`, too, since `help` is a builtin function in Python.

src/pretix/base/metrics.py
raphaelm: Yes, casting to set locally is totally fine for the sake of a good error message (you can even do it after the if statement so it is only done in the error case)

src/pretix/base/metrics.py
cherti: very good point, missed that one

src/pretix/base/metrics.py
cherti: No, the helpstring is not used (yet), but it is good practice to require one.

renamed.

src/pretix/base/metrics.py
```diff
@@ -0,0 +1,150 @@
...
+    Counters can only be increased, they can neither be set to a specific value
+    nor decreased.
+    """
+
+    def inc(self, amount=1, **kwargs):
+        """
+        Increments Counter by given amount for the labelset specified in kwargs.
+        """
+        if amount < 0:
+            raise MetricsError("Counter cannot be decreased.")
```
cherti: increase by negative value is a decrease, isn't it? no harm in reformulating, though.

cherti: changed

src/pretix/base/metrics.py
```diff
@@ -0,0 +1,150 @@
...
+            for labelname in self.labelnames:
+                metricname += '{0}="{1}",'.format(labelname, labelset[labelname])
+
+            return metricname[:-1] + "}"
+
+    def _inc_in_redis(self, key, amount):
+        """
+        Increments given key in Redis.
+        """
+        rkey = ns_prefix + key  # effective key for redis
```
cherti: with the new namespace-naming, this comment should indeed be obsolete.

src/pretix/base/metrics.py
cherti: done

src/pretix/base/metrics.py
```diff
@@ -0,0 +1,150 @@
...
+
+        fullmetric = self._construct_metric_identifier(self.name, kwargs)
+        self._set_in_redis(fullmetric, value)
+
+    def inc(self, amount=1, **kwargs):
+        """
+        Increments Gauge by given amount for the labelset specified in kwargs.
+        """
+        if amount < 0:
+            raise MetricsError("Amount must be greater than zero. Otherwise use dec().")
```
rixx: It might be worth considering to just call `dec()` here. cc/ @raphaelm

src/pretix/base/metrics.py
cherti: done

raphaelm:
One more thing: this all requires redis to be there but pretix can run without redis. You can find out if redis is present by using ``from django.conf import settings`` and then ``if settings.HAS_REDIS:``.

If redis is __not__ present, we need to find a nice way to:
* never call or import django_redis
* turn all methods of Metric objects to no-operations

src/pretix/base/metrics.py
raphaelm: I believe ``REDIS_KEY_PREFIX`` is clearer as a prefix is a well-defined concept while a namespaces does not exist in redis-world.

src/pretix/base/metrics.py
cherti: I disagree, increase and decrease are different operations. Besides that I don't think silently using one of them is a good idea, if I remember the spec correctly I don't think that this is allowed.

src/pretix/base/metrics.py
raphaelm: Yes it *is* a decrease, but if you just get an exception with "cannot be decreased" your first reaction will be "i did not call ``dec``!" and not "my amount was negative" ;)

src/pretix/base/metrics.py
rixx: Locally seems good enough for me:

```python
    raise MetricsError('Unknown labels used: {}'.format(set(labelset) - set(self.labelnames))
```

src/pretix/base/metrics.py
rixx: You are right, silently doing the other thing is not nice. I probably had the thought because the error message seemed a bit unclear to me.

src/pretix/base/metrics.py
cherti: done

src/pretix/base/metrics.py
```diff
@@ -0,0 +1,160 @@
+from django.conf import settings
+
+if settings.HAS_REDIS:
+    import django_redis
+
+REDIS_KEY_PREFIX = "pretix_metrics_"
+
+r = django_redis.get_redis_connection("redis")
```
rixx: I'd prefer a proper name, such as `REDIS` or `REDIS_CONNECTION`.

src/pretix/base/metrics.py
rixx: Maybe use similar messages to the one above? "Gauge cannot be increased by negative values. Use dec() to decrease."/"… decreased by negative values. Use inc() to increase."

src/pretix/base/metrics.py
rixx: `rc`, up until now, but it wasn't used as a global variable. Perfect time and location to start a good, new convention ;)

raphaelm: I normally use ``rc`` but I seldomly use it in a module-global namespace so far. Also, I don't feel mutable objects are constants, so I'd go for a simple ``redis`` here to be honest.

src/pretix/base/metrics.py
raphaelm: I belive if we say that this is an invalid thing to do, the error message is probably fine but gets even better with rixx' sugegstion.

raphaelm:
The ``MetricsError`` is currently only used if invalid data is passed to the methods, not for errors that happen during actual metrics operations. Shouldn't this be a plain ``ValueError``? @rixx what do you think?

rixx:
+1, `ValueError` is pythonic on unexpected input values.

raphaelm:
The ``HAS_REDIS`` part needs to be different, I'm sorry. The way you implemented it makes it to easy to get it wrong. For example, your stub misses ``Gauge.set`` and all the methods will throw errors when called with arguments. Also, you made the import conditional but not the connection creation.

I fail to come up with a really nice design but as you only call redis in _inc_in_redis and  _set_in_redis (and the exporter) it is maybe best to do it there directly. This has another benefit: The validation logic is also executed without redis (and thus during development and testing).

cherti:
@raphaelm 
now there is also a view for the metrics, from my point, if upstream has no further complaints, this can be merged. ;)

rixx:
Open points:
 - naming of `labelset`
 - `redis` as a global variable should be upper case, imo

Additional thoughts: 
 - It would be great to have basic tests, but we can add those later on, too
 - Personal preference: I'd love for a notification (INFO or WARNING) to be raised/logged when `_inc_in_redis` (and the others) just drop a command due to missing redis. @raphaelm?

raphaelm:
>  - `redis` as a global variable should be upper case, imo

I disagree here, because it is a mutable object and not a constant.

Another open point:
- We need authentication for the metrics endpoint before this gets deployed anywhere. In prometheus-world, HTTP Basic Auth is the thing to use which I believe is totally fine here. I'm not yet sure about how to share the password, I tend towards a ``settings.py``/``pretix.cfg`` approach in a similar way like we do with ``SECRET_KEY``.

> - It would be great to have basic tests, but we can add those later on, too

Yes, it would be nice to at least have the relevant codepaths exectued. As in our current setup the tests never run in presence of redis, I'd recommend to mock/monkeypatch the redis connection.

> - Personal preference: I'd love for a notification (INFO or WARNING) to be raised/logged when `_inc_in_redis` (and the others) just drop a command due to missing redis. @raphaelm?

I tend to disagree. If we want to use this to measure lots of cool things and have lots of statistics, we will call those methods *a lot*. Logging this when redis is unavailable spams the logfiles, degrades performance and most importantly irritates people running a development setup. I'd accept a warning that is raised *once* per thread (either by using python-level ``warning`` or by remembering it ourselves) if redis is not configured and ``DEBUG`` is off.

cherti:
@raphaelm 
Basic auth implemented in the view, ready for review.

src/pretix/base/views/metrics.py
```diff
@@ -0,0 +1,35 @@
+from django.conf import settings
+from http import HttpResponse
+from .. import metrics
+
+unauthedResponse = HttpResponse("<html><title>Forbidden</title><body>You are not authorized to view this page.</body></html>", mimetype="text/html")
+unauthedResponse["WWW-Authenticate"] = 'Basic realm="metrics"'
+unauthedResponse.status_code = 401
```
raphaelm: I believe it is a bad Idea to have this ``HttpResponse`` as a long-living object as it is mutable and might be mutated e.g. by middlewars. I suggest to replace it with a ``reponse_unauthenticated()`` method that returns this object.

src/pretix/base/views/metrics.py
```diff
@@ -0,0 +1,35 @@
...
+    if "HTTP_AUTHORIZATION" not in request.META:
+        return unauthedResponse
+
+    method, credentials = request.META["HTTP_AUTHORIZATION"].split(" ", 1)
+    if method.lower() != "basic":
+        return unauthedResponse
+
+    user, passphrase = credentials.strip().decode("base64").split(":", 1)
+
+    if user != settings.METRICS_USER:
```
raphaelm: There is no way to set this yet, right? We'd need to somehow read this from the config file in ``pretix/settings.py``. Also, we'll need documentation on this in ``doc/admin/config.rst``.

src/pretix/urls.py
```diff
@@ -4,7 +4,7 @@
 import pretix.control.urls
 import pretix.presale.urls
 
-from .base.views import cachedfiles, health, js_catalog, redirect
+from .base.views import cachedfiles, health, js_catalog, redirect, metrics
```
raphaelm: This causes an ``isort`` issue ;)

src/pretix/base/views/metrics.py
```diff
@@ -0,0 +1,35 @@
+from django.conf import settings
+from http import HttpResponse
+from .. import metrics
```
raphaelm: This import is neither correct (``http`` is inside django, not a top-level module) nor sorted the right way.

src/pretix/base/views/metrics.py
```diff
@@ -1,27 +1,33 @@
...
 
-unauthedResponse = HttpResponse("<html><title>Forbidden</title><body>You are not authorized to view this page.</body></html>", mimetype="text/html")
-unauthedResponse["WWW-Authenticate"] = 'Basic realm="metrics"'
-unauthedResponse.status_code = 401
+
+def unauthedResponse():
+    content = "<html><title>Forbidden</title><body>You are not authorized to view this page.</body></html>"
+    unauthedResponse = HttpResponse(content, mimetype="text/html")
+    unauthedResponse["WWW-Authenticate"] = 'Basic realm="metrics"'
+    unauthedResponse.status_code = 401
```
rixx:
```python
    return unauthedResponse
```
?

rixx: Also, am I missing something regarding the camelCasing here? `get_unauthed_response` and `unauthed_response` would be more in line with the code base.

src/pretix/base/views/metrics.py
```diff
@@ -3,31 +3,31 @@
...
 
-def unauthedResponse():
+def unauthed_response():
     content = "<html><title>Forbidden</title><body>You are not authorized to view this page.</body></html>"
-    unauthedResponse = HttpResponse(content, mimetype="text/html")
-    unauthedResponse["WWW-Authenticate"] = 'Basic realm="metrics"'
-    unauthedResponse.status_code = 401
+    response = HttpResponse(content, mimetype="text/html")
+    response["WWW-Authenticate"] = 'Basic realm="metrics"'
+    response.status_code = 401
```
cherti: indeed, thanks for the hint. after fixing that I should probably stop coding for today...

cherti:
tests for Counters are added, please comment if they are fine that way or should be changed somehow. Tests for Gauges will follow once the test for counters is accepted.

src/tests/base/test_metrics.py
```diff
@@ -0,0 +1,41 @@
+# pytest
+
+from django.test import override_settings
+
+from pretix.base import metrics
+
+
+class fake_redis(object):
```
raphaelm: Please use camel case (``FakeRedis``) for all class names :)

src/tests/base/test_metrics.py
```diff
@@ -0,0 +1,41 @@
...
+    monkeypatch.setattr(metrics, "redis", local_fake_redis, raising=False)
+
+    # now test
+    fullname = metrics.http_requests_total._construct_metric_identifier('http_requests_total', {"code": "200", "handler": "/foo", "method": "GET"})
+    metrics.http_requests_total.inc(code="200", handler="/foo", method="GET")
+    assert local_fake_redis.storage[metrics.REDIS_KEY_PREFIX + fullname] == 1
+    metrics.http_requests_total.inc(code="200", handler="/foo", method="GET")
+    assert local_fake_redis.storage[metrics.REDIS_KEY_PREFIX + fullname] == 2
+    metrics.http_requests_total.inc(7, code="200", handler="/foo", method="GET")
+    assert local_fake_redis.storage[metrics.REDIS_KEY_PREFIX + fullname] == 9
```
raphaelm: Could you add one more asserts that increments with e.g. a different ``method`` than ``GET`` and then checks that the number is **not** incremented?

cherti:
tests are finished so far. What is still missing is the test for Gauge, I'd add that once we actually use a gauge to not pollute the production-code with test-metrics.

If everything is fine so far, we could start distributing some of those metrics into pretix, right?

src/tests/base/test_metrics.py
```diff
@@ -0,0 +1,68 @@
...
+    metrics.http_requests_total.inc(code="200", handler="/foo", method="GET")
+    assert fake_redis.storage[metrics.REDIS_KEY_PREFIX + fullname_GET] == 2
+    metrics.http_requests_total.inc(7, code="200", handler="/foo", method="GET")
+    assert fake_redis.storage[metrics.REDIS_KEY_PREFIX + fullname_GET] == 9
+    metrics.http_requests_total.inc(7, code="200", handler="/foo", method="POST")
+    assert fake_redis.storage[metrics.REDIS_KEY_PREFIX + fullname_GET] == 9
+    assert fake_redis.storage[metrics.REDIS_KEY_PREFIX + fullname_POST] == 7
+
+
+@override_settings(HAS_REDIS=True,METRICS_USER="foo",METRICS_PASSPHRASE="bar")
```
raphaelm: Please add spaces after the commas :)

cherti: done

raphaelm:
Yay! Thanks! :)

PR #272, opened 2016-10-12
Remove AttendeeList view

rixx:
It could not even filter for attendee and you can do everything (and
better and more) in the OrdersList. And it's one menu entry less to
clutter the interface.


raphaelm:
Thanks! :)

Issue #161, opened 2016-08-06
Add "Empty Cart" button


rixx:
Just a feature proposal, feel free to disregard:

Maybe that's only relevant for testing, and not for actual users, but an "Empty Cart" button would have made my life a bit simpler today. Maybe just a small 'X' in a corner somewhere?

raphaelm:
Good idea! :) Should be easy to implement.

If someone wants to tackle this as his or her first issue in the project, do not hesitate to ask me questions if you get stuck!

sathe-aditya:
Hello, I'm a beginner to open-source, and would like to address this issue! Could anyone please guide me?

rixx:
Hi! There are several starting points you could choose. If you are not familiar with Django (the Python web framework used in pretix), you could follow the [official Django tutorial](https://docs.djangoproject.com/en/1.9/intro/tutorial01/) or the [Django Girls tutorial](http://tutorial.djangogirls.org/) to familiarize yourself with Django's overall structure.

If (or: once) you are familiar with Django, you might want to follow the [development setup guide](https://docs.pretix.eu/en/latest/development/setup.html), and play around a bit.

The template `src/pretix/presale/templates/pretixpresale/event/index.html` and the view `src/pretix/presale/views/cart.py` are probably the most relevant to this ticket. You could look at the existing `CartRemove` view, and maybe add a similar view named `CartRemoveAll` ...

Feel free to ask questions if you run into any problems, or if something is unclear to you.

sathe-aditya:
Sure! I am not familiar with Django; I'll start getting familiarized with it right away!
Thanks a lot for your help, I'll definitely ask you if I run into some problem!

FlaviaBastos:
@sathe-aditya hey! Are you planning to work on this or could I take a look?

sathe-aditya:
Hey there! Could you please take a look? Thanks.



On Fri, Oct 28, 2016 at 4:24 AM +0530, "Flavia Bastos" <notifications@github.com<mailto:notifications@github.com>> wrote:


@sathe-aditya<https://github.com/sathe-aditya> hey! Are you planning to work on this or could I take a look?

-

raphaelm:
THis has been fixed in the meantime :)

PR #279, opened 2016-10-20
Issue 173

FlaviaBastos:
First part of issue for feedback - add context variables to sendmail plugin:
- add placeholders {due_date}, {event}, {order}, {order_date}, {order_url} to message field

Comments/questions:
1. You mentioned 
   
   > I'd prefer to use the same syntax there, e.g. {order_url}
   > but the Email Settings has {url} for this specific example so I went with your suggestion, but please let me know and I can change any or all of them. 
2. I tried to squash my commits following [this](https://davidwalsh.name/squash-commits-git) but instead of a list of commits I got a `noop` (or something like that), so I probably did something wrong. :flushed: sorry.... If you want me to try again I will need to clarify the steps first. 
3. I'm stuck on the preview part: I'm not sure how to handle the new button action. I was looking into [this file](https://github.com/pretix/pretix/blob/master/src/pretix/control/views/event.py), starting on line 342 but I'm not sure how to translate that into this issue. Am I completely off?

Well, I think this is it for now. Thanks!


src/pretix/plugins/sendmail/views.py
```diff
@@ -40,7 +42,16 @@ def form_valid(self, form):
...
+                     {
+                         'event': o.event,
+                         'order': o.code,
+                         'order_date': date_format(o.datetime, 'SHORT_DATETIME_FORMAT'),
+                         'due_date': date_format(o.expires, 'SHORT_DATE_FORMAT'),
+                         'order_url': build_absolute_uri(o.event, 'presale:event.order', kwargs={
+                             'order': o.code,
+                             'secret': o.secret
+                         })
+                }, self.request.event, locale=o.locale, order=o)
```
raphaelm: The indentation of this and the above } should match the line of the opening one :)

src/pretix/plugins/sendmail/views.py
```diff
@@ -40,7 +42,16 @@ def form_valid(self, form):
         for o in orders:
             try:
                 mail(o.email, form.cleaned_data['subject'], form.cleaned_data['message'],
-                     None, self.request.event, locale=o.locale, order=o)
+                     {
+                         'event': o.event,
+                         'order': o.code,
+                         'order_date': date_format(o.datetime, 'SHORT_DATETIME_FORMAT'),
```
raphaelm: We *might* need to care about timezones here. Could you test that by setting the timezone in the event settings to something non-UTC and then test sending an email? If it is a problem, the solution should be along those lines:

```
tz = pytz.timezone(request.event.settings.timezone)
...
'order_date': date_format(o.datetime.astimezone(tz), 'SHORT_DATETIME_FORMAT')
```

FlaviaBastos: You were right! The order time needs the timezone adjustment. Already added it.

FlaviaBastos:
Hey, @raphaelm 
When you have some time would you please take a look at my latest  `src/pretix/plugins/sendmail/views.py`? I cannot find a way to return the request and the formatted string to the template. I tried to use render but that breaks bootstrap forms. HttpResponse gave me 'BadHeaderError' and so on. The only way I can keep the form populated when pushing "preview email" is with what I have now on line 50, but that doesn't allow my formatted text to go through. 
You will probably come back with a super simple answer but right now I cannot find it myself! :grin: 
Thank you!

src/pretix/plugins/sendmail/views.py
```diff
@@ -36,18 +39,35 @@ def form_valid(self, form):
...
-                     None, self.request.event, locale=o.locale, order=o)
-            except SendMailException:
-                failures.append(o.email)
-
-        if failures:
-            messages.error(self.request, _('Failed to send mails to the following users: {}'.format(' '.join(failures))))
-        else:
-            messages.success(self.request, _('Your message has been queued to be sent to the selected users.'))
+            if self.request.POST.get("action") == "preview":
+                message = self.request.POST.get('message_0')
```
raphaelm: You can use ``form.cleaned_data['message']`` here, so you don't need to care about the ``_0`` stuff, wherever that is coming from

raphaelm:
Okay, I see your problem. Unlike you expected, I do not immediately come up with an obvious answer. I'd go for ``return self.get(self.request)`` just like you, that is a good way of dealing with it (although ``return self.get(self.request, *self.args, **self.kwargs)`` is probably more correct).

As to how we get the formatted text to the view: We could save it to a property ``self.preview_text`` of our view class and then override ``get_context_data`` with something like:

```
def get_context_data(self, *args, **kwargs):
    ctx = super().get_context_data(self, *args, **kwargs)
    ctx['preview_text'] = getattr(self, 'preview_text, 'None')
    return ctx
```

src/pretix/plugins/sendmail/views.py
raphaelm: Actually, the ``_0`` is appended because this is the field of the first language (I didn't realize this when I commented here yesterday). If you change your event settings and enable German as a second language, you'll see that you get two textareas and a ``_1`` message. Obviously, it would be perfect to show the preview in all languages (you can get the list at ``request.event.settings.locales``).

As you noticed, you can't call ``format`` on a ``LazyI18nString``, which is a pretix-specific representation of a string that exists in multiple languages (there is documentation on it [here](https://docs.pretix.eu/en/latest/development/implementation/i18n.html#usage)). You can easily convert it into a string for example by ``form.cleaned_data['message'].localize('en')`` (or ``form.cleaned_data['message'].localize(locale)`` in the general case).

Does that help? :)

FlaviaBastos: It does help a lot! Thanks! :)
I had to find a way to deal with these new variations on the context dictionary now and I'm not sure I took the best approach, so please let me know if you have suggestions.

src/pretix/plugins/sendmail/views.py
```diff
@@ -36,13 +39,35 @@ def form_valid(self, form):
...
+        self.output = {}
         for o in orders:
-            try:
-                mail(o.email, form.cleaned_data['subject'], form.cleaned_data['message'],
-                     None, self.request.event, locale=o.locale, order=o)
-            except SendMailException:
-                failures.append(o.email)
+            if self.request.POST.get("action") == "preview":
+                self.locales_list = self.request.event.settings.locales
+                for l in self.locales_list:
```
raphaelm: There's no need for the class variable, ``for l in self.request.event.settings.locales`` is totally fine.

src/pretix/plugins/sendmail/views.py
```diff
@@ -36,13 +39,35 @@ def form_valid(self, form):
...
+                    self.output[l] = []
+                    self.output[l].append(form.cleaned_data['subject'].localize(l))
+                    message = form.cleaned_data['message'].localize(l)
+                    preview_text = (message.format(order='ORDER#', event='Your event name', order_date='2016-10-15', due_date='2016-10-31', order_url='link_to_order_url'))
+                    self.output[l].append(preview_text)
+                return self.get(self.request, *self.args, **self.kwargs)
```
raphaelm: You could improve here even further by acually switching to that locale before you generate the text:

```
from pretix.base.i18n import language
...

for l in self.request.event.settings.locales:
    with language(l):
        # Generate preview
```

This has the advantage that you can use actual data for some placeholders and it will be displayed in the correct language, eg.:

```
from datetime import timedelta
..

preview_text = message.format(
    order='ORDER1234', 
    event=self.request.event.name,
    due_date=date_format(now() + timedelta(days=7), 'SHORT_DATE_FORMAT'),
    ...
)
```

src/pretix/plugins/sendmail/views.py
```diff
@@ -54,3 +79,8 @@ def form_valid(self, form):
             event=self.request.event.slug,
             organizer=self.request.event.organizer.slug
         )
+
+    def get_context_data(self, *args, **kwargs):
+        ctx = super().get_context_data(*args, **kwargs)
+        ctx['output'] = getattr(self, 'output', 'None')
```
raphaelm: Use ``None`` here, not the string ``'None'`` :)

src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/send_form.html
```diff
@@ -10,12 +10,32 @@
...
             {% bootstrap_field form.subject layout='horizontal' %}
             {% bootstrap_field form.message layout='horizontal' %}
+            {% if request.method == "POST" %}
+            <fieldset>
+            <legend>{% trans "E-mail preview" %}</legend>
+
+                <div class="panel-body">
+                    {% for k,v in output.items %}
+                    <p>
+                        <strong>For locale: </strong>{{ k }}
```
raphaelm: We need to have this string translatable as well. This could be achieved like this:
```
<strong>{% blocktrans %}Preview for locale: {{ k }}{% endblocktrans %}</strong>
```
This will show something like ``Preview for locale: de``. It would obviously be even better to show ``Preview for locale: German``, but I don't see an easy way to achieve this (we'd need to somehow read this data from ``settings.py``), I suggest we keep it with ``de`` for now.

src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/send_form.html
```diff
@@ -10,12 +10,32 @@
             {% bootstrap_field form.sendto layout='horizontal' %}
             {% bootstrap_field form.subject layout='horizontal' %}
             {% bootstrap_field form.message layout='horizontal' %}
+            {% if request.method == "POST" %}
+            <fieldset>
+            <legend>{% trans "E-mail preview" %}</legend>
+
+                <div class="panel-body">
+                    {% for k,v in output.items %}
```
raphaelm: Prefer to use self-explanatory names, e.g. ``{% for locale, segments in output.items %}``

FlaviaBastos:
Hey, Raphael!

> I have to say that I didn't expect this to grow this complex when I assigned this ticket an "easy" flag. 


I know, right? :stuck_out_tongue: Me either!

> However, you're doing an amazing job so far :) 

Thanks! But let's give credit where credit is due: if it weren't for you guidance and help with chunks of code I would still be stuck waaay back. So, thanks! It's a combined effort. 

> flake8 is currently complaining about two code style issues ... but we can just defer this to the end and clean it up when this is ready for merge

Yes! I saw that too but I think I will fight that battle later on! :wink:

Now, let me say that although this has [grown more complex than we both expected](https://twitter.com/aras_p/status/791989711527182336), I am enjoying working on it a lot. 

I will take a look at your suggestions.
Thanks again!

FlaviaBastos:
flake8 is still complaining but I cannot see how to fix it: `./pretix/plugins/sendmail/views.py:76:26: E126 continuation line over-indented for hanging indent` :confused:

One thing I don't like but don't know how to fix is the following: I have the three locales enabled (en, de, de-informal) but if I leave one of the text message box blank, it will still "preview" the 3 locales, although one will be duplicated. :unamused: maybe we would have to check if the message is blank (or none) and not append that one... but then I wonder if this is really an issue, meaning, if you have more than one locale enabled, wouldn't you want to test all of them anyway? Why leaving one blank? #questions

raphaelm:
> flake8 is still complaining but I cannot see how to fix it: ./pretix/plugins/sendmail/views.py:76:26: E126 continuation line over-indented for hanging indent :confused:

That's a tough one. From the top of my head, try:

```
                         }, 
                         self.request.event, locale=o.locale, order=o
                     )
```

(Such that the closing ``)`` matches the indentation of the opening ``(``).

> One thing I don't like but don't know how to fix is the following: I have the three locales enabled (en, de, de-informal) but if I leave one of the text message box blank, it will still "preview" the 3 locales, although one will be duplicated. :unamused: maybe we would have to check if the message is blank (or none) and not append that one... but then I wonder if this is really an issue, meaning, if you have more than one locale enabled, wouldn't you want to test all of them anyway? Why leaving one blank? #questions

Good question. I think it is correct to show them anyway, even if they look duplicated to show what would actually happen if you sent out the mail. For example, if you only fill in ``en``, German users will receive the English text as well, but with German date formats. That's a good thing to show in a preview :)

FlaviaBastos:
flake8 put up a fight! Your suggestion raised yet another warning 
`./pretix/plugins/sendmail/views.py:76:26: E126 continuation line over-indented for hanging indent
./pretix/plugins/sendmail/views.py:78:25: E124 closing bracket does not match visual indentation`
Eventually I put the closing bracket where it is now and ran flake8 in a "this obviously won't work!" but it did! Don't ask me! :sweat: 

> Good question. I think it is correct to show them anyway, even if they look duplicated to show what would actually happen if you sent out the mail. For example, if you only fill in en, German users will receive the English text as well, but with German date formats. That's a good thing to show in a preview :)

Huh! That's one way to see it...

src/pretix/plugins/sendmail/views.py
```diff
@@ -36,13 +41,41 @@ def form_valid(self, form):
...
-                mail(o.email, form.cleaned_data['subject'], form.cleaned_data['message'],
-                     None, self.request.event, locale=o.locale, order=o)
-            except SendMailException:
-                failures.append(o.email)
+            if self.request.POST.get("action") == "preview":
+                for l in self.request.event.settings.locales:
+                    with language(l):
+                        self.output[l] = []
+                        self.output[l].append(
+                            form.cleaned_data['subject'].localize(l))
```
raphaelm: What would be your opinion on switching this to the following?
```
self.output[l].append(_('Subject: {subject}').format(subject=form.cleaned_data['subject'].localize(l)))
```

src/pretix/plugins/sendmail/views.py
```diff
@@ -36,13 +41,41 @@ def form_valid(self, form):
...
+                        message = form.cleaned_data['message'].localize(l)
+                        preview_text = message.format(
+                            order='ORDER1234',
+                            event=self.request.event.name,
+                            order_date=date_format(now(), 'SHORT_DATE_FORMAT'),
+                            due_date=date_format(now() + timedelta(days=7), 'SHORT_DATE_FORMAT'),
+                            order_url='link_to_order_url')
```
raphaelm: Let's use a more realistic example here:
```
order_url=build_absolute_uri(o.event, 'presale:event.order', kwargs={
    'order': 'ORDER1234',
    'secret': 'longrandomsecretabcdef123456'
}))
```

src/pretix/plugins/sendmail/views.py
FlaviaBastos: uh, I see. You are using the `ugettext_lazy` that you showed me the other day... Well, I wouldn't have thought of that! Looks much better :)

src/pretix/plugins/sendmail/views.py
raphaelm: Hey Flavia, did you see this? :) I think this is the only thing left here before I merge this.

FlaviaBastos: I did and I thought I changed it...  Do you see [the new code in here](https://github.com/pretix/pretix/pull/279/commits/1261f00eb60ce5a4bc5543d3d72be0c2b5283c6b) or is it gone?

FlaviaBastos:
Sorry about that again... :flushed: 
I just pushed new code. :)

raphaelm:
Thanks a lot! :)

FlaviaBastos:
It's merged! wow! :dancer:
This was a very interesting issue to work on. Thank you very much for all the guidance and for letting me work on it.

raphaelm:
You did a great job! I talked to a user of pretix today who had lots of ideas which lead to a bunch of new issues, two of them related to this mail-thingy, if you feel like digging here deeper. But feel free to do something completely different, if you don't. :)

FlaviaBastos:
I offered to work on #161 but yes, I would love to keep working on this plugin. I will look at these soon. Thanks! :)

PR #281, opened 2016-10-20
Corrected minor splelling mistakes

nelkenwelk:


raphaelm:
Good catch, thanks! :)

PR #283, opened 2016-10-23
Return boolean as method name suggests

rixx:


raphaelm:
Good catch. Are you using that method anywhere? Because I apparently don't :D

rixx:
Nope, only in interactive shells, but I'm not used to accessing anything called `*_set` anymore, so I wrapped them up ;)

PR #286, opened 2016-10-25
Admin settings

rixx:
This PR might still be somewhat WIP (please review!), but I think the most important things should be done: model, signal, form, view, template, access control for admin settings. Documentation via the signal docstring.

I'd like to use the admin settings as lowest-order fallback for settings objects (i.e. like organizer settings), but I'd understand if you don't want this. If you decide against it, I'd move and rename the `AdminDummyObject` to a more accessible place and it would basically serve as a singleton-ish accessor to admin settings in the database. (But I think providing those settings via eventsettings etc would be easier.)


src/pretix/base/models/settings.py
```diff
@@ -0,0 +1,33 @@
+from django.db import models
+
+
+class AdminSetting(models.Model):
+    """
+    An event option is a key-value setting which can be set for an
+    organizer. It will be inherited by the events of this organizer
+    """
+    key = models.CharField(max_length=255)
```
raphaelm: How do you feel about using this as the primary key of this model? There is no real reason for an ID.

src/pretix/base/signals.py
```diff
@@ -122,3 +122,8 @@ def send(self, sender: Event, **named) -> List[Tuple[Callable, Any]]:
 idempotent, i.e. it should not make a difference if this is sent out more often
 than expected.
 """
+register_admin_settings = django.dispatch.Signal()
```
raphaelm: Please put an empty line before to make clear what which docstring refers to.

src/pretix/base/signals.py
```diff
@@ -122,3 +122,8 @@ def send(self, sender: Event, **named) -> List[Tuple[Callable, Any]]:
...
+"""
+All plugins that are installed may send fields for the admin settings form, as
+a list of tuples (setting name, form field).
```
raphaelm: In other places where we have APIs like this (payment providers and ticket output providers), we expects dict-like collections of form fields (and all sample implementations use ``OrderedDict``). Do you think we should go for consistency here?

src/pretix/control/forms/admin.py
```diff
@@ -0,0 +1,45 @@
...
+            ('footer_link', I18nFormField(
+                widget=I18nTextInput,
+                required=False,
+                label=_("Additional footer link"),
+                help_text=_("Will be included as the link in the additional footer text.")
+            ))
+        ])
+        responses = register_admin_settings.send(self)
+        for r, response in responses:
+            for key in response:
```
raphaelm: Uhm, this looks to be as if you'd expect ``responses`` to be a dict-like object, like I suggested above but not like you documented it. Can you clarify this? If it is a dict-like object, I feel ``for key, value in response.items()`` is more idiomatic than iterating over keys and fetching values manually.

src/pretix/control/templates/pretixcontrol/admin.html
```diff
@@ -0,0 +1,18 @@
+{% extends "pretixcontrol/base.html" %}
+{% load i18n %}
+{% load bootstrap3 %}
+
+{% block title %}{% trans "Admin settings" %}{% endblock %}
+{% block content %}
+	<h1>{% trans "Admin settings" %}</h1>
```
raphaelm: Indentation got somehow mixed up here.

src/pretix/control/templates/pretixcontrol/admin.html
```diff
@@ -0,0 +1,18 @@
+{% extends "pretixcontrol/base.html" %}
+{% load i18n %}
+{% load bootstrap3 %}
+
+{% block title %}{% trans "Admin settings" %}{% endblock %}
```
raphaelm: Should we maybe call it "Global settings", as we don't really have a clear usage of the term "admin" yet (might be used for an *event admin* after the permissions refactor, I don't know yet)? I'm not sure about this.

src/pretix/base/signals.py
rixx: I forgot to adjust the docstring after noticing exactly this - yes, the settings form expects an OrderedDict, as it should.

src/pretix/control/forms/admin.py
rixx: I *think* it is this verbose because I tried using `.update()` only to discover that this will not preserve ordering, and then I jumped to the most verbose solution. Fixed.

src/pretix/control/templates/pretixcontrol/admin.html
rixx: Well, that was a fun rename … :ballot_box_with_check:

rixx:
I'm not quite happy with the `GlobalSettingsObject` class - I'd like to rename it to `GlobalSettings` and put it somewhere not in `control`, because you probably want to use its `settings` attribute to access global settings.

rixx:
Hm, I think I'll implement site settings as fallback settings between organizer settings and default settings - I'm fairly sure you'll want to access them via `event.settings` instead of getting the global settings object every time.

src/pretix/base/signals.py
```diff
@@ -122,3 +122,9 @@ def send(self, sender: Event, **named) -> List[Tuple[Callable, Any]]:
 idempotent, i.e. it should not make a difference if this is sent out more often
 than expected.
 """
+
+register_global_settings = django.dispatch.Signal()
+"""
+All plugins that are installed may send fields for the global settings form, as
+an OrderedDict of (setting name, form field).
```
raphaelm: I believe the correct API would be "as an object behaving like a dict", but since an OrderedDict is the most sensible thing to do, we can leave this like this.

src/pretix/base/models/organizer.py
```diff
@@ -57,7 +58,8 @@ def settings(self) -> SettingsProxy:
         """
         Returns an object representing this organizer's settings
         """
-        return SettingsProxy(self, type=OrganizerSetting)
+        from pretix.control.forms.global_settings import GlobalSettingsObject
```
raphaelm: Yeah, I believe we need to move that to ``pretix.base.settings``. Importing ``control`` from ``base`` should be avoided, just as you suggested.

src/pretix/control/views/global_settings.py
```diff
@@ -0,0 +1,17 @@
...
+class GlobalSettingsView(AdministratorPermissionRequiredMixin, FormView):
+    template_name = 'pretixcontrol/global_settings.html'
+    form_class = GlobalSettingsForm
+
+    def form_valid(self, form):
+        form.save()
+        return super().form_valid(form)
+
+    def get_success_url(self):
+        from django.shortcuts import reverse
```
raphaelm: I think this could be a global import

src/pretix/control/templates/pretixcontrol/global_settings.html
```diff
@@ -0,0 +1,18 @@
...
+{% load i18n %}
+{% load bootstrap3 %}
+
+{% block title %}{% trans "Global settings" %}{% endblock %}
+{% block content %}
+    <h1>{% trans "Global settings" %}</h1>
+    <form action="" method="post" class="form-horizontal">
+        {% csrf_token %}
+        {% bootstrap_form_errors form %}
+        {% bootstrap_form form %}
```
raphaelm: ``{% bootstrap_form form layout='horizontal' %}`` will make it look better :)

src/pretix/control/permissions.py
```diff
@@ -90,3 +90,33 @@ class OrganizerPermissionRequiredMixin:
     def as_view(cls, **initkwargs):
         view = super(OrganizerPermissionRequiredMixin, cls).as_view(**initkwargs)
         return organizer_permission_required(cls.permission)(view)
+
+
+def administrator_permission_required(permission):
```
raphaelm: If we don't use it, we can drop the permission parameter ;) In that case, we'll need to think about ``@administrator_permission_required`` vs ``@administrator_permission_required()``. Ideally, both should work, which is not easy to achieve, but IIRC ``functools`` provides a utility to support this.

src/pretix/control/permissions.py
```diff
@@ -90,3 +90,33 @@ class OrganizerPermissionRequiredMixin:
...
+        return wrapper
+    return decorator
+
+
+class AdministratorPermissionRequiredMixin:
+    """
+    This mixin is equivalent to the administrator_permission_required view decorator but
+    is in a form suitable for class-based views.
+    """
+    permission = ''
```
raphaelm: Same here :)

raphaelm:
This seems pretty solid to me now. Of course we still need to actually use the footer link setting, but this could also be done after the merge. If you do not see any more TODOs here, this is nearly complete. As ``pretix.base.settings`` is a critical part of pretix' ecosystem, I'd really want to have at least two basic tests in here:

* One test that calls the GlobalSettingsObject and saves and loads one or two settings
* One test that verifies that global setting are used as defaults of all organizer or event-level settings

Optionally, we could test the form itself, but this can come later, the settings object is more important.

rixx:
I'm not really happy about the marginally decreased coverage. And, yeah, footer links might be nice.

raphaelm:
> I'm not really happy about the marginally decreased coverage. 

This is due to the permission decorator. Calling the form once with a superuser and once with a non-superuser should suffice.

rixx:
I'll do that in my testing PR, I suppose.

raphaelm:
So shall I merge this and add the footer links myself or will you do so? Both is totally fine with me.

rixx:
@raphaelm, pushed footer links. The footer text and footer URL are both I18nStrings because you might want different links for each language (or maybe you only need to display some disclaimer link in one locale) - I hope that's fine with you?

I think I'm done here, otherwise ;)

src/pretix/presale/context.py
```diff
@@ -16,6 +17,17 @@ def contextprocessor(request):
     }
     _html_head = []
     _footer = []
+
+    with language(request.LANGUAGE_CODE):
```
raphaelm: I'm unsure whether this context manager is really necessary, it should not be, but it surely does not hurt.

raphaelm:
Thank you very much! :)

Issue #182, opened 2016-08-13
Add Open Sans webfont and use it everywhere


raphaelm:
Might fix #179 and obsoletes #167.

PR #290, opened 2016-10-29
add open sans webfont

FlaviaBastos:
Add Open Sans webfont
#182

Sass is new to me and I wonder if I made the correct changes....
Chrome renders correctly but Firefox shows 'Open Sans' in computed but Deja Vu in 'Fonts' :confused: Suggestions?


src/static/bootstrap/scss/bootstrap/_type.scss
```diff
@@ -8,6 +8,7 @@
 
 h1, h2, h3, h4, h5, h6,
 .h1, .h2, .h3, .h4, .h5, .h6 {
+  font-path: $font-path;
```
raphaelm: Why did you put this in here? (I think there is not even a CSS property ``font-path`` and you can just omit this)

src/static/bootstrap/scss/bootstrap/_variables.scss
```diff
@@ -43,7 +43,8 @@ $link-hover-decoration: underline !default;
 //
 //## Font, line-height, and color for body text, headings, and more.
 
-$font-family-sans-serif:  "Helvetica Neue", Helvetica, Arial, sans-serif !default;
+$font-path:        "../../../fonts" !default;
```
raphaelm: This file is part of bootstrap, we should not change it. Instead, we can set the variables in our ``_variables.scss`` to override them, e.g.:

```
$font-family-sans-serif:  "Open Sans", "OpenSans", "Helvetica Neue", Helvetica, Arial, sans-serif !default;
$font-path: static("fonts");
```

src/static/pretixcontrol/scss/main.scss
```diff
@@ -1,6 +1,7 @@
 @import "_variables.scss";
 @import "../../bootstrap/scss/_bootstrap.scss";
 $fa-font-path: static("fontawesome/fonts");
+$font-path: static("fonts");
```
raphaelm: Same here, this can move to ``_variables.scss``.

src/static/pretixpresale/scss/main.scss
```diff
@@ -1,6 +1,7 @@
 @import "../../pretixbase/scss/colors.scss";
 @import "../../bootstrap/scss/_bootstrap.scss";
 $fa-font-path: static("fontawesome/fonts");
+$font-path: static("fonts");
```
raphaelm: Same here, this can move to ``_variables.scss``.

raphaelm:
Sorry, I hit "Send" to early. I'll update my comment above in a few minutes with more details!

FlaviaBastos:
> Webfonts are not easy to get right on first try.

This is an understatement or you just being super nice (as usual), considering I got it all wrong! LOL :stuck_out_tongue_closed_eyes: 
Thank you very much for the pointers. I will see how far I get. :)

FlaviaBastos:
Let's try this again! :blush:  
I have a few questions, please:
1. Should I delete the original 'OpenSans' ttf files?

2. I put the new scss file with the `font-face` rules under `static/pretixbase`. Is that ok? (I initially had it under `static/pretixcontrol`, maybe because of `_variables.scss`, not sure, but I moved it. I can move back or anywhere else you think it makes more sense)

3. I'm stuck with naming the font face. It's only working because I called `$font-family-sans-serif:  "open_sansregular", ...` in `_variables`, but this is very specific, I know. It's not using all the other fancy stuff there. The more "generic" OpenSans name doesn't work and I don't know enough to know why. I found [this old article and am wondering if this solution](https://www.smashingmagazine.com/2013/02/setting-weights-and-styles-at-font-face-declaration/#style-linking) is something we should pursue. 

Thanks!

raphaelm:
Hi Flavia,

just to warn you: I'm at [Django under the hood](https://djangounderthehood.com/) till Sunday, so expect a bit slower response times as usual :)

> Should I delete the original 'OpenSans' ttf files?

Let's keep it for now, it is currently used for PDF file generation.

> I put the new scss file with the font-face rules under static/pretixbase. Is that ok? (I initially had it under static/pretixcontrol, maybe because of _variables.scss, not sure, but I moved it. I can move back or anywhere else you think it makes more sense)

I think that's fine. 

> I'm stuck with naming the font face. It's only working because I called $font-family-sans-serif: "open_sansregular", ... in _variables, but this is very specific, I know. It's not using all the other fancy stuff there. The more "generic" OpenSans name doesn't work and I don't know enough to know why. I found this old article and am wondering if this solution is something we should pursue.

Tough one. The article seems to be pretty good, do you understand what it says? 

We can probably just ignore (as in "throw out") everyting except Regular/Italic/Bold/BoldItalic, at least I currently don't see where we'd use other variants. This leaves us with four styles, which does not trigger the IE bug the article mentions. I don't really care about BlackBerry 9900 and the iPad 1, so I think we can go for [Style Linking](https://www.smashingmagazine.com/2013/02/setting-weights-and-styles-at-font-face-declaration/#style-linking), at least for now.

FlaviaBastos:
> just to warn you: I'm at Django under the hood till Sunday, so expect a bit slower response times as usual :)

oohhh, how exciting! Enjoy your time there! And thanks for the heads up :)

> so I think we can go for Style Linking, at least for now.

I will give this a try then. Thanks for the feedback.

FlaviaBastos:
As far as I can tell, style linking method worked out fine and was simple to implement. :open_mouth: 
I didn't delete the font folders for the variations we are not using but I can do that if you prefer.
Lastly, I know you are away. Don't worry about replying, please. :)

raphaelm:
I just deleted the unused files my self and merged this manually, so that we are done here! :) Thank you very much!

Issue #149, opened 2016-06-27
Copy answers into multiple tickets


raphaelm:
If a user buys more than one ticket, in many cases he has to enter multiple names and answer multiple questions defined by the admin. In some cases, it would be useful to have a butten that copies the answers from the first ticket into the other tickets.

andmckvr13:
Hello raphaelm, sorry if this is a stupid question, but are you using JavaScript for this or a different language? I'm a new developer (JS only for now)  and looking to gain more contributions.

raphaelm:
Hi @andmckvr13, yes, I would do this in JavaScript. We do not have a lot of JS in pretix right now, but this is one of the points where it makes sense.

If you want to to contribute to this, the best way to start is to [get pretix running on your machine](https://docs.pretix.eu/en/latest/development/setup.html) and play with it for a few minutes to get a feeling of how pretix works and where this is going. If you have any questions on the way, no matter how stupid they seem, please do not hestitate and ask, I'll be happy to guide you through the process!

andmckvr13:
Danke Schön

chrisdamba:
Hi @raphaelm Is this feature still needed? This would be my first pull request but I would like to give it a go?

raphaelm:
Hi @chrisdamba! I believe this would still be useful, so feel free to try yourself! If you get stuck in the process, please ask, I'll be happy to help :)

chrisdamba:
Thanks @raphaelm for the opportunity, let me set up pretix and will definitely give you a nudge should I bump into any problem :)

chrisdamba:
@raphaelm I've attached a screenshot of what I think may be the correct page to implement the enhancement, is that right? Am I looking into the right place, i.e. `pretix/control/templates/pretixcontrol/order/index.html`?
![pretix_1](https://cloud.githubusercontent.com/assets/2035056/19658130/b6e7e2d8-9a1e-11e6-8a16-3948c2ddb909.png)

raphaelm:
Yes, that is the correct file!

PR #292, opened 2016-10-30
Added copying multiple tickets answers enhancement (#149)

chrisdamba:
Enhancement for copying answers into multiple tickets with a checkbox state as UI

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/checkout_questions.html
+++ b/src/pretix/presale/templates/pretixpresale/event/checkout_questions.html
@@ -41,6 +41,11 @@
                     </div>
                 </div>
             {% endif %}
+            <div id="cp_copy" class="panel panel-default">
+                <div class="panel-body">
+                    <input type="checkbox" name="copy" id="copy_answers">  Copy Answers
+                </div>                
+            </div>
             {% for form in forms %}
                 <div class="panel panel-default">
                     <div class="panel-heading">
--- a/src/pretix/presale/templates/pretixpresale/event/checkout_questions.html
+++ b/src/pretix/presale/templates/pretixpresale/event/checkout_questions.html
@@ -54,7 +54,7 @@
                             </a>
                             {% if forloop.counter > 1 %}
                                 <span class="text-right pull-right">
-                                     <input type="checkbox" data-id="{{ forloop.counter0 }}" name="copy" class="js-copy-answers" id="copy_answers_{{ form.pos.id }}"> {% trans "Copy answers from above" %}
+                                     <input type="button" data-id="{{ forloop.counter0 }}" name="copy" class="js-copy-answers btn btn-block" id="copy_answers_{{ form.pos.id }}" value='{% trans "Copy answers from above" %}'> 
                                 </span>
                             {% endif %}                           
                         </h4>
--- a/src/pretix/presale/templates/pretixpresale/event/checkout_questions.html
+++ b/src/pretix/presale/templates/pretixpresale/event/checkout_questions.html
@@ -41,11 +41,7 @@
                     </div>
                 </div>
             {% endif %}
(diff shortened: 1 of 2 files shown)


src/pretix/presale/templates/pretixpresale/event/checkout_questions.html
```diff
@@ -41,6 +41,11 @@ <h4 class="panel-title">
                     </div>
                 </div>
             {% endif %}
+            <div id="cp_copy" class="panel panel-default">
+                <div class="panel-body">
+                    <input type="checkbox" name="copy" id="copy_answers">  Copy Answers
```
raphaelm: We need to have this translatable, you can use something like ``{% trans "Copy answers" %}`` for this.

src/static/pretixpresale/js/ui/main.js
```diff
@@ -22,8 +22,67 @@ $(function () {
...
+        }
+    })
 });
 
+function display_copy() {
+    var elements = $("input").filter(function(){
+        return this.id.match(/^id_(\d+)/);
+    });    
+    var shouldElementBeVisible = (elements.length > 2);    
+    $("#cp_copy").toggle(!!shouldElementBeVisible);
```
raphaelm: This is already boolean, I believe there is no reason for the `!!`.

chrisdamba:
Thanks @raphaelm for flagging the above. Of course I'm happy to revise and implement. On the UI, I felt a checkbox is a better control than a button as it allows toggling choices/states. I certainly agree with the suggested approach "copy answers from above" with the checkbox aligned right. Does this need to be on multiple tickets or on the last one?

Sure, will debug the other issues and also add support for other input types.

raphaelm:
> On the UI, I felt a checkbox is a better control than a button as it allows toggling choices/states. 

Yes! I did not even think about it a checkbox instead of the button before your PR, but I really like it :)

> I certainly agree with the suggested approach "copy answers from above" with the checkbox aligned right. Does this need to be on multiple tickets or on the last one?

I think ideally it should be on all of them (except the first one, obviously) and if you activate it, it should copy the answers from the **first** ticket to the one the checkbox got activated on.

chrisdamba:
Seems like I still have so much to learn from UX design :) It's actually starting to make sense if we remove the "unbind" which essentially renders the checkbox as a one-way action event and thus making way for a button! (the original design). Yeah, let's go for the button.

Still not sure why checkboxes and radios are not being picked, will have a look once again and let you know how that goes. Thanks for the feedback, really useful to me as it gradually points me in the right direction.

raphaelm:
Feel free to ping me if you get totally stuck with the checkboxes and radios :) But some googling on jQuery and checkbox handling should bring up some useful results.

chrisdamba:
@raphaelm Apologies for the slight delay in fix, been a bit snowed in with work lately but still very much into getting my first pull request changes done :)

raphaelm:
No worries, we're not in a hurry!

chrisdamba:
I am not sure what's causing the reduction in code coverage with the changes that I made in the PR. Can anyone help me to understand why that happened?

raphaelm:
As you don't even touch the python code… please ignore it. I have to turn off the coveralls commits at some point, they grow to be annoying :(

rixx:
It's neglegibly small; possibly based in your branch branching off from an earlier point of the master branch. Don't worry about it. :)

src/static/pretixpresale/js/ui/main.js
```diff
@@ -32,14 +31,6 @@ $(function () {
     });
 });
 
-function display_copy() {
```
chrisdamba: This was redundant, forgot to take it out after the requested changes.

chrisdamba:
Thanks @raphaelm ! :)

Issue #274, opened 2016-10-14
Disable weekends in datetimepicker for expiry


rixx:
… if `payment_term_weekdays` is set. The datetimepicker accepts a `daysOfWeekDisabled` option for this purpose.

chrisdamba:
Could pick this up if that's ok?

raphaelm:
Sure, go ahead! :)

chrisdamba:
@rixx This may sound silly but how do I pass the `payment_term_weekdays` context variable into the main.js file where the datetimepicker plugin is initialised?

raphaelm:
@chrisdamba it isn't, there is no obvious solution for that. But we had the problem before e.g. with the date formats and we solved it like this:

* in ``pretix/control/context.py`` we pass the variable to the template
* in the ``pretix/control/templates/pretixcontrol/base.html`` we set it as a data attribute to the ``<body>`` tag
* we then read it from there with javascript

chrisdamba:
Thanks @raphaelm for the workaround, that's quite helpful

PR #317, opened 2016-11-13
Fix for Disable weekends in datetimepicker for expiry #274

chrisdamba:


src/pretix/control/context.py
```diff
@@ -26,10 +26,14 @@ def contextprocessor(request):
             _html_head.append(response)
     ctx['html_head'] = "".join(_html_head)
 
+    _js_datetime_weekdays = '[]'
     _nav_event = []
     if hasattr(request, 'event'):
         for receiver, response in nav_event.send(request.event, request=request):
             _nav_event += response
+        if request.event.settings.get('payment_term_weekdays'):
+            _js_datetime_weekdays = '[0,6]'
```
raphaelm: When I first read this, I was very confused -- is this tha days that are disabled? Or the enabled ones? How about naming the variable (and the HTML attribute) something like ``payment_weekdays_disabled`` instead of ``datetimes_weekdays``?

raphaelm:
So if I see this correctly, we close this in favor of #320?

chrisdamba:
Yup, this can go

PR #320, opened 2016-11-17
Fix for Disable weekends in datetimepicker for expiry #274

chrisdamba:
Previous build on a different branch was failing so I managed to grab the latest changes from the main branch, merged the changes in a new branch, ran tests and get the new PR.

### DIFF
--- a/src/pretix/control/forms/orders.py
+++ b/src/pretix/control/forms/orders.py
@@ -15,7 +15,7 @@ class ExtendForm(I18nModelForm):
         widgets = {
             'expires': forms.DateInput(attrs={
                 'class': 'datepickerfield',
-                'data-payment-weekdays-disabled': '[]'
+                'data-is-payment-date': 'true'
             })
         }
 
--- a/src/pretix/control/forms/orders.py
+++ b/src/pretix/control/forms/orders.py
@@ -13,7 +13,10 @@ class ExtendForm(I18nModelForm):
         model = Order
         fields = ['expires']
         widgets = {
-            'expires': forms.DateInput(attrs={'class': 'datepickerfield'}),
+            'expires': forms.DateInput(attrs={
+                'class': 'datepickerfield',
+                'data-payment-weekdays-disabled': '[]'
+            })
         }
 
     def clean(self):
--- a/src/static/pretixcontrol/js/ui/main.js
+++ b/src/static/pretixcontrol/js/ui/main.js
@@ -127,7 +127,7 @@ $(function () {
     });
 
     $(".datepickerfield").each(function() {
-        $(this).datetimepicker({
+        var opts = {
             format: $("body").attr("data-dateformat"),
             locale: $("body").attr("data-datetimelocale"),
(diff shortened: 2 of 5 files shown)


chrisdamba:
@raphaelm The build is failing on JOB=style and I am a bit confused on why that's failing as the checks on local are OK. Would you have an idea?
pretix-1

rixx:
This is not something that happens due to your commits and should be solved after the merge into `master`, @chrisdamba.

Just for reference, though: You should be able to reproduce the failing style checks by running `flake8` in the `src` directory.

chrisdamba:
Thanks @rixx for the explanation :) I couldn't reproduce the failing style in the `src` directory (check attached previous screenshot), perhaps it's because I was in a different branch?

rixx:
Seems to be an issue with the newest flake8 (`pip install -U flake8` should give you the same errors). Since they are legitimate, I'll submit a PR fixing those, I guess.

rixx:
[pushed](https://github.com/pretix/pretix/pull/321)

chrisdamba:
Nice one, will update my branch with your fixes once the build passes

src/pretix/control/forms/orders.py
```diff
@@ -13,7 +13,10 @@ class Meta:
         model = Order
         fields = ['expires']
         widgets = {
-            'expires': forms.DateInput(attrs={'class': 'datepickerfield'}),
+            'expires': forms.DateInput(attrs={
+                'class': 'datepickerfield',
+                'data-payment-weekdays-disabled': '[]'
```
raphaelm: As this will *always* have this value, I again find the name misleading. How about ``data-is-payment-date=true`` instead to indicate it *is* a payment date? Because this is what this does, all other logic is in the ``body`` attribute.

src/static/pretixcontrol/js/ui/main.js
```diff
@@ -143,7 +143,10 @@ $(function () {
                 clear: 'fa fa-trash',
                 close: 'fa fa-remove'
             }
-        });
+        };
+        if ($(this).data().hasOwnProperty("paymentWeekdaysDisabled"))
```
raphaelm: I find jQuery's ``data()`` thing not very intuitive, how about ``$(this).is('[data-is-payment-date]')`` instead to keep with the HTML notation?

src/pretix/control/forms/orders.py
chrisdamba: I agree, `data-is-payment-date=true` is much more clearer. Let me update my branch with this suggestion. Thank you!

src/static/pretixcontrol/js/ui/main.js
chrisdamba: That should tally with the above suggestion so yeah, I don't see any reason why not. Will update accordingly

raphaelm:
Nice, thank you very much!

chrisdamba:
No worries, I appreciate the support

Issue #308, opened 2016-11-09
Add preview functionality to mail templates


raphaelm:
**Hey!** If you ever wanted to contribute to an Open Source project, but didn't know how, you just found a perfect place to start. We try to write this issue as instructive as possible, but as we're humans we almost certainly forgot to write something important in here. If you want to try to tackle this issue and get stuck at _any_ point, please do not hesitate and ask us, either here on GitHub or via email at support@pretix.eu. We'll be very happy to help! 

We think this issue is very suitable to people who are just starting with Python or Django development or who never contributed to Open Source before. We'd like to ask people with wide experience in these areas to leave this issue to beginners. Thanks! (If you want something more advanced, just drop us a line, we'll find something :wink:)

## How to get started

First of all, you need pretix running locally on your machine. We have a guide on the developer installation [over here](https://docs.pretix.eu/en/latest/development/setup.html). Play around with it and get a feeling for what pretix is and how it works :) Try to find the places in the interface mentioned in this issue where something needs to change. Maybe try to find the same places in the code? Just searching for characteristic words can be a powerful tool to find your way around.

## What this is about

If you go to Settings → Email in the control panel of a pretix event, you see a number of content text snippets that you can configure. For most of those texts, you can use placeholders like ``{total}`` for the order's total price to include individual values. It would be nice to provide a preview for those texts, so users can see that they used the placeholders correctly.

## How to do this

There are multiple approaches on this and I'd happily discuss other ideas, if you have any. The best user interface I've come up with this so far ist the following. It will require some knowledge about both Django and JavaScript/jQuery, but don't be scared, its simple usage of both and I'm here to help if you get stuck.

Add a "preview" button/link near the text that sends the current text in the textarea via AJAX to a to-be-created view that generates a preview by replacing the correct placeholders with sample data and returns the result that is then shown as a preview above or below the textarea.

I'd suggest you implement a first try for *one* of the text fields and create your PR to discuss the general method further,  before we implement it for all of them.

## I want to give it a try!

You're still reading, nice! If you want to try working on this, please comment below, so everyone else knows this is being looked at and can search for a different place to contribute. 

[...]

Issue #307, opened 2016-11-09
Sendmail: Log sent mails


raphaelm:
**Hey!** If you ever wanted to contribute to an Open Source project, but didn't know how, you just found a perfect place to start. We try to write this issue as instructive as possible, but as we're humans we almost certainly forgot to write something important in here. If you want to try to tackle this issue and get stuck at _any_ point, please do not hesitate and ask us, either here on GitHub or via email at support@pretix.eu. We'll be very happy to help! 

We think this issue is very suitable to people who are just starting with Python or Django development or who never contributed to Open Source before. We'd like to ask people with wide experience in these areas to leave this issue to beginners. Thanks! (If you want something more advanced, just drop us a line, we'll find something :wink:)

## How to get started

First of all, you need pretix running locally on your machine. We have a guide on the developer installation [over here](https://docs.pretix.eu/en/latest/development/setup.html). Play around with it and get a feeling for what pretix is and how it works :) Try to find the places in the interface mentioned in this issue where something needs to change. Maybe try to find the same places in the code? Just searching for characteristic words can be a powerful tool to find your way around.

## What this is about

If you activate the "sendmail" plugin in your pretix event, you'll be able to send out emails to all your event attendees at once (or all unpaid orders, etc).
However, you never again see what you sent. It would be nice to save those emails to the database and display them for later reference.

## How to do this

This requires to touch multiples areas of the Django framework. If you have no experience with Django, it will still be doable but probably require you to read up a bit on varios topics.

* Add a custom database model to the sendmail plugin. It should contain the email's subject, content and date as well as a many-to-many relationship to the Order model, specifiying which users exactly it has been sent to.

* Generate a database migration for this model

* Store all sent mails into the database using this model

* Show a list of the sent emails in the control panel, included or linked on the page with the mailing form

If you never heard of those before, read up on Django's models, migrations and views. Also, read how [plugins in pretix can add views](https://docs.pretix.eu/en/latest/development/api/customview.html). Be sure to also read [how to test mails without actually sending them](https://docs.pretix.eu/en/latest/development/setup.html#working-with-mails) in our developer documentation as it will make your life easier.

## I want to give it a try!

You're still reading, nice! If you want to try working on this, please comment below, so everyone else knows this is being looked at and can search for a different place to contribute. 

[...]

PR #323, opened 2016-11-18
Add empty cart button #161

FlaviaBastos:
Add 'Empty Cart' button to cart.
Add test_remove_all to tests

One note: I added a new view `CartRemoveAll` but the only difference from `Cart Remove` is the success message.... Is that ok or are they too similar to have both?

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/index.html
+++ b/src/pretix/presale/templates/pretixpresale/event/index.html
@@ -15,19 +15,32 @@
             </div>
             <div class="panel-body">
                 {% include "pretixpresale/event/fragment_cart.html" with cart=cart event=request.event editable=True %}
-                <div class="cart-row row">
-                    <div class="col-md-6 col-xs-12">
-                        <em id="cart-deadline" data-expires="{{ cart.first_expiry|date:"Y-m-d H:i:sO" }}">
-                            {% if cart.minutes_left > 0 %}
-                                {% blocktrans trimmed with minutes=cart.minutes_left %}
-                                    The items in your cart are reserved for you for {{ minutes }} minutes.
-                                {% endblocktrans %}
-                            {% else %}
-                                {% trans "The items in your cart are no longer reserved for you." %}
-                            {% endif %}
-                        </em>
+                <em id="cart-deadline" data-expires="{{ cart.first_expiry|date:"Y-m-d H:i:sO" }}">
+                    {% if cart.minutes_left > 0 %}
+                        {% blocktrans trimmed with minutes=cart.minutes_left %}
+                            The items in your cart are reserved for you for {{ minutes }} minutes.
+                        {% endblocktrans %}
+                    {% else %}
+                        {% trans "The items in your cart are no longer reserved for you." %}
+                    {% endif %}
+                </em>
+                <div class="row checkout-button-row">
+                    <div class="col-md-4 col-xs-12">
+                        <form method="post" data-asynctask action="{% eventurl request.event "presale:event.cart.remove.all" %}" >
+                            {% csrf_token %}
+                            {% for line in cart.positions %}
+                                {% if line.variation %}
+                                    <input type="hidden" name="variation_{{ line.item.id }}_{{ line.variation.id }}" value="{{ line.count }}" />
+                                    <input type="hidden" name="price_{{ line.item.id }}_{{ line.variation.id }}" value="{{ line.price }}" />
+                                {% else %}
+                                    <input type="hidden" name="item_{{ line.item.id }}" value="{{ line.count }}" />
+                                    <input type="hidden" name="price_{{ line.item.id }}" value="{{ line.price }}" />
+                                {% endif %}
+                            {% endfor %}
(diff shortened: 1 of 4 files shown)


raphaelm:
Hi Flavia,

I like very much how you built this re-using the code from ``CartRemove`` and just passing in values for all lines. I didn't even think of that ;) I see two ways to nicely get around duplicating code introduced by this:

1. Make ``CartRemoveAll`` a subclass of ``CartRemove`` that only defines the success message and inherits all other methods from its parent.

2. Drop the new view completely and improve ``CartRemove.get_success_message`` in a way that selects between the two messages depending on whether any more cart items exists: ``if CartPosition.objects.filter(cart_id=self.request.session.session_key).exists(): ...`` This will then also show the "is now empty" message if the user just deletes the last item manually, which is perfectly fine.

I'm slightly in favor of option two. What do you think?

We'll probably run into another problem here: There is the possibility to have multiple items in the cart that are to the same item and variation but are *not* grouped together in one cart line. This can happen if their price differs (e.g. because one has a voucher attached) or if there are questions defined and filled in etc. With your code, this would lead to the duplicate definition of a form field (which is fine in HTML, it will just get send twice) but probably won't be parsed correctly by ``_items_from_post_data()``. If you want, you can try to patch ``_items_from_post_data`` for this, but it probably won't be easy. I can do that as well. But I'd love for this PR to include at least a (failing) test that checks this so that we don't forget it and that we know it is solved once it is :) You can decorate that test with ``@pytest.mark.xfail`` ("expected failure") such that this PR stays green ;)

Best
Raphael

FlaviaBastos:
>  just passing in values for all lines. I didn't even think of that

Hm! Now I am curious: what would be your go to approach?

> I'm slightly in favor of option two. What do you think?

Yeah, let's go with 2. I will see how far I can go and I will let you know if I get stuck. I will also add the failing test as you requested. 

Thanks for the feedback! :)

raphaelm:
> Hm! Now I am curious: what would be your go to approach?

I would have probable made a view that simply executes ``CartPosition.objects.filter(cart_id=request.session.session_key).delete()``. Now that I've written that down, it looks very simple. But option two of the above doesn't add any new views and no new code that touches CartPositions, I like that.

FlaviaBastos:
I made the modifications you suggested above (implementing 2) and I can see the issue you pointed out: If I have a discounted t-shirt (with a voucher) and a regular priced t-shirt in the cart, the empty cart doesn't remove both. :unamused: 
But now I'm having a hard time writing a failing test. They all pass! :stuck_out_tongue: I am missing something for sure. Here's what I have now:
```
    def test_remove_all_same_variation_different_price(self):
        CartPosition.objects.create(
            event=self.event, cart_id=self.session_key, item=self.shirt, variation=self.shirt_red,
            price=14, expires=now() + timedelta(minutes=10)
        )
        v = Voucher.objects.create(item=self.shirt, variation=self.shirt_red, price=Decimal('10.00'), event=self.event)
        self.client.post('/%s/%s/cart/add' % (self.orga.slug, self.event.slug), {
            'variation_%d_%d_voucher' % (self.shirt.id, self.shirt_red.id): v.code,
        }, follow=True)
        response = self.client.post('/%s/%s/cart/remove' % (self.orga.slug, self.event.slug), {
            'variation_%d_%d' % (self.shirt.id, self.shirt_red.id): '2',
        }, follow=True)
        doc = BeautifulSoup(response.rendered_content, "lxml")
        self.assertIn('empty', doc.select('.alert-success')[0].text)
        self.assertFalse(CartPosition.objects.filter(cart_id=self.session_key, event=self.event).exists())
```
Can you help, please?

raphaelm:
Heh. You somehow outsmarted yourself when writing that test :wink:
```
        response = self.client.post('/%s/%s/cart/remove' % (self.orga.slug, self.event.slug), {
            'variation_%d_%d' % (self.shirt.id, self.shirt_red.id): '2',
        }, follow=True)
```
works, because CartRemove is clever enough to work with the ``2``. Your template, however, will generate something different. It won't generate a value of ``2`` but two values of ``1``:
```
<input type="hidden" name="variation_1_1" value="1">
<input type="hidden" name="variation_1_1" value="1">
```
So the test would probably (I didn't test this) need to look like:
```
        response = self.client.post('/%s/%s/cart/remove' % (self.orga.slug, self.event.slug), {
            'variation_%d_%d' % (self.shirt.id, self.shirt_red.id): ('1', '1')
        }, follow=True)
```

FlaviaBastos:
Where would you like me to go from here? Explore the `_items_from_post_data` and try to fix the issue I created :flushed: or do I leave it to you (you mentioned it will not be easy and I believe you!)?
If you prefer to take over the remaining part of this issue, then I guess I could look at #307 or #308 as you suggested... Thoughts?

raphaelm:
It's not actually that hard to *change*, but its rather hard to *find* what to change, so I did it and refactored the method a bit on the way. Sure, have a look at the other issues, I'd suggest #307 :)

FlaviaBastos:
Ah, nice! I see your code. 
I was a bit uneasy leaving things not working 100% :disappointed: Thanks for accepting my code anyway. :)
I will move to #307 then.

PR #326, opened 2016-11-20
[WIP] Reduce locking timeframes

raphaelm:


raphaelm:
Todo: check for race cond that allows duplicate voucher redemption

PR #331, opened 2016-11-24
Require live

rixx:


raphaelm:
If I'm not horribly mistaken, this will break the usage of ``@event_view`` without parameters. ``@event_view()`` will continue to work, but probably most usages are without the ``()``. This is *really* one of the nastiest things in Python to get right.

Fixing this requires weird workarounds like [this](http://stackoverflow.com/a/3931903/336784). If I recall correctly, there's also something in Fluent Python. [functools.wraps](https://docs.python.org/3/library/functools.html?highlight=functools#functools.wraps) helps, but also doesn't solve the problem.

For illustration, this is a stripped-down example of how it breaks:

```
In [4]: def decorator(flag=True):
   ...:     print("decorating with flag", flag)
   ...:     def decorator_wrapper(func, flag=flag):
   ...:         print("decorating", func, "with flag", flag)
   ...:         def wrap(*args, **kwargs):
   ...:             print("wrap called with", args, kwargs)
   ...:             return func(*args, **kwargs)
   ...:         return wrap
   ...:     return decorator_wrapper
   ...: 

In [5]: @decorator
   ...: def foo_without_flag():
   ...:     print("foo")
   ...:     
decorating with flag <function foo_without_flag at 0x7f2bbe6bc620>

In [6]: foo_without_flag()
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-6-b01cbe21516d> in <module>()
----> 1 foo_without_flag()

TypeError: decorator_wrapper() missing 1 required positional argument: 'func'

In [7]: @decorator(flag=False)
   ...: def foo_with_flag():
   ...:     print("foo")
   ...:     
decorating with flag False
decorating <function foo_with_flag at 0x7f2bbdc316a8> with flag False

In [8]: foo_with_flag()
wrap called with () {}
foo
```

rixx:
Wouldn't that break some tests then? I can try to go for one of the workarounds though, sure.

raphaelm:
Thanks! :)

PR #334, opened 2016-11-24
Fix 303

chrisdamba:


src/pretix/base/payment.py
```diff
@@ -190,6 +198,13 @@ def payment_form(self, request: HttpRequest) -> Form:
         form.fields = self.payment_form_fields
         return form
 
+    def _is_still_available(self):
+        availability_date = self.settings.get('_availability_date', as_type=datetime)
+        if availability_date is not None and datetime.now() > availability_date:
```
rixx: Please use `from django.utils.timezone import now`, and `now() > availability_date` which uses sensible timezone settings. :)

Also, `now` is a datetime object while the availability date is a date object, is that intended?

src/tests/base/test_payment.py
```diff
@@ -1,5 +1,6 @@
 from decimal import Decimal
 
+import datetime
```
rixx: As `now` is already imported in this file, use it here, too, please. 

(But just for reference, the style tests were failing because the datetime import is in the wong line - don't worry about that, though. In the future, you might want to run `isort -rc .` before committing if you add new imports. Just an aside, don't worry about it.)

src/pretix/base/payment.py
```diff
@@ -190,6 +198,13 @@ def payment_form(self, request: HttpRequest) -> Form:
...
+            return False
+        else:
+            return True
```
rixx: You might consider shortening this to
```
if available_date:
    return available_date >= now()
return True
```
making the logic more explicit.

src/pretix/base/payment.py
```diff
@@ -199,7 +214,7 @@ def is_allowed(self, request: HttpRequest) -> bool:
 
         The default implementation always returns ``True``.
         """
-        return True
+        return self._is_still_available()
```
rixx: Please update the docstring to reflect the new default implementation here.

src/pretix/base/payment.py
```diff
@@ -339,7 +354,7 @@ def order_change_allowed(self, order: Order) -> bool:
 
         :param order: The order object
         """
-        return True
+        return self._is_still_available()
```
rixx: Please update the docstring here, aswell.

src/tests/base/test_payment.py
chrisdamba: Thanks for the feedback, I'll definitely keep that in mind

raphaelm:
Hi,

I just reviewed and tested this extensively. The thing why I am so cautious is that this compares dates and is therefore prone to have timezone problems. The good news: **I can't find any, your code works fine** -- at least as long as the "current" timezone is the correct one, which depends on the code that *calls* this method. I prefer to make it explicit and always force the event's timezone to be used.

First, as rixx suggested, use Django's ``now()`` function instead of ``datetime.now`` and don't get something as a datetime from the settings store if you only ever stored a date in there:

```
import pytz
from django.utils.timezone import now
...

    def _is_still_available(self):
        tz = pytz.timezone(self.event.settings.timezone)
        availability_date = self.settings.get('_availability_date', as_type=date)
        if availability_date:
            return availability_date >= now().astimezone(tz).date()
        return True
```

Further, to make the code more testable, I suggest an optional argument for the current datetime. We do this in a lot of time-handling methods already:
```
    def _is_still_available(self, now_dt=None):
        now_dt = now_dt or now()
        tz = pytz.timezone(self.event.settings.timezone)
        availability_date = self.settings.get('_availability_date', as_type=date)
        if availability_date:
            return availability_date >= now_dt.astimezone(tz).date()
        return True
```

You can then easily add a test that tests some datetime edgecases, for example:

```
@pytest.mark.django_db
def test_availability_date_timezones(event):
    event.settings.set('timezone', 'US/Pacific')
    prov = DummyPaymentProvider(event)
    prov.settings.set('_availability_date', '2016-12-01')

    tz = pytz.timezone('US/Pacific')
    utc = pytz.timezone('UTC')
    with override(tz):
        assert prov._is_still_available(tz.localize(datetime.datetime(2016, 11, 30, 23, 0, 0)).astimezone(utc))
        assert prov._is_still_available(tz.localize(datetime.datetime(2016, 12, 1, 23, 59, 0)).astimezone(utc))
        assert not prov._is_still_available(tz.localize(datetime.datetime(2016, 12, 2, 0, 0, 1)).astimezone(utc))

```

chrisdamba:
Fantastic! I had not stretched my thinking cap that far with regards to timezones testing and your suggestions really make sense.

src/pretix/base/payment.py
```diff
@@ -190,16 +200,22 @@ def payment_form(self, request: HttpRequest) -> Form:
...
+        return True
+
     def is_allowed(self, request: HttpRequest) -> bool:
         """
         You can use this method to disable this payment provider for certain groups
         of users, products or other criteria. If this method returns ``False``, the
         user will not be able to select this payment method. This will only be called
         during checkout, not on retrying.
-
-        The default implementation always returns ``True``.
```
raphaelm: One more thing: How about instead of removing this line (and the similar one above) replacing it with something like "The default implementation checks for the ``_availability_date`` setting to be either unset or in the future.". This helps developers only reading the documentation to know how they should override this method, if they do.


We should also mention this field in the docstring of ``settings_form_fields`` for the same purpose.

chrisdamba: Sure, thanks for the pointer.

raphaelm:
Thanks a lot! :)

chrisdamba:
No worries, I'm actually learning a lot :)

chrisdamba:
Not related to this but do you have any plans in your road-map to incorporate machine learning in the app?

rixx:
Machine learning sounds like one of the best possible applications for a plugin, becaus it can be veeeery standalone. What would you want to learn?

raphaelm:
I currently haven't any plans in that direction. Do you have a good idea where it would be useful for pretix' users? If so, I'd be happy to talk about it! (Although I'd probably rather see something of that complexity in a plugin, as rixx suggested).

chrisdamba:
To be honest I don't have a specific use-case in mind but as a general idea an rough example I can think of (may not be applicable) could be anomaly detection where there can be some scripted attacks often associated with a noticeable increase in some measurable metric over a short period of time. For example, a sudden 5000% increase in ticket reservations in a particular event could be a result of excellent marketing, or fraud?

raphaelm:
Yes, fraud detection would be valuable -- and very hard. I believe our dataset is waaay to small until now to get this with usable quality.

PR #336, opened 2016-11-26
Change column layout in order overview. Closes #302

### LINKED ISSUE #302: Order overview: Improve columns
The labels of the columns are unclear and should be improved, maybe with explaining tooltips. Also, a reordering of the columns could help.

rixx:
2016-11-26-132808_1167x608_scrot

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/orders/overview.html
+++ b/src/pretix/control/templates/pretixcontrol/orders/overview.html
@@ -19,9 +19,11 @@
                     <th>{% trans "Product" %}</th>
                     <th>{% trans "Canceled" %}</th>
                     <th>{% trans "Refunded" %}</th>
+                    <th>{% trans "Expired" %}</th>
                     <th colspan="3">{% trans "Purchased" %}</th>
                 </tr>
                 <tr>
+                    <th></th>
                     <th></th>
                     <th></th>
                     <th></th>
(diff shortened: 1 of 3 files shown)


raphaelm:
Yay! Now that this looks so much nicer, what do you think about expired orders? I don't even remember correctly if they are in the "pending" column and added to the total or not, which is a pretty bad sign. :( Should we add them as their own column now?

rixx:
*sigh* We probably should.

rixx:
Done in second commit.

src/pretix/control/templates/pretixcontrol/orders/overview.html
```diff
@@ -17,63 +17,78 @@
...
+                                    {{ item.num_pending|togglesum }}
                                 </a>
                             </td>
                             <td>
                                 <a href="{{ listurl }}?item={{ item.id }}&amp;status=p&amp;provider={{ item.provider }}">
                                     {{ item.num_paid|togglesum }}
                                 </a>
                             </td>
+                            <td>
+                                <a href="{{ listurl }}?item={{ item.id }}&amp;provider={{ item.provider }}">
```
raphaelm: Shouldn't this now link to something like ``status=np`` that we would first need to create?

rixx: Hm, true - or we just don't add a link there. Seems sensible to me, too.

raphaelm:
Thanks! :)

PR #344, opened 2016-12-01
Use flexbox for better tiling ◻

rixx:


src/static/pretixcontrol/scss/_dashboard.scss
```diff
@@ -1,9 +1,26 @@
...
+  align-items: flex-start;
 }
-.dashboard > div {
+.dashboard .widget-container {
+  flex:1 0 auto;
+  align-self: stretch;
+  min-height: 1px;
   padding: 5px;
 }
+.dashboard .widget-container .full {
```
raphaelm: Should be ``.widget-container.full`` without the space, I guess.

PR #346, opened 2016-12-02
More extensive log display when changing an order email.

### LINKED ISSUE #345: Log old email when changing an order's email address
… because otherwise it's completely lost.

rixx:
Closes #345


raphaelm:
Thanks!

PR #348, opened 2016-12-04
Copy-to-clipboard for voucher bulk create

rixx:
- [x] Make it work
 - [x] Style it
 - [x] i18n


rixx:
I decided to go with a copy mechanism for only bulk creates for now. It's easier and more necessary.

raphaelm:
Sorry for taking so long here :cry: 

I like this a lot and it seems fine. While testing, I found one thing counter-intuitive: When the text box is still empty and you click the button, it shows "Copied!" even though my clipboard apparently still contains what it contained before. Do you see an easy way to fix this? A fix could be to show a different tooltip for example.  If it is not feasible with this library, I can also live with it like it is :)

Also, your .js file mixes tabs and spaces in the indentation. Could you… :blush: Thanks!

rixx:
Done, done and done.

raphaelm:
Thanks!

PR #353, opened 2016-12-09
Log emails sent with sendmail plugin #307

FlaviaBastos:
This logs to the database emails sent with sendmail plugin and displays them in the control backend
Also adds support to second-level nav to plugins

As I previously mentioned, I'm confused about *event*-level and *order*-level logs so I don't know if I am logging/displaying the right thing (probably not). I appreciate feedback. 
I'm displaying raw (uggly) data but 1- I wasn't sure I had the right thing and 2- I didn't know how to display it prettier. 

I didn't remove the code for the new Model either. Let me know if it needs to go.

### DIFF
--- a/src/.gitignore
+++ b/src/.gitignore
@@ -8,4 +8,3 @@ dist/
 *.egg-info/
 *.bak
 static/jsi18n/
-
--- a/src/pretix/plugins/sendmail/signals.py
+++ b/src/pretix/plugins/sendmail/signals.py
@@ -2,6 +2,7 @@ from django.core.urlresolvers import resolve, reverse
 from django.dispatch import receiver
 from django.utils.translation import ugettext_lazy as _
 
+from pretix.base.signals import logentry_display
 from pretix.control.signals import nav_event
 
 
@@ -19,5 +20,32 @@ def control_nav_import(sender, request=None, **kwargs):
             }),
             'active': (url.namespace == 'plugins:sendmail' and url.url_name == 'send'),
             'icon': 'envelope',
-        }
+            'children': [
+                {
+                    'label': _('Send email'),
+                    'url': reverse('plugins:sendmail:send', kwargs={
+                        'event': request.event.slug,
+                        'organizer': request.event.organizer.slug,
+                    }),
+                    'active': (url.namespace == 'plugins:sendmail' and url.url_name == 'send'),
+                },
+                {
+                    'label': _('Email history'),
+                    'url': reverse('plugins:sendmail:history', kwargs={
+                        'event': request.event.slug,
+                        'organizer': request.event.organizer.slug,
+                    }),
+                    'active': (url.namespace == 'plugins:sendmail' and url.url_name == 'history'),
+                },
+            ]
+        },
     ]
+
+
+@receiver(signal=logentry_display)
+def pretixcontrol_logentry_display(sender, logentry, **kwargs):
+    plains = {
+        'pretix.plugins.sendmail.sent': _('Email was sent'),
+    }
+    if logentry.action_type in plains:
+        return plains[logentry.action_type]
--- a/src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/history.html
+++ b/src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/history.html
@@ -16,31 +16,17 @@
                         {% if log.display %}
                             <br/><span class="fa fa-comment-o"></span> {{ log.display }}
                         {% endif %}
-                    </p>
-                    <p>
-                        <strong>Email address:</strong> {{ log.parsed_data.recipients }}
-                    </p>
-                    <p>
-                        <strong>Subject:</strong>
-                    </p>
-                    <p>
-                        {% for key,value in log.parsed_data.subject.items %}
-                            <p>
-                                {{ key }}: {{ value }}
-                            </p>
+                        <br/><span class="fa fa-shopping-cart"></span> {% trans "Sent to orders:" %}
+                        {% for status in log.parsed_data.sendto %}
+                            {{ status }}{% if forloop.revcounter > 1 %},{% endif %}
                         {% endfor %}
                     </p>
                     <p>
-                        <strong>Email content:</strong>
-                    </p>
-                    <p>
-                        {% for key,value in log.parsed_data.message.items %}
+                        {% for locale, value in log.pdata.locales.items %}
                             <p>
-                                {{ key }}:
+                                <strong>[{{ locale }}] {% trans "Subject:" %} {{ value.subject }}</strong>
                             </p>
-                            <p>
-                                {{ value|linebreaksbr }}
-                            </p>
+                            <pre>{{ value.message|linebreaksbr }}</pre>
                         {% endfor %}
                     </p>
                 </li>
--- a/src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/history.html
+++ b/src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/history.html
@@ -1,0 +1,51 @@
+{% extends "pretixcontrol/event/base.html" %}
+{% load i18n %}
+{% load bootstrap3 %}
+{% block title %}{% trans "Send out emails" %}{% endblock %}
+{% block content %}
+    <h1>{% trans "Email history" %}</h1>
+    <div>
+        <ul class="list-group">
+            {% for log in logs %}
+                <li class="list-group-item logentry">
+                    <p class="meta">
+                        <span class="fa fa-clock-o"></span> {{ log.datetime|date:"SHORT_DATETIME_FORMAT" }}
+                        {% if log.user %}
+                            <br/><span class="fa fa-user"></span> {{ log.user.get_full_name }}
+                        {% endif %}
+                        {% if log.display %}
+                            <br/><span class="fa fa-comment-o"></span> {{ log.display }}
+                        {% endif %}
+                    </p>
+                    <p>
+                        <strong>Email address:</strong> {{ log.parsed_data.recipients }}
+                    </p>
+                    <p>
+                        <strong>Subject:</strong>
+                    </p>
+                    <p>
+                        {% for key,value in log.parsed_data.subject.items %}
+                            <p>
+                                {{ key }}: {{ value }}
+                            </p>
+                        {% endfor %}
+                    </p>
+                    <p>
+                        <strong>Email content:</strong>
+                    </p>
+                    <p>
+                        {% for key,value in log.parsed_data.message.items %}
+                            <p>
+                                {{ key }}:
+                            </p>
+                            <p>
+                                {{ value|linebreaksbr }}
+                            </p>
+                        {% endfor %}
+                    </p>
+                </li>
+            {% endfor %}
+        </ul>
+    </div>
+    {% include "pretixcontrol/pagination.html" %}
+{% endblock %}
--- a/src/pretix/plugins/sendmail/views.py
+++ b/src/pretix/plugins/sendmail/views.py
@@ -79,8 +79,8 @@ class SenderView(EventPermissionRequiredMixin, FormView):
                             user=self.request.user,
                             data={
                                 'subject': form.cleaned_data['subject'],
-                                'content': form.cleaned_data['message'],
-                                'recipients': o.email
+                                'message': form.cleaned_data['message'],
+                                'recipient': o.email
                             }
                         )
                 except SendMailException:
@@ -110,7 +110,7 @@ class EmailHistoryView(EventPermissionRequiredMixin, ListView):
     permission = 'can_change_orders'
     model = LogEntry
     context_object_name = 'logs'
-    paginate_by = 15
+    paginate_by = 5
 
     def get_queryset(self):
         qs = LogEntry.objects.filter(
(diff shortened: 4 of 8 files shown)


src/.gitignore
```diff
@@ -8,4 +8,4 @@ dist/
 *.egg-info/
 *.bak
 static/jsi18n/
-
+pretix/plugins/sendmail/migrations/**
```
raphaelm: As commented, we don't need this line (it actually would be harmful, as the database tables wouldn't ever  be created for our users). However, as we don't need the model any more anyways...

src/pretix/plugins/sendmail/signals.py
```diff
@@ -19,5 +19,24 @@ def control_nav_import(sender, request=None, **kwargs):
             }),
             'active': (url.namespace == 'plugins:sendmail' and url.url_name == 'send'),
             'icon': 'envelope',
-        }
+            'children': True,
+            'second_level': [
```
raphaelm: I like that we can do this now! Thanks! :)

Not sure it is necessary here, but let's keep it for now. One minor comment on this, though: There is no reason for two attributes, ``children`` and ``second_level``, that just creates room for errors. Just drop the boolean property and just go with ``'children': […]``. You can still to ``if nav.children`` then, it will still work.

src/pretix/plugins/sendmail/views.py
```diff
@@ -95,3 +120,23 @@ def get_context_data(self, *args, **kwargs):
...
+    model = LogEntry
+
+    def get_queryset(self):
+        qs = LogEntry.objects.filter(
+            event=self.request.event
+        )
+        return qs
+
+    @receiver(signal=logentry_display)
+    def pretixcontrol_logentry_display(sender, logentry, **kwargs):
```
raphaelm: This method should move to ``signals.py`` :)

src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/history.html
```diff
@@ -0,0 +1,30 @@
+{% extends "pretixcontrol/event/base.html" %}
+{% load i18n %}
+{% load bootstrap3 %}
+{% block title %}{% trans "Send out emails" %}{% endblock %}
+{% block content %}
+    <h1>{% trans "Email history" %}</h1>
+    <div>
+        <ul class="list-group">
+            {% for log in request.event.all_logentries %}
+                {% if log.action_type == 'sendmail.email.sent' %}
```
raphaelm: This will work, but it will be slow for large events, as it first needs to load *all* logs from the databases to drop just a few then. The more elegant way to do this is to already do this in the ``get_queryset`` method and just use ``{% for log in logs %}`` here.

src/pretix/plugins/sendmail/views.py
```diff
@@ -95,3 +120,23 @@ def get_context_data(self, *args, **kwargs):
...
+
+
+class EmailHistoryView(EventPermissionRequiredMixin, ListView):
+    template_name = 'pretixplugins/sendmail/history.html'
+    permission = 'can_change_orders'
+    model = LogEntry
+
+    def get_queryset(self):
+        qs = LogEntry.objects.filter(
+            event=self.request.event
```
raphaelm: You can just extend this filter to restrict by type:
```
.filter(
    event=self.request.event,
    action_type='sendmail.email.sent'
)

File: src/pretix/plugins/sendmail/views.py [hunk 6]
```diff
@@ -95,3 +120,23 @@ def get_context_data(self, *args, **kwargs):
         ctx = super().get_context_data(*args, **kwargs)
         ctx['output'] = getattr(self, 'output', None)
         return ctx
+
+
+class EmailHistoryView(EventPermissionRequiredMixin, ListView):
+    template_name = 'pretixplugins/sendmail/history.html'
+    permission = 'can_change_orders'
+    model = LogEntry
```
raphaelm: If you add ``context_object_name = 'logs'`` here, you can just reference the list in the template as ``logs``. If not, it *probably* is ``logentry_list``, but let's better be explicit about it :)

raphaelm:
> Also adds support to second-level nav to plugins

Nice! :+1: 

> As I previously mentioned, I'm confused about event-level and order-level logs so I don't know if I am logging/displaying the right thing (probably not).

Yes, you display the right thing! To clarify things in your head: Many of our models have ``ForeignKey`` fields. This means that they *reference* a different model with that field, e.g. an order references the event it belongs to. These references are bound to a certain model type, so the Order.event field will always reference an event, and never for example a user. This is a concept that you'll find in any relational database. Our ``LogEntry`` model, however, has a ``GenericForeignKey``. This is a more django-specific thing that allows us to have a reference to *any* model type. This has a few drawbacks in terms of safety guarantees and performance, but it comes in very handy here: Think of our ``LogEntry`` as a sheet of paper, that can be attached to *any object*, like a [post-it](https://en.wikipedia.org/wiki/Post-it_note). The post-it also doesn't care what it is attached to.

Every time we change something important, we attach a post-it (or, a ``LogEntry``) containing *who* did *what* and *when* to *what object*. This can be "Raphael marked Order XYZ as paid now" or "Flavia changed the settings of the event yesterday". We would attach the first thing to a specific order object, and the second thing to the event object. 

In this case here, we're doing both: We're attaching one LogEntry to *every single* order that the email has been sent for **and** to the event object containing the raw text. That's what I meant by order-level or event-level. We want to *log* to *both*. On the history page, we only need to display the event-level entries.

> I didn't remove the code for the new Model either. Let me know if it needs to go.

There are things that would be easier *with* the model. For example, showing all orders that received email XYZ would be a very simple lookup while it is rather complicated without the model. On the other hand *dropping* the model would make things a lot easier, as we don't have yet-another logging mechanism but reuse the consistent method that is already there. I think we should drop it. :cry: 

If we need one of the other features later, I'd rather try to find a way to integrate it with ``LogEntry``.

> I'm displaying raw (uggly) data but 1- I wasn't sure I had the right thing and 2- I didn't know how to display it prettier.

Let's clean the model stuff (and my other comments up first) and then talk about making it pretty? :) We'll find a nice way, I'm sure :)

raphaelm:
I took a quick look and it looks good so far (had no time to test it in-depth so far, but will do so soon)!

Now towards making it more pretty: If you look at the ``data`` attribute of our ``LogEntry`` model, you'll see that it contains a JSON-encoded dictionary. We somehow need to access the data in there. To have this easily accessible from inside the template, I suggest adding a new dynamic property to the ``LogEntry`` model. We could use Django's [cached_property](https://docs.djangoproject.com/en/1.10/ref/utils/#django.utils.functional.cached_property) decorator to make sure the JSON only needs to be parsed once:

```
import json
from django.utils.functional import cached_property

class LogEntry:
    ...

    @cached_property
    def parsed_data(self):
        return json.loads(self.data)
```

With that in place, you can use something like ``log.parsed_data.subject`` inside your template :) Does that get you a bit further? :)

FlaviaBastos:
This helped A LOT! Thanks!
Just pushed new code.

raphaelm:
Wow, this looks really pretty 😍 

I just played around with it and found one remaining major problem: If you submit the mail, it still appears a hundred times in the history. To fix this, the ``self.request.event.log_action`` just needs to move to a different place in the code. Do you see why? ;)

The other thing is that as the list gets longer, it becomes hard to use and we should add some pagination, e.g. that after 15 mails the view will split into several pages. Luckily, all the heavy lifting for this is already in place: You just need to add a ``paginate_by = 15`` property to the view (like the ``context_object_name`` thing) and then include a ``{% include "pretixcontrol/pagination.html" %}`` in the template.

FlaviaBastos:
> this looks really pretty

I'm glad you liked! :)

I will take a look at these two fixes later on today. Thanks for the feedback!

FlaviaBastos:
> To fix this, the self.request.event.log_action just needs to move to a different place in the code. Do you see why? ;)

I see that the `self.request.event.log_action( 'sendmail.email.sent'`  on line 80 is logging one entry per address email sent, so 'per order', right? I was displaying that one in history.

Now I guess you really want the `self.request.event.log_action('pretix.plugins.sendmail.sent'` (line 41) displayed in history. You mentioned this a few days ago:

> On the history page, we only need to display the event-level entries.


In this case we won't have the email addresses but I guess this is the intended behaviour. 

So, I can either:
1) move ` 'sendmail.email.sent'` outside the for loop and display that - it will display only one entry but it wouldn't be logging 'per order' (and it would then became redundant code (?))

2) Adjust the history display to show `pretix.plugins.sendmail.sent` instead

So the conclusion I get to is that we want option 2. Am I getting this right?

The pagination thingy was breeze, as you mentioned! :thumbsup:

raphaelm:
Sorry, I didn't look very closely! My bad :( It's just a very minor glitch:

* The call in line 41 to ``self.request.event.log_action`` is fine. This is also the thing we want to display in the history.

* The call in line 80 is wrong. As this is the *order-level* logging, we actually want to attach the log entry to the order, so it should be calling ``o.log_action`` instead.

Also, for consistency, we should change the first parameter of **both** to ``pretix.plugins.sendmail.sent`` (and adjust the filter query in the history view).

FlaviaBastos:
> Sorry, I didn't look very closely! My bad :(

Oh, no problem! I think you do a very god job keeping track of everything. :)

I just pushed new code. This is what I did:
- adjust the history to display the event level log from `self.request.event.log_action`
- moved this call from line 41 to 88. In line 41 it was also logging the preview events. The display was duplicated (or triplicated, etc) depending on how many times you previewed the message before sending. On line 88 it's only logging once, even if you had previewed first.
- adjusted the other `log_action` to log orders instead by using `o.log_action` as you suggested. This is now on line 77
- for consistency I left the event-level call named as it was (`pretix.plugins.sendmail.sent`) and renamed the order-level one to `pretix.plugins.sendmail.order.email.sent` --> this is rather a long name but I wanted to make it really clear what it was. Is it too much?

What do you think of these changes?

src/pretix/plugins/sendmail/signals.py
```diff
@@ -19,5 +20,32 @@ def control_nav_import(sender, request=None, **kwargs):
...
+                },
+            ]
+        },
     ]
+
+
+@receiver(signal=logentry_display)
+def pretixcontrol_logentry_display(sender, logentry, **kwargs):
+    plains = {
+        'pretix.plugins.sendmail.sent': _('Email was sent'),
```
raphaelm: We now need to handle ``pretix.plugins.sendmail.order.email.sent`` here as well in some way, otherwise the log of an order will display "pretix.plugins.sendmail.order.email.sent" directly which does not look nice. Something like "Order received a mass email"?

src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/history.html
```diff
@@ -0,0 +1,51 @@
...
+                        <span class="fa fa-clock-o"></span> {{ log.datetime|date:"SHORT_DATETIME_FORMAT" }}
+                        {% if log.user %}
+                            <br/><span class="fa fa-user"></span> {{ log.user.get_full_name }}
+                        {% endif %}
+                        {% if log.display %}
+                            <br/><span class="fa fa-comment-o"></span> {{ log.display }}
+                        {% endif %}
+                    </p>
+                    <p>
+                        <strong>Email address:</strong> {{ log.parsed_data.recipients }}
```
raphaelm: This is kinda pointless now when we display event-level logs. However, it would be nice to display the recipients in some other way, e.g. "Recipients: paid, expired and pending orders". This data is already in ``log.parsed_data``. Do you find a way to display it? If not, no worry, this might not be obvious and I can think of something if you need help.

FlaviaBastos: Darn! You are absolutely right! :laughing: I meant to remove it but.... *sigh*

I looked around and `log.parsed_data` outputs an array of a one letter order status (in this case the recipients), such as `['p', 'c']` and I don't know of a way to convert these to the full order status as 'paid', 'canceled', etc in a decent way. 

I wrote a temporary code to deal with this in the template and display the full order status but it has two big problems: 1- it doesn't support translations and 2- if eventually one added a new order status one would have to fix this part too... :confused: I guess these things are not supposed to be dealt with in the template first place, right?
The code I wrote looks like this:
```
{% for status in log.parsed_data.sendto %}
    {% if status == 'n' %}pending orders,{% endif %}
    {% if status == 'p' %}paid orders,{% endif %}
    {% if status == 'e' %}expired orders,{% endif %}
    {% if status == 'c' %}canceled orders,{% endif %}
    {% if status == 'r' %}refunded orders,{% endif %}
{% endfor %}
```
So, I see three options:
1) remove the recipients altogether
2) use the single letters (this might not be helpful for users...)
3) use the chunk I wrote - well, unless you think of another way to process this

What do you think?

FlaviaBastos:
I just pushed code that adds the custom message to signals and removed the recipients from the history.
I am going offline for some time and I wanted to leave you with an option if you wanted to move this ahead.

If you decide you would like to try to expand the recipients thing, I'm ok with that too (I think it would be nice to have!) but I won't be able to follow up until sometime in January. 

Happy holidays! :)

raphaelm:
Hey there! :) I'll try to come up with something nice for the recipients, I think your try goes in a good direction. Happy holidays for now! :christmas_tree:

raphaelm:
Hi Flavia,

happy new year! :fireworks: Hope you're having a good time with some distance from the internet.

I just brought this up to date with pretix' current master and hacked around for a bit until it looked nice :) Thanks for all your work here – it is done now! :tada: 

Cheers
Raphael

FlaviaBastos:
Hey, Raphael! 
Happy new year to you too! :fireworks: 
I'm still enjoying my read-only mode :D I hope you have had some rest time also. :)

Thank you for picking this up and finishing it! I'm sure it's looking fantastic! :thumbsup:  I will make a note to go back and get acquainted with the way you completed this issue so I learn for next time. ;-)

PR #354, opened 2016-12-11
Display log entry for changed order secrets

rixx:


src/pretix/control/views/orders.py
```diff
@@ -531,17 +531,17 @@ def form(self):
     def post(self, *args, **kwargs):
         old_email = self.order.email
         if self.form.is_valid():
-            self.order.log_action('pretix.event.order.contact.changed', {
+            self.order.log_action('pretix.event.order.contact.changed', user=self.request.user, {
```
raphaelm: Travis reports a syntax error here: ``non-keyword arg after keyword arg`` :)

rixx: doobdoob, fixed

PR #362, opened 2016-12-17
Order Bankimports by their job's date

rixx:
Leads to having the most recent unresolved transfers on top instead of
in between.


raphaelm:
Uh, thanks. This was annoying.

PR #363, opened 2016-12-17
Added specific error message for no file uploaded for empty bank import.

brandon-pineda:
Now if you try to "Start upload" on Import bank data page without choosing any file, the error message reads "You must choose a file to import."


raphaelm:
Perfect, thanks a lot! :)

PR #371, opened 2017-01-07
Fixed Module not found: pretix.celery_app error

Nerade:
While setting up the pretix background worker daemon I faced an python NameError which was caused by a missing pretix.celery_app module. Looking into source folder I noticed a celery.py module. I updated the parameter of celery to point to pretix.celery and it worked as expected.


raphaelm:
Hey :) How did you install pretix, by ``pip install …``? You did just probably run into a "release race condition": The current module published on pip ist ``celery``, however it was renamed to ``celery_app`` very recently, so it will be ``celery_app`` in the next release -- and the documentation is already for the next one. I'll probably publish a new package to pip later today

raphaelm:
I just uploaded 1.0.0b2 to PyPI. Could you do a quick ``pip install --upgrade pretix`` and see if it now works with ``celery_app``?

Nerade:
I did not used pip to install/download pretix. But this could be a reason. Thanks for your fast reply.

PR #373, opened 2017-01-11
Added validation for voucher generation field

jfwiebe:
Pressing "Generate random vouchers" will result in a malformed AJAX request and an undefined error message.
image
I've added a check that the input field is not empty and changed the input field to `type=number`.

Please give it a short test before merging.

### DIFF
--- a/src/pretix/static/pretixcontrol/js/ui/main.js
+++ b/src/pretix/static/pretixcontrol/js/ui/main.js
@@ -104,8 +104,11 @@ $(function () {
                 $("#id_codes").val(data.codes.join("\n"));
             });
         } else {
-            alert("Please enter a number!");
+            $(".form-group:has(#voucher-bulk-codes-num)").addClass("has-error");
             $("#voucher-bulk-codes-num").focus();
+            setTimeout(function() {
+                $(".form-group:has(#voucher-bulk-codes-num)").removeClass("has-error");
+            }, 3000);
         }
     });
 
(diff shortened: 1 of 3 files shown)


src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -96,12 +96,17 @@ $(function () {
...
-        });
+        var num = $("#voucher-bulk-codes-num").val();
+        if (num != "") {
+            var url = $(this).attr("data-rng-url");
+            $("#id_codes").html("Generating...");
+            $.getJSON(url + '?num=' + num, function (data) {
+                $("#id_codes").val(data.codes.join("\n"));
+            });
+        } else {
+            alert("Please enter a number!");
```
rixx: Hm, an alert? Wouldn't a message next to the field be less intrusive?

jfwiebe:
Yeha it would. Currently it gives an alert so I wanted to clarify the message. Maybe also using Bootstrap Validation States (https://getbootstrap.com/css/#forms-control-validation) would look great.

jfwiebe:
@rixx what about this? or too lazy?

raphaelm:
I like it, I think it is clear enough.
One final note: Could you add another ``$(".form-group:has(#voucher-bulk-codes-num)").removeClass("has-error");`` into the first part of the ``if`` block? So that it does not stay red if someone submits something valid :)

rixx:
Looks good to me :)

raphaelm:
Top! Thank you very much!

PR #377, opened 2017-01-17
Allow superusers to view any event

rixx:


src/pretix/control/middleware.py
```diff
@@ -61,6 +61,16 @@ def process_request(self, request):
             return redirect_to_login(
                 path, resolved_login_url, REDIRECT_FIELD_NAME)
 
+        if request.user.is_superuser:
```
raphaelm: I don't think this is in the right place, it uses ``url.kwargs['event']`` but is **before** the ``if 'event' in url.kwargs…`` checks. Do I miss something?

PR #385, opened 2017-01-19
Changed pagination of organizer page

jfwiebe:
...to show more than one event per page. Maybe this value should become a configurable value in the future.

image


raphaelm:
Oops, that 1 is probably a debugging artifact. I think even 30 would be good, as the rows are very narrow.

jfwiebe:
I agree with you, thats right.

raphaelm:
Awesome, thank you! :)

PR #394, opened 2017-01-26
Correct wrong module name for Postgres in example configs

lutoma:
The example configs list `psycopg2` as the name for the Django postgres database module, but it's actually `postgresql_psycopg2`: https://github.com/django/django/tree/master/django/db/backends/postgresql_psycopg2


raphaelm:
Guilty as charged, I'm a MariaDB user most of the time... Thanks!

Issue #409, opened 2017-02-11
Sendmail plugin: false success message


FlaviaBastos:
In the sendmail plugin, if an order status that has 0 orders in that condition is selected, previewing that message will display the success 'Your message has been queued... ". It also logs that message as send in the email history.

Example:
I have no "expired" orders in my system. In the sendmail plugin I select "expired" in the "send to" field and fill in a subject and message. Clicking on preview will display the success message that the email will be sent.
As far as I could check, the email is NOT sent; it's however logged as sent...

PR #410, opened 2017-02-11
Fix false success message on sendmail plugin

FlaviaBastos:
This fixes issue #409

Add check to confirm if there are no orders with that status. In this case, display error message and keep form populated. Doesn't log any info.

### DIFF
--- a/src/pretix/plugins/sendmail/views.py
+++ b/src/pretix/plugins/sendmail/views.py
@@ -42,56 +42,60 @@ class SenderView(EventPermissionRequiredMixin, FormView):
 
         failures = []
         self.output = {}
-        for o in orders:
-            if self.request.POST.get("action") == "preview":
-                for l in self.request.event.settings.locales:
-                    with language(l):
-                        self.output[l] = []
-                        self.output[l].append(_('Subject: {subject}').format(subject=form.cleaned_data['subject'].localize(l)))
-                        message = form.cleaned_data['message'].localize(l)
-                        preview_text = message.format(
-                            order='ORDER1234',
-                            event=self.request.event.name,
-                            order_date=date_format(now(), 'SHORT_DATE_FORMAT'),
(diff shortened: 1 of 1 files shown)


src/pretix/plugins/sendmail/views.py
```diff
@@ -42,56 +42,60 @@ def form_valid(self, form):
...
-                    failures.append(o.email)
-        self.request.event.log_action('pretix.plugins.sendmail.sent',
-                                      user=self.request.user,
-                                      data=dict(form.cleaned_data))
-        if failures:
-            messages.error(self.request, _('Failed to send mails to the following users: {}'.format(' '.join(failures))))
+        if not orders:
+            messages.error(self.request, _('There are no orders matching this selection.'))
+            return self.get(self.request, *self.args, **self.kwargs)
         else:
```
raphaelm: To make the lines shorter, you can just drop this else statement and move everything below one tab to the left again. As the ``if not orders:`` block calls a ``return``, this will stop execution anyway.

FlaviaBastos: :+1:

FlaviaBastos:
> would not have thought of this

Me either! I found this by chance as I was trying to add tests to this plugin (and failing miserably... :unamused:).

I just changed the code as you suggested. I also noticed that if you have "mixed" order status checked (say a valid 'pending' and an non-existent 'expired'), it sends the mail and log as usual. How do you feel about this?

raphaelm:
Nice, merged! *high-five*

> Me either! I found this by chance as I was trying to add tests to this plugin (and failing miserably... 😒).

Oh! If you want to keep trying that, let me know if I can help :)

> I just changed the code as you suggested. I also noticed that if you have "mixed" order status checked (say a valid 'pending' and an non-existent 'expired'), it sends the mail and log as usual. How do you feel about this?

I think this is fine. The mail has been sent, and it has been sent to all users of that categories, even if one of them was empty. No reason to disturb the user.

Issue #389, opened 2017-01-20
Offer ical link/export for events


rixx:
Many events offer ical links - generating one from an event data automatically would be nice.

raphaelm:
Is best done after #381

jlwt90:
Hello, can i take this enhancement ticket?

I am planning to add the ical link to the event page.
http://localhost:8000/bigevents/xxxxxxx/

The link will dynamically generate an .ics file for download, e.g. 
http://www.pratie.com/lab/icalendar/

does it look good?

raphaelm:
Sure, go ahead! :) I believe the [icalendar](https://pypi.python.org/pypi/icalendar) package might be helpful to generate the file.

jlwt90:
@raphaelm sorry for the late. I am pretty new to django. Could you please give me some comments on the changes? 
https://github.com/jlwt90/pretix/commit/16bd3ccc3f8bb8115de02797e3039d0b39d5f5ef

I added a link on event index page which sends a GET request to url <organizer>/<event>/download . The EventDownload view will then create a CacheFile using icalendar library and return a FileResponse object. 

Thanks! :)

raphaelm:
Hi @jlwt90 no problem at all! Next time just create a pull request, that makes it easier to leave comments and discuss the changes. Overall, this looks very good! Thank you! I have two comments to beginn with:

* Could you name it ``EventIcalDownload`` instead of ``EventDownload`` and the URL ``/ical`` instead of ``/download``? Because you are downloading an ical, not an event, so this would be clearer :)
* You don't need to create a ``CachedFile``  here, that is only unneccesary work. Just drop that and directly output the ical :)

If any of this is unclear, feel free to ask!

PR #413, opened 2017-02-16
added event ical download feature

jlwt90:
Hello,

Here is the PR for #389 
Please feel free to leave me some comments!

### DIFF
--- a/src/pretix/presale/views/event.py
+++ b/src/pretix/presale/views/event.py
@@ -9,12 +9,14 @@ from django.db.models import Count, Prefetch, Q
 from django.http import Http404, HttpResponse
 from django.shortcuts import redirect
 from django.utils.decorators import method_decorator
+from django.utils.functional import cached_property
 from django.utils.timezone import now
 from django.utils.translation import ugettext_lazy as _
 from django.views import View
 from django.views.decorators.csrf import csrf_exempt
 from django.views.generic import TemplateView
 from icalendar import Calendar, Event
+from pytz import timezone
 
 from pretix.base.models import ItemVariation
 from pretix.multidomain.urlreverse import eventreverse
@@ -110,21 +112,36 @@ class EventIndex(EventViewMixin, CartMixin, TemplateView):
 
 
 class EventIcalDownload(EventViewMixin, View):
+
+    @cached_property
+    def event_timezone(self):
+        return timezone(self.request.event.settings.timezone)
+
     def get(self, request, *args, **kwargs):
         if not self.request.event:
             raise Http404(_('Unknown event code or not authorized to access this event.'))
 
         cal = Calendar()
         cal.add('version', '2.0')
-        cal.add('prodid', '-//{}//{}//'.format(self.request.event.organizer.name, self.request.event.name))
+        cal.add('prodid', '-//pretix//{}//'.format(settings.PRETIX_INSTANCE_NAME))
 
         event = Event()
-        event.add('summary', self.request.event.name)
-        event.add('dtstart', self.request.event.date_from)
-        event.add('dtend', self.request.event.date_to)
+        event.add('summary', str(self.request.event.name))
         event.add('dtstamp', datetime.now(pytz.utc))
-        event.add('location', self.request.event.location)
+        event.add('location', str(self.request.event.location))
         event.add('organizer', self.request.event.organizer.name)
+
+        if self.request.event.settings.show_times:
+            event.add('dtstart', self.request.event.date_from.replace(tzinfo=self.event_timezone))
+        else:
+            event.add('dtstart', self.request.event.date_from.date())
+
+        if self.request.event.settings.show_date_to:
+            if self.request.event.settings.show_times:
+                event.add('dtend', self.request.event.date_to.replace(tzinfo=self.event_timezone))
+            else:
+                event.add('dtend', self.request.event.date_to.date())
+
         cal.add_component(event)
 
         resp = HttpResponse(cal.to_ical(), content_type='text/calendar')
--- a/src/pretix/presale/views/event.py
+++ b/src/pretix/presale/views/event.py
@@ -145,7 +145,7 @@ class EventIcalDownload(EventViewMixin, View):
         cal.add_component(event)
 
         resp = HttpResponse(cal.to_ical(), content_type='text/calendar')
-        resp['Content-Disposition'] = 'attachment; filename="{}"'.format('pretixevent.ics')
+        resp['Content-Disposition'] = 'attachment; filename="pretixevent.ics"'
         return resp
 
 
--- a/src/pretix/presale/views/event.py
+++ b/src/pretix/presale/views/event.py
@@ -3,6 +3,7 @@ from datetime import datetime
 from importlib import import_module
 
 import pytz
+import vobject
 from django.conf import settings
 from django.core.exceptions import PermissionDenied
 from django.db.models import Count, Prefetch, Q
@@ -15,7 +16,6 @@ from django.utils.translation import ugettext_lazy as _
 from django.views import View
 from django.views.decorators.csrf import csrf_exempt
 from django.views.generic import TemplateView
-from icalendar import Calendar, Event
 from pytz import timezone
 
 from pretix.base.models import ItemVariation
@@ -121,31 +121,31 @@ class EventIcalDownload(EventViewMixin, View):
         if not self.request.event:
             raise Http404(_('Unknown event code or not authorized to access this event.'))
 
-        cal = Calendar()
-        cal.add('version', '2.0')
-        cal.add('prodid', '-//pretix//{}//'.format(settings.PRETIX_INSTANCE_NAME))
+        event = self.request.event
+        cal = vobject.iCalendar()
+        cal.add('prodid').value = '-//pretix//{}//'.format(settings.PRETIX_INSTANCE_NAME)
 
-        event = Event()
-        event.add('summary', str(self.request.event.name))
-        event.add('dtstamp', datetime.now(pytz.utc))
-        event.add('location', str(self.request.event.location))
-        event.add('organizer', self.request.event.organizer.name)
+        vevent = cal.add('vevent')
+        vevent.add('summary').value = str(event.name)
+        vevent.add('dtstamp').value = datetime.now(pytz.utc)
+        vevent.add('location').value = str(event.location)
+        vevent.add('organizer').value = event.organizer.name
 
         if self.request.event.settings.show_times:
-            event.add('dtstart', self.request.event.date_from.replace(tzinfo=self.event_timezone))
+            vevent.add('dtstart').value = event.date_from.astimezone(self.event_timezone)
         else:
-            event.add('dtstart', self.request.event.date_from.date())
+            vevent.add('dtstart').value = event.date_from.astimezone(self.event_timezone).date()
 
         if self.request.event.settings.show_date_to:
             if self.request.event.settings.show_times:
-                event.add('dtend', self.request.event.date_to.replace(tzinfo=self.event_timezone))
+                vevent.add('dtend').value = event.date_to.astimezone(self.event_timezone)
             else:
-                event.add('dtend', self.request.event.date_to.date())
+                vevent.add('dtend').value = event.date_to.astimezone(self.event_timezone).date()
 
-        cal.add_component(event)
-
-        resp = HttpResponse(cal.to_ical(), content_type='text/calendar')
-        resp['Content-Disposition'] = 'attachment; filename="pretixevent.ics"'
+        resp = HttpResponse(cal.serialize(), content_type='text/calendar')
+        resp['Content-Disposition'] = 'attachment; filename="{}-{}.ics"'.format(
+            self.request.organizer.slug, event.slug
+        )
         return resp
 
 
--- a/src/pretix/presale/views/event.py
+++ b/src/pretix/presale/views/event.py
@@ -122,29 +122,33 @@ class EventIcalDownload(EventViewMixin, View):
             raise Http404(_('Unknown event code or not authorized to access this event.'))
 
         event = self.request.event
+        creation_time = datetime.now(pytz.utc)
         cal = vobject.iCalendar()
         cal.add('prodid').value = '-//pretix//{}//'.format(settings.PRETIX_INSTANCE_NAME)
 
         vevent = cal.add('vevent')
         vevent.add('summary').value = str(event.name)
-        vevent.add('dtstamp').value = datetime.now(pytz.utc)
+        vevent.add('dtstamp').value = creation_time
         vevent.add('location').value = str(event.location)
         vevent.add('organizer').value = event.organizer.name
+        vevent.add('uid').value = '{}-{}-{}'.format(
+            event.organizer.slug, event.slug, creation_time.strftime('%Y%m%d%H%M%S%f')
+        )
 
-        if self.request.event.settings.show_times:
+        if event.settings.show_times:
             vevent.add('dtstart').value = event.date_from.astimezone(self.event_timezone)
         else:
             vevent.add('dtstart').value = event.date_from.astimezone(self.event_timezone).date()
 
-        if self.request.event.settings.show_date_to:
-            if self.request.event.settings.show_times:
+        if event.settings.show_date_to:
+            if event.settings.show_times:
                 vevent.add('dtend').value = event.date_to.astimezone(self.event_timezone)
             else:
                 vevent.add('dtend').value = event.date_to.astimezone(self.event_timezone).date()
 
         resp = HttpResponse(cal.serialize(), content_type='text/calendar')
         resp['Content-Disposition'] = 'attachment; filename="{}-{}.ics"'.format(
-            self.request.organizer.slug, event.slug
+            event.organizer.slug, event.slug
         )
         return resp
 
--- a/src/pretix/presale/views/event.py
+++ b/src/pretix/presale/views/event.py
@@ -1,9 +1,12 @@
 import sys
+from datetime import datetime
 from importlib import import_module
 
+import pytz
 from django.conf import settings
 from django.core.exceptions import PermissionDenied
 from django.db.models import Count, Prefetch, Q
+from django.http import Http404, HttpResponse
 from django.shortcuts import redirect
 from django.utils.decorators import method_decorator
 from django.utils.timezone import now
@@ -11,6 +14,7 @@ from django.utils.translation import ugettext_lazy as _
 from django.views import View
 from django.views.decorators.csrf import csrf_exempt
 from django.views.generic import TemplateView
+from icalendar import Calendar, Event
 
 from pretix.base.models import ItemVariation
 from pretix.multidomain.urlreverse import eventreverse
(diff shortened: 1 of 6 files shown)


src/pretix/presale/views/event.py
```diff
@@ -104,6 +108,29 @@ def get_context_data(self, **kwargs):
...
 
 
+class EventIcalDownload(EventViewMixin, View):
+    def get(self, request, *args, **kwargs):
+        if not self.request.event:
+            raise Http404(_('Unknown event code or not authorized to access this event.'))
+
+        cal = Calendar()
+        cal.add('version', '2.0')
+        cal.add('prodid', '-//{}//{}//'.format(self.request.event.organizer.name, self.request.event.name))
```
raphaelm: The prodid should be specific to the software, e.g. ``-//pretix.eu//pretix//EN``

src/pretix/presale/views/event.py
```diff
@@ -104,6 +108,29 @@ def get_context_data(self, **kwargs):
...
+
+        event = Event()
+        event.add('summary', self.request.event.name)
+        event.add('dtstart', self.request.event.date_from)
```
raphaelm: This will output the date in the wrong format, you will need to call ``strftime`` to bring it in the correct format. If you have trouble doing that, I can look up the correct arguments for you, it might be non-trivial with regard to timezones ;) This of course is valid for all date properties.

The other thing is that there are two settings that we should respect here:

* ``self.request.event.settings.show_times`` -- If this is ``False``, we should not show the time of day, so the icalendar should contain a full-day event. Can you look up how to do this?
* ``self.request.event.settings.show_date_to`` -- If this is ``False`` we should only show the beginning date. As the icalendar needs a end date, I do not see a good way to respect that currently and I think we can ignore it here.

rixx: In the second case, you might want to make it just a full-day event for the `date_from` value.

src/pretix/presale/views/event.py
```diff
@@ -104,6 +108,29 @@ def get_context_data(self, **kwargs):
...
+        event.add('dtend', self.request.event.date_to)
+        event.add('dtstamp', datetime.now(pytz.utc))
+        event.add('location', self.request.event.location)
```
rixx: As `location` is a `I18nCharField`, you will probably want to wrap it in a `str()` call to make sure the correct value lands in the calendar, independent of the calendar library's value handling.

(`I18nCharFields` are well-behaved, but it's probably better to make sure.)

src/pretix/presale/views/event.py
raphaelm: > In the second case, you might want to make it just a full-day event for the date_from value.

...which would lose the information of the time which might be very valuable, so I'm unsure here.

src/pretix/presale/views/event.py
jlwt90: @rixx thanks for your explanation. I think I should also warp `event.name`(summary) too coz it is also `I18nCharField`.

src/pretix/presale/views/event.py
jlwt90: @raphaelm should I get the value from settings, i.e. `PRETIX_INSTANCE_NAME`, `TIME_ZONE` in pretix/settings.py?

raphaelm: Yes, we could use the instance name for the prodid, e.g.

```

...

        cal.add('prodid', '-//pretix//{}//'.format(settings.PRETIX_INSTANCE_NAME))
 ```

For the timezone, we need to use ``self.request.event.settings.timezone`` as it can be different for different events.

File: src/pretix/presale/views/event.py (hunk 3 above)
Comment (raphaelm) [inline src/pretix/presale/views/event.py]: Good catch :)

File: src/pretix/presale/views/event.py (hunk 1 above)
Comment (jlwt90) [inline src/pretix/presale/views/event.py]: Sorry for replying late... I was quite busy recently. 
Regarding `prodid`, I included LANGUAGE_CODE as the last parameter. What do you think?
```
cal.add('prodid', '-//pretix//{}//{}'.format(settings.PRETIX_INSTANCE_NAME, 
                                                                    settings.LANGUAGE_CODE.upper()))
```

File: src/pretix/presale/views/event.py (hunk 2 above)
Comment (jlwt90) [inline src/pretix/presale/views/event.py]: According to icalendar document, I only need to provide datetime object for icalendar to generate the correct timestamp.[http://icalendar.readthedocs.io/en/latest/usage.html#components](doc)

However, you also mentioned timezone. My bad :( The original version is wrong!
Now I fixed it and it will output
```
DTSTART;TZID=Asia/Taipei;VALUE=DATE-TIME:20170220T073456
```
which is the same as the icalendar spec with timezone specified if it isn't UTC. (It's handled by `icalendar` automatically)
[https://icalendar.org/iCalendar-RFC-5545/3-6-1-event-component.html](DTSTART)
[https://icalendar.org/iCalendar-RFC-5545/3-2-19-time-zone-identifier.html](TZID)

I would like to commit my changes after confirming the behaviours of the event settings.
For `self.request.event.settings.show_times`, it's easy since `icalendar` supports date object too!

For `self.request.event.settings.show_date_to`, it's a little bit complicated.
I went through the specs of icalendar.
[https://icalendar.org/iCalendar-RFC-5545/3-6-1-event-component.html](DTSTART/DTEND)

**Case 1)**
If we provide datetime `DTSTART` without `DTEND`, end date will be the same as the start date. 
> For cases where a "VEVENT" calendar component specifies a "DTSTART" property with a DATE-TIME value type but no "DTEND" property, the event ends on the same calendar date and time of day specified by the "DTSTART" property.

So if `self.request.event.settings.show_times is True and self.request.event.settings.show_date_to is False`,
I guess we can just provide DTSTART so that the time can be kept.

**Case 2)**
If we provide date `DTSTART` without `DTEND`, the event will become an one-day event.
> For cases where a "VEVENT" calendar component specifies a "DTSTART" property with a DATE value type but no "DTEND" nor "DURATION" property, the event's duration is taken to be one day.

If `self.request.event.settings.show_times is False and self.request.event.settings.show_date_to is False`,
we will set `DTSTART` date and the event lasts for one day.

**Case 3)**
If `self.request.event.settings.show_times is False and self.request.event.settings.show_date_to is True`, we will ignore timezone and set `date` in both `DTSTART` and `DTEND`.

Does it sound good? Thanks for your comments :)

Comment (raphaelm) [inline src/pretix/presale/views/event.py]: Hey,

thank you for your patience. I see that I haven't thought about some of my comments thoroughly enough or haven't researched them properly, sorry for that. I'll take the time to review this in detail tomorrow or Wednesday.

File: src/pretix/presale/views/event.py (hunk 1 above)
Comment (raphaelm) [inline src/pretix/presale/views/event.py]: LANGUAGE_CODE is only the default setting of the installation and says nothing about the event's language(s). I can't find anything in the ical spec that says the PRODID should include the language, however most programs apparently do it. I'm not sure here, rather drop it.

File: src/pretix/presale/views/event.py (hunk 2 above)
Comment (raphaelm) [inline src/pretix/presale/views/event.py]: > However, you also mentioned timezone. My bad :( The original version is wrong!
> Now I fixed it and it will output

You did not upload that fix yet, did you? Timezones are always nasty :(

> I would like to commit my changes after confirming the behaviours of the event settings.
> For `self.request.event.settings.show_times`, it's easy since `icalendar` supports date object too!

Nice! Can you confirm that it then correctly outputs a full-day (or multiple-day) event?

> So if `self.request.event.settings.show_times is True and self.request.event.settings.show_date_to is False`,
I guess we can just provide DTSTART so that the time can be kept.

Yes, this seems like the correct thing, at least when looking at the spec. I have my doubts on how calendar applications out there will handle this, but let's go for this version for now.

> If `self.request.event.settings.show_times is False and self.request.event.settings.show_date_to is False`,
we will set `DTSTART` date and the event lasts for one day.

Seems about right.

> If `self.request.event.settings.show_times is False and self.request.event.settings.show_date_to is True`, we will ignore timezone and set `date` in both `DTSTART` and `DTEND`.

Yep.

> Does it sound good?

Yes it does! :) Just go ahead and tell me if you need any more help.

File: src/pretix/presale/views/event.py (hunk 1 above)
Comment (jlwt90) [inline src/pretix/presale/views/event.py]: this is the ical spec.
[Product Identifier Spec](https://icalendar.org/iCalendar-RFC-5545/3-7-3-product-identifier.html)

>This property specifies the identifier for the product that created the iCalendar object.

I think `prodid` should contain system information instead of event information? I couldn't find any restriction on each parameter of `prodid` in the spec., so we can drop it at this moment i guess? It's just the identifier of the ical object anyway.

Comment (raphaelm) [inline src/pretix/presale/views/event.py]: Yes, let's just drop it. The language is not a relevant piece of system information.

File: src/pretix/presale/views/event.py (hunk 2 above)
Comment (jlwt90) [inline src/pretix/presale/views/event.py]: I set up 4 different events for testing the above feature.
All events start at 2017-02-24 00:00:00 (Asia/Singapore, GMT+8) and end at 2017-02-26 00:00:00 (Asia/Singapore, GMT+8).

Since I am living in Japan, the calendar is using GMT+9.
I also attached the .ics files in case you want to import it to somewhere else.

**2-DAY-EVENT:** normal event which lasts for 2 days
```
DTSTART;TZID=Asia/Singapore;VALUE=DATE-TIME:20170224T000000
DTEND;TZID=Asia/Singapore;VALUE=DATE-TIME:20170226T000000
```
2-day-event
**2-DAY-EVENT-NO-TIME:** same event but time is not shown ---> full day event
```
DTSTART;VALUE=DATE:20170224
DTEND;VALUE=DATE:20170226
```
2-day-event-no-time
**2-DAY-EVENT-NO-ENDATE:** same event but end date is not shown ---> event with same start & end time
```
DTSTART;TZID=Asia/Singapore;VALUE=DATE-TIME:20170224T000000
```
2-day-event-no-enddate
**2-DAY-EVENT-NO-ENDATE-AND-TIME:** same event but time & end date are not shown 
```
DTSTART;VALUE=DATE:20170224
```
2-day-event-no-enddate-and-time


**Downloads:**
[event-ics.zip](https://github.com/pretix/pretix/files/794216/event-ics.zip)

Please let me know if you find any problem :)
I added a new commit for the above changes!

Comment (raphaelm) [inline src/pretix/presale/views/event.py]: Nice, this is looking very good! Thank you very much. As you noticed, we need to deal with a number of cases here, so some tests would be very good, as you suggested. If you need any help doing that, just ask!

**Comment by jlwt90** [comment]
Hello, I'll fix the tests first... please ignore the current version.  :(

**Comment by raphaelm** [comment]
Hereby officially ignored. Let me know if I can help :wink:

**Comment by jlwt90** [comment]
@raphaelm I think I need some helps. 
I tried to clone master branch and ran tests on my local machine.
I found lots of failed tests. I am sure that the problem is caused by locale. 

For example,
```
________________ LocaleDeterminationTest.test_event_not_allowed ________________

self = <tests.base.test_middleware.LocaleDeterminationTest testMethod=test_event_not_allowed>

    def test_event_not_allowed(self):
        self.event.settings.set('locales', ['en'])
        c = Client()
        cookies = c.cookies
        cookies[settings.LANGUAGE_COOKIE_NAME] = 'de'
        response = c.get('/dummy/dummy/')
        language = response['Content-Language']
>       self.assertEqual(language, 'en')
E       AssertionError: 'de' != 'en'
E       - de
E       + en
tests/base/test_middleware.py:108: AssertionError
___________________________ EventsTest.test_plugins ____________________________

self = <tests.control.test_events.EventsTest testMethod=test_plugins>

    def test_plugins(self):
        doc = self.get_doc('/control/event/%s/%s/settings/plugins' % (self.orga1.slug, self.event1.slug))
        self.assertIn("PayPal", doc.select(".form-plugins")[0].text)
>       self.assertIn("Enable", doc.select("[name=plugin:pretix.plugins.paypal]")[0].text)
E       AssertionError: 'Enable' not found in 'Aktivieren'
```

Did I forget to run something before test? Thanks!
[output.txt](https://github.com/pretix/pretix/files/813184/output.txt)

**Comment by jlwt90** [comment]
@raphaelm sorry, i think i messed up the local environment and it works now! I am going to fix my failed test cases now.

**Comment by jlwt90** [comment]
@raphaelm could you check my pull request? test cases are also ready now :D

**Comment by raphaelm** [comment]
Hi,

this is looking very good so far. I played around with it a bit and have two more comments:

* You somehow messed up your git branch, I'm not quite sure how. If you look at the commits that GitHub shows for this PR, they are a lot more than yours. If you want to, you can try rebasing your changes against the current master branches. If you don't, just don't, I'll fix it while merging, no worries.

* I'm not yet sure whether the timezone-things are correct, but maybe I'm just confused. Could you look at that again? I have an event that starts at 2016-12-26 10:00:00, Berlin time. The downloaded ICS says ``DTSTART;TZID=Europe/Berlin;VALUE=DATE-TIME:20161226T090000``, so it specified Europe/Berlin as a timezone correctly and then gives an UTC value for DATE-TIME. As I [read the spec](http://www.kanzaki.com/docs/ical/dateTime.html) it should contain local time there, e.g. ``DTSTART;TZID=Europe/Berlin;VALUE=DATE-TIME:20161226T090000``. I believe the ``replace(tzinfo...)`` call is wrong in that regard and the correct call would be ``self.request.event.date_from.astimezone(self.event_timezone)``. Of course, you also need a ``astimezone(self.event_timezone)`` in the test then. Feel free to correct me if I'm wrong!

Cheers
Raphael

**Comment by raphaelm** [comment]
Ah, just another thing: I think it would look nicer if the downloaded file is not named ``pretixevent.ics`` but something more unique, e.g.:

```
        resp['Content-Disposition'] = 'attachment; filename="{}-{}.ics"'.format(
            self.request.organizer.slug, self.request.event.slug
        )
```

**Comment by jlwt90** [comment]
> You somehow messed up your git branch, I'm not quite sure how. If you look at the commits that GitHub shows for this PR, they are a lot more than yours. If you want to, you can try rebasing your changes against the current master branches. If you don't, just don't, I'll fix it while merging, no worries.

I think I unintentionally did a git merge (not sure how as well)... Anyway, I think I just fixed it.

> I'm not yet sure whether the timezone-things are correct, but maybe I'm just confused. Could you look at that again?

Thank you for checking! You are absolutely correct.
My bad :( I just realised that I made a mistake when I showed you the result last week.

> I set up 4 different events for testing the above feature.
All events start at 2017-02-24 00:00:00 (Asia/Singapore, GMT+8) and end at 2017-02-26 00:00:00 (Asia/Singapore, GMT+8).

The actual start time should be **2017-02-24 08:00:00**, not 00:00:00.... I should have picked a better time. I'll fix it tomorrow!

**Comment by jlwt90** [comment]
@raphaelm Sorry I would like to confirm one thing before I commit my changes. 
In event settings page, are **event start time** and **event end time** local time (not utc time)?
I created the following event with ``Asia/Tokyo`` timezone and realised that the time stored in database is 08:57:52 (GMT+9), which means they are both local time.
screen shot 2017-03-05 at 9 07 15 pm

Regarding the timezone issue we discussed before, I went through the specs again.
[datetime format](http://www.kanzaki.com/docs/ical/dateTime.html)
[VTIMEZONE](https://tools.ietf.org/html/rfc5545)

From the 2nd link,
> An individual "VTIMEZONE" calendar component MUST be specified for each unique "TZID" parameter value specified in the iCalendar object.
The parameter MUST be specified on properties with a DATE-TIME value if the DATE-TIME is not either a UTC or a "floating" time. Failure to include and follow VTIMEZONE definitions in iCalendar objects may lead to inconsistent understanding of the local time at any given location.

It says VTIMEZONE must be exist if TZID is used in the file. The current datetime form only works if the calendar app has exactly the same timezone definition.
However, it's pretty hard to include VTIMEZONE in our ical file since I can't find any efficient way to create that VTIMEZONE object from pytz timezone (please correct me if I am wrong ;) ).

In this situation, I think we should just go for UTC time, i.e. form 2 in the 1st link.
What do you think?

**Comment by raphaelm** [comment]
> In event settings page, are event start time and event end time local time (not utc time)?

The rule of thumb in pretix (and Django) is: The database contains **always** UTC, the user **always** is shown and enters local time.

> I created the following event with Asia/Tokyo timezone and realised that the time stored in database is 08:57:52 (GMT+9), which means they are both local time.

So yes, this is correct.

>However, it's pretty hard to include VTIMEZONE in our ical file since I can't find any efficient way to create that VTIMEZONE object from pytz timezone (please correct me if I am wrong ;) ).
> In this situation, I think we should just go for UTC time, i.e. form 2 in the 1st link.

The problem is that if we go for UTC, some calendars might *show* time timezone difference which confuses users. Take for example your screenshots above where your calendar application shows two times next to each other. As users in many countries are not used to dealing with timezones, it would be nice to get around this.

I found this Django snippet: https://djangosnippets.org/snippets/10569/ (see also the first comment below), but it looks like a nasty thing to have in our codebase.

Apparently the [vobject](https://eventable.github.io/vobject/) library just deals with it automatically. Would you mind having a quick look if it would be worth it to swicht to this iCal implementation?

**Comment by jlwt90** [comment]
>found this Django snippet: https://djangosnippets.org/snippets/10569/ (see also the first comment below), but it looks like a nasty thing to have in our codebase.

Yes.. I think so too

>Apparently the vobject library just deals with it automatically. Would you mind having a quick look if it would be worth it to swicht to this iCal implementation?

Just have a quick test.
```
>>> import pytz
>>> from datetime import datetime
>>> import vobject
>>> import datetime
>>> from pytz import timezone
>>> cal = vobject.iCalendar()
>>> cal.add('vevent')
<VEVENT| []>
>>> cal.vevent.add('summary').value = "This is a note"
>>> start = cal.vevent.add('dtstart')
>>> start.value = datetime.datetime(2006, 2, 16, tzinfo = timezone('Asia/Tokyo'))
>>> cal.vevent.prettyPrint()
 VEVENT
    SUMMARY: This is a note
    DTSTART: 2006-02-16 00:00:00+09:00
>>> cal.serialize()
```
And it looks perfect :) 
> BEGIN:VCALENDAR
VERSION:2.0
PRODID:-//PYVOBJECT//NONSGML Version 1//EN
BEGIN:VTIMEZONE
TZID:Asia/Tokyo
BEGIN:STANDARD
DTSTART:20000101T000000
RRULE:FREQ=YEARLY;BYMONTH=1
TZNAME:JST
TZOFFSETFROM:+0900
TZOFFSETTO:+0900
END:STANDARD
END:VTIMEZONE
BEGIN:VEVENT
UID:20170306T150357Z - 19662@xxxx.local
DTSTART;TZID=Asia/Tokyo:20060216T000000
SUMMARY:This is a note
END:VEVENT
END:VCALENDAR

I'll go through the current issue list and document tomorrow to see if there is any potential issue but I think this implementation fits our needs! Thanks for your suggestion.

**Comment by raphaelm** [comment]
I've given the new implementation a quick look and I like it :) Shall I do some final testing and then merge it or do you see anything here left to do before that?

**Comment by jlwt90** [comment]
Thanks for checking! I think it's ready for final test.

The expected behaviour should be:
**Custom Timezone**
```
BEGIN:VTIMEZONE
..
END:VTIMEZONE
DTSTART;TZID=Asia/Tokyo:20170329T175752 #DTSTART;TZID=<timezone>:<local time>
```
**UTC**
```
DTSTART:20170329T175752Z
```
**Show date only**
```
DTSTART;VALUE=DATE:20170329
DTSTART;VALUE=DATE:20170331
```
**Show start date/datetime only**
```
DTSTART;VALUE=DATE:20170329
#no DTEND
```

Please let me know if you find any bug/improvement.

**Comment by raphaelm** [comment]
Okay, I played around a bit. I'm sorry that I again have very last remarks:

* Currently, vobject generates a random UID that contains the server's hostname, e.g. ``cal.vevent.add('uid').value = '{}-{}-{}'.format(event.organizer.slug, event.slug, time.time())``
* Our ``setup.py`` also needs to include the new dependency on ``vobject``.

Otherwise, I'm happy! :)

**Comment by jlwt90** [comment]
Oops, my bad! I completely forgot that UID thing (was going to ask for your opinion).
The default UID is actually recommended by icalendar spec coz UID uniqueness can be guaranteed (most likely) but I don't think it's a good idea to include hostname. In the last commit, I use 
```
vevent.add('uid').value = '{}-{}-{}'.format(
        event.organizer.slug, event.slug, creation_time.strftime('%Y%m%d%H%M%S%f')
)
```
I think `datetime` should be better than `time` if we aim for uniqueness!

> Our setup.py also needs to include the new dependency on vobject.
didn't know that ;) thanks for your reminder!

raphaelm:
It is beautiful! Thank you very much, your contribution is highly appreciated 🎈

PR #417, opened 2017-02-22
add test for sendmail plugin

FlaviaBastos:
Start adding tests to sendmail plugin.

Hi, @raphaelm! I didn't make much progress on my own. When you have some time, would you please help me move this forward?
Thanks!

### DIFF
--- a/src/tests/plugins/test_sendmail.py
+++ b/src/tests/plugins/test_sendmail.py
@@ -1,0 +1,69 @@
+import datetime
+
+import pytest
+
+from django.utils.timezone import now
+
+from pretix.base.models import (
+    Event, EventPermission, Item, ItemCategory, Order, OrderPosition,
+    Organizer, OrganizerPermission, User,
+)
+
+
+@pytest.fixture
+def event():
+    '''Returns an event instance'''
+    o = Organizer.objects.create(name='Dummy', slug='dummy')
+    event = Event.objects.create(
+        organizer=o, name='Dummy', slug='dummy',
+        date_from=now(),
+        plugins='pretix.plugins.sendmail,tests.testdummy',
+    )
+    return event
+
+
+@pytest.fixture
+def item(event):
+    '''Returns an item instance'''
+    return Item.objects.create(name='Test item', event=event, default_price=13)
+
+
+@pytest.fixture
+def item_category(event):
+    '''Returns an item category instance'''
+    return ItemCategory.objects.create(event=event)
+
+
+@pytest.fixture
+def order(item):
+    '''Returns an order instance'''
+    o = Order.objects.create(event=item.event, status=Order.STATUS_PENDING,
+                             expires=now() + datetime.timedelta(hours=1),
+                             total=13, code='DUMMY', email='dummy@dummy.test',
+                             datetime=now(), payment_provider='banktransfer')
+    OrderPosition.objects.create(order=o, item=item, price=13)
+    return o
+
+
+@pytest.fixture
+def logged_in_client(client, event):
+    '''Returns a logged client'''
+    user = User.objects.create_superuser('dummy@dummy.dummy', 'dummy')
+    OrganizerPermission.objects.create(organizer=event.organizer, user=user, can_create_events=True)
+    EventPermission.objects.create(event=event, user=user, can_change_items=True,
+                                   can_change_settings=True, can_change_orders=True, can_view_orders=True)
+    client.force_login(user)
+    return client
+
+
+@pytest.mark.parametrize('url,expected', [
+    ('/control/event/{orga}/{event}/sendmail/', 200),
+])
+@pytest.mark.django_db
+def test_sendmail_view(logged_in_client, url, expected, event):
+    url = url.format(
+        event=event.slug, orga=event.organizer.slug,
+    )
+    response = logged_in_client.get(url)
+
+    assert response.status_code == expected
--- a/src/tests/plugins/test_sendmail.py
+++ b/src/tests/plugins/test_sendmail.py
@@ -98,7 +98,7 @@ def test_sendmail_simple_case(logged_in_client, sendmail_url, event, order):
 
 
 @pytest.mark.django_db
-def test_sendmail_email_not_sent_to_order(logged_in_client, sendmail_url, event, order):
+def test_sendmail_email_not_sent_if_order_not_match(logged_in_client, sendmail_url, event, order):
     djmail.outbox = []
     response = logged_in_client.post(sendmail_url,
                                      {'sendto': 'p',
--- a/src/tests/plugins/test_sendmail.py
+++ b/src/tests/plugins/test_sendmail.py
@@ -9,7 +9,6 @@ from pretix.base.models import (
     Event, EventPermission, Item, ItemCategory, Order, OrderPosition,
     Organizer, OrganizerPermission, User,
 )
-from pretix.base.services.mail import mail
 
 
 @pytest.fixture
@@ -75,16 +74,16 @@ def test_sendmail_view(logged_in_client, sendmail_url, expected=200):
 
 
 @pytest.mark.django_db
-def test_sendmail_one_message(logged_in_client, sendmail_url, event, order, expected=200):
+def test_sendmail_simple_case(logged_in_client, sendmail_url, event, order):
     djmail.outbox = []
     response = logged_in_client.post(sendmail_url,
-                                     {'sendto': ('c'),
-                                      'subject': 'Test subject',
-                                      'message': 'This is a test file for sending mails.'
-                                      })
-    assert response.status_code == expected
-
-    mail('dummy@dummy.test', 'Test subject', 'mailtest.txt', {}, event)
+                                     {'sendto': 'n',
+                                      'subject_0': 'Test subject',
+                                      'message_0': 'This is a test file for sending mails.'
+                                      },
+                                     follow=True)
+    assert response.status_code == 200
+    assert 'alert-success' in response.rendered_content
 
     assert len(djmail.outbox) == 1
     assert djmail.outbox[0].to == [order.email]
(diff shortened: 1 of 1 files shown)


raphaelm:
Hi Flavia,

I'm really sorry this reply took me so long :cry:

As far as I can see, you successfully managed to create a test file with a few fixtures and a first (more or less empty) test. Nice!

First of all, drop the
```
@pytest.mark.parametrize('url,expected', [
    ('/control/event/{orga}/{event}/sendmail/', 200),
])
```
it will only be confusing, we don't need parametrization here.

Now, with the test skeleton set up the next question is: **What to test?** It requires some training to get into the mindset that you need to come up with test ideas, so don't worry if it feels hard at first. The first rule of testing is good news: Any test is better than no test :wink: 

In our case, the most basic test would be:

1. Do a POST request to the send mail form containing a message etc.
2. Check that an email has been sent
3. Do a GET request to the mail history and check that the mail appears

Once that runs, you could go to more sophisticated tests, e.g.:

* Create an order that is not paid, do a POST request that says the message should only be sent to paid orders, and check that an email has *not* been sent.
* Create an order that only asks for a preview and check that no mail is sent.
* Submit invalid data (missing subject, missing message, ...) to the form and check that no mail is sent and an error is returned.
* Submit a multi-lingual message and create orders with different locales and check that the sent mails have the correct language.

You see, the thought process is basically going over your code line by line and asking yourself "if this line was wrong or missing, what would I do to notice?".

Looking at my suggestions above, you only need a very overseeable set of tools to build the tests:

* You need to know how to do GET and POST requests with the ``logged_in_client``. You can find documentation on that [here](https://docs.djangoproject.com/en/1.10/topics/testing/tools/#the-test-client) and some easy examples [here](https://github.com/pretix/pretix/blob/master/src/tests/control/test_waitinglist.py#L47).

* You need to know how to check for sent mails: While testing, pretix won't actually send emails but just store them into a list. You can find documentation on that [here](https://docs.djangoproject.com/en/1.10/topics/testing/tools/#email-services) and a number of examples [in test_mail.py](https://github.com/pretix/pretix/blob/master/src/tests/base/test_mail.py).

* You need to know how to create orders: Look at your own file :wink: 

I hope this helps you to get going. As always, please ask if you have any further questions. I'm quite busy at the moment but will reply as soon as I can :)

Best
Raphael

FlaviaBastos:
Hi, @raphaelm 

> I'm really sorry this reply took me so long :cry:

Oh, please. No worries at all! I never expected to have an immediate reply. You are always very kind to read and answer my questions. I am in no position to be mad at you! :laughing: 

As always, thank you very much for taking the time to write all this. I really (really!) appreciate. I will spend some time chewing this and will come back with something - hopefully code! :rofl:

FlaviaBastos:
Hi, @raphaelm!

I am stuck and was wondering if you could please help out when you have some spare time.

I am trying to make a `get request` to mail history and assert that I see my expected text there ("Test subject"). For this I am following the example you pointed out on [waiting_list](https://github.com/pretix/pretix/blob/master/src/tests/control/test_waitinglist.py#L47) using
`response.rendered_content`. When I run my test, I get an error that looks like I am getting a modal screen as response. This is the error message:

```
>       assert 'Test subject' in response.rendered_content
E       assert 'Test subject' in '\n\n\n\n<!DOCTYPE html>\n<html>\n\t<head>\n\t\t<title>Send out emails :: \n            pretix.de</title>\n\t\t<link r...   If this takes longer than a few minutes, please contact us.\n            </p>\n        </div>\n\t</body>\n</html>\n'
E        +  where '\n\n\n\n<!DOCTYPE html>\n<html>\n\t<head>\n\t\t<title>Send out emails :: \n            pretix.de</title>\n\t\t<link r...   If this takes longer than a few minutes, please contact us.\n            </p>\n        </div>\n\t</body>\n</html>\n' = <TemplateResponse status_code=200, "text/html; charset=utf-8">.rendered_content
```

I ran the test on `waiting_list` and that pass just fine but I am failing to identify what is
missing in my test, or what is the difference in the `rendered_content` call.

Thanks!

raphaelm:
Hi Flavia,

you just built a great example for something everyone needs to learn about testing. Let's pick your test apart together.

```
def test_sendmail_one_message(logged_in_client, sendmail_url, event, order, expected=200):
```

Just a minor note here, drop the ``expected=200``, it does not make a lot of sense here, just use ``200`` directly later where you now use ``expected``.

```
    djmail.outbox = []
    response = logged_in_client.post(sendmail_url,
                                     {'sendto': ('c'),
                                      'subject': 'Test subject',
                                      'message': 'This is a test file for sending mails.'
                                      })
    assert response.status_code == expected
```

Here is where it gets interesting. You submit the sendmail form and you check that it returns a 200 code. Yay, it does, test passes so far. But is this a useful test? No, because the form never got submitted, you sent invalid data, but you don't check for the error messages in ``response`` ;) A helpful way to see this might be to add a ``print(response.rendered_content)`` in there to view the output.

We can fix this in two ways. Best is to do both. Number one would be to check for the success message, e.g. ``assert 'alert-success' in response.rendered_content``. The other way is to check if an email has actually been sent. This is what you try by doing…

```
    assert len(djmail.outbox) == 1
    assert djmail.outbox[0].to == [order.email]
    assert djmail.outbox[0].subject == 'Test subject'
    assert 'This is a test file for sending mails.' in djmail.outbox[0].body
```

…but apparently it did not work. It did not work, because you submitted invalid data to the form, but during your debugging you inserted…

```
    mail('dummy@dummy.test', 'Test subject', 'mailtest.txt', {}, event)
```

… which of course *does* send the email, but not because the form worked, because you sent it manually ;) Don't do that in a test! ;)

Obviously, if the form didn't work, the history won't work as well. 

Now to the interesing question: **Why does the form not work?** You send the fields ``sendto``, ``subject`` and ``message``. If you open pretix in the browser and open the developer tools (F12 in most browsers) in the "network" tab and then submit the form you will see the POST request and click it for details. You will see that the actual fields are named differently, e.g. ``subject_0``. The ``_0`` comes from the internationalization-thing, if it would be a multi-language event, the field for the second language would be named ``_1`` and so on.

I hope my point is clear from this long post. If not, just ask and I'll try to be clearer on some details. To debug such failures, it generally helps to ``print`` a lot of things and to compare with the dev tools in the browser.

FlaviaBastos:
Thank you for the detailed explanation. It helps a lot! As you can see, testing is still a mystery to me..... :disappointed_relieved:
I normally debug with print statements but I couldn't make them work in the test files. I didn't think of checking dev tools, though. Thanks!

rixx:
@FlaviaBastos When running pytest, the `-s` flag tells it to not capture output, which comes in very handy. :)

src/tests/plugins/test_sendmail.py
```diff
@@ -94,5 +93,84 @@ def test_sendmail_one_message(logged_in_client, sendmail_url, event, order, expe
     url = sendmail_url + 'history/'
     response = logged_in_client.get(url)
 
-    assert response.status_code == expected
+    assert response.status_code == 200
     assert 'Test subject' in response.rendered_content
+
+
+@pytest.mark.django_db
+def test_sendmail_email_not_sent_to_order(logged_in_client, sendmail_url, event, order):
```
rixx: Hmm, maybe give the method a name that describes *why* the mail is not sent. (`_not_sent_if_…`)

src/tests/plugins/test_sendmail.py
```diff
@@ -94,5 +93,84 @@ def test_sendmail_one_message(logged_in_client, sendmail_url, event, order, expe
...
+                                      },
+                                     follow=True)
+
+    assert 'help-block' in response.rendered_content
+
+    assert len(djmail.outbox) == 0
+
+
+@pytest.mark.django_db
+def test_sendmail_multi_locales(logged_in_client, sendmail_url, event, item):
```
rixx: Nice!

src/tests/plugins/test_sendmail.py
FlaviaBastos: Right! I will do that.
Hey, sorry I only addressed Raphael. I didn't know you were going to review. Thanks, by the way! :+1:

rixx: No worries - when I have some time I look over pretix stuff and so far Raphael doesn't seem to mind. I really like that you're getting into testing, by the way ;)

FlaviaBastos: > I really like that you're getting into testing, by the way ;)

Me too! :) Testing can be so scaring... it's just easier to postpone and never do it.
Thanks for the encouragement.

raphaelm: > and so far Raphael doesn't seem to mind

I can't see the day coming on which I would mind ;)

Flavia, I'm gonna try to review this tonight, let's see if I actually get around to do it :D

src/tests/plugins/test_sendmail.py
```diff
@@ -0,0 +1,176 @@
...
+@pytest.mark.django_db
+def test_sendmail_invalid_data(logged_in_client, sendmail_url, event, order):
+    djmail.outbox = []
+    response = logged_in_client.post(sendmail_url,
+                                     {'sendto': 'n',
+                                      'subject_0': 'Test subject',
+                                      },
+                                     follow=True)
+
+    assert 'help-block' in response.rendered_content
```
raphaelm: I think ``help-block`` is always there, better check for ``has-error``.

src/tests/plugins/test_sendmail.py
```diff
@@ -0,0 +1,176 @@
...
+
+from pretix.base.models import (
+    Event, EventPermission, Item, ItemCategory, Order, OrderPosition,
+    Organizer, OrganizerPermission, User,
+)
+
+
+@pytest.fixture
+def event():
+    '''Returns an event instance'''
```
raphaelm: Just a very minor style thing here, please use ``"""`` instead of ``'''`` for docstrings.

rixx: (… because there is a [PEP](https://www.python.org/dev/peps/pep-0257/) specifying docstring conventions.)

FlaviaBastos: :flushed: Thank you for pointing this out! Yet another lesson for me...

FlaviaBastos:
I just pushed another version that with fixes from your latest feedback.
Thanks! :)

FlaviaBastos:
yay! :tada: Thank you!

Issue #372, opened 2017-01-09
Plugin hook for "Copy from event"


raphaelm:

raphaelm:
e.g. for the Pages plugin

Issue #309, opened 2016-11-09
Add an interest list


raphaelm:
Add the possibility to register *before* the presale starts to be notified when it starts. This allows the organizer to estimate the interest in the event.

Issue #297, opened 2016-11-02
pretixdroid: Show metrics in the control panel


raphaelm:
The pretixdroid Android app allows organizers to scan the tickets at the entrance. The date and time of those check-ins are saved and it would be nice to do at least the following things with them:

- [x] Add checkin to order log
- [x] Show list of checkins by ticket
- [x] Show number of checkins on dashboard
- [x] List of people already checked in / not yet checked in

raphaelm:
Hi @jlwt90, if you're interested in this, please tell me if the scope of this is clear, if not, I can sure write one of my lengthier issue descriptions here :wink:

jlwt90:
Hi @raphaelm, sorry for replying late. I just got some time checking the finished tasks in this ticket.
Could I confirm whether my direction is correct?

The api pretixdroid is calling is provided by `src/pretix/plugins/pretixdroid`. The checkin information is stored by using the following code block:
```
op.order.log_action('pretix.plugins.pretixdroid.scan', data={
                        'position': op.id,
                        'positionid': op.positionid,
                        'first': True,
                    })
```
So basically I will play around `OrderPosition`, `Checkin` & `LogEntry` and of course the event dashboard.

Could you mind explaining the last task `List of people already checked in / not yet checked in` in details? I guess it should be a separate page or sth. for admin to keep track of everyone's check-in status. When admin clicks on the `checkins` block in dashboard, he/she will be redirected to the check-in detail page. Is that what you want to achieve?

Thanks!

raphaelm:
Hi,

> So basically I will play around OrderPosition, Checkin & LogEntry and of course the event dashboard.

I believe you won't need ``LogEntry``, ``Checkin`` should already contain all required information.

> Could you mind explaining the last task List of people already checked in / not yet checked in in details? I guess it should be a separate page or sth. for admin to keep track of everyone's check-in status. When admin clicks on the checkins block in dashboard, he/she will be redirected to the check-in detail page. Is that what you want to achieve?

Yep, that's exactly what I imagined. I think there should be a page with the check-in status of all attendees (maybe later it could be searchable etc.) and that would be perfect to link to from the dashboard tile.

This should not need to live in the pretixdroid plugin, as there will probably be more methods to generate ``Checkin`` objects later. It can just be a regular page around here: 
![2017-04-25-092259_255x283_scrot](https://cloud.githubusercontent.com/assets/64280/25373154/cdc3ce54-2998-11e7-8cee-f534d96ef56d.png)

PR #432, opened 2017-03-11
fix #428: ignore timezone when saving event

### LINKED ISSUE #428: Event time with timezone
It's actually a follow-up for #413.

>The rule of thumb in pretix (and Django) is: The database contains always UTC, the user always is shown and enters local time.

I asked this question because I found something interesting when creating event.
The scenarios are as follows:
1) Create new event using timezone `Asia/Tokyo`
create-tokyo
After save:
saved-tokyo
Both start and end time were changed with respect to the timezone I set. 
From **11:19:59 (GMT+9)** to **20:19:59 (GMT+9)**.
I guess we consider the start and end time as UTC time by default regardless of the timezone we set.

2) Change time only
timezone-unchanged
As expected, it works well.

3) Change timezone of an existing event
change timezone
This time I changed the timezone from `Asia/Tokyo`(GMT+9) to `Asia/Singapore`(GMT+8).
saved timezone
Both times were changed from **08:50:56** to **07:50:56**.

I am quite new to pretix and not sure whether this kind of behaviour is expected.
From my opinion, I think user will only set event time using local time, which means it's better to include timezone value before we save data to database.

jlwt90:
Hi, this is the PR for fixing #428 .

**event creation**
This has been fixed by adding timezone to date related values in `clean` method in `EventWizardBasicsForm`.

**event update**
I still have no idea how I can fix it. (Sorry I'm pretty new to django..)
>From the top of my head, I would go neither for option 1 nor for option 2. EventWizard and EventUpdate are views, which are rather to high-level for this kind of thing and Event is a model which is rather to low-level for this. In my opinion, the golden middle road would be the forms involved, which are EventWizardBasicsForm and EventUpdateForm. We could do it in the save() methods there but I think an even better way might be the clean() method in Django's validation cycle.

@raphaelm I just learnt the django's validation cycle you mentioned in the comment from #428. Thanks! 👍 👍  However, I am wondering whether we can use the same technique in event update.
Since `timezone` is stored in `EventSettingsForm` while `date_to` (as well as the other 3 fields) are stored in `EventUpdateForm`, I can't use (more like I don't know how to use... ) the timezone to update all 4 fields in `clean` method. Since `EventUpdate` view is the bridge between the two forms, should I also change part of the logic there?

### DIFF
--- a/src/tests/control/test_events.py
+++ b/src/tests/control/test_events.py
@@ -1,7 +1,9 @@
 import datetime
 from decimal import Decimal
 
+import pytz
 from i18nfield.strings import LazyI18nString
+from pytz import timezone
 from tests.base import SoupTest, extract_form_fields
 
 from pretix.base.models import (
@@ -45,7 +47,6 @@ class EventsTest(SoupTest):
         doc = self.get_doc('/control/event/%s/%s/settings/' % (self.orga1.slug, self.event1.slug))
         doc.select("[name=date_to]")[0]['value'] = "2013-12-30 17:00:00"
         doc.select("[name=settings-max_items_per_order]")[0]['value'] = "12"
-        print(extract_form_fields(doc.select('.container-fluid form')[0]))
 
         doc = self.post_doc('/control/event/%s/%s/settings/' % (self.orga1.slug, self.event1.slug),
                             extract_form_fields(doc.select('.container-fluid form')[0]))
@@ -60,7 +61,6 @@ class EventsTest(SoupTest):
         doc.select("[name=settings-timezone]")[0]['value'] = "Asia/Tokyo"
         doc.find('option', {"value": "Asia/Tokyo"})['selected'] = 'selected'
         doc.find('option', {"value": "UTC"}).attrs.pop('selected')
-        print(extract_form_fields(doc.select('.container-fluid form')[0]))
 
         doc = self.post_doc('/control/event/%s/%s/settings/' % (self.orga1.slug, self.event1.slug),
                             extract_form_fields(doc.select('.container-fluid form')[0]))
@@ -69,6 +69,11 @@ class EventsTest(SoupTest):
         assert doc.select("[name=date_to]")[0]['value'] == "2013-12-30 17:00:00"
         assert doc.find('option', {"value": "Asia/Tokyo"})['selected'] == "selected"
         assert doc.select("[name=settings-max_items_per_order]")[0]['value'] == "12"
+
+        self.event1.refresh_from_db()
+        # Asia/Tokyo -> GMT+9
+        assert self.event1.date_to.strftime('%Y-%m-%d %H:%M:%S') == "2013-12-30 08:00:00"
+        assert self.event1.settings.timezone == 'Asia/Tokyo'
 
     def test_plugins(self):
         doc = self.get_doc('/control/event/%s/%s/settings/plugins' % (self.orga1.slug, self.event1.slug))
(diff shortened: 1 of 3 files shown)


raphaelm:
> Since timezone is stored in EventSettingsForm while date_to (as well as the other 3 fields) are stored in EventUpdateForm, I can't use (more like I don't know how to use... ) the timezone to update all 4 fields in clean method. Since EventUpdate view is the bridge between the two forms, should I also change part of the logic there?

Your observation is very correct, I missed that fact. Yes, I think the view is a good place then.

jlwt90:
I updated both `EventUpdate` view and `EventWizardBasicsForm` and added another test case for checking timezone & date time.
However, I couldn't find any test case for event creation (please tell me if there is any). I think I'll create another test case with custom timezone in `EventWizardBasicsForm` later.

src/tests/control/test_events.py
```diff
@@ -52,6 +52,23 @@ def test_settings(self):
...
         assert doc.select("[name=settings-max_items_per_order]")[0]['value'] == "12"
 
+    def test_settings_timezone(self):
+        doc = self.get_doc('/control/event/%s/%s/settings/' % (self.orga1.slug, self.event1.slug))
+        doc.select("[name=date_to]")[0]['value'] = "2013-12-30 17:00:00"
+        doc.select("[name=settings-max_items_per_order]")[0]['value'] = "12"
+        doc.select("[name=settings-timezone]")[0]['value'] = "Asia/Tokyo"
+        doc.find('option', {"value": "Asia/Tokyo"})['selected'] = 'selected'
+        doc.find('option', {"value": "UTC"}).attrs.pop('selected')
+        print(extract_form_fields(doc.select('.container-fluid form')[0]))
```
raphaelm: Can you remove the ``print`` statement here?

src/tests/control/test_events.py
```diff
@@ -52,6 +52,23 @@ def test_settings(self):
...
+
+        doc = self.post_doc('/control/event/%s/%s/settings/' % (self.orga1.slug, self.event1.slug),
+                            extract_form_fields(doc.select('.container-fluid form')[0]))
+        assert len(doc.select(".alert-success")) > 0
+        # date_to should not be changed even though the timezone is changed
+        assert doc.select("[name=date_to]")[0]['value'] == "2013-12-30 17:00:00"
+        assert doc.find('option', {"value": "Asia/Tokyo"})['selected'] == "selected"
+        assert doc.select("[name=settings-max_items_per_order]")[0]['value'] == "12"
```
raphaelm: Could you not only check what is shown in the form after saving but additionaly also perform a `self.event1.refresh_from_db()` and then check the time that is saved in the database? Just to be double-sure nothing gets lost in the form layer.

jlwt90: nice idea! thanks!

raphaelm:
Nice! This is good work! :clap: Thank you very much for your contribution!

If you want to continue working on pretix, I think you are ready for something beyond the "easy" scope. The issues #308, #372, #297 or #309 come to my mind here. They are of different size but all have a well-defined goal and I can provide you with help on how to do it. Is there one among them that would interest you?

jlwt90:
@raphaelm thanks! I'll certainly continue working on pretix (it's an interesting project!).
I went through the issues you mentioned above and I think I can help on #308, if you don't mind working with newbie.

Since I still have another task at hand, I think I'll probably start working on it 2~3 days later.
Thank you again for your feedback and suggestions! 👏

raphaelm:
Sure, take your time and just ask if anything is unclear on #308 :)

Issue #102, opened 2015-10-04
Add multiple-choice question types


raphaelm:
pretix currently allows the admin to define questions which can be answered when ordering a ticket. For example, the admin might ask for the attendee's age. Currently, only questions that ask for text, numbers or Yes/No are supported. It would be nice if we would also support asking to choose one of multiple options
- [x] Checkboxes (choose multiple)
- [x] Selects (choose one)
- [ ] "Other"-option

**Update**: We now have multiple-choice questions that allow the user to select one of many options. The next step would be to (optionally) provide an "Other" option that allows the user to enter custom input in a text field. This would need a new model field for the question and some javascript for the fillout process. A new field on the answer object should not be necessary.

simoneott:
Hello,  I'm new to the  github and I would like to work on the multiple question types. Is it ok if I start working on that or is there an official process I need to go through before jumping in. Don't want to step on anyone's toes but would love to get working on a project and this sounds particularly interesting to me.

raphaelm:
Hello! You are absolutely welcome to start working on this. As this project is currently very small in terms of contributors, we have no formal or official processes. I'd be glad to help you through the technical process, though.

Sadly, it is currently not as easy to set up a local version of pretix as it should be. We have a guide in [our documentation](http://docs.pretix.eu/en/latest/development/setup.html) but it might be incomplete or out of date. Please *do not hesitate* to ask if you get stuck!

In order to tackle this ticket, a few things need to be done:

* In the admin area, called ``pretix control`` in all technical documentation, there is an area in which it is possible to define the mentioned questions and select which type of answer is expected. We have to add two new types of answers here (choose multiple/choose one) as well as a possibility to add choices to those questions. For this, both the ``QuestionForm`` and the respective templates have to be modified.
* The ``Question`` *data model* (living in ``pretix/base/models.py``) needs to be extended in order to store those options. A very simple way would be to store them in a text field, separated by linebreaks; a more advanced way would be to create a separate model ``QuestionOption`` to store them.
* In the frontwend, called ``pretix presale``, when actually ordering one of the products assigned to the question, the user has to be presented a selection field instead of a free input field. This has to be done in ``pretix/presale/forms/checkout.py``.

I can't repeat it often enough, please ask if you get stuck, might be that I can help you with a very simple pointer in the right direction. Also, it is great feedback for me if I hear what is easy and what is hard to understand about the codebase.

Have you used Python and Django before?

simoneott:
Hello!

[...]

Issue #100, opened 2015-10-01
Create an embeddable version for event webpages


raphaelm:
Most ticketing softwares provide a way to integrate the ticket shop into an existing event website. This is a very useful thing and we want to support that too. There are generally two ways this can be done.

* Using a ``<iframe>`` that contains the whole shop. This is how [Eventbrite](https://www.eventbrite.com/support/articles/en_US/How_To/how-to-sell-eventbrite-tickets-registrations-on-your-website-using-embeddable-widgets?lg=en_US) and XING Events do it.
* Using a javascript widget that is embedded into the page. This is how [Tito](https://ti.to/docs/widget) does it. Their docs contain beautiful examples of it.

I think it is clear that the javascript approach looks *way* better than the iframe approach and allows the website owner to apply custom CSS to style the product list. However, there is a strong reason against copying Tito's behaviour: We'd need to maintain two complete sets of UI for all of the ordering process. We can't re-use the existing UI and we can't transform the existing UI to a JavaScript one because we're proud of the fact that pretix is completely usable even without JavaScript.

This leads to the following "middle-ground" suggestion:

1. Build a JavaScript widget that inserts a list of products into an arbitrary webpage that can be styled via custom CSS
2. Once a user clicks the "Add to cart" button, an overlay will be open (like a lightbox that covers e.g. 80-90% of the screen) that includes an ``<iframe>`` with the steps needed to complete the order.

This requires us only to maintain the list of products twice, not the entire process. Open questions:

* How does the order process end? Do we show the order confirmation inside the iframe? The problem with this is that this way, the user will not be seeing the URL in the address bar that they need later to return to the confirmation page (e.g. in case they don't get the email) and reloading the page is not really possible. I think the cleanest solution would be to solve this via the [history API](https://developer.mozilla.org/en-US/docs/Web/API/History_API) and a custom thing like ``#pretix-order-ABCDE-secret`` appended to the address bar.
* How do we deal with mobile devices? Tito just opens the normal shop in a new tab for them which is a good behaviour I guess.

Current state: There is initial work in the ``embeddable`` branch that contains a dependency-free JavaScript snippet that shows part of the product list. It does not yet support variations, vouchers or free-price items; it doesn't have a default CSS style yet and it does not open the iframe popup yet. And it's probably not that well-written JS code yet ;)

felixrindt:
Isn't this already implemented as seen on https://pretix.eu/about/en/features/shop ?

raphaelm:
[...]

PR #438, opened 2017-03-25
Issue #308: Preview for email templates

jlwt90:
Hello, this is the PR for #308 .

I modified the behaviour of the first email template: *Placed Order*. 
screen shot 2017-03-25 at 4 35 31 pm

There are two tabs at the bottom, *edit* and *preview*.
When you click on the *preview* tab, an ajax request will be fired to `/control/event/**/***/settings/email/preview` endpoint.

screen shot 2017-03-25 at 4 35 57 pm

I haven't implemented the backend logic at this moment (I'll add more commits later). 
The dummy response I am currently using is:
```
{
            'item': 'order_placed',
            'msgs': {
                'mail_text_order_placed_0': 'English version!',
                'mail_text_order_placed_1': 'Deutsch version!'
            }
        }
```
What do you think about this UI? :D
PS: Please also help check whether I can improve my code too. I don't have much front-end experience. Comments are welcome :)

*Task:*
- [x] Draft of UI change
- [x] UI Modification based on comments
- [x] Remove dummy response in backend view
- [x] Work on the remaining templates
- [x] Add tests
- [x] Fix locale index issue
- [x] Improve dummy data
- [x] Display date based on locale


src/pretix/control/templates/pretixcontrol/event/mail.html
```diff
@@ -21,9 +21,29 @@ <h4 class="panel-title">
                             </a>
                         </h4>
                     </div>
-                    <div id="order_placed" class="panel-collapse collapse">
+                    <div id="order_placed" class="panel-collapse collapse preview-panel">
                         <div class="panel-body">
-                            {% bootstrap_field form.mail_text_order_placed layout="horizontal" %}
+                            <div>
+                                <label class="col-md-3 control-label">Text</label>
```
raphaelm: ``{% trans "Text" %}`` please, just so we don't forget later :)

src/pretix/static/pretixcontrol/js/ui/mail.js
```diff
@@ -0,0 +1,83 @@
+function preview_task_callback(data, jqXHR, status) {
+    "use strict";
+    if (data.item){
```
raphaelm: Just a formatting nitpick, but a space between ``)`` and ``{``

jlwt90:
@raphaelm Thanks!

>but if they send the form to save, the textareas will be sent twice to the server, one time with the changed content and one time with the unchanged content which might lead to unexpected behaviour

I also have the same concern when I wrote that part. My original idea is to keep the preview textarea with exactly the same style so that the transition from `edit` mode to `preview` mode looks really smooth. I was struggling whether I should include the dynamic block `{% bootstrap_field %}` twice so as to prevent duplicated code in `mail.html`. The code in current version is cleaner but kinda risky (I totally agree with you!). 

>Don't use {% bootstrap_field but construct the preview fields manually in the template

Let's pick this one. Since preview feature is just for improving user experience, I don't think it's appropriate to change backend form implementation (option 2).

raphaelm:
> My original idea is to keep the preview textarea with exactly the same style so that the transition from edit mode to preview mode looks really smooth

I appreciate this thought! Although I think that it wouldn't be that much of a problem if the preview looked different than the editor. But I think it should be possible to replicate the textareas nevertheless.

jlwt90:
>I appreciate this thought! Although I think that it wouldn't be that much of a problem if the preview looked different than the editor

After I replaced `textarea` with `p`, I feel the same. :)
I think this one looks better?
Here is the screenshot:
screen shot 2017-03-27 at 1 07 59 am

Regarding the corresponding view, could you give me some directions?
I read some code in some other preview feature, e.g. `TicketSettingsPreview`. Apparently I should create a dummy event and order with transaction rollback, then call a method for generating mail content (`src/pretix/base/services/mail.py`?). 
Thanks!

raphaelm:
> After I replaced textarea with p, I feel the same. :)
> I think this one looks better?

Yes, looks good! I believe in the sendmail plugin we even use ``<pre>`` for the previews, but I'm not sure if that would look pretty here. (You can activate the sendmail plugin at Settings > Plugins, which will create a new item in the menu on the left featuring an email form with preview).

> I read some code in some other preview feature, e.g. TicketSettingsPreview. Apparently I should create a dummy event and order with transaction rollback, then call a method for generating mail content (src/pretix/base/services/mail.py?).

Nooooo! That is an ugly, ugly hack :see_no_evil: We need it for tickets, because tickets might be generated by plugins that know nothing about previews. We don't have this problem here, so we don't need to replicate this :D

Instead, it's pretty simple. Just take the email text sent by the user, and call ``.format()`` on it with useful sample values and… done! The available placeholders are listed in the help texts of the form and you could have a look at Flavia's work in PR #279 which did a similar thing to the sendmail plugin.

jlwt90:
>Yes, looks good! I believe in the sendmail plugin we even use \<pre\> for the previews, but I'm not sure if that would look pretty here. (You can activate the sendmail plugin at Settings > Plugins, which will create a new item in the menu on the left featuring an email form with preview).

Thanks! I'll take a look!

>Nooooo! That is an ugly, ugly hack 🙈 We need it for tickets, because tickets might be generated by plugins that know nothing about previews. We don't have this problem here, so we don't need to replicate this :D

Also good to know :D I was thinking why I need to play with database just for rendering the content! Indeed I thought there is an existing method in services layer which simply generates email content based on given event & order (and template for sure) without sending a mail. If this preview feature uses the same logic as the real one, we can guarantee that the preview will be exactly the same as the real mail. We also need to maintain both `MailPreviewView` and mail sending logic if we add/remove new fields. Is my understanding correct?

Thanks again for your detailed explanation!

raphaelm:
> I thought there is an existing method in services layer which simply generates email content based on given event & order (and template for sure) without sending a mail. If this preview feature uses the same logic as the real one, we can guarantee that the preview will be exactly the same as the real mail. We also need to maintain both MailPreviewView and mail sending logic if we add/remove new fields. Is my understanding correct?

Sadly: yes, we have this logic twice then. The ``mail`` module in the services layer just takes a template and a dictionary of things and combines them. Maintaining that logic twice is no problem, as its only a ``.format`` call. The interesting logic resides within the *calls* to ``mail``. It would be interesting to refactor that in a way that prevents us from keeping the logic twice -- but maybe that's a target for your next pull request? 😉

jlwt90:
Hello @raphaelm , this commit contains
- \<pre\> preview block
- skeleton of backend preview logic

Please take a look if you have time :D
*Edit:*
screen shot 2017-03-29 at 1 09 40 am

*Preview:*
screen shot 2017-03-29 at 1 09 31 am

src/pretix/control/templates/pretixcontrol/event/mail.html
```diff
@@ -23,7 +23,29 @@ <h4 class="panel-title">
...
+                                    <ul class="nav nav-pills pull-right">
+                                        <li role="presentation" class="active">
+                                            <a data-toggle="pill" type="edit" href="#mail_text_order_placed_edit"><i class="fa fa-pencil-square-o fa-fw"></i> {% trans "Edit" %}</a>
+                                        </li>
+                                        <li role="presentation">
+                                            <a data-toggle="pill" type="preview" href="#mail_text_order_placed_preview"><i class="fa fa-tv fa-fw"></i> {% trans "Preview" %}</a>
+                                        </li>
+                                    </ul>
+                                </div>
+                            </div>
```
raphaelm: Just a thought: We will need all of this boilerplate for every single mail template. I think once the basic setup is clear, we should put this into a seperate template (i.e. ``mail_settings_fragment.html``) that we [{% include %}](https://docs.djangoproject.com/en/1.10/ref/templates/builtins/#include).

src/pretix/control/views/event.py
```diff
@@ -399,6 +403,56 @@ def post(self, request, *args, **kwargs):
             return self.get(request)
 
 
+class MailSettingsPreview(EventPermissionRequiredMixin, View):
+    permission = 'can_change_settings'
+
+    # return the origin text if key is missing in dict
+    class SafeDict(dict):
+        def __missing__(self, key):
+            return '{' + key + '}'
```
raphaelm: Nice approach! :+1:

src/pretix/control/views/event.py
```diff
@@ -399,6 +403,56 @@ def post(self, request, *args, **kwargs):
...
+                'invoice_company': str(self.request.event.organizer.name)
+            }
+        }
+
+    def post(self, request, *args, **kwargs):
+        preview_item = request.POST.get('item', '')
+        if preview_item not in self.items:
+            raise HttpResponseBadRequest(_('invalid item'))
+
+        regex = r"^" + re.escape(preview_item) + r"_\d{1}$"
```
raphaelm: I don't think it's a good idea to hardcode that we'll never have more than 10 languages in an event (although it is unlikely, I agree).

src/pretix/control/views/event.py
```diff
@@ -399,6 +403,56 @@ def post(self, request, *args, **kwargs):
...
+    def items(self):
+        return ['mail_text_order_placed', 'mail_text_order_paid', 'mail_text_order_free', 'mail_text_resend_link',
+                'mail_text_resend_all_links', 'mail_text_order_changed', 'mail_days_order_expire_warning',
+                'mail_text_order_expire_warning', 'mail_text_waiting_list']
+
+    @cached_property
+    def dummy_data(self):
+        return {
+            'mail_text_order_placed': {
+                'event': str(self.request.event.name),
```
raphaelm: If you not cast this to string here, it should still work fine -- and will make things easier if we want the localized versions later.

src/pretix/control/templates/pretixcontrol/event/mail.html
```diff
@@ -92,7 +114,7 @@ <h4 class="panel-title">
                             </a>
                         </h4>
                     </div>
-                    <div id="order_expirew" class="panel-collapse collapse">
+                    <div id="order_expire" class="panel-collapse collapse">
```
raphaelm: ``order_expirew`` was supposed to be a (apparently bad) abbreviation for "expire *w*arning". If you change it, change the reference in the ``href=`` argument above as well.

src/pretix/control/templates/pretixcontrol/event/mail.html
jlwt90: Nice idea! Had a quick look at django template engine doc.
`{% include "name_snippet.html" with person="Jane" greeting="Hello" %}`
I think we can probably use this technique to render all the previews with parameter `preview_item`, e.g. `mail_text_order_placed, mail_text_order_paid`.

Or even more aggressive approach, we can use `{% include %}` to avoid writing duplicated `panel` block, i.e.
```
<div class="panel panel-default">
    <div class="panel-heading">
        <h4 class="panel-title">
            <a data-toggle="collapse" href="#order_paid">
                <strong>{% trans "Paid order" %}</strong>
            </a>
        </h4>
    </div>
    <div id="order_paid" class="panel-collapse collapse">
       <div class="panel-body">
           {% bootstrap_field form.mail_text_order_paid layout="horizontal" %}
       </div>
    </div>
</div>
```
The only differences between two mail panels are:
- Panel title, i.e. `{% trans "Paid order" %}`
- Panel id
- form element, i.e. `form.mail_text_order_paid`

What's your opinion? (I'll give a try tomorrow 👍 )

src/pretix/control/templates/pretixcontrol/event/mail.html
jlwt90: oops! I thought it's just a typo 😭 
Thanks for your reminder, I forgot to change the ref. as well. I'll just change it back!

src/pretix/control/views/event.py
jlwt90: Yes, you are right! Apparently only this line failed if `str` is not used.
`'invoice_name': _('Invoice for %s.') % {str(self.request.event.name)},`

src/pretix/control/views/event.py
jlwt90: :D agree, I'll accept all valid `preview_item` followed by any number.

src/pretix/control/templates/pretixcontrol/event/mail.html
raphaelm: Please, go for the "aggressive" option with less repetition!

src/pretix/control/views/event.py
raphaelm: You can fix that by using new-style formats here as well:

```
'invoice_name': _('Invoice for {}.').format(self.request.event.name)
```

They can deal with the internationalized strings. In the long run, we eventually want to use ``.format()`` everywhere.

jlwt90:
@raphaelm Hello, could you review again? This version contains `mail_preview_fragment.html` and custom filter `mail_settings_preview.py`. 

1) The fragment template consists of a collapsable panel with a set of preview panels.
When we include the template, the following key arguments should be provided:
- **pid**: panel ID
- **title**: name display on panel header
- **items**: item names delimited by `,`, e.g. `mail_text_resend_link,mail_text_resend_all_links` for **Resend link** panel

2) Added locale in dummy data generation
As discussed, I assume that `mail_text_resend_link_0` is mapped to the default locale, and the indices are the same as the template, e.g. 0 for en, 1 for de

src/pretix/control/templates/pretixcontrol/event/mail.html
```diff
@@ -13,106 +14,13 @@
...
+                {% include "pretixcontrol/event/mail_settings_fragment.html" with pid="order_paid" title="Paid order" items="mail_text_order_paid" %}
+                {% include "pretixcontrol/event/mail_settings_fragment.html" with pid="order_free" title="Free order" items="mail_text_order_free" %}
+                {% include "pretixcontrol/event/mail_settings_fragment.html" with pid="resend_link" title="Resend link" items="mail_text_resend_link,mail_text_resend_all_links" %}
+                {% include "pretixcontrol/event/mail_settings_fragment.html" with pid="order_changed" title="Order changed" items="mail_text_order_changed" %}
+                {% include "pretixcontrol/event/mail_settings_fragment.html" with pid="order_expirew" title="Payment reminder" items="mail_days_order_expire_warning,mail_text_order_expire_warning" exclude="mail_days_order_expire_warning" %}
+                {% include "pretixcontrol/event/mail_settings_fragment.html" with pid="waiting_list" title="Waiting list notification" items="mail_text_waiting_list" %}
```
raphaelm: You surprised me with the ``|getattr:``, I was wondering how you're going to do that but well, it seems to work and I don't have a cleaner solution right now ;)

However, we need to treat the title differently so the strings show up correctly in the translation, for example:
```
{% blocktrans asvar the_title %}Placed order{% endblocktrans %}
{% include "pretixcontrol/event/mail_settings_fragment.html" with pid="order_placed" title=the_title items="mail_text_order_placed" %}
```

Also, if you look at the ``resend_link`` part, there are two different text fields that are currently both labeled with ``Text``, we probably need to pass alternative labels here. Maybe you try to just use the ``label`` attribute of the form field? it should be filled correctly.

src/pretix/control/views/event.py
```diff
@@ -399,6 +404,79 @@ def post(self, request, *args, **kwargs):
...
+        regex = r"^" + re.escape(preview_item) + r"_(?P<idx>[\d+])$"
+        msgs = {}
+        for k, v in request.POST.items():
+            # only accept allowed fields
+            matched = re.search(regex, k)
+            if matched is not None:
+                idx = int(matched.group('idx'))
+                if idx < len(self.request.event.settings.locales):
+                    with translation.override(self.request.event.settings.locales[idx]):
+                        msgs[k] = v.format_map(self.placeholders(preview_item))
```
raphaelm: Looks good on first sight

jlwt90:
>Also, pleased delete the 0052_auto_20170311_0634.py file that you accidentally created when branching of, it currently causes migrations to fail if this is merged into master.

Oops! I didn't notice that I committed migration file too. Thanks!

> Note that I'll be travelling to DjangoCon Europe starting tomorrow, so over the next week I might be slower with reviewing and replying.

No problem! Enjoy DjangoCon 👍 
Yes, I'll fix the problem you mentioned and add tests 😄

src/pretix/control/templates/pretixcontrol/event/mail.html
jlwt90: Thanks! I completely forgot about the translation part.

> Also, if you look at the resend_link part, there are two different text fields that are currently both labeled with Text, we probably need to pass alternative labels here. Maybe you try to just use the label attribute of the form field? it should be filled correctly.

I'll try using the default label attribute and see if the css works well

jlwt90:
added tests too. I'm sorry that I somehow messed up the commits when I tried to rebase with master. It's now fixed but a list of commits appears at the bottom of this pull request..

src/pretix/control/views/event.py
```diff
@@ -400,6 +405,78 @@ def post(self, request, *args, **kwargs):
...
+        return {
+            'event': self.request.event.name,
+            'total': 100,
+            'currency': self.request.event.currency,
+            'date': date_format(now() + timedelta(days=7), 'SHORT_DATE_FORMAT'),
+            'paymentinfo': _('Please transfer money to this bank account: BANK-ACCOUNT-0001'),
+            'payment_info': _('Please transfer money to this bank account: BANK-ACCOUNT-0001'),
+            'url': build_absolute_uri('presale:event.index', kwargs={
+                'event': self.request.event.slug,
+                'organizer': self.request.event.organizer
```
raphaelm: This needs to be ``self.request.event.organizer.slug``

src/pretix/control/views/event.py
```diff
@@ -400,6 +405,78 @@ def post(self, request, *args, **kwargs):
...
+            }),
+            'invoice_name': _('Invoice for {}.').format(self.request.event.name),
+            'invoice_company': self.request.event.organizer.name,
+            'orders': 'ORDER-0001\nORDER-0002\nORDER-0003\nORDER-0004',
```
raphaelm: If you look at ``pretix/presale/views/user.py``, you see that this is actually structured in the format:

```
 - EVENTSLUG-ABCDE - https://link/to/order
 - EVENTSLUG-HIJKL - https://link/to/order
```

src/pretix/control/views/event.py
```diff
@@ -400,6 +405,78 @@ def post(self, request, *args, **kwargs):
...
+            'currency': self.request.event.currency,
+            'date': date_format(now() + timedelta(days=7), 'SHORT_DATE_FORMAT'),
+            'paymentinfo': _('Please transfer money to this bank account: BANK-ACCOUNT-0001'),
+            'payment_info': _('Please transfer money to this bank account: BANK-ACCOUNT-0001'),
+            'url': build_absolute_uri('presale:event.index', kwargs={
+                'event': self.request.event.slug,
+                'organizer': self.request.event.organizer
+            }),
+            'invoice_name': _('Invoice for {}.').format(self.request.event.name),
+            'invoice_company': self.request.event.organizer.name,
```
raphaelm: This should be the name and company parts of the invoice address the *user* inputed, e.g. "John Doe" and "Sample Corporation"

src/pretix/control/views/event.py
```diff
@@ -400,6 +405,78 @@ def post(self, request, *args, **kwargs):
...
+            'mail_text_order_changed': ['event', 'url', 'invoice_name', 'invoice_company'],
+            'mail_text_order_expire_warning': ['event', 'url', 'expire_date', 'invoice_name', 'invoice_company'],
+            'mail_text_waiting_list': ['event', 'url', 'product', 'hours', 'code']
+        }
+
+    @cached_property
+    def dummy_data(self):
+        return {
+            'event': self.request.event.name,
+            'total': 100,
```
raphaelm: Let's use a Decimal like 42.23 here for showing a more realistic example.

src/pretix/control/views/event.py
```diff
@@ -400,6 +405,78 @@ def post(self, request, *args, **kwargs):
...
+            'currency': self.request.event.currency,
+            'date': date_format(now() + timedelta(days=7), 'SHORT_DATE_FORMAT'),
```
raphaelm: We also need to evaluate this in some lazy way because the date format depends of the locale. Do you have a good idea?

src/pretix/control/views/event.py
```diff
@@ -400,6 +405,78 @@ def post(self, request, *args, **kwargs):
...
+            'paymentinfo': _('Please transfer money to this bank account: BANK-ACCOUNT-0001'),
+            'payment_info': _('Please transfer money to this bank account: BANK-ACCOUNT-0001'),
+            'url': build_absolute_uri('presale:event.index', kwargs={
+                'event': self.request.event.slug,
+                'organizer': self.request.event.organizer
+            }),
+            'invoice_name': _('Invoice for {}.').format(self.request.event.name),
+            'invoice_company': self.request.event.organizer.name,
+            'orders': 'ORDER-0001\nORDER-0002\nORDER-0003\nORDER-0004',
+            'expire_date': date_format(now() + timedelta(days=15), 'SHORT_DATE_FORMAT'),
```
raphaelm: Same as above

src/pretix/control/views/event.py
```diff
@@ -400,6 +405,78 @@ def post(self, request, *args, **kwargs):
...
+            'product': _('Admission Ticket'),
+            'hours': 12,
```
raphaelm: We could read the actual value from the waiting list settings here.

src/pretix/control/views/event.py
jlwt90: I can probably separate static data with locale related data I think.
The line`with translation.override('de'):` should be able to handle the locale based date format if we specify the formats in `pretix/helpers/formats/de/formats.py` etc.

We currently have format mappings for js file only.
```
./pretix/static/jsi18n/de/djangojs.js:    "SHORT_DATE_FORMAT": "d.m.Y",
./pretix/static/jsi18n/de-informal/djangojs.js:    "SHORT_DATE_FORMAT": "d.m.Y",
./pretix/static/jsi18n/en/djangojs.js:    "SHORT_DATE_FORMAT": "Y-m-d",
```

jlwt90:
@raphaelm I'll be on leave for a few days.  I'll update my source code later 😄

src/pretix/control/views/event.py
raphaelm: That javascript is a special artifact, I'm not even sure why it's there. The translation from SHORT_DATE_FORMAT to the different formats is done within Django, not on our side.

jlwt90: got it! I realised that the problem comes from `@cached_property`. Since I cached the placeholder `{date}` and `{expire_date}` as `str`, the date value didn't change accordingly.

src/pretix/control/views/event.py
jlwt90: thanks for your reminder! I checked almost all the event settings and couldn't find anything in `control` namespace. I thought they were event organiser's name & company (but I couldn't differentiate name & company in this case lol)

src/pretix/control/views/event.py
jlwt90: Although it sounds a little bit naive, I would like to ask a question regarding this setting.
I couldn't find the corresponding key-value pair in `pretixbase_eventsetting` table for `self.request.event.settings.waiting_list_hours`. How do I know where this setting is stored?
Would you mind giving me the direction?

jlwt90:
>Did I already mention that I like your code a lot? 😉 I did not yet get around to do a final review of all the code and the tests but I pulled the branch and played around with it.

Thanks! 😃 

>I found a few misunderstandings about the placeholders. They are none of your fault, its just that we documented them very poorly so far. That's also on my todo list to change…

Yes... I was actually quite confused with the meaning of some placeholders. Thank you for your clarification.

>Also, apparently my assumptions about the languages were not all correct. I have an event with en and de_Informal activated and the fields are named mail_text_resend_all_links_0 and mail_text_resend_all_links_2. Therefore, your checks don't work. If you see an easy solution to this, go ahead, but otherwise I'll deal with this mess ;)

My bad 😞  I didn't test with `en` and `de-informal`. I added the following lines to create a simple mapping for locale indices and name.
```
@cached_property
    def supported_locale(self):
        locales = {}
        for idx, val in enumerate(settings.LANGUAGES):
            if val[0] in self.request.event.settings.locales:
                locales[str(idx)] = val[0]
        return locales
```
Now we can check whether the locale is valid for this event but then I realised that my `<pre>` block generation didn't work coz the logic relied on `for loop counter`. Therefore, I changed the JSON with this format:
```
{
    'item': 'mail_text_order_free',
    'msgs': {
        'en': 'xxx',
        'de-informal': 'xxx'
    }
}
```
and some minor changes in js & html.
I guess the fix is kinda simple! 🔨 🔨  Is there any concerning point in this part? Feel free to let me know.

src/pretix/control/views/event.py
raphaelm: It's not naive at all :) It will only be written into the ``pretixbase_eventsetting`` when it is changed for the first time. Until then, ``self.request.event.settings.waiting_list_hours`` will give the default value.

To be correct, the default value is determined in a multi-stage way. We just this week removed this concept from pretix itself and put it into a separate library. If you want to read more about it, go here: https://django-hierarkey.readthedocs.io/en/latest/concept.html

raphaelm:
Everything you say in your comment sounds perfectly fine. I'll try to find time later today or tomorrow for a final review! :)

raphaelm:
Hey @jlwt90,

I played around with it for a while now and everything works like a charm :sparkles: 

You did an amazing job here. This is a very good patch on so many levels, and I honestly believe I could not have done this any better, instead, it would probably would have turned out worse if I had done it. Thank you so much!

In case you are interested in taking on another issue, go have a look around if something fits your taste, otherwise I have a few suggestions: #297 and and #54 would be interesting django-y things that aren't too large. Finishing #102 could be a rather quick fix for someone as experienced as you seem to be. On the frontend side, there is the huge task of #100 where I could take some help, although I wouldn't expect from anyone to complete it, rather to take it small steps forward.

All the best,
Raphael

jlwt90:
>You did an amazing job here. This is a very good patch on so many levels, and I honestly believe I could not have done this any better, instead, it would probably would have turned out worse if I had done it. Thank you so much!

Thanks! I won't be able to finish this feature without your great help 😄 
#297 looks interesting to me but I don't have any experience on android application (no problem with java coz I mostly work on java applications in my work). Do you think the task is suitable for me?

Issue #444, opened 2017-03-31
Add alternative text to the top-right navigation


raphaelm:
**Hey!** If you ever wanted to contribute to an Open Source project, but didn't know how, you just found a perfect place to start. We try to write this issue as instructive as possible, but as we're humans we almost certainly forgot to write something important in here. If you want to try to tackle this issue and get stuck at _any_ point, please do not hesitate and ask us, either here on GitHub or via email at support@pretix.eu. We'll be very happy to help! 

We think this issue is very suitable to people who are just starting with Python or Django development or who never contributed to Open Source before. We'd like to ask people with wide experience in these areas to leave this issue to beginners. Thanks! (If you want something more advanced, just drop us a line, we'll find something :wink:)

## How to get started

First of all, you need pretix running locally on your machine. We have a guide on the developer installation [over here](https://docs.pretix.eu/en/latest/development/setup.html). Play around with it and get a feeling for what pretix is and how it works :) Try to find the places in the interface mentioned in this issue where something needs to change. Maybe try to find the same places in the code? Just searching for characteristic words can be a powerful tool to find your way around.

## What this is about

If you look at pretix' backend (at /control/), you see in the top-right corner a link to your profile settings and a logout link. Also, plugins can add more links there. Currently, only the Log Out link has a ``title=""`` attribute that helps users who do not understand the icons (or screenreaders) to understand where the link goes. It would be nice to have this for all of them!

## How to do this

* Go to ``pretix/control/templates/pretixcontrol/base.html``. As a warm-up, add a title attribute to the user page link.
* Look at the ``{% for nav in nav_topbar %}`` loop. Change it such that the ``<a>`` tags have a ``title`` attribute based on the ``nav.title`` property.
* Go to ``pretix/control/signals.py`` and change the documentation of ``nav_topbar`` to indicate that this attribute can now be set by plugins.

## I want to give it a try!

You're still reading, nice! If you want to try working on this, please comment below, so everyone else knows this is being looked at and can search for a different place to contribute. 

[...]

PR #457, opened 2017-04-06
Add alternative text to the top-right navigation: Issue #444

morrme:


raphaelm:
Perfect, thank you very much!

PR #458, opened 2017-04-06
Include some missing security headers

BenBE:
This change adds the following security headers:
* X-Content-Type-Options to prevent content type sniffing
* Referrer-Policy to prevent leaking referrer information when navigating away from the instance

### DIFF
--- a/deployment/docker/nginx.conf
+++ b/deployment/docker/nginx.conf
@@ -22,9 +22,11 @@ http {
 
 	include /etc/nginx/mime.types;
 	default_type application/octet-stream;
+	add_header X-Content-Type-Options nosniff;
 
 	access_log /var/log/nginx/access.log private;
 	error_log /var/log/nginx/error.log;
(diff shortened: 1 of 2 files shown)


raphaelm:
Thank you! :)

Would you mind also adding them to the nginx config snippet in ``doc/admin/installation/manual_smallscale.rst``? Docker is not the only way people use pretix.

BenBE:
There's substantial difference between the Docker sample and the snippet provided in the configuration guide. E.g. the custom log format and explicit compression rules are missing in the manual configuration.

raphaelm:
> There's substantial difference between the Docker sample and the snippet provided in the configuration guide. E.g. the custom log format and explicit compression rules are missing in the manual configuration.

That's partly because one is meant as a complete config and one as a snippet. However, people use the snipped as a complete config, so you've got a point there.

Anyway, thanks for this PR!

Issue #449, opened 2017-03-31
Show and change the locale of an order


raphaelm:
**Hey!** If you ever wanted to contribute to an Open Source project, but didn't know how, you just found a perfect place to start. We try to write this issue as instructive as possible, but as we're humans we almost certainly forgot to write something important in here. If you want to try to tackle this issue and get stuck at _any_ point, please do not hesitate and ask us, either here on GitHub or via email at support@pretix.eu. We'll be very happy to help! 

We think this issue is very suitable to people who are just starting with Python or Django development or who never contributed to Open Source before. We'd like to ask people with wide experience in these areas to leave this issue to beginners. Thanks! (If you want something more advanced, just drop us a line, we'll find something :wink:)

## How to get started

First of all, you need pretix running locally on your machine. We have a guide on the developer installation [over here](https://docs.pretix.eu/en/latest/development/setup.html). Play around with it and get a feeling for what pretix is and how it works :) Try to find the places in the interface mentioned in this issue where something needs to change. Maybe try to find the same places in the code? Just searching for characteristic words can be a powerful tool to find your way around.

## What this is about

Every ticket order in pretix has a locale associated. This is the language that was active for the user when the user created the order and it is the language that will be used whenever emails are sent to the user etc.

Currently, it is neither possible for the organizer to know which language that is, nor to change it. Both would be very useful, for example if the organizer wants to communicate with a user through another channel.

## How to do this

* Display the order locale in the order details section of ``pretix/control/templates/pretixcontrol/order/index.html``.
* Create a new view and a new form that allow to change the locale. You can get inspiration from ``OrderContactChange`` (``pretix/control/views/orders.py``) which changes the email address of an order and is therefore quite similar.

## I want to give it a try!

You're still reading, nice! If you want to try working on this, please comment below, so everyone else knows this is being looked at and can search for a different place to contribute. 

First, [fork](https://github.com/pretix/pretix/fork) the repository to your own GitHub account. Then clone it locally (or [set it as your remote](https://help.github.com/articles/configuring-a-remote-for-a-fork/) if you already cloned it earlier) and prepare your pull requests. There are a number of tutorials on the Pull Request workflow online, for example this [short version](https://guides.github.com/introduction/flow/) by GitHub itself and this [more in-depth tutorial](https://www.digitalocean.com/community/tutorials/how-to-create-a-pull-request-on-github). We're excited to see your pull request! 

[...]

PR #459, opened 2017-04-07
Issue #449: Display and change order locale

rahmonov:
Hi! This is the PR for #449 . 

Added ability to display the order locale in the order details section:

image

Added ability to change the order locale:

image

Feedback and comments are much appreciated. Thanks :)


src/pretix/control/forms/orders.py
```diff
@@ -118,3 +118,9 @@ class OrderContactForm(forms.ModelForm):
     class Meta:
         model = Order
         fields = ['email']
+
+
+class OrderLocaleForm(forms.ModelForm):
+    class Meta:
+        model = Order
+        fields = ['locale']
```
rixx: This results in `locale` being presented as a normal text input, so that I could put down `asdfghklö` as locale. I think it would be preferable to present a `ChoiceField` similar to the one of the [EventWizardBaseForm](https://github.com/pretix/pretix/blob/master/src/pretix/control/forms/event.py#L45).

src/pretix/control/templates/pretixcontrol/order/change_locale.html
```diff
@@ -0,0 +1,27 @@
...
+{% load i18n %}
+{% load bootstrap3 %}
+{% block title %}
+    {% trans "Change locale information" %}
+{% endblock %}
+{% block content %}
+    <h1>
+        {% trans "Change locale information" %}
+    </h1>
+
```
rixx: I think it would be nice to add a short explanation to the view, explaining the results of changing an order locale. I don't think that this is immediately obvious to a user.

src/tests/control/test_orders.py
```diff
@@ -105,6 +105,18 @@ def test_order_set_contact(client, env):
...
 @pytest.mark.django_db
+def test_order_set_locale(client, env):
+    q = Quota.objects.create(event=env[0], size=0)
+    q.items.add(env[3])
+    client.login(email='dummy@dummy.dummy', password='dummy')
+    client.post('/control/event/dummy/dummy/orders/FOO/locale', {
+        'locale': 'de'
+    })
+    o = Order.objects.get(id=env[2].id)
+    assert o.locale == 'de'
```
rixx: Yay, tests! A second one showing the result of `POST`ing an illegal locale would be nice, too.

rahmonov: @rixx Thank you for the comments! Will update the PR soon

src/pretix/control/forms/orders.py
raphaelm: Yes. The choices need to come from ``event.settings.locales``, which makes this a bit more complicated, you will probably need a custom ``__init__`` method on that form that sets the choices on the field (or creates the field dynamically).

src/pretix/control/templates/pretixcontrol/order/change_locale.html
raphaelm: 👍

src/pretix/control/forms/orders.py
rahmonov: @raphaelm where is `event.settings.locales`? I don't seem to find it or understand what that is

raphaelm: Given an event ``event``, ``event.settings.locales`` is a list of languages that are supported for this event.

You could do for example the following (untested):

    class OrderLocaleForm(forms.ModelForm):
        class Meta:
            model = Order
            fields = ['locale']

            self.fields['locale'].choices = self.instance.event.settings.locales

rahmonov:
@rixx @raphaelm I updated the code as you guys suggested. Please, take a look and let me know what you think

raphaelm:
Hey @rahmonov,

thank you very much!

Cheers
Raphael

PR #461, opened 2017-04-10
Wrong checkbox order

jfwiebe:
The boxes vor adding and editing vouchers were switched for new users.


raphaelm:
:man_facepalming: Thanks!

PR #462, opened 2017-04-10
Added missing semicolon to docker nginx config

jfwiebe:
...which results in docker container setup not starting properly.
You can see the log message setting up the container and entering
`docker exec -it --user root pretix.service cat /var/log/nginx/error.log`


raphaelm:
Good catch, thanks.

PR #466, opened 2017-04-13
Add attendee email field

raphaelm:
- [x] Include field in exports
- [x] Tests
- [ ] i18n
- [ ] Later: Ticket download links?


src/pretix/control/forms/event.py
```diff
@@ -274,6 +289,11 @@ def clean(self):
             raise ValidationError({
                 'attendee_names_required': _('You cannot require specifying attendee names if you do not ask for them.')
             })
+        if data['attendee_emails_required'] and not data['attendee_emails_asked']:
+            raise ValidationError({
+                'attendee_emails_required': _('You cannot require specifying attendee emails if you do not ask for '
```
rixx: "You have to ask for attendee emails if you want to make them required." or something like that?

src/pretix/control/forms/event.py
```diff
@@ -240,6 +240,21 @@ class EventSettingsForm(SettingsForm):
         help_text=_("Require customers to fill in the names of all attendees."),
         required=False
     )
+    attendee_emails_asked = forms.BooleanField(
+        label=_("Ask for attendee e-mails"),
+        help_text=_("Ask for an e-mail address for all tickets which include admission to the event. Important: For "
```
rixx: I don't think this is very clear … maybe more of a positive wording?

```
Attendee emails addresses are required regardless of this setting for confirmation links. This setting adds the possiblity to ask for a second email address, eg for group orders.
```

… and/or rename the setting label - if you have to explain that you mean something different, it might be better to rename. `Ask for secodary email address` or something like that?

raphaelm: Hmmm, I get your point and this should totally be improved but I believe your suggestion is even more unclear. It's hard to get the distinction between **purchaser** emails and **attendee** emails across and I think "Attendee emails addresses are required regardless of this setting for confirmation links" makes it even more confusing.

rixx: I get what you're saying - I think my version was confusing because I was kind of confused about the purpose of this (which adds to my point: maybe the label "Ask for attendee e-mails" itself should be changed, too).

raphaelm: Hmmm. I see, but do you have an idea for a new label? "Ask for email addresses of all attendees" isn't much clearer, maybe "Ask for additional attendee e-amil addresses"?

PR #464, opened 2017-04-12
Upgrade to Django 1.11

rixx:
WIP! Needs at the very least a `django-i18nfield` release.

Issue #456, opened 2017-04-05
Allow products to be excluded from ticket-generation


pc-coholic:
It would be nice to have a checkbox for every product to dis-/enable the generation of tickets (PDF, passbook, etc.).

Usecase: In addition to an entrance-ticket for an event, also offer a donation-product. When generating the downloadable tickets, the donations do not trigger a (useless) printout with a barcode which no one is going to use.

So basically we could either add a "is a downloadable ticket"-property to each product or add a "generate only admission tickets"-checkbox to the ticket generation process.

raphaelm:
>  "generate only admission tickets"-checkbox to the ticket generation process.

I believe this is the way to go.

hd719:
Hello I am new to Open source can I work on this?

raphaelm:
Sure!

First of all, you need pretix running locally on your machine. We have a guide on the developer installation [over here](https://docs.pretix.eu/en/latest/development/setup.html). Play around with it and get a feeling for what pretix is and how it works :) Try to find the places in the interface mentioned in this issue where something needs to change. Maybe try to find the same places in the code? Just searching for characteristic words can be a powerful tool to find your way around.

If you have any questions, just let me know!

hd719:
Hey @raphaelm sorry, but I cannot work on this.

raphaelm:
Hi @hd719, sure, we'll find someone else :)

aph3rson:
https://github.com/pretix/pretix/issues/443#issuecomment-298418044
@raphaelm I can take a crack at it. The description is a bit vague, though - what I'm getting is that you'd like products to have a property that tickets should be generated for them (so donations/etc. don't generate tickets). I'll need to look through the code a bit more to determine the program flow around when tickets are generated, though.

raphaelm:
@aph3rson 

> The description is a bit vague, though - what I'm getting is that you'd like products to have a property that tickets should be generated for them (so donations/etc. don't generate tickets)

Nearly correct: There already *is* a property ``Item.admission`` that we want to re-use here. You would now need to:

* Add a setting to ``TicketSettingsForm`` that says "Generate tickets for non-admission products". For backwards compatibility, you should set the default to True in ``pretix/base/settings.py``.
* Disable the download button in ``pretix/presale/templates/pretixpresale/event/fragment_cart.html`` if the setting is unchecked and a product is not an admission product
* Enforce the new setting in ``OrderDownload`` (``pretix/presale/views/order.py``) when downloading a single ticket
[...]

PR #481, opened 2017-04-30
#297: pretixdroid: Show metrics in the control panel

jlwt90:
Hi,
This is the PR for #297.

I added dashboard widget and checkin page under orders section in pretix control. 
There are 2 points to confirm:
1) I added filter to check whether the order position links to an item related to admission ticket. Is my understanding correct?
https://github.com/jlwt90/pretix/blob/602274fb6e2fe456e0a87d3b53cbe5ead64d7423/src/pretix/control/views/checkin.py#L27

2) I used a dict to apply multiple filters to the query to avoid [this](http://stackoverflow.com/a/28253623) issue. Is there any better way to do it?
https://github.com/jlwt90/pretix/blob/602274fb6e2fe456e0a87d3b53cbe5ead64d7423/src/pretix/control/views/checkin.py#L16

Feel free to leave some comments ;) Thanks!

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/checkin/index.html
+++ b/src/pretix/control/templates/pretixcontrol/checkin/index.html
@@ -12,7 +12,15 @@
                 <option value="1" {% if request.GET.status == "1" %}selected="selected"{% endif %}>{% trans "Checked in" %}</option>
                 <option value="0" {% if request.GET.status == "0" %}selected="selected"{% endif %}>{% trans "Not checked in" %}</option>
             </select>
-            <input type="text" name="item" class="form-control" placeholder="{% trans "Search item" %}" value="{{ request.GET.item }}">
+            <select name="item" class="form-control">
+                <option value="">{% trans "All products" %}</option>
+                {% for item in items %}
+                    <option value="{{ item.id }}"
+                            {% if request.GET.item|add:0 == item.id %}selected="selected"{% endif %}>
+                        {{ item.name }}
+                    </option>
+                {% endfor %}
+            </select>
             <input type="text" name="user" class="form-control" placeholder="{% trans "Search user" %}" value="{{ request.GET.user }}">
             <button class="btn btn-primary" type="submit">{% trans "Filter" %}</button>
         </form>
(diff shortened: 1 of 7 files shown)


src/pretix/control/templates/pretixcontrol/checkin/index.html
```diff
@@ -0,0 +1,74 @@
+{% extends "pretixcontrol/event/base.html" %}
+{% load i18n %}
+{% load eventurl %}
+{% load urlreplace %}
+{% block title %}{% trans "Check-Ins" %}{% endblock %}
+{% block content %}
+    <h1>{% trans "Check-ins" %}</h1>
```
raphaelm: Here it's 'Check-ins', two lines above you used a capital I. I think we go for lowercase I consistently?

src/pretix/control/templates/pretixcontrol/checkin/index.html
```diff
@@ -0,0 +1,74 @@
...
+                    </thead>
+                    <tbody>
+                    {% for e in entries %}
+                        {% with e.checkins.first as checkin %}
+                        <tr>
+                            <td>
+                                <strong><a href="{% url "control:event.order" event=request.event.slug organizer=request.event.organizer.slug code=e.order.code %}"
+                        >{{ e.order.code }}</a></strong>
+                            </td>
+                            <td>{{ e.order.email }}</td>
```
raphaelm: Let's also display an attendee_name here, if it is set.

src/pretix/control/views/dashboards.py
```diff
@@ -169,6 +169,21 @@ def shop_state_widget(sender, **kwargs):
 
 
 @receiver(signal=event_dashboard_widgets)
+def checkin_widget(sender, **kwargs):
+    size = OrderPosition.objects.filter(order__event=sender, order__status='p', item__admission=True).count()
+    checked = Checkin.objects.filter(position__order__event=sender, position__order__status='p', position__item__admission=True).count()
```
raphaelm: In theory, we could have more than one checkin per order position (although the API currently doesn't allow that), in which case this could return something like 20/10. Should we rather count OrderPositions in both cases and assert ``checkins__isnull=False`` in one case?

src/pretix/control/templates/pretixcontrol/checkin/index.html
jlwt90: oh i didn't notice that 😛  thanks!

jlwt90:
>Yes and no. In theory scanning makes most sense for admission tickets, but in practice one might also scan tickets for example when giving out pre-ordered t-shirts which are no admission tickets. I'm not sure what to do here, maybe disable the filter for now and re-introduce it based on the setting that will be introduced in #456. Eventually, we will move to something more complicated which I briefly described in this comment, but that will take another while.

Got it! I added `item__name` to the checkin table in this commit coz I think it would be useful if we show more than one line for an order.

>The way you chose is absolutely fine. However, did you actually experience the linked problem? I haven't played around with it, but from the code I believe we wouldn't be affected because checkins__isnull is the single only lookup to a (One|Many)ToMany field in the query.

My bad. Made a mistake when I was constructing the queryset :)

jlwt90:
added tests as well :)

raphaelm:
Hi,

thank you very much! This is going to be the best-tested list view we have :) Code looks fine for me so far.

Some remarks:
* Somehow, the tests for ordering fail, but only on PostgreSQL. I'm not yet sure why. If you want to, I can debug that.
* I think I'd prefer the item filter to be a ``<select>`` list, similarly to the order view. We normally don't have that many products and people shouldn't be forced to remember their exact names :wink: What do you think?
* We [recently](https://pretix.eu/about/de/blog/20170502-release-130/) introduced add-on products. A consequence of this is that in some cases, the name of an attendee is not saved directly on the ``OrderPosition`` but on its "parent". Therefore, we might want to display that here too in case we know it. This would require a ``select_related('addon_to')`` on the queryset and then something like ``p.attendee_name|default:p.addon_to.attendee_name`` in the template. For ordering, we use ``qs.order_by(Coalesce('attendee_name', 'addon_to__attendee_name'))`` elsewhere already.

Let me know if you have any questions!

Cheers
Raphael

jlwt90:
@raphaelm I successfully set up postgreSQL on my local and found the same error.
The mismatches in expected order for the 7 tests are caused by null value ordering. By default, null value will be put at the end
e.g. for the test `test_checkins_list_mixed[status=&item=Ticket&user=&ordering=checkins__datetime-expected3]`, 
> E         Full diff:
E         - ['A3Ticket', 'A1Ticket', 'A2Ticket'] <- result
E         + ['A2Ticket', 'A3Ticket', 'A1Ticket'] <- expected

since `A2Ticket` has null `checkins__datetime`, it was placed at the end of the result set.
Apparently there is no good way to keep the same ordering for different db at this moment. (Please tell me if there is! 😄 ) I find that this will be a new feature in django 1.11.
https://docs.djangoproject.com/en/dev/releases/1.11/
>Added the nulls_first and nulls_last parameters to Expression.asc() and desc() to control the ordering of null values.

Maybe i should skip those order checking at this moment?

raphaelm:
> Maybe i should skip those order checking at this moment?

Yeah, comment it out. I'll leave a note in #464 so we don't forget it.

As I mentioned above already, I just merged #483. I'd suggest to limit both the dashboard counter and the list view to items with ``admission=True`` if and only if the newly introduced ``event.settings.ticket_download_nonadm`` setting is false. I think that makes sense for most usecases.

jlwt90:
@raphaelm all done! Could you check whether the changes look good?

raphaelm:
It's perfect, thank you very much!

raphaelm:
Should you want to continue, there is a small part of #102 that hasn't been picked up for years but would be a quite useful feature without too much effort :)

jlwt90:
Thanks! I'll take that :)

PR #483, opened 2017-05-05
Allow products to be excluded from ticket-generation

### LINKED ISSUE #456: Allow products to be excluded from ticket-generation

aph3rson:
Per #456, allow products designated as non-admission to be excluded from ticket generation.

Includes the following:  
* Events setting `ticket_generation_nonadm` for generating tickets for non-admission products (defaults to true for backwards compatibility)  
* Logic to remove download button for non-relevant line items in order view  
* Error for non-admission items (if not allowed) in single ticket download and multi-ticket download.

Thanks to @raphaelm with https://github.com/pretix/pretix/issues/456#issuecomment-298421117 for assistance on what needed to be done here.
Closes #456.

### DIFF
--- a/src/pretix/base/settings.py
+++ b/src/pretix/base/settings.py
@@ -163,6 +163,10 @@ DEFAULTS = {
         'default': 'False',
         'type': bool
     },
+    'ticket_download_nonadm': {
+        'default': 'True',
+        'type': bool
+    },
     'last_order_modification_date': {
         'default': None,
         'type': datetime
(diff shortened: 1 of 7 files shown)


raphaelm:
This looks really great, no comments on the code! :) Could you also add the check to ``pretix/plugins/ticketoutputpdf/ticketoutput.py``? It overrides ``generate_order`` because it generates multi-page PDFs instead of ZIP files.

aph3rson:
I've gone ahead and added this in 93ee8da.

raphaelm:
Thank you very much, looks great!

PR #487, opened 2017-05-05
Marked webfonts as binary files

### LINKED ISSUE #486: [Git] Cloning on Windows doesn't recognize .woff/etc as binary
As in the title, web font files aren't explicitly marked as binary in the `.gitattributes` file. This should be a quick fix, and only deals with cloning to a Windows-based environment.

aph3rson:
`.gitattributes` is used in the cloning process for, among other things, determining if a file is text, and needs to have line-endings converted. Doing this to the binary webfont files will cause issues with using them.
When merged, fixes #486.


raphaelm:
Thanks, good to know!

PR #488, opened 2017-05-06
fixed bug for same dates, added unit check for daterange

scabux:
Hey,

in this commit I fixed a bug in `daterange.py`. `daterange(df,dt)` now does what it should if the two dates are equal: It returns the string for one day.
I also added a (probably _slightly_ overkill) unit test.

Cheers
Marvin

### DIFF
--- a/src/pretix/helpers/daterange.py
+++ b/src/pretix/helpers/daterange.py
@@ -6,12 +6,16 @@ def daterange(df, dt):
     lng = get_language()
 
     if lng.startswith("de"):
-        if df.year == dt.year and df.month == dt.month:
+        if df.year == dt.year and df.month == dt.month and df.day == dt.day:
+            return "{}".format(_date(df, "j. F Y"))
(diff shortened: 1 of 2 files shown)


raphaelm:
Awesome! Unit tests are never overkill ;) However, they *might* have a small problem: ``translation.activate('de')`` changes the language of the global execution context. Depending on the test that runs *after* the last test of this set, this might or might not be a problem. To be safe, we can re-write it like this::
```
def test_same_day_german():
    with translation.override('de'):
        df = date(2003, 2, 1)
        assert daterange(df, df) == "1. Februar 2003"
```

scabux:
> `translation.activate('de')` changes the language of the global execution context
Oops, that was stupid. 🙈 Thank you very much! I changed the lines accordingly.

raphaelm:
> Oops, that was stupid

Not at all, how could you know. In the end, we change a lot of global things in the tests and in some cases, py.test knows how to revert them and in some cases it doesn't…

PR #494, opened 2017-05-17
Added configurable text for organizer page

scabux:
This PR adds a configurable text that can be displayed under an organizer's name on the corresponding page. Things to follow in near future:

- organizers might be given language options (overwritten by events' language options) allowing language selection at an organizer's front end page
- unit tests will be added after some more work on the organizer page
- `rich_text` filter tag should recognize single newlines


raphaelm:
Thanks! :)

PR #497, opened 2017-05-18
added a missing string and translation for it

eth3rgeist:
Found a not translated string. Should be translated now.


raphaelm:
Top, thanks! :)

PR #500, opened 2017-05-21
Fix calling of object attribute on tuple

lhhong:
Test data creation was broken as t was a tuple instead of a Team object.


raphaelm:
Thank you!

PR #501, opened 2017-05-21
Refactor paymentinfo to payment_info, resolve #499

### LINKED ISSUE #499: Fix placeholder confusion
There is payment_info in some mails and paymentinfo in others. We should unify this to payment_info.

lhhong:


raphaelm:
Hi,

thanks a lot! However, we have to deal with backwards compatibility as there are many events in the system that have the now-wrong placeholder in their email templates. Therefore, we need to either

1. Write a [data migration](https://docs.djangoproject.com/en/1.11/topics/migrations/#data-migrations) that replaces the placeholders in the settings of all events
2. Make sure the old one continues to work in all places, i.e. is still being replaced and passes validation

I'd prefer to go for option 1 to keep things clean. Would you like to tackle this as well or shall I go forward with it?

lhhong:
Alright, I will tackle this.

lhhong:
Included the data migration. I believe the inconsistency was only in mail_text_order_placed. Hope I didn't miss out other fields.

raphaelm:
This looks good, thank you very much!

PR #509, opened 2017-05-26
Read payment details from MT940 NS tags too

cluck:
Payment details from raiffaisen.ch are only available in non-SWIFT (NS) tags. Thus added support for these tags in mt940. Now use the newer version and also parse these fields when available.


raphaelm:
Thanks a lot! :)

PR #513, opened 2017-06-02
Read-only REST API

raphaelm:
- [x] Authentication
- [x] Make DRF docs generation work with hierarchical docs
- [x] viewset-level permission handling
- [x] Add read-only viewsets for all important models
  - [x] Event
  - [x] Organizer
  - [x] Item
  - [x] Invoice
  - [x] ItemCategory
  - [x] ItemVariation
  - [x] Order
  - [x] Question
  - [x] Quota
  - [x] Voucher
  - [x] WaitingListEntry
  - [x] OrderPosition
    - [x] Attendee names of addon products
  - [x] CheckIn
- [x] Allow plugins to hook themselves into the API
- [ ] put event property into all serializers?
- [x] Ticket download
- [x] Invoice download
- [x] Deal with i18nfields (read-only)
- [x] Fill docs with details
  - [x] Details on auth
  - [x] Details on pagination
  - [x] Details on versioning
  - [x] Style docs (tables, see items and questions)
  - [x] disable drf docs
  - [x] events
  - [x] organizers
  - [x] categories
  - [x] items
  - [x] invoices
  - [x] orders
  - [x] orderpositions
  - [x] questions
  - [x] quotas
  - [x] vouchers
  - [x] waitinglistentries
- [x] Tests
  - [x] auth
  - [x] permissions
  - [x] events
  - [x] organizers
  - [x] categories
  - [x] items
  - [x] invoices
  - [x] orders
  - [x] orderpositions
  - [x] questions
  - [x] quotas
  - [x] vouchers
  - [x] waitinglistentries


src/pretix/api/serializers/event.py
```diff
@@ -0,0 +1,11 @@
+from rest_framework import serializers
+
+from pretix.base.models import Event
+
+
+class EventSerializer(serializers.ModelSerializer):
+    class Meta:
+        model = Event
+        fields = ('name', 'slug', 'live', 'currency', 'date_from',
```
rixx: `organizer` is not included, is that intentional? Since you don't provide `id` in `Event` and `Organizer`, you'll probably want either a nested serializer or a `SlugRelatedField` here (and on every other model referencing `Event`, too).

raphaelm: Yes, it's intentional. As I'm currently working mostly offline, what you see here is not my latest state. In the current version, you currently can *only* get events as a nested resource, e.g. ``/organizers/<slug>/events/`` and ``/organizers/<slug>/events/<slug>/items/<id>/``. There are some very convincing reasons for this (i.e. its what you want most of the time and the required lookups to enforce permissions would be very compley otherwise), but there are of course some downsides as well. This is one of quite a few API design questions I wanted to talk to you about after the long weekend ;)

PR #518, opened 2017-06-13
Allow to add a prefix when generating voucher codes

rixx:
2017-06-13-104537_1387x394_scrot

### DIFF
--- a/src/pretix/base/models/vouchers.py
+++ b/src/pretix/base/models/vouchers.py
@@ -13,8 +13,10 @@ from .event import Event
 from .items import Item, ItemVariation, Quota
 
 
-def _generate_random_code():
+def _generate_random_code(prefix=None):
     charset = list('ABCDEFGHKLMNPQRSTUVWXYZ23456789')
+    if prefix:
+        return prefix + get_random_string(length=settings.ENTROPY['voucher_code'], allowed_chars=charset)
     return get_random_string(length=settings.ENTROPY['voucher_code'], allowed_chars=charset)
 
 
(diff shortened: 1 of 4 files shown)


src/pretix/base/models/vouchers.py
```diff
@@ -13,8 +13,10 @@
 from .items import Item, ItemVariation, Quota
 
 
-def _generate_random_code():
+def _generate_random_code(prefix=None):
     charset = list('ABCDEFGHKLMNPQRSTUVWXYZ23456789')
+    if prefix:
+        return prefix + get_random_string(length=settings.ENTROPY['voucher_code'] - len(prefix), allowed_chars=charset)
```
raphaelm: I'm not sure if we should subtract here or just keep the length. The default length is only 16 characters, so entropy gets quite low with a medium-long prefix already.

src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -105,11 +105,11 @@ $(function () {
     // Vouchers
     $("#voucher-bulk-codes-generate").click(function () {
         var num = $("#voucher-bulk-codes-num").val();
+        var prefix = $('#voucher-bulk-codes-prefix').val();
         if (num != "") {
-            $(".form-group:has(#voucher-bulk-codes-num)").addClass("has-error");
             var url = $(this).attr("data-rng-url");
             $("#id_codes").html("Generating...");
-            $.getJSON(url + '?num=' + num, function (data) {
+            $.getJSON(url + '?num=' + num + '&prefix=' + prefix, function (data) {
```
raphaelm: Should we escape the prefix in the URL in case it contains e.g. an ``&``?

src/pretix/base/models/vouchers.py
rixx: Yeah, I wasn't quite sure about that either, and I resolved that you'd probably expect a voucher code of the standard length, if you were just using that method. I'll bow to your wishes, though ;)

rixx: Adjusted.

rixx:
Updated to reflect your requests – tested with & and = in prefixes.

raphaelm:
Thanks a lot! :)

PR #520, opened 2017-06-15
Clarify mail subject prefix.

rixx:
I'll update translations once you've merged the other PR, to prevent rebase/merge conflicts.

### DIFF
--- a/src/pretix/control/forms/event.py
+++ b/src/pretix/control/forms/event.py
@@ -471,8 +471,8 @@ class InvoiceSettingsForm(SettingsForm):
 class MailSettingsForm(SettingsForm):
     mail_prefix = forms.CharField(
         label=_("Subject prefix"),
-        help_text=_("This will be prepended to the subject of all outgoing emails. This could be a short form of "
-                    "your event name."),
+        help_text=_("This will be prepended to the subject of all outgoing emails, formatted as [prefix]. "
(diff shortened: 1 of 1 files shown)


src/pretix/control/forms/event.py
```diff
@@ -471,8 +471,8 @@ class InvoiceSettingsForm(SettingsForm):
 class MailSettingsForm(SettingsForm):
     mail_prefix = forms.CharField(
         label=_("Subject prefix"),
-        help_text=_("This will be prepended to the subject of all outgoing emails. This could be a short form of "
-                    "your event name."),
+        help_text=_("This will be prepended to the subject of all outgoing emails, formatted as [prefix]. "
```
raphaelm: That's better than before, but I feel that one could still read this as "add the brackets your self". What do you think, fine-tune the help text even more ("This will be automatically formatted within brackets, e.g.") and/or add a RegexValidator that complains if you put them yourselves.

raphaelm: Hmmm, yeah, maybe the validator is too much. The JavaScript solution you suggested is also probably overkill (but it's nice, I'd merge it, but I wouldn't do it myself in your place). Maybe just replace the "formatted as [prefix]" with "The subject will be formatted as '[prefix] Subject…'" or something like this.

PR #523, opened 2017-06-18
Allow to pass prefix to generate_code, too

rixx:
Useful in plugins if you don't want to replicate the uniqueness loop.


raphaelm:
Good point!

Issue #517, opened 2017-06-12
Change navpill look in email template editor


rixx:
When editing email templates, there are nav pills provided below the i18n input, to switch between edit and preview mode.

To me, they looked like buttons, and I assumed at first that the "Edit" button was meant to save progress. I think this is because normally a very similar looking button is provided at the same relative location (lower right hand corner), and nav pills are normally left-aligned and to the top.

Off the top of my head, I'd probably expect an explicit switcher/toggle button between the two modes.

![2017-06-12-152940_1342x260_scrot](https://user-images.githubusercontent.com/2657547/27036069-f8a3d3a6-4f83-11e7-974f-e6a660fa00ed.png)

raphaelm:
Of something tab-like *above* the input

tydonk:
I'd like to give it a try.  I'm running in dev mode but I couldn't find this page, in browser or source code.  Could someone point me in the direction to start?  Thanks

rixx:
Under Settings > E-Mail > E-Mail content you can edit and preview the templates; that is where my screenshot is from.

The template is at `control/templates/pretixcontrol/event/mail_settings_fragment.html`

tydonk:
Thanks for the path.  I'm still not sure how to view this in the browser.  When I run the server in development mode and login at `localhost:8000/control/login` it takes me to the dashboard.  How do I view the front end in dev mode?  Sorry for the stupid question, I read through the Dev docs and didn't see an answer.

rixx:
It's not a stupid question at all! Since this is an event specific setting, you'll need to create a test event from the dashboard. You can use `python make_testdata.py` to generate an event with reasonable settings, products etc configured.

Also, if you see a way to make this clearer in the docs, don't hesitate to put it in a pull request!

tydonk:
As usual, user error was the issue :) I wasn't logging in as admin@localhost.  Should be good to go now.  Thanks!

tydonk:
Which do you like better?  IMO with the positioning in the top left, the nav pills look better than the tabs.

Tabs - ![screen shot 2017-06-21 at 11 49 02 pm](https://user-images.githubusercontent.com/13215301/27420821-7f10e5dc-56dc-11e7-986e-0d0f132ae1d3.png)

Pills - ![screen shot 2017-06-21 at 11 49 33 pm](https://user-images.githubusercontent.com/13215301/27420825-81b0b3e4-56dc-11e7-882c-2b163ae7200e.png)

[...]

PR #533, opened 2017-06-26
Changed navpills to navtabs in email template editor (#517)

tydonk:


raphaelm:
Hi!

First of all, thank you very much! I tried it, and I liked it a lot. There are two more things that would make it perfect:

* There is a unnecessarily large margin below the content of the new box and the bottom border of the new box. Can you get rid of this? (Be sure to test with both only one and with multiple languages active)
* I feel that the new border needs do be round, at least on all edges except the top-left one (which melds with the tabs)

Let me know if that causes any trouble :)

tydonk:
No problem!  Should be all set now.

raphaelm:
Yep, looks great. Thank you so much! :sparkles:

Issue #528, opened 2017-06-20
Make configuration path configurable


pennylane0815:
The configuration path /etc/pretix/pretix.cfg seems to be hardcoded.
As I am currently trying to install pretix on FreeBSD, where all user configuration is done in /user/local/etc, I placed the pretix configuration in /usr/local/etc/pretix. 
It did not use my configuration until i symlinked /usr/local/etc/pretix to /etc/pretix.
Having a --config flag or similar would greatly enhance that experience!

raphaelm:
A --config flag would be impractical in many situations (we'd somehow need to integrate it into various commands of django), but it would totally merge an option to override the config file location by an environment variable.

Volcyy:
Hello!
I'd be interested in this issue, but looking at the [source code of settings.py](https://github.com/pretix/pretix/blob/master/src/pretix/settings.py#L15-L20) it appears that if an environment variable `PRETIX_CONFIG_FILE` is set, it is used instead of the default of `/etc/pretix/pretix.cfg`. 

Would a documentation update be needed, mainly in [config.rst](https://github.com/pretix/pretix/blob/master/doc/admin/config.rst)?

raphaelm:
Oops. You totally caught me. Yep, a documentation update would be what is needed then. :D

PR #539, opened 2017-07-02
config.rst: Include environment variable in list of paths

SanketDG:
Closes https://github.com/pretix/pretix/issues/528


raphaelm:
Thanks!

PR #542, opened 2017-07-05
Feature/remove flag as language

mfattoru:
The language now is displayed with the local translation of the language

screenshot-20170705095318-1148x246

### DIFF
--- a/src/pretix/presale/context.py
+++ b/src/pretix/presale/context.py
@@ -63,5 +63,6 @@ def contextprocessor(request):
     ctx['html_foot'] = "".join(_html_foot)
     ctx['footer'] = _footer
     ctx['site_url'] = settings.SITE_URL
+    ctx['languages'] = settings.LANGUAGES
 
     return ctx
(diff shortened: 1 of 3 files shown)


src/pretix/presale/context.py
```diff
@@ -53,6 +54,8 @@ def contextprocessor(request):
             ctx['css_file'] = default_storage.url(request.event.settings.presale_css_file)
         ctx['event_logo'] = request.event.settings.get('logo_image', as_type=str, default='')[7:]
         ctx['event'] = request.event
+        for code in request.event.settings.locales:
+            ctx.setdefault('languages',[]).append( get_language_info(code) )
```
raphaelm: Hmm, I'm not really a fan of calling ``setdefault`` that many times, although it probably doesn't matter. I think a list comprehension would be cleaner:
```
ctx['languages'] = [
    get_language_info(code) for code in request.event.settings.locales
]
```
What do you think?

mfattoru: yes, your style looks cleaner than mine

raphaelm:
Wonderful! Congratulations on your very first pull request and thank you very much! :sparkles:

mfattoru:
Thank you for assisting me in my first pull request,it really put me in the mood of contributing to more projects. Didn't expect to find friendly and helpful people assisting.
Thank you so much

Issue #35, opened 2015-03-16
Microdata for search engines


raphaelm:
**Hey!** If you ever wanted to contribute to an Open Source project, but didn't know how, you just found a perfect place to start. We try to write this issue as instructive as possible, but as we're humans we almost certainly forgot to write something important in here. If you want to try to tackle this issue and get stuck at _any_ point, please do not hesitate and ask us, either here on GitHub or via email at support@pretix.eu. We'll be very happy to help! 

We think this issue is very suitable to people who are just starting with Python or Django development or who never contributed to Open Source before. We'd like to ask people with wide experience in these areas to leave this issue to beginners. Thanks! (If you want something more advanced, just drop us a line, we'll find something :wink:)

## How to get started

First of all, you need pretix running locally on your machine. We have a guide on the developer installation [over here](https://docs.pretix.eu/en/latest/development/setup.html). Play around with it and get a feeling for what pretix is and how it works :) Try to find the places in the interface mentioned in this issue where something needs to change. Maybe try to find the same places in the code? Just searching for characteristic words can be a powerful tool to find your way around.

## What this is about

Every event organized by pretix gets a small web page representing the event -- it's ticket shop. In the HTML code of this page we want to add microdata markup. Microdata is a structured way to tell search engines what the given page is about and allows them to understand the page better and do fancy things like showing the event location or date directly in the search results.

## How to do this

First, you need to read up a bit about structured data. For example, there is the [JSON-LD](http://json-ld.org/) data format that we probably want to use. Don't get discouraged by complicated-sounding words, just keep to the examples, they should be pretty clear in many cases. Google also has a lot of documentation and tools on the topic, for example:

* [A short JSON-LD guide](https://developers.google.com/schemas/formats/json-ld)
* [A number of examples specifically for events](https://developers.google.com/search/docs/data-types/events)
* [A test tool to check if your structured data is good](https://search.google.com/structured-data/testing-tool/u/0/)

Once you know how it should look like, you basically only have to put it in the right place, which is ``src/pretix/presale/templates/pretixpresale/event/index.html``, and fill it up with the correct data via template variables.

## I want to give it a try!

You're still reading, nice! If you want to try working on this, please comment below, so everyone else knows this is being looked at and can search for a different place to contribute. 

[...]

PR #546, opened 2017-07-08
Addition of all basics required for microdata

Joepriesto:
Microdata for event added to the head of the base event html template. #35


src/pretix/base/models/event.py
```diff
@@ -330,6 +330,21 @@ def get_payment_providers(self) -> dict:
                 providers[pp.identifier] = pp
         return providers
 
+
```
raphaelm: flake8 complains here, use only one blank line :)

src/pretix/base/models/event.py
```diff
@@ -330,6 +330,21 @@ def get_payment_providers(self) -> dict:
...
+        eventdict = {"@context": "http://schema.org", "@type": "Event"}
+        eventdict["location"] = {"@type": "Place",
+                                 "address": str(self.location)}
+        eventdict["startDate"] = str(self.date_from)
+        eventdict["name"] = str(self.name)
+        eventdict["endDate"] = str(self.date_to)
+
+        return json.dumps(eventdict)
+
+
```
raphaelm: same again, only one blank line :)

src/pretix/base/models/event.py
```diff
@@ -330,6 +330,21 @@ def get_payment_providers(self) -> dict:
                 providers[pp.identifier] = pp
         return providers
 
+
+    @property
+    def get_event_microdata(self):
```
raphaelm: Drop the ``@property`` or name it to ``event_microdata``. I don't really care, both are equally fine, but a property should never be named with a verb like ``get_…``.

src/pretix/base/models/event.py
```diff
@@ -330,6 +330,21 @@ def get_payment_providers(self) -> dict:
...
+        import json
+
+        eventdict = {"@context": "http://schema.org", "@type": "Event"}
+        eventdict["location"] = {"@type": "Place",
+                                 "address": str(self.location)}
+        eventdict["startDate"] = str(self.date_from)
```
raphaelm: 1. We should be specific about the dateformat. As I see it,  in this place ``self.date_from.isoformat()`` would be correct.
2. According to Google's documents, we should only specify the date if the time is not known. We have a setting for that:
````
if self.settings.show_times:
    eventdict["startDate"] = self.date_from.isoformat()
else:
    eventdict["startDate"] = self.date_from.date().isoformat()
```

src/pretix/base/models/event.py
```diff
@@ -330,6 +330,21 @@ def get_payment_providers(self) -> dict:
...
+        eventdict["name"] = str(self.name)
+        eventdict["endDate"] = str(self.date_to)
```
raphaelm: We should only set this property if ``self.settings.show_date_to`` is True and ``self.date_to`` is not ``None``, otherwise we do not know the end date, i.e.
````
if self.date_to and self.settings.show_date_to:
    if self.settings.show_times:
        eventdict["endDate"] = self.date_to.isoformat()
    else:
        eventdict["endDate"] = self.date_to.date().isoformat()
```

src/pretix/presale/templates/pretixpresale/base.html
```diff
@@ -30,6 +30,13 @@
         <script type="text/javascript" src="{% static "pretixpresale/js/ui/typocheck.js" %}"></script>
         <script type="text/javascript" src="{% static "lightbox/js/lightbox.min.js" %}"></script>
     {% endcompress %}
+    {% if event %}
+    {% autoescape off %}
```
raphaelm: To make it easier to read, indent blocks like
````
{% if event %}
    {% autoescape off %}
        ...
````

File: src/pretix/presale/templates/pretixpresale/base.html [hunk 7]
```diff
@@ -30,6 +30,13 @@
         <script type="text/javascript" src="{% static "pretixpresale/js/ui/typocheck.js" %}"></script>
         <script type="text/javascript" src="{% static "lightbox/js/lightbox.min.js" %}"></script>
     {% endcompress %}
+    {% if event %}
+        {% autoescape off %}
+        <script type="application/ld+json">
+            {{ event.event_microdata }}
+        </script>
+    {   % endautoescape %}
```
raphaelm: There is a syntax error, could you check those spaces?

src/pretix/base/models/event.py
raphaelm: This still persists :)

src/pretix/base/models/event.py
```diff
@@ -330,6 +330,28 @@ def get_payment_providers(self) -> dict:
...
+    @property
+    def event_microdata(self):
+        import json
+
+        eventdict = {"@context": "http://schema.org", "@type": "Event"}
+        eventdict["location"] = {"@type": "Place",
+                                 "address": str(self.location)}
+        if self.settings.show_times:
+            eventdict["startDate"] = str(self.date_from.isoformat())
+            if self.settings.show_date_to & self.date_to is not None:
```
raphaelm: ``&`` is not ``and``, please use ``and``. ``&`` is the binary number operator.

src/pretix/base/models/event.py
```diff
@@ -330,6 +330,28 @@ def get_payment_providers(self) -> dict:
...
+
+    @property
+    def event_microdata(self):
+        import json
+
+        eventdict = {"@context": "http://schema.org", "@type": "Event"}
+        eventdict["location"] = {"@type": "Place",
+                                 "address": str(self.location)}
+        if self.settings.show_times:
+            eventdict["startDate"] = str(self.date_from.isoformat())
```
raphaelm: You don#t need ``str`` any more, ``isoformat()`` already creates a string

Joepriesto:
I'm guessing this fail is because I'm the unlucky one that pushed the coverage down. But I should have put in tests. So I am adding a new test class to the event test module.

raphaelm:
Hi @Joepriesto, yes, a simple test would not be strictly required, but still great. Otherwise, the code is fine! Let me know if you need any help.

raphaelm:
Thanks @Joepriesto and congratulations for fixing pretix' oldest open issue! :sparkles:

Issue #543, opened 2017-07-06
Allow to send email from order


raphaelm:
Currently, we can send emails automatically (e.g. on payment) or to many buyers at once (via the sendmail plugin). However, it would also be very useful to send emails directly to one buyer.

This could be a new button in the backend's order detail view, e.g. next to the "mark as paid" or "refund order" buttons, or next to the "change email address" button. The button would take you to a new page with a new form, similar to the form in the sendmail plugin, and allow you to send an email to the user.

danlls:
I'll work on this :)

PR #550, opened 2017-07-11
Ref #543 -- Allow send mail from order

danlls:
- Add send email directly for order
- Add email history (from mass and custom) to each specific order

** Need help to identify issue on testing, mail doesn't seems to be sent
  in the test cases. However, testing it using a browser with a running
  python debugging SMTP server works perfectly fine as the email is printed
  on the shell stdout and alert-success appear on redirected page.


src/pretix/control/forms/orders.py
```diff
@@ -178,3 +179,31 @@ def __init__(self, *args, **kwargs):
...
+            help_text=_("Available placeholders: {due_date}, {event}, {order}, {order_date}, {order_url}, "
+                        "{invoice_name}, {invoice_company}"),
+            validators=[PlaceholderValidator(['{due_date}', '{event}', '{order}', '{order_date}', '{order_url}',
+                                              '{invoice_name}', '{invoice_company}'])]
+        )
+        self.fields['sendto'] = I18nFormField(
+            widget=I18nTextInput, required=True,
+            initial=order.email, locales=event.settings.get('locales')
+        )
+        self.fields['sendto'].widget.attrs['readonly'] = 'readonly'
```
raphaelm: We don't really need I18nFormFields here, since when we write to one attendee only, we will only use one language. You can just use plain CharFields/TextFields/EmailFields and then you can define them like any other django form field

danlls:
@raphaelm Thanks! it make sense to use plain fields because the email only needs to be localized to the user but not the author. Everything is resolved now using plain django form fields.

src/pretix/control/logdisplay.py
```diff
@@ -96,6 +96,7 @@ def pretixcontrol_logentry_display(sender: Event, logentry: LogEntry, **kwargs):
         'pretix.event.order.payment.changed': _('The payment method has been changed.'),
         'pretix.event.order.expire_warning_sent': _('An email has been sent with a warning that the order is about '
                                                     'to expire.'),
+        'pretix.event.order.mail_sent': _('The order received a custom email.'),
```
raphaelm: Better: A custom email has been sent. An order is not something that can receive emails ;)

danlls:
@raphaelm Thanks for the feedback, fixed point 1 and 3. I agree with point 4 to include other mails in history but in another PR.

Other modification:
- Renamed placeholder and email_context keys to be similar with other existing placeholders for consistency.

danlls:
@raphaelm latest master branch is failing pytest in my local machine.

rixx:
It is fixed now :)

src/pretix/control/templates/pretixcontrol/order/mail_history.html
```diff
@@ -0,0 +1,33 @@
+{% extends "pretixcontrol/event/base.html" %}
+{% load i18n %}
+{% load bootstrap3 %}
+{% block title %}{% trans "Email history" %}{% endblock %}
+{% block content %}
+    <h1>{% trans "Email history" %}</h1>
+    <div class="callout callout-info">
+        <p>Includes sent custom and mass emails(<em>plugin</em>) history only.</p>
+    </div>
```
raphaelm: Please use ``alert alert-info`` instead of defining a new UI pattern and use ``{% trans "…" %}`` for the text.

(Also, I'd drop the "(plugin)", I don't think its helpful)

src/pretix/control/forms/orders.py
```diff
@@ -215,3 +215,30 @@ def __init__(self, *args, **kwargs):
...
+            label=_("Recipient"),
+            required=True,
+            initial=order.email
+        )
+        self.fields['sendto'].widget.attrs['readonly'] = 'readonly'
+        self.fields['message'] = forms.CharField(
+            label=_("Message"),
+            required=True,
+            widget=forms.Textarea,
+            initial=order.event.settings.mail_text_order_custom_mail,
```
raphaelm: we should probably use ``order.event.settings.mail_text_order_custom_mail.localize(order.locale)`` here

danlls:
As you've mention we should log the message in one language in `SenderView`, which make logical sense as each order only receives the message in their specified locale. Then, event would log the message in multiple locales. 
 
The logs you're looking at are the old logs. Try sending new mass mail it should work correctly.

raphaelm:
> The logs you're looking at are the old logs. Try sending new mass mail it should work correctly.

:see_no_evil: You caught me. 

Of course, real users also have old logs out there, but I'll find a way to deal with that.

Thank you very much for this high-quality pull request! :heart:

PR #551, opened 2017-07-12
Fix #516 -- Prevent gaps in plugin list

### LINKED ISSUE #516: Prevent gaps in plugin list
flexbox in column direction would be nice, but would kind of destroy ordering.

2017-06-11-151115_1457x420_scrot

danlls:
Use flexbox to fix misalignment when column height is imbalanced.


raphaelm:
Top, thanks! :heart:

PR #558, opened 2017-07-14
Fix #556 - Max. redemptions instead of amount of vouchers

### LINKED ISSUE #556: Voucher not being counted correctly
In the voucher by tag overview, the number sometimes just don't add up ;-)

image

So the progressbar is showing nothing, because it's class is set to 200%. Okay, that's kinda expected to not work ;-)

Explanation for the numbers:
- 2: Two vouchers are created with this tag
- 4: Those two vouchers have been redeem twice each.

So I guess we have to decide if the overview should show the amount of vouchers or the amount of possible voucher redemptions.

pc-coholic:
Another 5-character fix today ;-)


raphaelm:
Top, thanks! :)

Issue #552, opened 2017-07-14
Add icons to expansible panels


danlls:
Currently, expansible panels headers has the same style as the non-expansible except having strong text. It could mislead users thinking it is just a normal header.

Suggestion:
-  `fa-angle-down` icon can be used as indication
![panel-collapsed](https://user-images.githubusercontent.com/28383090/28199910-a30f6c44-689b-11e7-8c95-f2beca8dbd5d.png)

- rotated when expanded
![panel-expanded](https://user-images.githubusercontent.com/28383090/28199919-ab0ac9fc-689b-11e7-8e4b-8a671b514120.png)

- make the whole block of header clickable instead of just the text

Collapsible panels are currently used in multiple pages: 
`event/mail_setting_fragment`, `event/payment` , `presale/event/checkout_questions` and more ..

raphaelm:
Sounds great! Do you want to create a pull request yourself?

(There is the other topic of expanding it automatically whenever there is a form error inside it. checkout_questions does it, but mail_setting_fragment and payment do not, and I'm not sure what a good way to do this might be)

danlls:
Yeah, I'll work on this. 

The last time I tried, the forms in `checkout_questions` doesn't expand upon error. It display error messages only when the panel is not collapsed. Maybe you meant the invoice panel will expand if it is set as required instead of optional. I have no good way to do this atm, gotta research.

raphaelm:
>  The last time I tried, the forms in checkout_questions doesn't expand upon error. It display error messages only when the panel is not collapsed. Maybe you meant the invoice panel will expand if it is set as required instead of optional. I have no good way to do this atm, gotta research.

Hmm, ok. You can also leave the error handling part to me or for a separate PR and get started with the symbol first, it is helpful in itself already.

PR #560, opened 2017-07-17
Ref #552 -- Add icon to expansible panels

danlls:
Offer informative feedback to user by showing icons to indicate
expansible panels.

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/checkout_addons.html
+++ b/src/pretix/presale/templates/pretixpresale/event/checkout_addons.html
@@ -15,10 +15,10 @@
                     <div class="panel-heading">
                         <h4 class="panel-title">
                             <a data-toggle="collapse" href="#cp{{ form.cartpos.pk }}">
-                                <strong>{{ form.item.name }}</strong>
-                                {% if form.variation %}
+                                <strong>{{ form.item.name }}{% if form.variation %}
                                     – {{ form.variation }}
-                                {% endif %}
+                                {% endif %}</strong>
+                                <i class="fa fa-angle-down collapse-indicator"></i>
                             </a>
                         </h4>
                     </div>
--- a/src/pretix/presale/templates/pretixpresale/event/order_modify.html
+++ b/src/pretix/presale/templates/pretixpresale/event/order_modify.html
@@ -22,10 +22,10 @@
                     <div class="panel-heading">
                         <h4 class="panel-title">
                             <a data-toggle="collapse" href="#invoice" data-parent="#questions_accordion">
-                                <strong>{% trans "Invoice information" %}</strong>
-                                {% if not event.settings.invoice_address_required %}
+                                <strong>{% trans "Invoice information" %}{% if not event.settings.invoice_address_required %}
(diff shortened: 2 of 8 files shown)


src/pretix/static/pretixpresale/scss/main.scss
```diff
@@ -176,3 +176,31 @@ body.loading .container {
...
+    padding: 10px 15px;
+    margin: -10px -15px;
+    align-items: center;
+    justify-content: space-between;
+    outline: 0;
+    text-decoration: none;
+}
+
+.panel-title a[data-toggle="collapse"]:hover {
+    text-shadow: 1px 1px #999;
```
raphaelm: I'm not sure the text shadow is a good idea, since we have it *nowhere* in the system, it doesn't look fitting to me. I believe [Bootstrap's default](https://getbootstrap.com/javascript/#collapse-example-accordion) was a problem because it also underlined the symbol?

Maybe use ``<strong>`` everywhere and just use ``.panel-title a[data-toggle="collapse"]:hover strong {text-decoration: underline}``?

src/pretix/presale/templates/pretixpresale/event/checkout_addons.html
```diff
@@ -14,11 +14,11 @@
                 <div class="panel panel-default">
                     <div class="panel-heading">
                         <h4 class="panel-title">
-                            <a data-toggle="collapse" href="#cp{{ form.cartpos.pk }}">
-                                <strong>{{ form.item.name }}</strong>
-                                {% if form.variation %}
+                            <a class="collapsed" data-toggle="collapse" href="#cp{{ form.cartpos.pk }}">
```
raphaelm: These are expanded by default, so the ``collapsed`` class is not needed

src/pretix/presale/templates/pretixpresale/event/order_modify.html
```diff
@@ -21,11 +21,11 @@
                 <div class="panel panel-default">
                     <div class="panel-heading">
                         <h4 class="panel-title">
-                            <a data-toggle="collapse" href="#invoice" data-parent="#questions_accordion">
-                                <strong>{% trans "Invoice information" %}</strong>
-                                {% if not event.settings.invoice_address_required %}
+                            <a class="collapsed" data-toggle="collapse" href="#invoice" data-parent="#questions_accordion">
```
raphaelm: Those are expanded by default, so no ``collapsed``

src/pretix/static/pretixpresale/scss/main.scss
danlls: Hmm.. what about `background-color: #eeeeee;` just like the current sidebar? or both :D

raphaelm: A background color would work, if the *complete* panel head would change (and not only the text). Is that possible?

danlls: Yep, it is. Then, I will do that instead

raphaelm:
Thank you very much, I like this a lot!

Issue #475, opened 2017-04-21
Add to quota / create quota directly from the product creation form


raphaelm:
**Hey!** If you ever wanted to contribute to an Open Source project, but didn't know how, you just found a perfect place to start. We try to write this issue as instructive as possible, but as we're humans we almost certainly forgot to write something important in here. If you want to try to tackle this issue and get stuck at _any_ point, please do not hesitate and ask us, either here on GitHub or via email at support@pretix.eu. We'll be very happy to help! 

We think this issue is very suitable to people who are just starting with Python or Django development or who never contributed to Open Source before. We'd like to ask people with wide experience in these areas to leave this issue to beginners. Thanks! (If you want something more advanced, just drop us a line, we'll find something :wink:)

## How to get started

First of all, you need pretix running locally on your machine. We have a guide on the developer installation [over here](https://docs.pretix.eu/en/latest/development/setup.html). Play around with it and get a feeling for what pretix is and how it works :) Try to find the places in the interface mentioned in this issue where something needs to change. Maybe try to find the same places in the code? Just searching for characteristic words can be a powerful tool to find your way around.

## What this is about

If you want to sell tickets with pretix, you first need to create a product. In the code, that is often called an "item" (sorry for this confusion). Then, you need to add it to a quota, so the system know how often the product can be sold.

It would be nice to be able to select either "Add this to existing quota […]" or "Create a new quota for this product of size […]" right when creating a product, so you don't need to think of the second step.

## How to do this

You need to add new, custom fields to ``ItemCreateForm`` in ``pretix/control/forms/item.py``. Then, you need to display them in a clear way in the template ``pretix/control/templates/pretixcontrol/item/create.html``. For saving the data, you might need to add some logic to the ``form_valid`` method of ``ItemCreate`` in ``pretix/control/views/item.py``.

## I want to give it a try!

You're still reading, nice! If you want to try working on this, please comment below, so everyone else knows this is being looked at and can search for a different place to contribute. 

[...]

PR #562, opened 2017-07-17
Ref #475 -- add to existing quota / add new quota from product form

parkcl:
Ref #475 -- add to existing quota / add new quota from product form

### DIFF
--- a/src/pretix/control/forms/item.py
+++ b/src/pretix/control/forms/item.py
@@ -141,6 +141,39 @@ class ItemCreateForm(I18nModelForm):
             required=False
         )
 
+        self.fields['quota_option'] = forms.ChoiceField(
+            label=_("Quota options"),
+            widget=forms.Select,
+            choices=(
+                (0, 'None'), 
+                (1, "Add this to existing quota"),
+                (2, "Add this to new quota")
+            ),
+            required=False
+        )
+
+        self.fields['quota_add_existing'] = forms.ModelChoiceField(
+            label=_("Add to existing quota"),
+            widget=forms.Select(),
+            queryset=self.instance.event.quotas.all(),
+            required=False
+        )
+
+        self.fields['quota_add_new_name'] = forms.CharField(
+            label=_("Name"),
+            max_length=200,
+            widget=forms.TextInput(attrs={'placeholder':'New quota name'}),
+            required=False
+        )
+
+        self.fields['quota_add_new_size'] = forms.IntegerField(
+            min_value=0,
+            label=_("Size"),
+            widget=forms.TextInput(attrs={'placeholder':'New quota size'}),
+            required=False
+        )
+        
+
     def save(self, *args, **kwargs):
         if self.cleaned_data.get('copy_from'):
             self.instance.description = self.cleaned_data['copy_from'].description
@@ -155,6 +188,20 @@ class ItemCreateForm(I18nModelForm):
             self.instance.position = (self.event.items.aggregate(p=Max('position'))['p'] or 0) + 1
 
         instance = super().save(*args, **kwargs)
+
+        if not self.cleaned_data.get('quota_option') == '0':
+            if self.cleaned_data.get('quota_add_existing'):
+                quotaName = self.cleaned_data.get('quota_add_existing').name
+                quota = self.instance.event.quotas.filter(name=quotaName).get()
+                quota.items.add(self.instance)
+            else:
+                quotaName = self.cleaned_data.get('quota_add_new_name')
+                quotaSize = self.cleaned_data.get('quota_add_new_size')
+
+                quota = Quota.objects.create(
+                    event=self.event, name=quotaName, size=quotaSize
+                )
+                quota.items.add(self.instance)
 
         if self.cleaned_data.get('has_variations'):
             if self.cleaned_data.get('copy_from') and self.cleaned_data.get('copy_from').has_variations:
@@ -171,6 +218,15 @@ class ItemCreateForm(I18nModelForm):
                 question.items.add(instance)
 
         return instance
+
+    def clean(self):
+        cleaned_data = super().clean()
+
+        if self.cleaned_data.get('quota_option') == '2':
+            if not self.cleaned_data.get('quota_add_new_name'):
+                raise forms.ValidationError({'quota_add_new_name' : ['Quota name is required.']})
+
+        return cleaned_data
 
     class Meta:
         model = Item
--- a/src/pretix/control/forms/item.py
+++ b/src/pretix/control/forms/item.py
@@ -191,7 +191,6 @@ class ItemCreateForm(I18nModelForm):
 
         if not self.cleaned_data.get('quota_option') == '0':
             if self.cleaned_data.get('quota_option') == '1' and self.cleaned_data.get('quota_add_existing') is not None:
-                print("here for some reason")
                 quotaName = self.cleaned_data.get('quota_add_existing').name
                 quota = self.instance.event.quotas.filter(name=quotaName).get()
                 quota.items.add(self.instance)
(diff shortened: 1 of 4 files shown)


src/pretix/control/forms/item.py
```diff
@@ -141,6 +141,38 @@ def __init__(self, *args, **kwargs):
             required=False
         )
 
+        self.fields['quota_option'] = forms.ChoiceField(
+            label=_("Quota options"),
+            widget=forms.Select,
+            choices=(
+                (0, 'None'),
+                (1, "Add this to existing quota"),
+                (2, "Add this to new quota")
```
raphaelm: First, the choices should also be translatable, e.g. ``_("Add this to existing quota").
Second, I'd find the following texts more clear:
* Do not add to a quota now
* Add product to an existing quota
* Create a new quota for this product

Also, I'd recommend you to use strings like ``none``, ``existing`` and ``new`` instead of 0,1,2. it makes the code easier to read

src/pretix/control/forms/item.py
```diff
@@ -141,6 +141,38 @@ def __init__(self, *args, **kwargs):
...
+            label=_("Add to existing quota"),
+            widget=forms.Select(),
+            queryset=self.instance.event.quotas.all(),
+            required=False
+        )
+
+        self.fields['quota_add_new_name'] = forms.CharField(
+            label=_("Name"),
+            max_length=200,
+            widget=forms.TextInput(attrs={'placeholder': 'New quota name'}),
```
raphaelm: Same here, placeholder should be translatable with ``_('…')``

src/pretix/control/forms/item.py
```diff
@@ -156,6 +188,20 @@ def save(self, *args, **kwargs):
 
         instance = super().save(*args, **kwargs)
 
+        if not self.cleaned_data.get('quota_option') == '0':
```
raphaelm: This ``if`` clause does not change anything, does it? I think you can drop it

src/pretix/control/forms/item.py
```diff
@@ -156,6 +188,20 @@ def save(self, *args, **kwargs):
...
+            if self.cleaned_data.get('quota_option') == '1' and self.cleaned_data.get('quota_add_existing') is not None:
+                quota_name = self.cleaned_data.get('quota_add_existing').name
+                quota = self.instance.event.quotas.filter(name=quota_name).get()
+                quota.items.add(self.instance)
```
raphaelm: Why don't you just use the quota object?
```
quota = self.cleaned_data.get('quota_add_existing')
quota.items.add(self.instance)
```
Getting it again by name is not only useless, but also dangerous (there could be two with the same name for example)

src/pretix/control/forms/item.py
```diff
@@ -172,6 +218,17 @@ def save(self, *args, **kwargs):
 
         return instance
 
+    def clean(self):
+        cleaned_data = super().clean()
+
+        if self.cleaned_data.get('quota_option') == '2':
+            if not self.cleaned_data.get('quota_add_new_name'):
+                raise forms.ValidationError(
+                    {'quota_add_new_name': ['Quota name is required.']}
```
raphaelm: Same here, should be translatable. Also, shouldn't the size be required in this case as well?

Also, if an existing quota is used, that field should be required.

src/pretix/control/forms/item.py
parkcl: yes, I am not sure why I included this, can be dropped.

src/pretix/control/forms/item.py
parkcl: I left the size not required, because that's how the Quota form functions now--you can leave size blank and it will just create a Quota of size _unlimited_. Maybe I should add help text for that? 

Ok, I will make existing quota required, if selected.

raphaelm: > I left the size not required, because that's how the Quota form functions now--you can leave size blank and it will just create a Quota of size unlimited. Maybe I should add help text for that?

:see_no_evil: I'm always surprised how bad I know my own system. You're obviously right.

src/pretix/control/forms/item.py
```diff
@@ -132,6 +132,11 @@ class ItemCreateForm(I18nModelForm):
     def __init__(self, *args, **kwargs):
         self.event = kwargs['event']
         super().__init__(*args, **kwargs)
+
+        self.NONE = 'none'
+        self.EXISTING = 'existing'
+        self.NEW = 'new'
```
raphaelm: You can just specify them in the class body, e.g.
```
class ItemCreateForm(I18nModelForm):
    NONE = 'none'
...
```

src/tests/control/test_items.py
```diff
@@ -445,3 +445,36 @@ def test_create_copy(self):
...
+
+    def test_add_to_existing_quota(self):
+        q = Quota.objects.create(event=self.event1, name="New Test Quota", size=50)
+
+        doc = self.get_doc('/control/event/%s/%s/items/add' % (self.orga1.slug, self.event1.slug))
+        form_data = extract_form_fields(doc.select('.container-fluid form')[0])
+        form_data['name_0'] = 'Existing'
+        form_data['default_price'] = '2.00'
+        form_data['quota_option'] = 'existing'
+        form_data['quota_add_existing'] = '1'
```
raphaelm: You should use ``str(q.pk)`` here instead ``"1"``. It's not guaranteed that this ID will be always 1, which is why the tests currently fail on PostgreSQL and MySQL.

src/pretix/control/forms/item.py
```diff
@@ -141,6 +146,39 @@ def __init__(self, *args, **kwargs):
             required=False
         )
 
+        self.fields['quota_option'] = forms.ChoiceField(
+            label=_("Quota options"),
+            widget=forms.Select,
```
raphaelm: For choice fields with only three options, I feel that ``widget=forms.RadioSelect`` is better because the options are more clear to the user. What do you think? (This would also require small changes to the JavaScript part)

src/pretix/static/pretixcontrol/js/ui/hidequota.js
```diff
@@ -0,0 +1,30 @@
+$(document).ready(function() {
+  hideDeselected();
+});
+
+function hideDeselected() {
+  var v = $("#id_quota_option").val();
+
```
raphaelm: I think we can try adding in soft animation and also make the code a bit clearer and a lot shorter like the following (untested):
```
var v = $("#id_quota_option").val();
$("#existing-quota-group").slideToggle(v === "existing");
$("#new-quota-group").slideToggle(v === "new");
```
We don't even need ``hideAll`` then.

src/pretix/control/forms/item.py
parkcl: Yes, I agree with your point--it is clearer to the user with RadioSelect since we have just 3 static options.

src/tests/control/test_items.py
parkcl: Thanks for making me aware of this!

src/pretix/static/pretixcontrol/js/ui/hidequota.js
parkcl: I like your proposal here for a soft animation instead of a `hide()`. Unfortunately, it seems as though `slideToggle` doesn't currently support a boolean condition you can pass in like that. So I still think we'd need `hideAll`, to initialize the fields to be hidden when coming to the page.

raphaelm:
Sorry for the late reply. This is solid work, thank you very much! :sparkles:  Congratulations on your first contribution!

parkcl:
Thanks Raphael, for your help all throughout! I hope to help out again in the future.

raphaelm:
@parkcl Sure! If you want to take a look at something simple, #572 and #586 are currently not taken ;)

Issue #293, opened 2016-10-31
Download reminder email


raphaelm:
**Hey!** If you ever wanted to contribute to an Open Source project, but didn't know how, you just found a perfect place to start. We try to write this issue as instructive as possible, but as we're humans we almost certainly forgot to write something important in here. If you want to try to tackle this issue and get stuck at _any_ point, please do not hesitate and ask us, either here on GitHub or via email at support@pretix.eu. We'll be very happy to help! 

We think this issue is very suitable to people who are just starting with Python or Django development or who never contributed to Open Source before. We'd like to ask people with wide experience in these areas to leave this issue to beginners. Thanks! (If you want something more advanced, just drop us a line, we'll find something :wink:)

## How to get started

First of all, you need pretix running locally on your machine. We have a guide on the developer installation [over here](https://docs.pretix.eu/en/latest/development/setup.html). Play around with it and get a feeling for what pretix is and how it works :) Try to find the places in the interface mentioned in this issue where something needs to change. Maybe try to find the same places in the code? Just searching for characteristic words can be a powerful tool to find your way around.

## What this is about

It would be nice to -- optionally -- send out emails a few days before the event that remind people to download their tickets.

## How to do this

There are basically three things that need to be done:

* If you go to Settings → Email in the control panel of a pretix event, you see a number of content text snippets that you can configure. You will need to add a new text there, find out how ;) If you don't find it, just ask, I can provide more detailled guidance.
* We need to have a new [model](https://docs.djangoproject.com/en/1.10/topics/db/models/) field in the ``Order`` model that stores if the mail has already been sent out
* We need to actually send out the emails. This can be very done very similar to the ``send_expiry_warnings`` in ``pretix/base/services/orders.py``.

## I want to give it a try!

You're still reading, nice! If you want to try working on this, please comment below, so everyone else knows this is being looked at and can search for a different place to contribute. 

First, [fork](https://github.com/pretix/pretix/fork) the repository to your own GitHub account. Then clone it locally (or [set it as your remote](https://help.github.com/articles/configuring-a-remote-for-a-fork/) if you already cloned it earlier) and prepare your pull requests. There are a number of tutorials on the Pull Request workflow online, for example this [short version](https://guides.github.com/introduction/flow/) by GitHub itself and this [more in-depth tutorial](https://www.digitalocean.com/community/tutorials/how-to-create-a-pull-request-on-github). We're excited to see your pull request! 

[...]

PR #567, opened 2017-07-19
Add ticket downloading reminder

SanketDG:
Closes https://github.com/pretix/pretix/issues/293


src/pretix/base/services/orders.py
```diff
@@ -515,6 +515,42 @@ def send_expiry_warnings(sender, **kwargs):
             else:
                 o.log_action('pretix.event.order.expire_warning_sent')
 
+@receiver(signal=periodic_task)
+def send_download_reminders(sender, **kwargs):
+    eventcache = {}
+    today = now().replace(hour=0, minute=0, second=0)
+
+    for o in Order.objects.filter(expires__gte=today + timdelta(days=5), download_reminder_sent=False,
```
SanketDG: @raphaelm Hi, I am still not sure what should be done here, some help would be appreciated

src/pretix/base/settings.py
```diff
@@ -359,6 +359,17 @@
...
+        'type': LazyI18nString,
+        'default': LazyI18nString.from_gettext(ugettext_noop("""Hello,
+
+you bought a ticket for {event}.
+
+This is to remind you that your ticket is ready for downloading!
+
+Best regards,
+Your {event} team"""))
+    },
```
raphaelm: We'd also need a setting ``mail_days_download_reminder`` as in "how many days before the event should we sent this". By default, it should be None, i.e. by default the mail should not be sent.

src/pretix/base/services/orders.py
raphaelm: Sure. Sorry that it took me so long to reply.

What this currently does:

* Find all orders which
  * expire in the next five days and
  * have not been sent a reminder yet and
  * are in status pending
* For each of these orders
  * Look into the event settings for the configured number of days N (i.e. 3)
  * Check if the order expires in the next N days
  * If so, send an email

Here, we need to loop over a different set of orders and check different things. To make it easier, you can split it into two for loops and throw out all that ``eventcache`` stuff in there. It coult then look like:

* Find all events which
  * happen in the future (``date_from__gte=today``)
* For each of these events
  * Check if ``mail_days_download_reminder`` is not None, ``continue`` to skip the event otherwise
  * Calculate the date for reminders (``reminder_date = (event.date_from - timedelta(days=event.settings.mail_days_download_reminder)).replace(hour=0, minute=0, second=0)``)
  * Check if that date is already reached (``today > reminder_date``). If not, ``continue`` to skip this event
  * Find all orders which
    * Belong to this event (``event.orders.filter(…)``)
    * Have the status paid
    * Have not received the reminder email yet
  * For each of these orders
     * Send the email

Does that make it more clear?

SanketDG:
@raphaelm I have updated the changes although I suspect there are more changes.

I am also not sure if the codecov issues are related to my change.

raphaelm:
> @raphaelm I have updated the changes although I suspect there are more changes.

I think it is fine for now. Thank you very much!

>I am also not sure if the codecov issues are related to my changes.

Well, they are, in the way that you did not write any unit tests for the new code. But I think it's fine, I will do them while integrating the changes.

Issue #569, opened 2017-07-20
Includes other email to order-specific email history


danlls:
As mentioned in https://github.com/pretix/pretix/pull/550#pullrequestreview-49534315 , it would be useful to add other order-specific emails to email history (e.g. expiry_warning, order_paid, ..)

Todo:
- Some emails are currently sent without being logged such as `mark_order_paid`, `_cancel_order`. Add logs  when email is sent.
- Currently, email logs are inconsistent in naming such as `pretix.plugins.sendmail.order.email.sent` and `pretix.event.order.mail_sent`. We can standardize it with an infix `order.email` for all email sent to a specific order:
`pretix.plugins.sendmail.order.email.sent` to `pretix.plugins.sendmail.order.email.mass_sent`
`pretix.event.order.mail_sent` to `pretix.event.order.email.custom_sent`
and the new ones can be 
`pretix.event.order.email.order_paid`
`pretix.event.order.email.order_canceled`

This makes `OrderEmailHistory` view's queryset filtering easier using `action_type__contains="order.email"` or by regular expression without a bunch of OR `|` statements.

PR #570, opened 2017-07-21
Ref #569 -- Include other email to history

danlls:
Log all other order-specified emails and include them into email history. In addition, I've also added `SendMailException` handling to several mail() calls.


raphaelm:
Hi,

thanks for taking this on! First glance looks good, however there's one thing to consider here and one thought:

* I'm all in for a more consistend naming, but if we rename ``pretix.event.order.mail_sent``, we still need to support the old value e.g. in the log display handler (or perform a data migration). There's tons of existing logs out there, we don't them to suddenly look bad.
* You had to move all of the template contexts out of the ``mail()`` calls (which I actually like more than before), but it feels very error-prone to render the body (i.e. the format_map call) in both ``mail()`` as well as the function that calls ``mail()``. Wouldn't it be better and easier to maintain if we called ``log_action`` **within** ``mail`` whenever the ``order=`` keyword argument is set? We could introduce a new argument to specify the type of mail, if we want.

What do you think!

Cheers
Raphael

danlls:
- At first I was thinking it would be good if `mail()` could return the rendered email to avoid rendering twice(once in `mail()` and view). Well, since `mail()` is always rendering it, adding log_action within `mail()` would be great as it reduced so much unnecessary codes everywhere. However, introducing new argument require mail service to now watch for appropriate tags to specify the mail, would that increase noises in `mail.py` service as it should only focus on providing email sending functionality? Otherwise, I'm all in for it! 😉 

- As for older value `pretix.event.order.mail_sent`, would it be sufficient if we just retain the log display handler? What do you suggest?

Cheers,
Daniel

raphaelm:
Apologies for the late replies. I'm quite busy at the moment since I'll be travelling next week and have to get a few things done before. (Ergo, I'll also be slow with replying next week :wink:)

> At first I was thinking it would be good if mail() could return the rendered email to avoid rendering twice(once in mail() and view). Well, since mail() is always rendering it, adding log_action within mail() would be great as it reduced so much unnecessary codes everywhere. However, introducing new argument require mail service to now watch for appropriate tags to specify the mail, would that increase noises in mail.py service as it should only focus on providing email sending functionality? Otherwise, I'm all in for it! :wink:

Yes, it would increase the noise in mail(), but save us a *LOT* of noise in other places. But there might be a middle ground:

* Create a new method in ``mail.py`` called ``render_mail(template, context)`` that just does the rendering part
* Use ``render_mail()`` in ``mail()``
* Add a new method ``send_mail(subject, template, context, headers, log_action_type='pretix.order.mail.sent', email=None, sender=None)`` or similar on the ``Order`` model that creates the log entry and calls ``mail()``. You don't need to pass ``order``, ``locale`` and ``event`` there, as the order object already know them and use the ``order.email`` attribute as default for the email argument.

This approach could lead to both ``mail()`` and the callers of ``mail()`` getting *less* cluttered instead of *more*.

> As for older value pretix.event.order.mail_sent, would it be sufficient if we just retain the log display handler? What do you suggest?

Yes, I think that is sufficient.

danlls:
@raphaelm Updated. Let me know how you like it 😄

raphaelm:
Great work. Thank you!

Issue #406, opened 2017-02-07
Waiting list: Allow to reorder manually


raphaelm:
Allow the admin to manually move people up or down on the list to enhance their chances to get a ticket.

danlls:
I can work on this. Since the ordering should persist, should I create a new column in `WaitingListEntry` that specific the priority of each entry?

raphaelm:
Hi, yes, that would be required.

danlls:
Should this function be limited to only those entry without voucher yet?

raphaelm:
Yes.

Also, having up/down buttons like in the category/item list might be really cumbersome if you have many many entries. Normally, you probably only want to move important people up. So it might be worth considering actually allowing the user to edit the priority field directly instead of hiding it.

danlls:
How about instead of letting them edit, having an additional button which send an entry to the topmost? Because wouldn't editing the number directly be cumbersome too if the list is huge?

raphaelm:
I also thought about the button to set it to the top, but that gives really little flexibility.

The thing with editing the number directly is that one wouldn't need to fill in "1,2,3,4,5,6…" you could just use it as a real priority field like "1000 for the people I like best, 100 for some other people, 1 for the rest" and keep the natural order (date) within the ones with same priority. We could even give the priorities names (1-very low, 2-low, 3-medium, …) but I think just having an input field works as well if we explain it properly.

danlls:
Alright! I guess allowing them to fill in arbitrary numbers would maximize flexibility. They can use it in all variation: "1,2,3,4,5,6, ..." or grouped like "1000, 100, 10, 1" or mix grouping and individual "1, 10, 100, 101, 1000".

raphaelm:
Yep, I think that's good

PR #575, opened 2017-07-29
Ref #406 -- Allow manual reorder in waiting list

danlls:
- Add model column 'priority' as sorting criteria
- Add a new page to reorder priority of entries which has not been
  assigned a voucher.


danlls:
Not sure why build is failing on postgresql config only. Any insight?

rixx:
When restarting the build of one of the failing scenarios, it errored out due to conflicting migrations – would you mind rebasing against the commit 86f1709 which is the last passing commit on the master branch?

Rebasing against a specific commit is done by checking out your feature branch and running `git rebase 86f1709`. – If you run into issues or don't feel comfortable with this, I could do it, too.

Afterwards you'll probably need to fix your migration to depend on the last migration in the current pretix line, which means changing the `dependencies` list in your migration to read `('pretixbase', '0071_auto_20170729_1616')`.

Afterwards, tests should be able to run again, and we can have a look at the postgres issue if it persists.

danlls:
Weird, rebased and still failing 😢

danlls:
Alright, thanks for your help!

src/pretix/base/migrations/0072_auto_20170803_1006.py
```diff
@@ -0,0 +1,37 @@
...
+
+
+def set_priority(app, schema_editor):
+    wles = app.get_model('pretixbase', 'WaitingListEntry')
+    for wle in wles.objects.all():
+        if wle.voucher:
+            wle.priority = None
+        else:
+            wle.priority = 10
+        wle.save()
```
michael-k: Number of Database Queries: 2 * #WaitlingListEntry + #WaitingListEntryWithVoucher

`wles.objects.filter(voucher__isnull=True).update(priority=10)`

More details:
- `wle.voucher` leads to a db query (check `wle.voucher_id` instead or use `select_related`)
- `wle.priority = None` is the default, no need to call save (=> hint that `filter` should be used)
- `wle.save()` sends all data to the database; consider using `save`'s `update_fields` instead

src/pretix/base/models/waitinglist.py
```diff
@@ -136,3 +142,10 @@ def send_voucher(self, quota_cache=None, user=None):
                 self.event,
                 locale=self.locale
             )
+
+    def save(self, *args, **kwargs):
+        if self.voucher is not None:
```
michael-k: Consider checking `self.voucher_id` instead.

src/pretix/base/migrations/0072_auto_20170803_1006.py
danlls: Thanks for the neat suggestions to reduce database hit count. I've learned much! 😄

raphaelm:
I debugged the postgres failure: The test ``test_delete_single`` tries to delete an entry from the list. It just randomly grabs the first one. Due to your new default ``ordering`` and the internal ordering for postgres, this now accidentally happens to be the one list entry that already has a voucher and therefore can no longer be deleted. Just changing the respective line to ``wle = WaitingListEntry.objects.filter(voucher=None).first()`` in the test helps.


Otherwise, I see one open question here. You now have implemented a separate page that shows a form field for the priority of every entry on one page. This is very convenient if you want to change multiple ones and certainly a good idea *until* you have more than 30 entries and the list begins to paginate. Suddenly, it probably becomes inconvenient because (a) you need to know that you must first save, then switch to the first page and (b) the changes you made on the first page might change the order of everything such that the second page will *not* be what you expect, since you might see some of the ones you already changed again and miss some others completely.

I'm not sure what the good solution is here. The *easy* solution certainly is to switch back to a simple update view that just allows changing one entry at a time. Do you have other ideas how we could approach this?

danlls:
I'm not sure if the following idea will work, but here it is:

Maybe instead of using `ModelFormSet`, we can use a normal `FormSet` to act as a proxy form.

1. Fields of formset initial value will be set to be as `wle.priority`
2. Any modification to each field will be saved/recorded using ajax or on pagination to ensure persistence when user go back and forth pages.
3. These modification won't directly modify the database, list ordering won't change upon pagination.
4. Once user actually clicked on `Save`, then copy all changes and apply it on the `WaitingListEntry` model.

raphaelm:
Yep, I guess that would work, but it feels like more effort that it's worth.

PR #579, opened 2017-07-29
Add optional help_text to Question objects. Closes #574.

### LINKED ISSUE #574: Add help texts to questions
When creating a question, it should optionally be possible to insert a help text that explains the question in more detail. This requires a new model field on the ``Question`` model that is then added to the respective forms.

Also, the auto-generated form fields during checkout should then use that as their help text.

If someone wants to take this on, let me know if help is needed, I'll gladly provide more information!

rixx:


raphaelm:
migrations and isort 🙄 looks great otherwise :D

rixx:
Fixed. You could exclude migrations in isort though, couldn't you? ._.

raphaelm:
Yes, I probably could… Thanks!

PR #589, opened 2017-08-02
Organizer color & settings propagation

scabux:
Added a "primary color option" for organizer pages (cf. event).
The selected color will not yet be set as default value for the organizers' events, because I still have to figure out an elegant way to do so.
I will continue working on the organizer page adding a language selector and restructuring the control interface in order to replicate the workflow for editing events. In the process I will deal with the default value issue.
Also, there will be unit tests for the new features.

### DIFF
--- a/src/pretix/control/forms/organizer.py
+++ b/src/pretix/control/forms/organizer.py
@@ -1,6 +1,7 @@
 from django import forms
 from django.conf import settings
 from django.core.exceptions import ValidationError
+from django.core.validators import RegexValidator
 from django.utils.translation import ugettext_lazy as _
 from i18nfield.forms import I18nFormField, I18nTextarea
 
@@ -118,6 +119,16 @@ class OrganizerSettingsForm(SettingsForm):
         label=_("Use languages"),
         widget=forms.CheckboxSelectMultiple,
         help_text=_('Choose all languages that your organizer homepage should be available in.')
+    )
+
+    organizer_primary_color = forms.CharField(
+        label=_("Primary color"),
+        required=False,
+        validators=[
+            RegexValidator(regex='^#[0-9a-fA-F]{6}$',
+                           message=_('Please enter the hexadecimal code of a color, e.g. #990000.'))
+        ],
+        widget=forms.TextInput(attrs={'class': 'colorpickerfield'})
     )
 
     organizer_homepage_text = I18nFormField(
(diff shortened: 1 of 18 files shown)


raphaelm:
Hi! Thanks a lot! I'm currently on vacation so I'll probably only be able to take a deep look next week.

> The selected color will not yet be set as default value for the organizers' events, because I still have to figure out an elegant way to do so.

If you use ``primary_color`` instead of ``organizer_primary_color`` as a settings key, this will work fully automatically, due to the way [hierarkey](https://django-hierarkey.readthedocs.io/en/latest/) works. However, there are two problems:

1. The setting will be inherited by the events, but the CSS for the events will not automatically be generated (this only affects this specific setting)

2. Once the user first submits the settings form of the event, the value will be written for the event and later changes to the organizer value will not be propagated to the event, which might be confusing.

I have a not-that-simple idea in my mind how we could solve (2) that we can discuss, but not one for (1) yet.

> I will continue working on the organizer page adding a language selector and restructuring the control interface in order to replicate the workflow for editing events. In the process I will deal with the default value issue.

Nice, I'm excited to see these changes!

rixx:
> Once the user first submits the settings form of the event, the value will be written for the event and later changes to the organizer value will not be propagated to the event, which might be confusing.

Automatic propagation might be just as confusing and an active change to running events.

scabux:
> Automatic propagation might be just as confusing and an active change to running events.

I totally agree with rixx, which is why I used `organizer_primary_color` instead of `primary_color` in the first place.

rixx:
The way I see it, the default hierarkey behaviour makes a lot of sense here: Propagate initially but not if an event's color has been explicitly set. That's something you can put in a comprehensive, short explanation aswell.

CSS generation … oof. Any way to trigger this from inside pretix, @raphaelm?

scabux:
> CSS generation … oof. Any way to trigger this from inside pretix

Just calling `regenerate_css` should do the trick or did I get something wrong?

raphaelm:
> Automatic propagation might be just as confusing and an active change to running events.

I disagree, automatic propagation should -- at least for superficial settings like design and colors -- be the expected way of things. Of course we need to ensure that (a) the organizer-level form makes clear that it propagates and (b) that the event-level form makes clear that it overrides things. I have a design idea in my mind, just haven't got around to sketch it yet, will do so asap.

> Just calling regenerate_css should do the trick or did I get something wrong?

Yep, we'd need to call it for all events every time the value changes on organizer level.

Issue #531, opened 2017-06-24
Placeholders in subjectline are not replaced


pc-coholic:
When sending custom mails using ``/control/event/<organizer>/<event>/sendmail/``, the placeholders ({due_date}, {event}, {order}, {order_date}, {order_url}, {invoice_name}, {invoice_company}) are not replaced when placed in the subject line.

rahmonov:
Hey, can I be assigned to this?

raphaelm:
Sure, go ahead!

raphaelm:
Hi @rahmonov, did you already have time to take a look at this? :)

rahmonov:
@raphaelm hey, I am sorry but I haven't had a chance. I should have informed earlier. Sorry once again. Please, remove the assigned label so that another person can fix this issue. Apologies.

SanketDG:
Can I take this up?

raphaelm:
@SanketDG Sure!

SanketDG:
Hi, due to the limited capabilities of testing locally, I am not able to replicate this. Can I get some help on this?

raphaelm:
Hi, in what way can you not test this locally? You just need to somehow create an order, then you should be able to try this out. There is a note in the developer docs regarding sending emails locally without actually sending them. Let me know if you need more help, though I am on vacation for a week from today and will answer slowly. :)

PR #594, opened 2017-08-12
Make placeholders replace in subject

SanketDG:
Placeholders in subject were not being replaced because there was
no `.format()` called on the subject.

This commit creates a context dict that is used for both the body
and the subject. It is then replaced using `.format_map()`

Fixes https://github.com/pretix/pretix/issues/531

### DIFF
--- a/src/pretix/plugins/sendmail/views.py
+++ b/src/pretix/plugins/sendmail/views.py
@@ -78,24 +78,33 @@ class SenderView(EventPermissionRequiredMixin, FormView):
 
         if self.request.POST.get("action") == "preview":
             for l in self.request.event.settings.locales:
+
                 with language(l):
-                    self.output[l] = []
-                    self.output[l].append(
-                        _('Subject: {subject}').format(subject=form.cleaned_data['subject'].localize(l)))
-                    message = form.cleaned_data['message'].localize(l)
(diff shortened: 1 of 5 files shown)


SanketDG:
@rixx Which test file do I need to change?

rixx:
`base/tests/test_mail.py` probably.

SanketDG:
@rixx I made the changes to `test_sendmail.py` that felt like the right test file.

rixx:
You'll probably have to wait some days for feedback from Raphael, who's away atm, but the PR looks good to me. :)

src/pretix/plugins/sendmail/views.py
```diff
@@ -78,24 +78,32 @@ def form_valid(self, form):
 
         if self.request.POST.get("action") == "preview":
             for l in self.request.event.settings.locales:
+
+                context_dict = {
```
raphaelm: You can't move this out of the ``with language(l):`` block, otherwise the date formats etc. will be wrong.

SanketDG:
@raphaelm Sorry I got busy!

It's updated now, I am testing for stuff in `test_mail` now.

raphaelm:
Thank you very much! :tada:

PR #620, opened 2017-09-04
Check format of stripe API keys in settings 

### LINKED ISSUE #619: stripe: check format of API keys in settings
The current settings form allows the stripe secret key to be submitted as the publishable keys. This can easily happen by accident (publishable key starts with `pk_` => "private key"). Since publishable keys always start with `pk_` and secret keys always start with `sk_`, a simple validator should do the trick.

luto:
This checks if the publishable and secret keys given in the stripe settings survive a basic sanity check: do they start with `pk_` and `sk_`? The goal is to prevent users from swapping the keys accidentally.

Please let me know what you think about this solution as well as the tests. If you would like anything in the PR to be solved differently, just let me know and I am happy to adapt the code. Thanks!

Fixes #619.

### DIFF
--- a/src/pretix/plugins/stripe/payment.py
+++ b/src/pretix/plugins/stripe/payment.py
@@ -33,6 +33,24 @@ class RefundForm(forms.Form):
     )
 
 
+class StripeKeyValidator():
+    def __init__(self, prefix):
+        assert isinstance(prefix, str)
+        assert len(prefix) > 0
+        self._prefix = prefix
+
+    def __call__(self, value):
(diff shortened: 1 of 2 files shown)


raphaelm:
Hi,

both the code and the tests look fine! I'd have intuitively taken a RegexValidator instead of a new validator class, but I'm not sure it would have actually been any better.

Thank you for your contribution and congrats for your first PR to pretix! :sparkles: 

Cheers
Raphael

luto:
🎉 Thanks!

PR #621, opened 2017-09-05
Allow markdown rendering in transaction comments.

rixx:
This commit allows transaction comments to display newlines and URLs in
a useful way, helping when additional data (such as a reference to a
ticket system or a longer discussion) is required.
This PR also prevents pretix from having to bring its own chat system ;)


raphaelm:
Is it that easy? The comments can be edited via inline JavaScript – I assume markdown is not rendered directly after saving

raphaelm:
Hi :) Since I'm not sure how our last discussion ended, just give me a quick signal if this is now ready-to-merge from your side!

rixx:
Nope, it works for ~2/3 cases, still on it.

rixx:
Ready to merge on my end.

raphaelm:
Thanks! :)

Issue #623, opened 2017-09-13
docker-compose


Speedy1991:
Hi!
I try to run Pretix via docker-compose and if i get it running i will open a PR if you are interested.
The current state is:
- compose is starting without errors
- database doesn't get migrated
- website isn't reachable

I hope you can help me figure out the problems.
The isn't any error to see except  `pretix_1  | 2017-09-13 19:14:04,988 INFO exited: pretixtask (exit status 1; not expected)`

Here are my files + config + docker output

docker-compose.yml:
```
version: '2'
services:
  db:
    image: mariadb:latest
    environment:
      - MYSQL_DATABASE=pretix
      - MYSQL_USER=pretix
      - MYSQL_PASSWORD=pretix
      - MYSQL_RANDOM_ROOT_PASSWORD=1

  redis:
    image: redis:latest

  pretix:
    image: pretix/standalone
    volumes:
      - pretix-data:/data
      - ./etc:/etc/pretix/
    ports:
      - "0.0.0.0:8345:8345"
    depends_on:
      - db
      - redis
      
  
volumes:
  pretix-data:

```

pretix.cfg:
```

[pretix]
instance_name=My pretix installation
url=localhost
currency=EUR
; DO NOT change the following value, it has to be set to the location of the
; directory *inside* the docker container
datadir=/data

[database]
; Replace mysql with postgresql_psycopg2 for PostgreSQL
backend=mysql
name=pretix
user=pretix
password=pretix
; Replace with host IP address for PostgreSQL
host=db

[mail]
; See config file documentation for more options
from=tickets@yourdomain.com
; This is the default IP address of your docker host in docker's virtual
; network. Make sure postfix listens on this address.
host=172.17.0.1

[redis]
location=redis?db=0
; Remove the following line if you are unsure about your redis' security
; to reduce impact if redis gets compromised.
sessions=true

[celery]
backend=redis+socket://redis?virtual_host=1
broker=redis+socket://redis?virtual_host=2
```

Full output of docker:
```
$ docker-compose up  --force-recreate --build
Recreating pretix_db_1 ...
Recreating pretix_db_1
Recreating pretix_redis_1 ...
Recreating pretix_redis_1 ... done
Recreating pretix_pretix_1 ...
Recreating pretix_pretix_1 ... done
Attaching to pretix_db_1, pretix_redis_1, pretix_pretix_1
[...]

PR #625, opened 2017-09-14
docker-compose

Speedy1991:
Added docker-compose example setup
see #623 
Maybe someone can check the files if i missed anything - i am new to pretix :)


raphaelm:
Thanks!

One question: Have you tested the cronjob thingy? I believe ``/etc/init.d/cron start`` shouldn't be very effective because later in the docker container only supervisord runs, not the debian init system.

Also, the directory could maybe have a short README file explaining a bit what it does, what is included and what not, and how to use it :)

Speedy1991:
cron: I'll check it tonight/tomorrow

Yea i ll add a README too :)

Speedy1991:
- cronjob: nice catch - i created a workaround and faking a cronjob via supervisord and a cron.py (tested)
- README: Added readme. My english is not the best, feel free to edit it to fit your needs

Speedy1991:
added memcached + fixed the redis socket issue
review plz :)

raphaelm:
Hi!

* I'm not 100% happy with the cron-hack. Using supervisord is already a hack in itself, in an ideal docker world, you would spin up four containers, one for pretixweb, one for pretixtask, one for crond and one for nginx. The standalone container entrypoint should accept an argument ("webworker", "taskworker"), so it would be easy to isolate pretixweb and pretixtask, it gets harder with nginx. If you want to play with it, do it, if not, it's okay for now.

* ``run.sh`` should be executable (chmod +x)

* You can drop memcached altogether. redis is good enough to do both tasks.

* I get a lot of "command at '/etc/cron.py' is not executable", also a chmod +x problem maybe?

* I get a lot of "pretixtask (exit status 1; not expected)" and I'm not sure how to debug this :wink: 

FYI, I'll be on vacation next week, so expect slower replies than usual

Speedy1991:
- added requested changes
- I replaced gunicorn with uwsgi for running the cron

deployment/docker-compose/pretix/etc/pretix.cfg
```diff
@@ -0,0 +1,20 @@
...
+[database]
+backend=mysql
+name=pretix
+user=pretix
+password=pretix
+host=db
+
+[redis]
+location=redis://redis/0
+session=true
```
almereyda: I think this needs to be sessions, according to https://github.com/pretix/pretix/blob/3c3e59e932080dffe9dca18156693a8b6573a9dc/src/pretix/settings.py#L182

Also see https://github.com/ecobytes/pretix/blob/master/deployment/docker-compose/pretix/etc/pretix.cfg#L33

almereyda:
Thank you for this excellent example to run Pretix.

We merged it with master and learned a lot from it while adapting to our environment with https://github.com/pretix/pretix/compare/master...ecobytes:master 

The adaptations allow to configure transactional email via SMTP, introduce configuration for a load balancer and switch to host-mounted volumes for speed increase.

Yet I feel a deployment Docker Compose configuration should be lightweight and decoupled from the application source code, why I copied big parts this into https://github.com/ecobytes/docker-pretix

Speedy1991:
@almereyda Thanks for your feedback and glad i could help you with this PR :)

> host-mounted volumes

I made the decision for named docker volumes cause:
 - trouble with the host uid's/guid's if these doesn't match the docker uid's/guid's
 - dump the containers into a tar file and have a full backup with a single command

>  load balancer

I didn't add any loadbalancer (nginx could be one, but in this example it is just a simple reverse proxy). My approach was more like:  Just keep things simple, this isn't intend to run in production.

(OFF TOPIC: ... because this compose can be easily converted to a kubernetes config file -> replicas + loadbalancing comes out of the box)

Never the less, i will update my personal pretix instance soon (maybe this weekend) to the newest version and will improve this PR as soon as possible. I also will update the compose to the newest version with separated networks and maybe less boilerplate stuff.

deployment/docker-compose/pretix/etc/pretix.cfg
Speedy1991: true, nice catch :+1:

Speedy1991:
updated and hardend the setup.
I get some errors like
`Invalid template /usr/local/lib/python3.6/site-packages/django_otp/templates/otp/admin17/login.html: 'admin_static' is not a registered tag library. Must be one of:`
when running manage.py compress
Looks like the _django-otp_ is using the deprected _{% admin_static %}_ instead of the _{% static %}_ tagloader.

agowa338:
@Speedy1991: Thanks for your docker-compose, but I'm experiencing the following problems with it:

- Only reachable from 127.0.0.1, when using the external IP, django throws `'django.security.DisallowedHost exception Unknown host:'`
- When clicking on "Add to cart" pretix gets stuck at "We're processing your request"

deployment/docker-compose/pretix/Dockerfile
```diff
@@ -0,0 +1,37 @@
+FROM python:3.6
+
```
agowa338: `ENV DEBIAN_FRONTEND=noninteractive` should be added or the action starting in line 9 will silently fail. It also suppresses some errors about not specified TERM variable and interactive promts.

deployment/docker-compose/pretix/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+RUN apt-get update
+RUN apt-get install -y git libxml2-dev libxslt1-dev python-dev python-virtualenv locales \
+    libffi-dev build-essential python3-dev zlib1g-dev libssl-dev gettext libpq-dev \
+    libmysqlclient-dev libjpeg-dev
```
agowa338: `libmysqlclient-dev` cannot be found. In the referenced ubuntu version it is called `default-libmysqlclient-dev` for some reasons. Former and later versions call it `libmysqlclient-dev`.

deployment/docker-compose/pretix/pretix/etc/pretix.cfg
```diff
@@ -0,0 +1,28 @@
+[pretix]
+instance_name=My pretix installation
+url=http://127.0.0.1/
```
agowa338: There should be a note in the Readme to change this line to allow access from external. Otherwise the user receives a django Host Denial `django.security.DisallowedHost exception Unknown host:`

deployment/docker-compose/pretix/pretix/etc/pretix.cfg
```diff
@@ -0,0 +1,28 @@
...
+currency=EUR
+datadir=/data
+
```
agowa338: The pretix.cfg should contain:
```ini
[pretix]
registration=False
long_sessions=True
```
Or everyone can create a account in the backend, which may not be desirable.

deployment/docker-compose/pretix/pretix/etc/pretix.cfg
```diff
@@ -0,0 +1,28 @@
...
+backend=redis://redis/1
+broker=redis://redis/2
+
+[uwsgi]
+http =  0.0.0.0:8000
+master = True
+processes = 4
+threads = 2
+module = pretix.wsgi
+cron = -30 -1 -1 -1 -1 cd /pretix/src && python3 -m pretix runperiodic
```
agowa338: This file should also include the following section:
```ini
[django]
secret=ooShah3UpeiCohngoop0phaeghequah0aor5ohpa3xeephie9E
debug=off
```
where secret is the django encryption secret.

deployment/docker-compose/pretix/README.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+No volumes are shared with the host. Named Volumes will be used.
+
+Run without nginx container:
+----------------------------
+
+```bash
+cd docker-compose/pretix
+mkdir static data # Create the dirs by hand, otherwise they will belong to root
```
agowa338: Why are these no named volumes? You're using them in your nginx example, why not also here. This would also simplify your nginx compose file, as you would not need to define them there.

deployment/docker-compose/pretix/README.md
```diff
@@ -0,0 +1,48 @@
...
+chown -R www-data:www-data static data
```
agowa338: This needs some more explanation. If the user and group ids of the dockerhost and container diverge it is still wrong. You should use the numeric values instead if it is really necessary.

For example I'm using ArchLinux and I don't have the www-data user nor the corresponding group.

deployment/docker-compose/pretix/Dockerfile
```diff
@@ -0,0 +1,37 @@
...
+COPY pretix .
+COPY pretix-entrypoint.sh .
+COPY task-entrypoint.sh .
+
+RUN chmod +x pretix-entrypoint.sh task-entrypoint.sh
+
+RUN pip3 install -r requirements/production.txt
+RUN pip3 install -r requirements/mysql.txt
+RUN pip3 install -r requirements/redis.txt
+RUN chown -R www-data:www-data /pretix /etc/pretix /data /static
```
agowa338: This line should maybe also be added to the entrypoint, as these directories could be on bind mounted volumes like with the current docker-compose.yml

deployment/docker-compose/pretix/README.md
```diff
@@ -0,0 +1,48 @@
...
+cp -r ../../../src src_tmp # Due to security reasons Docker cant build from parent directories
+docker-compose up --force-recreate --build
```
agowa338: you may want to run `docker-compose build --no-cache --pull --compress` for each containing service before, as `--force-recreate` still uses the build cache. So it may be unnecessary to use it here.
also the `docker-compose up` should contain the parameter `--abort-on-container-exit`.

While I tested this docker-compose I run into the problem, that the pretix got stuck on an endless ajax request loop, as the taskrunner image crashed for reasons stated in another comment.

agowa338:
@Speedy1991 Only one thing, I've noticed, that you have mergecommits in your branch. In order for automerges to work you should remove them and instead rebase your commits on those.
For example you could roll back your branch to the time before you merged in master and then cherrypick your commits. And if you want to get all changes to master into your branch, you could than use `docker rebase`.
Your work gave me a great starting point for my own docker compose setup. And it allowed me to "get up and running" more quickly. Thanks for it, I hope this gets merged asap, so that also others not looking at the pull requests can also benefit from it.
If you want to have a look at my changes: https://github.com/agowa338/pretix/tree/feature/compose2

mdehollander:
Just a quick feedback message.
I gave this docker-compose approach also a try. With the ecobytes/docker-pretix repo from @Speedy1991 I got also some errors, including the mysql client not found reported by @agowa338. I did not look further into this approach.
When I tried the feature/compose2 branch of @agowa338, it works very smooth. In a few minutes pretix is installed and I logged into to the admin area. Great! Next, configuring it.

raphaelm:
See #1068 :)

PR #642, opened 2017-10-18
Fix #630 -- manual check-in of attendees

### LINKED ISSUE #630: Manual check-in of attendees
It is useful e.g. for statistical reasons to be able to mark participant as checked-in manually after the event, e.g. when check-in was done offline with a paper list.

I imagine the best way to do this would be to have checkboxes in the first column of the view at "Orders" -> "Check-ins" in the event admin interface that allow to bulk-change the check-in status of attendees.

koebi:
This enables manual check-in of attendees.  The post-code was heavily
copied from the APIRedeemView of the pretixdroid, thus so far this seems
to be working, but I have a few questions:

The checkin-Objects generated by the pretixdroid-app have a nonce.
Should the checkin object generated here have a nonce, too?

Should the result of the check-in be noted in any other way than by the
change of the status?

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/checkin/index.html
+++ b/src/pretix/control/templates/pretixcontrol/checkin/index.html
@@ -52,6 +52,8 @@
                 <table class="table table-condensed table-hover">
                     <thead>
                     <tr>
+                      <th>{% trans "CheckIn?" %}
+                      </th>
                         <th>{% trans "Order code" %} <a href="?{% url_replace request 'ordering' '-code'%}"><i class="fa fa-caret-down"></i></a>
                       <a href="?{% url_replace request 'ordering' 'code'%}"><i class="fa fa-caret-up"></i></a></th>
                         <th>{% trans "Item" %} <a href="?{% url_replace request 'ordering' '-item'%}"><i class="fa fa-caret-down"></i></a>
@@ -74,6 +76,11 @@
                     {% for e in entries %}
                         {% with e.checkins.first as checkin %}
                         <tr>
+                          <td> 
+                            <input type="checkbox" name="checkin"
+                                                   id="id_checkin" class=""
+                                                                   value="{{e.order.code}}"/>
+                          </td>
                             <td>
                                 <strong><a href="{% url "control:event.order" event=request.event.slug organizer=request.event.organizer.slug code=e.order.code %}">{{ e.order.code }}</a></strong>
                             </td>
@@ -106,6 +113,11 @@
                     {% endfor %}
                     </tbody>
                 </table>
+            </div>
+            <div class="form-group submit-group">
+              <button type="submit" class="btn btn-primary btn-save">
+                Check-In selected
+              </button>
             </div>
         </form>
     {% endif %}
--- a/src/pretix/control/templates/pretixcontrol/checkin/index.html
+++ b/src/pretix/control/templates/pretixcontrol/checkin/index.html
@@ -52,8 +52,7 @@
                 <table class="table table-condensed table-hover">
                     <thead>
                     <tr>
-                      <th>{% trans "CheckIn?" %}
-                      </th>
+                        <th />
                         <th>{% trans "Order code" %} <a href="?{% url_replace request 'ordering' '-code'%}"><i class="fa fa-caret-down"></i></a>
                       <a href="?{% url_replace request 'ordering' 'code'%}"><i class="fa fa-caret-up"></i></a></th>
                         <th>{% trans "Item" %} <a href="?{% url_replace request 'ordering' '-item'%}"><i class="fa fa-caret-down"></i></a>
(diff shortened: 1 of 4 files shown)


src/pretix/control/templates/pretixcontrol/checkin/index.html
```diff
@@ -52,6 +52,8 @@
                 <table class="table table-condensed table-hover">
                     <thead>
                     <tr>
+                      <th>{% trans "CheckIn?" %}
```
raphaelm: I think we consistently write it as ``Check-In``, although I'd just drop this label and just leave the checkboxes unlabeled for two reasons: (a) the column can be narrower (b) we can later add multiple actions that operate on the selected rows by just adding multiple buttons, if we want

src/pretix/control/templates/pretixcontrol/checkin/index.html
```diff
@@ -107,6 +114,11 @@
                     </tbody>
                 </table>
             </div>
+            <div class="form-group submit-group">
+              <button type="submit" class="btn btn-primary btn-save">
+                Check-In selected
```
raphaelm: We need to translate it, and I'd put it as ``{% trans "Check-in selected attendees" %}

src/pretix/control/templates/pretixcontrol/checkin/index.html
```diff
@@ -74,6 +76,11 @@
                     {% for e in entries %}
                         {% with e.checkins.first as checkin %}
                         <tr>
+                          <td> 
+                            <input type="checkbox" name="checkin"
+                                                   id="id_checkin" class=""
+                                                                   value="{{e.order.code}}"/>
```
raphaelm: This won't work: Imagine you buy two tickets with the same order: They will turn up as two rows here, but both will have the same code. I think we can just use ``e.pk`` here.

src/pretix/control/views/checkin.py
```diff
@@ -69,6 +70,34 @@ def get_context_data(self, **kwargs):
                            or "subevent" in self.request.GET)
         return ctx
 
+    def post(self, request, *args, **kwargs):
+        for c in request.POST.getlist('checkin'):
+            created = False
+            op = OrderPosition.objects.select_related('item', 'variation', 'order', 'addon_to').get(
+                order__event=self.request.event,
+                order__code=c
+            )
```
raphaelm: See above, replace by ``pk=c``. We could also optimize this by restructuring it the following way:

```
 positions = OrderPosition.objects.select_related('item', 'variation', 'order', 'addon_to').filter(
    order__event=self.request.event,
    pk__in=request.POST.getlist('checkin')
)
for op in positions:
```
This only requires one SQL query instead of ``N``.

src/pretix/control/views/checkin.py
```diff
@@ -69,6 +70,34 @@ def get_context_data(self, **kwargs):
...
+                    'first': True,
+                    'datetime': now()
+                })
+            else:
+                op.order.log_action('pretix.control.views.checkin', data={
+                    'position': op.id,
+                    'positionid': op.positionid,
+                    'first': False,
+                    'datetime': now()
+                })
```
raphaelm: It seems like we can just simplify it to
```
              op.order.log_action('pretix.control.views.checkin', data={
                   'position': op.id,
                   'positionid': op.positionid,
                   'first': created,
                   'datetime': now()
               })
```
and drop the if statement

src/pretix/control/views/checkin.py
```diff
@@ -69,6 +70,34 @@ def get_context_data(self, **kwargs):
...
+
+        return self.get(request, *args, **kwargs)
```
raphaelm: The problem with calling ``self.get`` here is, that two checkins will be created if the user now refreshs the page (as the form will be sent again). We should instead store a message to the user session and redirect back:

```
messages.success(request, _('The selected tickets have been marked as checked in.'))
return redirect(reverse('control:event.orders.checkins') + '?' + request.GET.urlencode())
```

raphaelm:
One more thing: It'd be great if you could write one or two unit tests for this new feature as well :) Let me know if you need any help with that!

koebi:
Adressed the comments, one question remains: should the datetime of the manual check-in be noted in the log entry? The entry is created at basically the same time as the login, so i feel it is not neccessary… Also, should double checkin be mentioned in the log?

raphaelm:
> Adressed the comments, one question remains: should the datetime of the manual check-in be noted in the log entry? The entry is created at basically the same time as the login, so i feel it is not neccessary… 

You're right. We could keep it for future compatibility and compatibility with the pretixdroid logs, though. For apps like pretixdroid it matters, since you might have checked in at 11:15 but only at 11:20 was the phone able to notify the server about it, so the two times would differ. In the future, I want to have a unified API for all this instead of duplicate code here and in pretixdroid so from my side, just let it be.

> Also, should double checkin be mentioned in the log?

I think so, yes.

Looks great otherwise! I'll try to do a final test and review in the coming days. It'd be great if you could add 1 or 2 simple unit tests for the view. Let me know if you need a hand at that! :)

raphaelm:
Hi! :) Thank you very much! This is really good work. Giving it a more thorough test, I only found very minor issues. As I plan to create a new pretix release in the next days and want to include this change in this release, I took the liberty of doing the following myself:

* Rebasing the branch against current master
* A slight layout change to the template that puts the button nearer to the checkboxes
* Add the current user to the logging call to make it more transparent
* Change the unit test to make it a bit easier to read and to check if a CheckIn object is actually created
* Merge it. Grats! :100:

Issue #643, opened 2017-10-18
Order Details: History


domke:
The order history as the text "_An email has been sent to notify user order has been received and require payment._" Which does not sound like like proper english.

Please change that into something like "_An email has been sent to notify the user that the order has been received and requires payment._"

scperkins:
@domke Opened a PR for this [here](https://github.com/pretix/pretix/pull/644). Thanks!

PR #644, opened 2017-10-19
Fixes the order history text as mentioned in issue #643

scperkins:
https://github.com/pretix/pretix/issues/643


raphaelm:
Thank you very much! :sparkles:

PR #645, opened 2017-10-19
Fix #598 -- Use zxcvbn for password validation

### LINKED ISSUE #598: Better password strength validation
* We want to use [zxcvbn](https://github.com/dropbox/zxcvbn) to estimate password strength reasonably
* We want to display visual feedback on the password strengh while creating accounts or changing passwords.

It might be worth checking if https://pypi.python.org/pypi/django-zxcvbn-password/ is a good way to achieve this.

aimanparvaiz:
…s/prod.txt. Adding zxcvbn_password.ZXCVBNValidator to AUTH_PASSWORD_VALIDATORS for the password strength validation
Adding zxcvbn_password.ZXCVBNValidator to  AUTH_PASSWORD_VALIDATORS would allow for better password strength. Feedback going back to the user is "Add another word or two. Uncommon words are better., Avoid repeated words and characters."


aimanparvaiz:
I see some failing tests:
`         - tests/control/test_auth.py:194 RegistrationFormTest.test_success
         - tests/control/test_auth.py:376 PasswordRecoveryFormTest.test_recovery_valid_token_success
         - tests/control/test_user.py:72 UserSettingsTest.test_change_password_success`
I am very new to tests, let me read up and fix this.

raphaelm:
If you look at the three failing tests, you'll probably see that they try to create a user or change a user's password. Apparently, the passwords used there are now not god enough any more ;)

aimanparvaiz:
Thanks for the response @raphaelm , how should we proceed from here, would you reject this PR and I will get back to work and send out a new one ?

raphaelm:
No need to reject the PR, you can just add more commits to the branch and it will get updated :)

raphaelm:
I added the client-side integration to all forms. I'm not merging it currently since I'd like to display feedback on *why* the password is bad and there is no easy way to internationalize it currently. See https://github.com/dropbox/zxcvbn/pull/124 as well

raphaelm:
Closing for inactivity

PR #649, opened 2017-10-22
Change of wording of log display

domke:
This PR corrects some further wording as an extension to issue #643 and proposes to harmonize the use of the words "changed" and "modified".


raphaelm:
Thank you! :) :sparkles:

PR #655, opened 2017-10-29
Show Buttons on Cancelled Order [Fix #641]

### LINKED ISSUE #641: Buttons missing after order cancelation
When an order is in the state _canceled_ the menu bar on top of the order details page in the admin backend is missing. 
It would be nice if the buttons _View order as user_ and _View email history_ would still be available when an order is canceled (the other buttons don't make sense in this state).

BenHagan:
Modified order template to show "View order as user" and "View email history" buttons on orders with cancelled status in the admin backend.

Fix #641.

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/order/index.html
+++ b/src/pretix/control/templates/pretixcontrol/order/index.html
@@ -18,7 +18,7 @@
         {% include "pretixcontrol/orders/fragment_order_status.html" with order=order class="pull-right" %}
     </h1>
     {% if 'can_change_orders' in request.eventpermset %}
-        {% if order.status == 'n' or order.status == 'p' or order.status == 'e' %}
+        {% if order.status == 'n' or order.status == 'p' or order.status == 'e' or order.status == 'c' %}
             <form action="{% url "control:event.order.transition" event=request.event.slug organizer=request.event.organizer.slug code=order.code %}"
                     method="post">
                 {% csrf_token %}


src/pretix/control/templates/pretixcontrol/order/index.html
```diff
@@ -18,7 +18,7 @@
         {% include "pretixcontrol/orders/fragment_order_status.html" with order=order class="pull-right" %}
     </h1>
     {% if 'can_change_orders' in request.eventpermset %}
-        {% if order.status == 'n' or order.status == 'p' or order.status == 'e' %}
+        {% if order.status == 'n' or order.status == 'p' or order.status == 'e' or order.status == 'c' %}
```
raphaelm: Could you instead just drop this ``if`` clause alltogether? It doesn't to anything useful any more since all relevant buttons have their own if clause :)

BenHagan: I think the reason I added the 'c' case to the `if` statement is that we still have the 'r' case out there.
If we're ok with the 'r' case having the same two button behavior as 'c' (and why not), then I'll make the change.

raphaelm: Yep, I think ``r`` should have the same behavior as ``c``

BenHagan:
Hey thanks for having me!  If we want to move forward with the changes discussed in the code comments, I'll need to make a new PR (deleted the fork, newbie mistake).

raphaelm:
> Hey thanks for having me! If we want to move forward with the changes discussed in the code comments, I'll need to make a new PR (deleted the fork, newbie mistake).

Ah, no worries :)

BenHagan:
Closing in favor of PR #658.

Issue #654, opened 2017-10-27
Writable API


raphaelm:
This is to track the whole progress on adding write operations to our REST API.

- [ ] Orders
  - [x] create (#599, #911)
  - [x] mark paid
  - [x] mark pending
  - [x] mark expired
  - [x] mark canceled
  - [x] mark refunded?
  - [x] extend order
  - [x] change comment
  - [x] create invoice
  - [x] resend link
  - [x] change email address
  - [x] regenerate secrets
  - [x] change locale
  - [x] change invoice address
  - [x] change names, question answers, … (#1904)
  - [x] change positions (ie all orderchangemanager operations)
    - [x] changing the product or variation of a position (#2552)
    - [x] changing the subevent of a position (#2552)
    - [x] changing the seat of a position (#2552)
    - [x] changing the price of a position (#2552)
    - [x] changing the tax rule of a position (#2552)
    - [ ] changing the membership used for a position
    - [x] cancelling a position (already implemented)
    - [x] adding a position (top-level or add-on) (#2552)
    - [x] splitting the order (#2552)
    - [x] changing the value of a fee (#2552)
    - [x] cancelling a fee (#2552)
    - [x] regenerating the secret of a position (#2552)
    - [x] re-calculating all taxes from product and tax rule data (#2552)

- [x] events (#756)
  - [x] create (#756)
  - [x] update (#756)
  - [x] delete
  - [x] settings (#1378)
  - [x] blocked seats
- [x] subevents
  - [x] create (#1217)
  - [x] update (#1217)
  - [x] delete (#1217)
- [x] invoices
  - [x] regenerate
  - [x] reissue
- [x] tax rules
  - [x] create
  - [x] update
  - [x] delete
- [x] categories (#818)
  - [x] create (#818)
  - [x] update (#818)
  - [x] delete (#818)
- [x] items (#676)
  - [x] create (#676)
  - [x] update (#676)
  - [x] delete (#676)
- [x] questions (#813)
  - [x] create (#813)
  - [x] update (#813)
  - [x] delete (#813)
- [x] vouchers
  - [x] create
  - [x] update
  - [x] delete
- [x] waiting list entries (#885)
  - [x] create (#885)
  - [x] update (#885)
  - [x] delete (#885)
  - [x] assign voucher (#885)
- [x] quotas
  - [x] create
  - [x] update
  - [x] delete
- [x] teams
- [x] devices

turegjorup:
[...]

PR #657, opened 2017-10-30
Added quotas API write methods

turegjorup:
Thank you for your advice in #654. 

This is my first attempt at implementing create/update/delete for the `quotas` endpoint. There are a couple of things I'm unsure of, so I have marked the PR as WIP:

* I haven't found any validation rules for quotas besides the model validation - is this correct?
* I have added a couple of tests, but I'm unsure of what the edge cases would be, so this could definitely be expanded upon.

Code checks and unit tests _should_ pass.


doc/api/resources/quotas.rst
```diff
@@ -106,6 +106,134 @@ Endpoints
...
+        "subevent": null
+      }
+
+   :param organizer: The ``slug`` field of the organizer to create a voucher for
+   :param event: The ``slug`` field of the event to create a voucher for
+   :statuscode 201: no error
+   :statuscode 400: The voucher could not be created due to invalid submitted data.
+   :statuscode 401: Authentication failure
+   :statuscode 403: The requested organizer/event does not exist **or** you have no permission to create this resource.
+   :statuscode 409: The server was unable to acquire a lock and could not process your request. You can try again after a short waiting period.
```
raphaelm: 409 can not happen for this operation, so we can drop it.

doc/api/resources/quotas.rst
```diff
@@ -106,6 +106,134 @@ Endpoints
...
+        "size": 200,
+        "items": [1, 2],
+        "variations": [1, 4, 5, 7],
+        "subevent": null
+      }
+
+   :param organizer: The ``slug`` field of the organizer to create a voucher for
+   :param event: The ``slug`` field of the event to create a voucher for
+   :statuscode 201: no error
+   :statuscode 400: The voucher could not be created due to invalid submitted data.
```
raphaelm: voucher → quota

doc/api/resources/quotas.rst
```diff
@@ -106,6 +106,134 @@ Endpoints
...
+      }
+
+   :param organizer: The ``slug`` field of the organizer to modify
+   :param event: The ``slug`` field of the event to modify
+   :param id: The ``id`` field of the quota rule to modify
+   :statuscode 200: no error
+   :statuscode 400: The quota could not be modified due to invalid submitted data
+   :statuscode 401: Authentication failure
+   :statuscode 403: The requested organizer/event does not exist **or** you have no permission to change this resource.
+   :statuscode 409: The server was unable to acquire a lock and could not process your request. You can try again after a short waiting period.
```
raphaelm: Same here, 409 won't happen

src/pretix/api/views/item.py
```diff
@@ -77,18 +78,58 @@ class Meta:
...
     ordering = ('id',)
     permission = 'can_change_items'
+    write_permission = 'can_change_items'
 
     def get_queryset(self):
         return self.request.event.quotas.all()
 
+    def create(self, request, *args, **kwargs):
+        with request.event.lock():
+            return super().create(request, *args, **kwargs)
```
raphaelm: We don't need the lock here as we do with vouchers and can therefore drop this method completely.

src/pretix/api/views/item.py
```diff
@@ -77,18 +78,58 @@ class Meta:
...
+        )
+
+    def get_serializer_context(self):
+        ctx = super().get_serializer_context()
+        ctx['event'] = self.request.event
+        return ctx
+
+    def update(self, request, *args, **kwargs):
+        with request.event.lock():
+            return super().update(request, *args, **kwargs)
```
raphaelm: We don't need the lock here as we do with vouchers and can therefore drop this method completely.

src/pretix/api/views/item.py
```diff
@@ -77,18 +78,58 @@ class Meta:
...
+
+    def perform_update(self, serializer):
+        serializer.save(event=self.request.event)
+        serializer.instance.log_action(
+            'pretix.event.quota.changed',
+            user=self.request.user,
+            api_token=(self.request.auth if isinstance(self.request.auth, TeamAPIToken) else None),
+            data=self.request.data
+        )
```
raphaelm: In the form view, we do a quite elaborate logging based on if the associated subevent changed. The reason is that we do not only want the log to show up in the quota detail view but in the subevent's log as well. I just saw that we only do this during updating, not during deletion or creation (which is kinda inconsistent), but we should at least keep API and form behaviour consistent. Reference: https://github.com/pretix/pretix/blob/master/src/pretix/control/views/item.py#L685

Also, we should call ``serializer.instance.rebuild_cache()`` after saving here.

raphaelm:
Hi,

thank you very much for the high-quality PR! :sparkles:  I've added a few minor comments with pretix-specific things you couldn't have known and I have some answers to your questions.

> I haven't found any validation rules for quotas besides the model validation - is this correct?

Heh. Yes and no. This is why I suggested this as a starter, because it has no real explicit validation rules. The form does however have some *implicit* validation baked into its structure that is really non-obvious. As I see it, they are:

* It must be forbidden to add a variation to the ``variations`` list that does not belong to an item contained in the ``items`` list
* It must be forbidden to remove an item from the ``items`` list if one of its ``variations`` is still in the ``variations`` list
* It must be forbidden to add an item to the ``items`` ist if it *has* variations but none of its variations is in the ``variations`` list
* (most important one) It must be forbidden to link an item or variation that belongs to a different event – this would allow to hijack other organizer's ticket shops otherwise
* It must be forbidden to link a subevent that belongs to a different event

> I have added a couple of tests, but I'm unsure of what the edge cases would be, so this could definitely be expanded upon.

Your tests are fine, but with the additional validation constraints I outlined above it should be clear how to come up with some interesting edge cases :wink: 

>Code checks and unit tests should pass.

I'm really unsure why Travis does not build this PR :confused: 
If it doesn't start to do so, I'll check manually as soon as this is done with.

Let me know if you have any more questions! :)

raphaelm:
Also, forgot that to mention, you should add the new URLs to the test parameter set in ``tests/api/test_permissions.py`` to make sure the permission requirements are tested. You can look at the voucher lines: The fourth parameter is the expected return code for "permission granted" which is 400/404 in this case since the test will send invalid data (does not matter for checking the permissions).

turegjorup:
Thank you for your feedback @raphaelm. I have attempted to address most of it:
* `tests/api/test_permissions.py` updated with new URLs
* validation rules for items/variations added

I still need to address logging and subevents

src/pretix/base/models/items.py
```diff
@@ -914,3 +914,11 @@ def clean_items(event, items, variations):
...
                 if bool(selected_vars) is False:
                     raise ValidationError(_('One or more items has variations but none of these are in the variations list'))
+
+    @staticmethod
+    def clean_subevent(event, subevent):
+        if event.has_subevents:
+            if not subevent:
+                raise ValidationError(_('Subevent cannot be null for event series'))
+            if event != subevent.event:
+                raise ValidationError(_('The subevent does not belong to this event'))
```
raphaelm: At the same time, we could check that subevent *is* null if event.has_subevents is False :)

turegjorup:
Hi @raphaelm - I believe the last commits fixes the issues you have mentioned. If not please advise where further improvements are needed :-)

The only thing I'm not completely satisfied with is the fixtures in the tests. I think we should consider how best to structure them going forward. So far, I have used names like "Event2" and "Event3" but this doesn't really communicate either the content or the intended usage.

I think we should consider these questions:
* Should we name fixtures after content, e.g. "event_with_subevent", "item_with_variations", etc?
* Can/shall fixtures be shared among tests, e.g. `conftest.py` or should we work towards more dedicated fixtures for individual tests? The former will give less code duplication but probably a more complex fixtures structure. The alter will cause us to duplicate fixtures but might enable more focused/less complex fixtures for the individual tests.

Tests pass with a few exceptions:
```   
1729 passed
1 xpassed
1 failed
- tests/control/test_settings.py:301 MailSettingPreviewTest.test_localised_payment_info
1 skipped
3 rerun
```
Fail:
```
E       AssertionError: assert 'Please transfer money to this bank account: 9999-9999-9999-9999' != 'Please transfer money to this bank account: 9999-9999-9999-9999'

tests/control/test_settings.py:313
```
xpassed:
```
tests/api/test_orders.py
```
In both cases it seems unrelated to the PR so I'm unsure how to address this?

raphaelm:
Hey! :) Thanks, I'll look into it more deeply as soon as I get around to.

> The only thing I'm not completely satisfied with is the fixtures in the tests. I think we should consider how best to structure them going forward. So far, I have used names like "Event2" and "Event3" but this doesn't really communicate either the content or the intended usage.

I can relate to that, this absolutely something I'm not satisfied with in the pretix codebase. This is probably something to do a large refactoring about but I'm also really unsure how the final design should look. We could go for more unified fixtures in conftest files or for using [FactoryBoy](https://factoryboy.readthedocs.io/en/latest/) instead of py.test fixtures in places where we need very individual fixtures. 

> The former will give less code duplication but probably a more complex fixtures structure.

This is exactly the trade-off I'm struggling to make. I don't know, yet, so for now, I don't care about it too much for the scope of this PR.

> Should we name fixtures after content, e.g. "event_with_subevent", "item_with_variations", etc?

Yeah, that's probably better. I don't like pytests fixture system here as it leads to long parameter and therefore long variable names which makes the tests harder to write, but as long as it's easier to read, then… ;) If you only need a very simple fixture once or twice I'd probably drop the fixture and inline it to the test.

> Fail: tests/control/test_settings.py:313

This is just because you did not compile translation files (``make localecompile``)

> xpassed: tests/api/test_orders.py

This is my fault and can be ignored for now :wink:

turegjorup:
> This is exactly the trade-off I'm struggling to make. I don't know, yet, so for now, I don't care about it too much for the scope of this PR.

OK - I'll leave it as it is for now. Looking forward to further feedback 😃

doc/api/resources/quotas.rst
```diff
@@ -106,6 +106,132 @@ Endpoints
...
 
+.. http:post:: /api/v1/organizers/(organizer)/events/(event)/quotas/
+
+   Creates a new quota
+
+   **Example request**:
+
+   .. sourcecode:: http
+
+      POST /api/v1/organizers/bigevents/events/sampleconf/quotas/1/ HTTP/1.1
```
raphaelm: this is only to ``/quotas/``, not ``/quotas/1/``, I guess :wink:

src/pretix/api/views/item.py
```diff
@@ -77,18 +78,71 @@ class Meta:
...
+        if serializer.instance.subevent:
+            serializer.instance.subevent.log_action(
+                'pretix.subevent.quota.changed',
+                user=self.request.user,
+                api_token=(self.request.auth if isinstance(self.request.auth, TeamAPIToken) else None),
+                data=self.request.data
+            )
```
raphaelm: This still misses the subtle case in which *changing* the subevent should log an *added* on the new subevent and an *deleted* on the old one.

src/pretix/base/models/items.py
```diff
@@ -897,3 +897,31 @@ def _position_lookup(self) -> Q:
...
+        for variation in variations:
+            if variation.item not in items:
+                raise ValidationError(_('All variations must belong to an item contained in the items list'))
+                break
+
+    @staticmethod
+    def clean_items(event, items, variations):
+        for item in items:
+            if event != item.event:
+                raise ValidationError(_('One or more items does not belong to this event'))
```
raphaelm: plural items => "do not belong" instead of "does"

src/pretix/base/models/items.py
```diff
@@ -897,3 +897,31 @@ def _position_lookup(self) -> Q:
 
     class QuotaExceededException(Exception):
         pass
+
+    @staticmethod
+    def clean_variations(items, variations):
+        for variation in variations:
+            if variation.item not in items:
+                raise ValidationError(_('All variations must belong to an item contained in the items list'))
```
raphaelm: Could you end the error messages with a full stop, if they are a full sentence? :)

src/pretix/base/models/items.py
```diff
@@ -897,3 +897,31 @@ def _position_lookup(self) -> Q:
...
+                break
+
+    @staticmethod
+    def clean_items(event, items, variations):
+        for item in items:
+            if event != item.event:
+                raise ValidationError(_('One or more items does not belong to this event'))
+            if item.has_variations:
+                selected_vars = set(item.variations.all()).intersection(variations)
+                if bool(selected_vars) is False:
```
raphaelm: I find this hard to read and understand. Also, the query to item.variations.all() is not really required. How about:

``
if not any(var.item == item for var in variations):
``

doc/api/resources/quotas.rst
turegjorup: Correct - and body shouldn't contain `"id": 1,` - I'll have a word with the copy-paste monkey 🙈

src/pretix/base/models/items.py
turegjorup: Much better - thanks. Still getting up to speed in python. Didn't know that construct.

src/pretix/api/views/item.py
turegjorup: This should be addressed by commit [79c3579](https://github.com/pretix/pretix/pull/657/commits/79c3579db8e20a04ec3a49b874a8c1f399d353d7)

turegjorup:
I think my last commits should address the remaining issues so I have removed the WIP label. 

If there you find additional problems please advice.

PR #658, opened 2017-10-31
Fix #641 -- Show Buttons on Cancelled or Refunded Order

### LINKED ISSUE #641: Buttons missing after order cancelation

BenHagan:
Changes template to show "View order as user" and "View email history"
buttons on orders in refunded or cancelled status in control backend.

Fix #641.


BenHagan:
Replacement for PR #655.

raphaelm:
Top, thank you so much! :) :sparkles:

PR #659, opened 2017-10-31
Fix #489 -- Handle Vouchers With Expired Items

### LINKED ISSUE #489: Handle vouchers with expired items
Hi! When redeeming a voucher for a product where `available_until` is before today, the list for available products is empty and there is no error message.
image
As a user, I would expect
- to buy the ticket regardless of `available_until`, or
- to get an error message

Thanks for your awesome work and with best regards,
Jan Felix

BenHagan:
* Fix #489 -- Handle Vouchers With Expired Items

* Add regression test

### DIFF
--- a/src/pretix/presale/views/cart.py
+++ b/src/pretix/presale/views/cart.py
@@ -395,6 +395,8 @@ class RedeemView(NoSearchIndexViewMixin, EventViewMixin, TemplateView):
                     err = error_messages['voucher_redeemed']
                 if self.voucher.valid_until is not None and self.voucher.valid_until < now():
                     err = error_messages['voucher_expired']
+                if self.voucher.item is not None and self.voucher.item.is_available() is False:
+                    err = error_messages['voucher_item_not_available']
 
                 redeemed_in_carts = CartPosition.objects.filter(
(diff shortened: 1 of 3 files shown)


BenHagan:
Travis plugin job failed as such: `An error occurred while generating the build script.`

Kinda weird, ran tests locally before pushing to Github without any major issues.

src/pretix/presale/views/cart.py
```diff
@@ -395,6 +395,10 @@ def dispatch(self, request, *args, **kwargs):
                     err = error_messages['voucher_redeemed']
                 if self.voucher.valid_until is not None and self.voucher.valid_until < now():
                     err = error_messages['voucher_expired']
+                if self.voucher.item is not None:
+                    if self.voucher.item.available_until is not None and (
+                            self.voucher.item.available_until < now()):
+                        err = error_messages['ended']
```
raphaelm: I believe this message is misleading, we should rather create a new message that says
"Your voucher is valid for a product that is currently not for sale."

BenHagan:
image

Updated and ready for review.  Thanks for the tip on `Item.is_available()`.  Very helpful.

PR #660, opened 2017-10-31
occured -> occurred

koebi:
All resources I could find listed this as misspelled, but I wasn't too
sure…
Also, it should be checked if all changes to the .po-files are respected
in the corresponding src-files.

### DIFF
--- a/src/pretix/base/services/orders.py
+++ b/src/pretix/base/services/orders.py
@@ -53,7 +53,7 @@ error_messages = {
                  'the quantity you selected. Please see below for details.'),
     'price_changed': _('The price of some of the items in your cart has changed in the '
                        'meantime. Please see below for details.'),
-    'internal': _("An internal error occured, please try again."),
+    'internal': _("An internal error occurred, please try again."),
     'empty': _("Your cart is empty."),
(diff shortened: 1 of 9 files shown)


koebi:
I split these spelling corrections into three parts, because I am only about ~90% sure on the occur-thing and for the german corrections, you don't have to check the corresponding files, so I think that this might make things a bit easier… If it doesn't, let me know so I can learn how to proceed with these sorts of things in the future and how to handle git correctly :)

raphaelm:
> I split these spelling corrections into three parts, because I am only about ~90% sure on the occur-thing and for the german corrections, you don't have to check the corresponding files, so I think that this might make things a bit easier… If it doesn't, let me know so I can learn how to proceed with these sorts of things in the future and how to handle git correctly :)

I don't now what the best workflow here is but as long as I get simple PRs with clean, readable diffs, what else could I do than merge? ;) However, travis complains here because of:

```
CommandError: Execution of msgfmt failed: /home/travis/build/pretix/pretix/src/pretix/locale/de/LC_MESSAGES/djangojs.po:172
```

Apparently I *did* spell it correctly in at least *one* case :D Could you have a look at that? :)

raphaelm:
Same here :)

PR #661, opened 2017-10-31
correct typos in formal german translation

koebi:
I think the only "critical" fix here is the change from "Zahlmethoden"
to "Zahlungsmethoden", but that is the word used in the rest of the
translations, so I figured it should be changed here as well.


raphaelm:
:blush: I think its clear that I often write those translations under time pressure… :D I now fixed the same issues with the informal translation as well! Thanks for finding them!

raphaelm:
:blush: I think its clear that I often write those translations under time pressure… :D I now fixed the same issues with the informal translation as well! Thanks for finding them!

PR #664, opened 2017-11-01
[WIP] Add Danish locale (translation)

rimi-itk:
We're in the process of adding Danish translations to pretix, and this is our progress so far. Our/my experience with pretix and Django is very limited, and any comments on the way we're doing it will be much appreciated.

I've made it possible to override translations in DATA_DIR/locale to make local tweaks to translations. Is this a hack or a viable solution?

It would be nice to be able to limit the locales available in the pretix installation using configuration, e.g. to make only Danish and English available we could use something like
```
[locale]
default=da
enabled_locales=da,en
```
and then filter `LANGUAGES` in `settings.py` accordingly. Any comments on this?


raphaelm:
> We're in the process of adding Danish translations to pretix, and this is our progress so far.

Nice! :)

> Our/my experience with pretix and Django is very limited, and any comments on the way we're doing it will be much appreciated.

The most interesting question for me would be: Are there any strings in there that are hard to you to translate because it's not clear what they mean just from the English text? I'd be more than happy to add in some translator comments, if I know where :)

> I've made it possible to override translations in DATA_DIR/locale to make local tweaks to translations. Is this a hack or a viable solution?

Something in between, I'd like to keep the need for local tweaks as low as possible. May I ask why you need it?

> It would be nice to be able to limit the locales available in the pretix installation using configuration, e.g. to make only Danish and English available we could use something like

Is there any reason to do this? You could just not create any events with German enabled and German will show up nowhere except in the "create event" dialog, therefore I think this is unnecessary configuration.

rimi-itk:
> The most interesting question for me would be: Are there any strings in there that are hard to you to translate because it's not clear what they mean just from the English text? I'd be more than happy to add in some translator comments, if I know where :)

Most of the translation work is straightforward, but sometimes requires some knowledge of the inner workings of pretix. Real issues only come up with stuff like "reverse charges" and "taxes". However, this is not related to pretix, but to the general complexity of these concepts and regional differences (e.g "Mehrwertsteuer" is called "moms" in Danish, but we don't call it a "skat" ["Steuer"]; it's an "afgift" ["Gebühr"], but "gebyr" in Danish is something else).

> May I ask why you need [to override translations in DATA_DIR/locale]?

In Danish we can call "An event" many things, e.g. "Arrangement" (the one I'm currently using in the Danish translation), "Begivenhed", or even "Event". Being able to tweak translations per installation would be useful for this, but the amount of translations that have to be customized may make this infeasible. I'll remove this to keep things clear. 

> Is there any reason to [be able to limit available locales]? You could just not create any events with German enabled and German will show up nowhere except in the "create event" dialog, therefore I think this is unnecessary configuration.

Good point! This will be indeed be unnecessary configuration.

Thanks for your comments and remarks!

raphaelm:
Hi @rimi-itk!

I'm going close this PR for the simple reason that we today launched a new workflow for translating pretix. You can read more about it in [our blogpost](https://pretix.eu/about/de/blog/20180308-translation-process/). The blogpost also explains why we prefer this process over separate Pull Requests.

Don't worry, your work is still much appreciated and remains just as useful. You can just take the .po(s) file from this pull request and upload it [here](https://translate.pretix.eu/projects/pretix/pretix/da/#upload) and [here (djangojs)](https://translate.pretix.eu/projects/pretix/pretix-js/da/#upload) after you signed up on our translation platform. I can do this for you, if you want, but did not do so in order to not take credit for the translation submission :wink: I migrated your code fixes to the master branch already.

You can either do this now and continue translating on the platform itself (and others might join you!), or you can upload one now and once later, or you can just upload it after you're done. It's up to you, although I think uploading now would be great to make the progress visible and encourage others to join translating.

Best
Raphael

rimi-itk:
Hi @raphaelm!

This is great news. The new translation workflow is really cool. I've uploaded my current translations to translate.pretix.eu and will continue translation there.

Thanks!

Best regards,
Mikkel

PR #675, opened 2017-11-08
Fix #634 -- Do not allow deleting the last date of an event series

### LINKED ISSUE #634: Error message if an event series has no date
Hi!

I am using the latest "stable" pretix release from today. When an event series has no scheduled dates, opening the "event selector" (top left corner) leads into an "internal server error".

image

image

With best regards
Jan Felix

aimanparvaiz:


src/tests/control/test_events.py
```diff
@@ -683,8 +684,13 @@ def test_delete(self):
         doc = self.get_doc('/control/event/ccc/30c3/subevents/%d/delete' % self.subevent1.pk)
         assert doc.select("button")
         doc = self.post_doc('/control/event/ccc/30c3/subevents/%d/delete' % self.subevent1.pk, {})
-        assert doc.select(".alert-success")
+        if self.event1.subevents.count() > 1:  # we created 2 subevents. Number of subevents left after one is deleted
```
raphaelm: What is this if statement doing here? This should always be the case, so its unnecessary. Even worse: If it is *not* true, the further checks are skipped and the test will pass! Better use ``assert self.event1.subevents.count() > 1`` if you really want to check this instead :)

src/pretix/control/views/subevents.py
```diff
@@ -103,6 +103,9 @@ def delete(self, request, *args, **kwargs):
             messages.error(request, pgettext_lazy('subevent', 'A date can not be deleted if orders already have been '
                                                   'placed.'))
             return HttpResponseRedirect(self.get_success_url())
+        if self.get_object().event.subevents.count() == 1:  # checking if this is the last date in the event series
+            messages.error(request, pgettext_lazy('subevent', 'Last date of an event series can not be deleted'))
```
raphaelm: If we put "The last date […] can not be deleted." (with a dot at the end and a The in the beginning) it looks a bit better, I guess :)

src/tests/control/test_events.py
aimanparvaiz: Yeah, you are right, since we are creating 2 subevents, this would always be true. Will fix it soon,

src/pretix/control/views/subevents.py
aimanparvaiz: sounds good.

aimanparvaiz:
@raphaelm took care of both the points :) Thanks.

src/tests/control/test_events.py
```diff
@@ -684,13 +684,12 @@ def test_delete(self):
         doc = self.get_doc('/control/event/ccc/30c3/subevents/%d/delete' % self.subevent1.pk)
         assert doc.select("button")
         doc = self.post_doc('/control/event/ccc/30c3/subevents/%d/delete' % self.subevent1.pk, {})
-        if self.event1.subevents.count() > 1:  # we created 2 subevents. Number of subevents left after one is deleted
-            assert doc.select(".alert-success")
+        assert doc.select(".alert-success")
         if self.event1.subevents.count() == 1:  # last one, can't be deleted
```
raphaelm: The same as for the other if statement holds true here again :wink: This is never false, if everything goes well. Replace by ``assert self.event1.subevents.count() == 1`` :)

src/tests/control/test_events.py
```diff
@@ -684,13 +684,12 @@ def test_delete(self):
...
             doc = self.post_doc('/control/event/ccc/30c3/subevents/%d/delete' % self.subevent2.pk, {})
             assert doc.select(".alert-danger")
+            assert not SubEventItem.objects.filter(pk=self.subevent2.pk).exists()
```
raphaelm: Why ``assert not …exists()``? Shouldn't it be ``assert … exists()`` as the deletion should fail? :)

src/tests/control/test_events.py
aimanparvaiz: yep, sorry man! I am new to unit tests, still trying to get a hold of it.

src/tests/control/test_events.py
aimanparvaiz: 👍

PR #676, opened 2017-11-09
Refs #654 -- API: Writable item endpoints

turegjorup:
Starting work on the `items` API endpoint. Work is far from complete, but I already see one issue that should be discussed. In the current `get`  endpoint `variations` and `addons` are embedded. If we maintain that when we implement create and update we need to think carefully about how `PUT/PATCH` should work:

> What is the correct behavior of a partial update with nested items? Well, these are some questions you need to answer:
>  * When passing a list of nested items where the list does not contain all items that currently exist in my database, should I delete the not present items or just ignore them?
>  * For the items that are present, should I require complete items or allow updating only some fields?
> * If some of the items are invalid but others are valid, should the valid items still be updated, or should the whole request be invalidated?
> * If some of the nested items are invalid, should the main object (Invoice) still be updated, or should the whole request be invalidated?

(From [reddit discussion](https://www.reddit.com/r/django/comments/4jgov5/djangorestframework_updating_nested_serializer/))

First decision is if we maintain the current layout where they are embedded, or do we switch to use relations by primary key as in the quotas endpoint? Using primary keys might be easier to implement, but it will be a _[backwards-incompatible](https://docs.pretix.eu/en/latest/api/fundamentals.html#compatibility)_ change and as such require a new version of the API.

Any thoughts on this will be greatly appreciated 😃


raphaelm:
> Starting work on the items API endpoint. Work is far from complete, but I already see one issue that should be discussed. In the current get endpoint variations and addons are embedded. If we maintain that when we implement create and update we need to think carefully about how PUT/PATCH should work:

Glad you asked :wink: I feared that this will come up.

**If** we keep them embedded, I'd answer the questions as follows:

> When passing a list of nested items where the list does not contain all items that currently exist in my database, should I delete the not present items or just ignore them?

They should be deleted because otherwise there would not be a way to delete them.

> For the items that are present, should I require complete items or allow updating only some fields?

I think some fields would be fine, as it would still be clearly defined.

> If some of the items are invalid but others are valid, should the valid items still be updated, or should the whole request be invalidated?
> If some of the nested items are invalid, should the main object (Invoice) still be updated, or should the whole request be invalidated?

In both cases, the whole request should be invalidated, since all requests our API currently are kinda "atomic" – if the serializers fail to validate, no action is ever performed, a 400 request can always be fixed and retried.

> First decision is if we maintain the current layout where they are embedded, or do we switch to use relations by primary key as in the quotas endpoint?

I think the cleanest design would be to have a nested endpoint, e.g.

```
GET /api/v1/organizers/(organizer)/events/(event)/items/(id)/variations/
POST /api/v1/organizers/(organizer)/events/(event)/items/(id)/variations/
PATCH /api/v1/organizers/(organizer)/events/(event)/items/(id)/variations/(varid)/
DELETE /api/v1/organizers/(organizer)/events/(event)/items/(id)/variations/(varid)/
```

and similarly for addons. Such a nested design is not really supported by Django REST Framework, but we could simulate it using a separate URL router (see ``api/urls.py``) the same way as we nest objects into the Organizer and Event resources already. In my eyes, this works well enough that we don't need a new complex dependency like [drf-nested-routers](https://github.com/alanjds/drf-nested-routers)

What do you think about this?

> Using primary keys might be easier to implement, but it will be a backwards-incompatible change and as such require a new version of the API.

Actually, it wouldn't. As long as the item endpoint still includes the ``variations`` and ``addons`` lists in its **output**, we stay compatible. This is not only desirable for compatibility, it is probably also very convenient for API clients. We don't need to accept it as an input field here as well.

turegjorup:
> They should be deleted because otherwise there would not be a way to delete them.

Well, there would be if we distinguish between `PUT` and `PATCH`. For `PATCH` partial updates would be allowed, for `PUT` the full object would be required. This way it would be possible to delete using `PUT` but not using `PATCH`. I think that if we go with embedded write this would be the best way. I do not like mixing behavior, e.g. that you can leave out fields to not update them, but if you leave out embedded object they are removed.

>  think the cleanest design would be to have a nested endpoint

Agreed, so the above discussion is not that relevant.

> Actually, it wouldn't. As long as the item endpoint still includes the variations and addons lists in its output, we stay compatible.

True, and probably the best option. I wasn't completely clear on why I considered it a breaking change. The purist in me would like for a design where valid output is also valid input. Hence any data that you receive with a `GET` from `/id/` should be valid in a `PUT` to `/id/`. Doing this would exclude the combined approach you recommend. However I think it's a valid compromise you suggest so this is what I'll work to implement. 

So to summarize I will leave the `GET` endpoints for `/items` unchanged and implement the additional endpoints for `variations` and `addons` unless there are additional feedback 😄

turegjorup:
Need a bit of guidance here if you have the time @raphaelm. I have tried following your advice in order to implement `GET /api/v1/organizers/(organizer)/events/(event)/items/(id)/variations/`.
But I can't seem to get it working. 

I have added an `item_router`, `ItemVariationViewSet` and `ItemVariationSerializer`. The router seems to work but the Item doesn't seem to be fully hydrated like the event in the request. I get `'RelatedManager' object has no attribute 'variations'` error.

Any hints as to what I'm missing would be greatly appreciated.

raphaelm:
> The purist in me would like for a design where valid output is also valid input.

The purist in me as well, but pretix is written by the pragmatist in me ;)

> Hence any data that you receive with a GET from /id/ should be valid in a PUT to /id/.

It could be *valid*, but the ``variations`` part would be ignored, so it wouldn't be sufficient.

> So to summarize I will leave the GET endpoints for /items unchanged and implement the additional endpoints for variations and addons unless there are additional feedback :smile:

Perfect!

> I have added an item_router, ItemVariationViewSet and ItemVariationSerializer. The router seems to work but the Item doesn't seem to be fully hydrated like the event in the request. I get 'RelatedManager' object has no attribute 'variations' error.

The problem is here:
```
    def get_queryset(self):
        return self.request.event.items.variations.all()
```
This does not work, we need to tell which item we are referring to. Unlike events, this is not already conveniently made available on the request object by our middleware ;)
This could look either like this:
```
    def get_queryset(self):
        return ItemVariation.objects.filter(item__pk=self.kwargs['item'], item__event=self.request.event)
```
or like this:
```
    def get_queryset(self):
        try:
            item = self.request.event.items.get(pk=self.kwargs['item'])
        except Item.DoesNotExist:
            # throw django rest frameworks error 404
        return item.variations.all()
```
While the latter uses one more SQL query and is longer, I believe it is more correct as it returns a 404 for the list of variations of an unknown product, instead of an empty list.

turegjorup:
> This does not work, we need to tell which item we are referring to. Unlike events, this is not already conveniently made available on the request object by our middleware

I figured as much but couldn't locate where this happens for Events. Just out of curiosity could you point to where this is? Also, wasn't sure what the best option was. To extend the middleware to  make items available on the request object or to do the load in the API. 

> This could look either like this....

Agreed on the `404` part. Any reason to not just use `get_object_or_404()` like this?
```    
def get_queryset(self):
    item = get_object_or_404(Item, pk=self.kwargs['item'])
    return item.variations.all()
```

raphaelm:
> I figured as much but couldn't locate where this happens for Events. Just out of curiosity could you point to where this is?

Uh, in case of the API it actually doesn't happen in a middleware (like it does for the normal interface), but in the permission layer: https://github.com/pretix/pretix/blob/master/src/pretix/api/auth/permission.py#L27

> Also, wasn't sure what the best option was. To extend the middleware to make items available on the request object or to do the load in the API.

To do it in the API. For events, we do it on a different level because we basically need that information *everywhere* and also need it for permission checking already.

> Agreed on the 404 part. Any reason to not just use get_object_or_404() like this?

Yes, ``get_object_or_404`` will throw Django's 404 error page (a nice html page), but we want to throw Rest Frameworks JSON-aware error page.

turegjorup:
> Yes, get_object_or_404 will throw Django's 404 error page (a nice html page), but we want to throw Rest Frameworks JSON-aware error page.

Are you sure? When I test it I get this response:
```
HTTP/1.0 404 Not Found
Date: Wed, 22 Nov 2017 10:43:01 GMT
Server: WSGIServer/0.2 CPython/3.6.3
Content-Type: application/json
Allow: GET, POST, HEAD, OPTIONS
Vary: Accept-Language, Cookie, Host
Content-Language: en
X-Frame-Options: DENY
Content-Length: 23

{"detail":"Not found."}
```

raphaelm:
> Are you sure?

No, I'm not. Sometimes I'm just speculating. You are apparently correct, go ahead and use get_object_or_404 ;)

turegjorup:
OK - next question. How to deal with the position field on variations. I have been trying to add a clean positions method to check for duplicate positions. But I have trouble making the hydrated item available to the serializer.

So before work more on that we should decide how stringent we want to be regarding position:
* Do we allow duplicate positions?
* Do we allow non-sequential positions for the variations on an item, e.g 1,2,5?

raphaelm:
> Do we allow duplicate positions?
> Do we allow non-sequential positions for the variations on an item, e.g 1,2,5?

Actually, with the nested endpoint instead of the nested resource, we *need* to allow at least one of these two violations since the API does not have a concept of transactions. If we strictly enforce both, how would you insert a variation somewhere in the middle? You could neither "create a gap" nor "slide it in".

I'd say we can be relaxed about this with no harm. Non-sequential positions (1,2,5) are well defined in their behaviour when sorting the list of variations. They will get overridden by sorting operations performed through the web interface, but I think that is fine. Duplicate positions lead to non-predictable but stable sorting behaviour, so I think the damage here is very small as well. I think we can let the API user decide whether they care about this, so a warning in the docs should be enough.

turegjorup:
Apologies for the absence. Other projects took priority for a while. Still trying to implement the endpoint for `items` and nested endpoints for `variations` and `addons`. I believe the overall structure is in place but I'm struggling with some specific problems. Hoping you can provide some pointers to solutions again @raphaelm.

For the `DEL variations` endpoint I have checks to make sure that variations with order positions or cart positions can't be deleted. The data model ensures this, but it doesn't give a proper HTTP status code, just a `500`. I have tried to raise a validation error similar to how I have done for quotas. However, this also gives a `500`, not a `400` as expected. (Also, should this be a `400` or something else like such as `403 Forbidden` or `409 Conflict`?)

For the addons I'm missing something relative to the relation to the addon_category. When calling POST I this error:
```
Got a `TypeError` when calling `ItemAddOn.objects.create()`. This may be because 
you have a writable field on the serializer class that is not a valid argument to 
`ItemAddOn.objects.create()`. You may need to make the field read-only, or override 
the ItemAddOnSerializer.create() method to handle this correctly.
```
I have tried to look for tricks for how to handle relations but haven't been able to spot anything I have missed.

Last issue is regarding images and how to handle uploads over a rest api? Should we make this another nested endpoint under items like `/api/v1/organizers/bigevents/events/sampleconf/items/1/picture/` ? I have looked at [django-rest-imageupload-example](https://github.com/ChristianKreuzberger/django-rest-imageupload-example) but haven't had the time to try an actual implementation yet.

doc/api/resources/variations.rst
```diff
@@ -0,0 +1,256 @@
...
+      HTTP/1.1 200 OK
+      Vary: Accept
+      Content-Type: application/json
+
+      {
+            "id": 3,
+            "value": {
+                  "en": "Student"
+            },
+             "default_price": "10.00",
```
raphaelm: Indentation is off here, but that's a minor issue :wink:

doc/api/resources/variations.rst
```diff
@@ -0,0 +1,256 @@
...
+
+      POST /api/v1/organizers/(organizer)/events/(event)/items/(id)/variations/ HTTP/1.1
+      Host: pretix.eu
+      Accept: application/json, text/javascript
+      Content: application/json
+
+      {
+          "value": {"en": "Student"},
+          "default_price": "10.00",
+          "price": "10.00",
```
raphaelm: ``price`` is a read-only property, we should mention that and change the sample accordingly.

doc/api/resources/variations.rst
```diff
@@ -0,0 +1,256 @@
...
+   :statuscode 401: Authentication failure
+   :statuscode 403: The requested organizer/event does not exist **or** you have no permission to create this resource.
+
+.. http:patch:: /api/v1/organizers/(organizer)/events/(event)/items/(id)/variations/(id)/
+
+   Update a variation. You can also use ``PUT`` instead of ``PATCH``. With ``PUT``, you have to provide all fields of
+   the resource, other fields will be resetted to default. With ``PATCH``, you only need to provide the fields that you
+   want to change.
+
+   You can change all fields of the resource except the ``id`` field.
```
raphaelm: and the price field, since that one is generated

src/pretix/api/views/item.py
```diff
@@ -28,18 +29,149 @@ class Meta:
...
+    def perform_update(self, serializer):
+        serializer.save(event=self.request.event)
+        serializer.instance.log_action(
+            'pretix.event.item.changed',
+            user=self.request.user,
+            api_token=(self.request.auth if isinstance(self.request.auth, TeamAPIToken) else None),
+            data=self.request.data
+        )
+
+    def perform_destroy(self, instance):
```
raphaelm: As for variations, we need a check here that the item has not been ordered so far

src/pretix/api/views/item.py
```diff
@@ -28,18 +29,149 @@ class Meta:
...
+    serializer_class = ItemVariationSerializer
+    queryset = ItemVariation.objects.none()
+    filter_backends = (DjangoFilterBackend, OrderingFilter,)
+    ordering_fields = ('id', 'position')
+    ordering = ('id',)
+    permission = 'can_change_items'
+    write_permission = 'can_change_items'
+
+    def get_queryset(self):
+        item = get_object_or_404(Item, pk=self.kwargs['item'])
```
raphaelm: We need to check that this is of the correct event, e.g. ``get_object_or_404(Item, pk=self.kwargs['item'], event=self.request.event)``

src/pretix/api/views/item.py
```diff
@@ -28,18 +29,149 @@ class Meta:
...
+        return item.variations.all()
+
+    def perform_create(self, serializer):
+        item = get_object_or_404(Item, pk=self.kwargs['item'])
```
raphaelm: same here, check for event

src/pretix/api/views/item.py
```diff
@@ -28,18 +29,149 @@ class Meta:
...
+        serializer.save(item=item)
+        item.log_action(
+            'pretix.event.item.variation.added',
+            user=self.request.user,
+            api_token=(self.request.auth if isinstance(self.request.auth, TeamAPIToken) else None),
+            data={**self.request.data, **{'ORDER': serializer.instance.position}, **{'id': serializer.instance.pk}}
+        )
+
+    def perform_update(self, serializer):
+        variation = get_object_or_404(ItemVariation, pk=self.kwargs['pk'])
```
raphaelm: Shouldn't this already be in ``serializer.instance`` (*with* the event check already)?

src/pretix/api/views/item.py
```diff
@@ -28,18 +29,149 @@ class Meta:
...
+        serializer.save(variation=variation)
+        variation.item.log_action(
+            'pretix.event.item.variation.changed',
+            user=self.request.user,
+            api_token=(self.request.auth if isinstance(self.request.auth, TeamAPIToken) else None),
+            data={**self.request.data['value'], **{'id': self.kwargs['pk']}}
+        )
+
+    def destroy(self, request, *args, **kwargs):
+        variation = get_object_or_404(ItemVariation, pk=self.kwargs['pk'])
```
raphaelm: I suggest overriding ``perform_destroy`` instead, then you don't need to get the variation yourself

src/pretix/api/views/item.py
```diff
@@ -28,18 +29,149 @@ class Meta:
...
+    ordering = ('id',)
+    permission = 'can_change_items'
+    write_permission = 'can_change_items'
+
+    def get_queryset(self):
+        item = get_object_or_404(Item, pk=self.kwargs['item'])
+        return item.addons.all()
+
+    def perform_create(self, serializer):
+        item = get_object_or_404(Item, pk=self.kwargs['item'])
```
raphaelm: same as for variations

src/pretix/api/views/item.py
```diff
@@ -28,18 +29,149 @@ class Meta:
...
+    serializer_class = ItemAddOnSerializer
+    queryset = ItemAddOn.objects.none()
+    filter_backends = (DjangoFilterBackend, OrderingFilter,)
+    ordering_fields = ('id', 'position')
+    ordering = ('id',)
+    permission = 'can_change_items'
+    write_permission = 'can_change_items'
+
+    def get_queryset(self):
+        item = get_object_or_404(Item, pk=self.kwargs['item'])
```
raphaelm: same as for variations

src/pretix/api/views/item.py
```diff
@@ -28,18 +29,149 @@ class Meta:
...
+        serializer.save(item=item, addon_category=category)
+        item.log_action(
+            'pretix.event.item.addons.added',
+            user=self.request.user,
+            api_token=(self.request.auth if isinstance(self.request.auth, TeamAPIToken) else None),
+            data={**self.request.data, **{'ORDER': serializer.instance.position}, **{'id': serializer.instance.pk}}
+        )
+
+    def perform_update(self, serializer):
+        addon = get_object_or_404(ItemAddOn, pk=self.kwargs['pk'])
```
raphaelm: same as for variations

src/pretix/api/views/item.py
```diff
@@ -28,18 +29,149 @@ class Meta:
...
+        serializer.save(addon=addon)
+        addon.item.log_action(
+            'pretix.event.item.addons.changed',
+            user=self.request.user,
+            api_token=(self.request.auth if isinstance(self.request.auth, TeamAPIToken) else None),
+            data={**self.request.data['value'], **{'id': self.kwargs['pk']}}
+        )
+
+    def perform_destroy(self, serializer):
+        addon = get_object_or_404(ItemAddOn, pk=self.kwargs['pk'])
```
raphaelm: the signature of this is `` def perform_destroy(self, instance):``, not ``serializer``, you don't need to retrieve the addon yourself

src/pretix/base/models/items.py
```diff
@@ -476,6 +498,19 @@ def __lt__(self, other):
             return self.id < other.id
         return self.position < other.position
 
+    @staticmethod
+    def clean_order_positions(variation):
+        if not len(variation.orderposition_set.all()):
```
raphaelm: ``if not variation.orderposition_set.exists()`` is more efficient **here**. It might be less efficient in other places since ``.all()`` might already be cached due to ``prefetch_related`` calls, but that doesn't affect us here.

src/pretix/base/models/items.py
```diff
@@ -476,6 +498,19 @@ def __lt__(self, other):
...
+            raise ValidationError('The variation cannot be deleted because it has already been ordered by a user. '
+                                  'Please set the variation as "inactive" instead.')
+
+    @staticmethod
+    def clean_cart_positions(variation):
+        if len(variation.cartposition_set.all()):
```
raphaelm: same

doc/api/resources/variations.rst
```diff
@@ -0,0 +1,256 @@
+Variations
```
raphaelm: We could call this page ``Item variations`` to make it clearer to see the relationship in the table of contents.

raphaelm:
> Apologies for the absence. Other projects took priority for a while.

No need to worry, happy to see you back :)

> Still trying to implement the endpoint for items and nested endpoints for variations and addons. I believe the overall structure is in place but I'm struggling with some specific problems. Hoping you can provide some pointers to solutions again @raphaelm.

I'll try my best!

> For the DEL variations endpoint I have checks to make sure that variations with order positions or cart positions can't be deleted. The data model ensures this, but it doesn't give a proper HTTP status code, just a 500. I have tried to raise a validation error similar to how I have done for quotas. However, this also gives a 500, not a 400 as expected.

Good question, I'm not really sure why. My guess would be that the concept of validation (and thus, ValidationErrors) are not usually used for *deletions*. I'd propose to implement this in the same way as I did for TaxRules, by adding a ``Variation.allow_delete()`` method that can also be used by the respective view but doesn't raise an error, see here: https://github.com/pretix/pretix/blob/master/src/pretix/api/views/event.py#L67

> (Also, should this be a 400 or something else like such as 403 Forbidden or 409 Conflict?)

I believe it shouldn't be a 400 since the request in itself is valid and it shouldn't be a 409 because it is unresolvable, so 403 sounds good to me. But I guess that is a matter of taste.

> For the addons I'm missing something relative to the relation to the addon_category. When calling POST I this error:
> Got a `TypeError` when calling `ItemAddOn.objects.create()`. This may be because 
you have a writable field on the serializer class that is not a valid argument to 
`ItemAddOn.objects.create()`. You may need to make the field read-only, or override 
the ItemAddOnSerializer.create() method to handle this correctly.
I have tried to look for tricks for how to handle relations but haven't been able to spot anything I have missed.

The "wrong" argument isn't ``addon_category``. Instead, `` serializer.save(item=item, addon_category=category)`` should be `` serializer.save(base_item=item, addon_category=category)`` because…well…naming things is hard? :confused: 

> Last issue is regarding images and how to handle uploads over a rest api? Should we make this another nested endpoint under items like /api/v1/organizers/bigevents/events/sampleconf/items/1/picture/ ? I have looked at django-rest-imageupload-example but haven't had the time to try an actual implementation yet.

Excellent question. I don't know. We can either have a separate endpoint for every image, or we could to it like Stripe and have one global endpoint that allows uploading files that can later be referenced in a POST/PATCH call: https://stripe.com/docs/file-upload

Having separate endpoints is probably easier, especially with regards to getting rid of unused files.

User file handling in pretix is far from optimal right now. If you want, implement it, but I'd also happily merge the PR if you just document that the picture is  read-only via the API right now, I think that would be fine for a start as well.

turegjorup:
Hi @raphaelm 

First of, happy new year 🎊 

As for this PR I think we're getting close. I believe it is feature complete now, however there are still a couple of questions:

I have one failing test I can't figure out. In `test_items_with_order_position_not_delete` the status code returns a `405`, not a `403` however when I test manually with a rest client I get the `403`. Also `test_items_with_cart_position_not_delete` works fine with a comparable setup. My guess is that it somehow relates to the `order_position` fixture, but I'm unsure how to fix it. Running in debug mode reveals that [`allow_delete`](https://github.com/pretix/pretix/pull/676/files#diff-7db8f68a3fef932b5de07f4528fdf0a1R372) on `Item` returns true when I run the tests, but false when I test manually. It seems that the [`order_position`](https://github.com/pretix/pretix/pull/676/files#diff-9b8ded6effe3d55668eca3effeb20088R43) fixtures has not run yet when `allow_delete` is called.

I have added checks for the correct event as you suggested and I have expanded on the URLs in `event_permission_urls` in test_permissions.py but I'm not sure if we cover all the access scenarios for the new nested endpoints?

As for the [addon/add-on question](https://github.com/pretix/pretix/issues/727) I have attempted to use "add-on" in all descriptions and error texts but I have not changed field names in the API, and I have opted to use "addon" as the nested endpoint so that it corresponds to the field name.

src/pretix/api/views/item.py
```diff
@@ -28,18 +31,155 @@ class Meta:
...
+
+    def perform_create(self, serializer):
+        item = get_object_or_404(Item, pk=self.kwargs['item'], event=self.request.event)
+        serializer.save(item=item)
+        item.log_action(
+            'pretix.event.item.variation.added',
+            user=self.request.user,
+            api_token=(self.request.auth if isinstance(self.request.auth, TeamAPIToken) else None),
+            data={**self.request.data, **{'ORDER': serializer.instance.position}, **{'id': serializer.instance.pk},
+                  **{'value': serializer.instance.value}}
```
raphaelm: We currently still support Python 3.4 (and will continue to do so until Django stops doing so with Django 2.1). The ``{**…}`` syntax is not supported before 3.5. We therefore unfortunately have to fall back on a solution like

doc/api/resources/item_variations.rst
```diff
@@ -0,0 +1,255 @@
...
+      HTTP/1.1 200 OK
+      Vary: Accept
+      Content-Type: application/json
+
+      {
+        "id": 3,
+        "value": {
+              "en": "Student"
+        },
+         "default_price": "10.00",
```
raphaelm: Indentation seems off here

doc/api/resources/item_variations.rst
```diff
@@ -0,0 +1,255 @@
...
+          }
+        ]
+      }
+
+   :query integer page: The page number in case of a multi-page result set, default is 1
+   :query boolean active: If set to ``true`` or ``false``, only items with this value for the field ``active`` will be
+                          returned.
+   :param organizer: The ``slug`` field of the organizer to fetch
+   :param event: The ``slug`` field of the event to fetch
+   :param id: The ``id`` field of the item
```
raphaelm: id → item

doc/api/resources/item_variations.rst
```diff
@@ -0,0 +1,255 @@
...
+   :statuscode 200: no error
+   :statuscode 401: Authentication failure
+   :statuscode 403: The requested organizer/event/item does not exist **or** you have no permission to view this resource.
+
+.. http:get:: /api/v1/organizers/(organizer)/events/(event)/items/(id)/variations/(id)/
```
raphaelm: Using ``(id)`` twice here seems confusing, better replace the first one with ``item``

doc/api/resources/item_variations.rst
```diff
@@ -0,0 +1,255 @@
...
+
+   :param organizer: The ``slug`` field of the organizer to fetch
+   :param event: The ``slug`` field of the event to fetch
+   :param id: The ``id`` field of the item
+   :param id: The ``id`` field of the variation to fetch
+   :statuscode 200: no error
+   :statuscode 401: Authentication failure
+   :statuscode 403: The requested organizer/event does not exist **or** you have no permission to view this resource.
+
+.. http:post:: /api/v1/organizers/bigevents/events/sampleconf/items/1/variations/
```
raphaelm: This is a headline and should be
``/api/v1/organizers/(organizer)/events/(event)/items/(item)/variations/``

doc/api/resources/item_variations.rst
```diff
@@ -0,0 +1,255 @@
...
+         "default_price": "10.00",
+         "price": "10.00",
+         "active": true,
+         "description": null,
+         "position": 0
+      }
+
+   :param organizer: The ``slug`` field of the organizer to fetch
+   :param event: The ``slug`` field of the event to fetch
+   :param id: The ``id`` field of the item
```
raphaelm: id → item

doc/api/resources/item_variations.rst
```diff
@@ -0,0 +1,255 @@
...
+
+.. http:post:: /api/v1/organizers/bigevents/events/sampleconf/items/1/variations/
+
+   Creates a new variation
+
+   **Example request**:
+
+   .. sourcecode:: http
+
+      POST /api/v1/organizers/(organizer)/events/(event)/items/(id)/variations/ HTTP/1.1
```
raphaelm: This is a sample and should be ``/api/v1/organizers/bigevents/events/sampleconf/items/1/variations/``

doc/api/resources/item_variations.rst
```diff
@@ -0,0 +1,255 @@
...
+          "default_price": "10.00",
+          "price": "10.00",
+          "active": true,
+          "description": null,
+          "position": 0
+      }
+
+   :param organizer: The ``slug`` field of the organizer of the event/item to create a variation for
+   :param event: The ``slug`` field of the event to create a variation for
+   :param id: The ``id`` field of the item to create a variation for
```
raphaelm: id → item

doc/api/resources/item_variations.rst
```diff
@@ -0,0 +1,255 @@
...
+   :statuscode 201: no error
+   :statuscode 400: The variation could not be created due to invalid submitted data.
+   :statuscode 401: Authentication failure
+   :statuscode 403: The requested organizer/event does not exist **or** you have no permission to create this resource.
+
+.. http:patch:: /api/v1/organizers/(organizer)/events/(event)/items/(id)/variations/(id)/
```
raphaelm: first id → item

doc/api/resources/item_add-ons.rst
```diff
@@ -0,0 +1,241 @@
+Item add-ons
+=====
+
+Resource description
+--------------------
+
+With add-ons, you can specify products that can be bought as an addition to this product. For example, if you host a
+conference with a base conference ticket and a number of workshops, you could define the workshops as add-ons to the
+conference ticket. With this configuration, the workshops cannot be bought on their own but only in combination with a
+conference ticket. You can here specify categories of products that can be used as add-ons to this product. You can also
```
raphaelm: this product → a specific product

doc/api/resources/item_add-ons.rst
```diff
@@ -0,0 +1,241 @@
...
+              "position": 1,
+              "price_included": true
+            }
+          ]
+        }
+
+   :query integer page: The page number in case of a multi-page result set, default is 1
+   :param organizer: The ``slug`` field of the organizer to fetch
+   :param event: The ``slug`` field of the event to fetch
+   :param id: The ``id`` field of the item
```
raphaelm: id → item

doc/api/resources/item_add-ons.rst
```diff
@@ -0,0 +1,241 @@
...
+   :statuscode 200: no error
+   :statuscode 401: Authentication failure
+   :statuscode 403: The requested organizer/event/item does not exist **or** you have no permission to view this resource.
+
+.. http:get:: /api/v1/organizers/(organizer)/events/(event)/items/(id)/addons/(id)/
```
raphaelm: first (id) → (item)

doc/api/resources/item_add-ons.rst
```diff
@@ -0,0 +1,241 @@
...
+
+   **Example response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Vary: Accept
+      Content-Type: application/json
+
+        {
```
raphaelm: I think it is confusing if the JSON part of the response is further indented than the headers (goes for all the samples).

doc/api/resources/item_add-ons.rst
```diff
@@ -0,0 +1,241 @@
...
+          "addon_category": 1,
+          "min_count": 0,
+          "max_count": 10,
+          "position": 1,
+          "price_included": true
+        }
+
+   :param organizer: The ``slug`` field of the organizer to fetch
+   :param event: The ``slug`` field of the event to fetch
+   :param id: The ``id`` field of the item
```
raphaelm: id → item

doc/api/resources/item_add-ons.rst
```diff
@@ -0,0 +1,241 @@
...
+
+.. http:post:: /api/v1/organizers/bigevents/events/sampleconf/items/1/addons/
+
+   Creates a new add-on
+
+   **Example request**:
+
+   .. sourcecode:: http
+
+      POST /api/v1/organizers/(organizer)/events/(event)/items/(id)/addons/ HTTP/1.1
```
raphaelm: id → item

doc/api/resources/item_add-ons.rst
```diff
@@ -0,0 +1,241 @@
...
+          "addon_category": 1,
+          "min_count": 0,
+          "max_count": 10,
+          "position": 1,
+          "price_included": true
+        }
+
+   :param organizer: The ``slug`` field of the organizer of the event/item to create a add-on for
+   :param event: The ``slug`` field of the event to create a add-on for
+   :param id: The ``id`` field of the item to create a add-on for
```
raphaelm: id → item

doc/api/resources/item_add-ons.rst
```diff
@@ -0,0 +1,241 @@
...
+   :statuscode 201: no error
+   :statuscode 400: The add-on could not be created due to invalid submitted data.
+   :statuscode 401: Authentication failure
+   :statuscode 403: The requested organizer/event does not exist **or** you have no permission to create this resource.
+
+.. http:patch:: /api/v1/organizers/(organizer)/events/(event)/items/(id)/addon/(id)/
```
raphaelm: id → item

doc/api/resources/item_add-ons.rst
```diff
@@ -0,0 +1,241 @@
...
+          "addon_category": 1,
+          "min_count": 0,
+          "max_count": 10,
+          "position": 1,
+          "price_included": true
+        }
+
+   :param organizer: The ``slug`` field of the organizer to modify
+   :param event: The ``slug`` field of the event to modify
+   :param id: The ``id`` field of the item to modify
```
raphaelm: id → item

src/pretix/api/serializers/item.py
```diff
@@ -16,11 +16,34 @@ class Meta:
...
+        fields = ('id', 'value', 'active', 'description',
+                  'position', 'default_price', 'price')
+
+
 class InlineItemAddOnSerializer(serializers.ModelSerializer):
     class Meta:
         model = ItemAddOn
         fields = ('addon_category', 'min_count', 'max_count',
-                  'position')
+                  'position', 'price_included')
```
raphaelm: Would you mind adding this new field to the documentation and examples in ``items.rst`` as well (together with a ``versionchanged`` block)? :)

doc/api/resources/item_add-ons.rst
```diff
@@ -0,0 +1,241 @@
...
+===================================== ========================== =======================================================
+Field                                 Type                       Description
+===================================== ========================== =======================================================
+id                                    integer                    Internal ID of the add-on
+addon_category                        integer                    Internal ID of the item category the add-on can be
+                                                                 chosen from.
+min_count                             integer                    The minimal number of add-ons that need to be chosen.
+max_count                             integer                    The maximal number of add-ons that can be chosen.
+position                              integer                    An integer, used for sorting
+===================================== ========================== =======================================================
```
raphaelm: Please add a 
```
.. versionchanged:: 1.12

   This resource has been added.
```
for consistency :)

doc/api/resources/item_variations.rst
```diff
@@ -0,0 +1,255 @@
...
+default_price                         money (string)             The price set directly for this variation or ``null``
+price                                 money (string)             The price used for this variation. This is either the
+                                                                 same as ``default_price`` if that value is set or equal
+                                                                 to the item's ``default_price`` (read-only).
+active                                boolean                    If ``False``, this variation will not be sold or shown.
+description                           multi-lingual string       A public description of the variation. May contain
+                                                                 Markdown syntax or can be ``null``.
+position                              integer                    An integer, used for sorting
+===================================== ========================== =======================================================
+
```
raphaelm: Please add a 
```
.. versionchanged:: 1.12

   This resource has been added.
```
for consistency :)

src/pretix/api/serializers/item.py
```diff
@@ -44,6 +67,18 @@ class Meta:
                   'require_voucher', 'hide_without_voucher', 'allow_cancel',
                   'min_per_order', 'max_per_order', 'checkin_attention', 'has_variations',
                   'variations', 'addons')
+        read_only_fields = ('variations', 'addons')
```
raphaelm: Should we declare ``picture`` read-only for now and document that? Also, ``has_variations`` is read-only, this is a calculated property that can't be changed (and should be documented as such).

src/pretix/base/models/items.py
```diff
@@ -369,10 +369,40 @@ def check_quotas(self, ignored_quotas=None, count_waitinglist=True, subevent=Non
         return min([q.availability(count_waitinglist=count_waitinglist, _cache=_cache) for q in check_quotas],
                    key=lambda s: (s[0], s[1] if s[1] is not None else sys.maxsize))
 
+    def allow_delete(self):
+        from pretix.base.models.orders import CartPosition, OrderPosition
+
+        return (
+            not OrderPosition.objects.filter(item=self).exists()
+            and not CartPosition.objects.filter(item=self).exists()
+        )
```
raphaelm: We could also use this in ``pretix/control/views/item.py:1026``, don't you think?

src/pretix/base/models/items.py
```diff
@@ -369,10 +369,40 @@ def check_quotas(self, ignored_quotas=None, count_waitinglist=True, subevent=Non
...
+
     @cached_property
     def has_variations(self):
         return self.variations.exists()
 
+    @staticmethod
+    def clean_per_order(min_per_order, max_per_order):
+        if min_per_order is not None and max_per_order is not None:
+            if min_per_order > max_per_order:
+                raise ValidationError(_('max_per_order must be greater than min_per_order (or null for no limitation).'))
```
raphaelm: In case we also want to use this for the frontend, let's rephrase it more human-friendly (that is also understood by a user who does not know the field names) as e.g. ``The maximum number per order can not be lower than the minimum number per order``.

src/pretix/base/models/items.py
```diff
@@ -369,10 +369,40 @@ def check_quotas(self, ignored_quotas=None, count_waitinglist=True, subevent=Non
...
+
+    @staticmethod
+    def clean_category(category, event):
+        if category is not None and category.event is not None and category.event != event:
+            raise ValidationError(_('The items category must belong to the same event as the item.'))
```
raphaelm: items  → item's

src/pretix/base/models/items.py
```diff
@@ -369,10 +369,40 @@ def check_quotas(self, ignored_quotas=None, count_waitinglist=True, subevent=Non
...
+
+    @staticmethod
+    def clean_tax_rule(tax_rule, event):
+        if tax_rule is not None and tax_rule.event is not None and tax_rule.event != event:
+            raise ValidationError(_('The items tax_rule must belong to the same event as the item.'))
```
raphaelm: items → item's

src/pretix/base/models/items.py
```diff
@@ -369,10 +369,40 @@ def check_quotas(self, ignored_quotas=None, count_waitinglist=True, subevent=Non
...
+
+    @staticmethod
+    def clean_available(from_date, until_date):
+        if from_date is not None and until_date is not None:
+            if from_date > until_date:
+                raise ValidationError(_('The available_from date must be before the available_until date (or null).'))
```
raphaelm: Same here, I'd suggest e.g. ``The item's availability cannot end before it starts.``

src/pretix/base/models/items.py
```diff
@@ -527,7 +565,18 @@ class Meta:
...
     def clean(self):
-        if self.max_count < self.min_count:
+        self.clean_max_min_numbers(self.max_count, self.min_count)
+
+    @staticmethod
+    def clean_categories(item, addon, new_category):
+        if addon is None or addon.addon_category != new_category:
+            for addon in item.addons.all():
+                if addon.addon_category == new_category:
+                    raise ValidationError(_('The item all ready have an add-on of this category.'))
```
raphaelm: → "The item already has an add-on of this category."

src/pretix/api/views/item.py
```diff
@@ -28,18 +31,155 @@ class Meta:
...
+        )
+
+    def perform_destroy(self, instance):
+        if not instance.allow_delete():
+            raise PermissionDenied('This variation cannot be deleted because it has already been ordered '
+                                   'by a user or currently is in a users\'s cart. Please set the variation as '
+                                   '"inactive" instead.')
```
raphaelm: We need to also check that this is not the last variation of this item. It is illegal to make a product with variations into a product without variations.

src/pretix/api/views/item.py
```diff
@@ -28,18 +31,155 @@ class Meta:
...
+    permission = 'can_change_items'
+    write_permission = 'can_change_items'
+
+    def get_queryset(self):
+        item = get_object_or_404(Item, pk=self.kwargs['item'], event=self.request.event)
+        return item.variations.all()
+
+    def perform_create(self, serializer):
+        item = get_object_or_404(Item, pk=self.kwargs['item'], event=self.request.event)
+        serializer.save(item=item)
```
raphaelm: We need to check that this item already has variations, otherwise it's illegal to add one.

doc/api/resources/item_add-ons.rst
```diff
@@ -0,0 +1,241 @@
...
+   :param id: The ``id`` field of the item to create a add-on for
+   :statuscode 201: no error
+   :statuscode 400: The add-on could not be created due to invalid submitted data.
+   :statuscode 401: Authentication failure
+   :statuscode 403: The requested organizer/event does not exist **or** you have no permission to create this resource.
+
+.. http:patch:: /api/v1/organizers/(organizer)/events/(event)/items/(id)/addon/(id)/
+
+   Update an add-on. You can also use ``PUT`` instead of ``PATCH``. With ``PUT``, you have to provide all fields of
+   the resource, other fields will be resetted to default. With ``PATCH``, you only need to provide the fields that you
```
raphaelm: resetted → reset

doc/api/resources/item_variations.rst
```diff
@@ -0,0 +1,255 @@
...
+   :param id: The ``id`` field of the item to create a variation for
+   :statuscode 201: no error
+   :statuscode 400: The variation could not be created due to invalid submitted data.
+   :statuscode 401: Authentication failure
+   :statuscode 403: The requested organizer/event does not exist **or** you have no permission to create this resource.
+
+.. http:patch:: /api/v1/organizers/(organizer)/events/(event)/items/(id)/variations/(id)/
+
+   Update a variation. You can also use ``PUT`` instead of ``PATCH``. With ``PUT``, you have to provide all fields of
+   the resource, other fields will be resetted to default. With ``PATCH``, you only need to provide the fields that you
```
raphaelm: resetted → reset

doc/api/resources/item_variations.rst
```diff
@@ -0,0 +1,255 @@
+Item variations
+=====
+
+Resource description
+--------------------
+
+Variations of items can be use for products (items) that are available in different sizes, colours or other variations
```
raphaelm: colour → color

src/pretix/api/serializers/item.py
turegjorup: I'll skip trying to implement image uploads for now.  So I have marked both as read-only in the table view for the endpoint. Not sure if there should be a more explicit way to document read-only fields?

src/pretix/base/models/items.py
turegjorup: Done. However it seemed that check was only for order_positions, not for cart_positions?

turegjorup:
Thank you for the thorough review. Most of the issues should be fixed with my latest commits. However there are a few questions left.

> I have one large issue: Many reports and queries in pretix are built on the assumption that `has_variations` never changes, i.e. that an item either always has variations or never has. 

I understand your reasoning and indeed adding a variation automatically might be the best option. However, the problem tempted me to attempt to implement a writeable nested endpoint: 

```
class ItemSerializer(I18nAwareModelSerializer):
    addons = InlineItemAddOnSerializer(many=True, required=False)
    variations = InlineItemVariationSerializer(many=True, required=False)
    tax_rate = ItemTaxRateField(source='*', read_only=True)

    class Meta:
        model = Item
        fields = ('id', 'category', 'name', 'active', 'description',
                  'default_price', 'free_price', 'tax_rate', 'tax_rule', 'admission',
                  'position', 'picture', 'available_from', 'available_until',
                  'require_voucher', 'hide_without_voucher', 'allow_cancel',
                  'min_per_order', 'max_per_order', 'checkin_attention', 'has_variations',
                  'variations', 'addons')
        read_only_fields = ('has_variations', 'picture')

    def create(self, validated_data):
        variations_data = validated_data.pop('variations')
        item = Item.objects.create(**validated_data)
        for variation_data in variations_data:
            ItemVariation.objects.create(item=item, **variation_data)
        return item

    @atomic
    def update(self, validated_data):
        variations_data = validated_data.pop('variations')
        item = Item.objects.create(**validated_data)
        for variation_data in variations_data:
            if variation_data['id'] is not None:
                variation = get_object_or_404(ItemVariation, pk=variation_data['id'])
                for attr, value in validated_data.items():
                    setattr(variation, attr, value)
                variation.save()
            else:
                ItemVariation.objects.create(item=item, **variation_data)
        return item

    def validate(self, data):
        data = super().validate(data)
        event = self.context['event']

        Item.clean_per_order(data.get('min_per_order'), data.get('max_per_order'))
        Item.clean_available(data.get('available_from'), data.get('available_until'))
        Item.clean_category(data.get('category'), event)
        Item.clean_tax_rule(data.get('tax_rule'), event)

        return data
```

This works for `POST`, but for `PATCH` it fails with an `TypeError at /api/v1/organizers/bigevents/events/2020/items/8/ update() takes 2 positional arguments but 3 were given` that I haven't been able to solve. But before I spend too much time on it I wanted to get your take on the idea? I know we initially opted not to do this but I think this problem warrants that we give it another look. This way we could avoid having the unexpected behavior.

Only other issue is for the tests. I now get an error from test_permissions.py L41 where '('post', 'can_change_items', 'items/1/addons/', 400' returns an 404, not a 400. When testing manually it works as expected.

src/pretix/api/serializers/item.py
raphaelm: I think for now it is fine.

src/pretix/base/models/items.py
raphaelm: Yeah, but that was probably more of an oversight. I'm not even sure what the "right" approach is, since blocking because of CartPositions is confusing since the admin cannot see them anywhere. On the other hand, once the item is "out there" it is okay if you can't delete it since you can still disable it.

raphaelm:
> I understand your reasoning and indeed adding a variation automatically might be the best option. However, the problem tempted me to attempt to implement a writeable nested endpoint:
> …
> This works for POST, but for PATCH it fails with an TypeError at /api/v1/organizers/bigevents/events/2020/items/8/ update() takes 2 positional arguments but 3 were given that I haven't been able to solve. But before I spend too much time on it I wanted to get your take on the idea? I know we initially opted not to do this but I think this problem warrants that we give it another look. This way we could avoid having the unexpected behavior.

Well, we initially opted not to do nested resources since their behaviour is completely undefined and unintuitive for PATCH operations. For POST, they are well-defined. What do you think about only supporting them for ``create`` as a way of solving that chicken-and-egg-problem, but continuing to only allow using the sub-resource at a different path for ``update``? Is that too confusing? It probably is less confusing then automatically creating a variation if we document it properly.

> Only other issue is for the tests. I now get an error from test_permissions.py L41 where '('post', 'can_change_items', 'items/1/addons/', 400' returns a 404, not a 400. When testing manually it works as expected.

The reason is that ``get_serializer_context`` is apparently executed before ``serializer.valid()`` and therefore
```
ctx['item'] = get_object_or_404(Item, pk=self.kwargs['item'], event=self.request.event)
```
throws a 404 even before it is noticed that the POST body is invalid. I would argue that the 404 is even correct since we are accessing a subresource of an item that doesn't exist in the test, so we might even want to return a 404 for the respective variations-endpoint as well for consistency.

turegjorup:
> What do you think about only supporting them for create as a way of solving that chicken-and-egg-problem, but continuing to only allow using the sub-resource at a different path for update?

I gave it some thought and discussed it internally in the team. We agreed that this was indeed the best compromise. This is now implemented. For consistency I have opted to do a similar implementation for `add-ons`. The only thing I'm not completely satisfied with is the validation part for the nested add-ons. I followed the example from http://www.django-rest-framework.org/api-guide/relations/#writable-nested-serializers however this seems to bypass the validation defined in the nested serializer. I tried and failed to resolve this and get DRF to validate the nested objects using the `ItemAddOnSerializer` so I ended up calling the validation methods directly from the `ItemSerializer`. 

> I would argue that the 404 is even correct since we are accessing a subresource of an item that doesn't exist in the test, so we might even want to return a 404 for the respective variations-endpoint as well for consistency.

I agree, so this is now the behavior for both endpoints. 

I consider this feature complete now, so I have removed the WIP label. Depending on the outcome of your review it should be ready to be merged 😄.

raphaelm:
Thank you again for this great pull request. Thank you even more for the extensive test cases for each and every error case – this makes the review **so** much easier and creates so much trust in the code, it's really appreciated. The failure on Travis is my fault, I committed something broken to master yesterday, the fix is on its way. Let's be done with this! 🎉

Issue #670, opened 2017-11-03
ux issue: bank details shown too early, position in ordering process unclear


stapelberg:
3 different people (!) had issues ordering tickets for this year’s RGB2R. All of them thought their order was completed because they saw the bank details. This resulted in much unnecessary confusion and unhappy participants who were too late for our shirt order.

A few ideas which would likely help:
* An option to hide the bank details until _after_ the order was accepted
* A clear indicator telling the user that the order is not yet saved

cc @koebi

PR #677, opened 2017-11-09
Implement progress indicator during checkout

raphaelm:
as part of a reaction to #670 

Minimal case:
download 2
download 3

Maximal case:
download

On Mobile:
2017-11-09-164147_320x154_scrot

What do you all think? 

@rixx @stapelberg @breunigs @janx2 opinions on this? :)

### DIFF
--- a/src/pretix/presale/checkoutflow.py
+++ b/src/pretix/presale/checkoutflow.py
@@ -6,7 +6,7 @@ from django.http import HttpResponseNotAllowed, JsonResponse
 from django.shortcuts import redirect
 from django.utils import translation
 from django.utils.functional import cached_property
-from django.utils.translation import ugettext_lazy as _
+from django.utils.translation import pgettext_lazy, ugettext_lazy as _
 from django.views.generic.base import TemplateResponseMixin
 
 from pretix.base.models import Order
@@ -33,6 +33,7 @@ from pretix.presale.views.questions import QuestionsViewMixin
 
 class BaseCheckoutFlowStep:
     requires_valid_cart = True
+    icon = 'pencil'
 
     def __init__(self, event):
         self.event = event
@@ -41,6 +42,10 @@ class BaseCheckoutFlowStep:
     @property
     def identifier(self):
         raise NotImplementedError()
+
+    @property
+    def label(self):
+        return pgettext_lazy('checkoutflow', 'Step')
 
     @property
     def priority(self):
@@ -137,6 +142,11 @@ class TemplateFlowStep(TemplateResponseMixin, BaseCheckoutFlowStep):
         kwargs.setdefault('step', self)
         kwargs.setdefault('event', self.event)
         kwargs.setdefault('prev_url', self.get_prev_url(self.request))
+        kwargs.setdefault('checkout_flow', [
+            step
+            for step in self.request._checkout_flow
+            if step.is_applicable(self.request)
+        ])
         return kwargs
 
     def render(self, **kwargs):
@@ -166,6 +176,8 @@ class AddOnsStep(CartMixin, AsyncAction, TemplateFlowStep):
     task = set_cart_addons
     known_errortypes = ['CartError']
     requires_valid_cart = False
+    label = pgettext_lazy('checkoutflow', 'Add-on products')
+    icon = 'puzzle-piece'
 
     def is_applicable(self, request):
         if not hasattr(request, '_checkoutflow_addons_applicable'):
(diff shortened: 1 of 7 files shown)


rixx:
**Question:** You call it "maximum case" – but can't several plugins add steps?

**Feedback:** I don't like the "done" step, and I don't think it adds anything valuable. Having another circle there implies a step that requires an action, but "done" implies no action required at all. I'd think that being done is clear when all circles show green check marks.

stapelberg:
Thanks for working on this. I think this will be a massive improvement :)

raphaelm:
> Question: You call it "maximum case" – but can't several plugins add steps?

Sure they can. "Maximum currently achievable case with all currently existing plugins activated" I should have called it.

> Feedback: I don't like the "done" step, and I don't think it adds anything valuable. Having another circle there implies a step that requires an action, but "done" implies no action required at all. I'd think that being done is clear when all circles show green check marks.

Well, yeah, that's basically the open question here. I added it in last-minute before pushing this for the following reason: The problem that lead to this proposal was that there were people abandoning their carts too early in the checkout process thinking it was complete. I don't know exactly where they did that, but the confirm page is a good candicate, especially with e.g. PayPal payments where you already have been redirected back from PayPal.  If we now have "Confirm" as the last step, it will already be blue on the confirm page, and even though the progress bar is not filled completely, it might *look* like you're done on first sight – which is exactly what we're trying to avoid.

I'm not really sure about this either.

turegjorup:
> Feedback: I don't like the "done" step, and I don't think it adds anything valuable.

I can follow you thinking but I disagree. While it might seem redundant I believe the last step should be the page that doesn't require an action. Otherwise there will be users who fail to confirm their order thinking they are done. 

Have looked around the web to see if there are any clear recommendations but it seems people are doing it both ways with regard to done/last step. A couple of examples for instipration:
* https://uxplanet.org/progress-trackers-in-ux-design-4319cef1c600
* https://semantic-ui.com/elements/step.html

However, I have two other suggestions for improvements:
* Change the display so that the line between the circles doesn't extend beyond (to the right of) the last circle. To clearly indicate when to the user when they are done.
* Change "Done" to "Get tickets" or just "Tickets". This will make it seem less redundant and clearly signal that you need to go to this step. However I don't know if there people using pretix to sell items without there being an admission ticket. If so, this might not work.

rixx:
I see the point you both are making, and I'm the first to admit that my judgement on UX issues is not the finest, so I'll bow out here.

Regarding the "Done" phrasing, I must disagree, though – not only is pretix versatile enough to be used without admission tickets – it is also possible to finish a booking and only receive the actual ticket weeks or months later, so that a "Tickets" point would be actively misleading. pretix could, of course, try to figure out if there will be a ticket waiting for download, but in my eyes, this would be needlessly complex and (due to plugins, addons, etc) error-prone.

raphaelm:
> Change the display so that the line between the circles doesn't extend beyond (to the right of) the last circle. To clearly indicate when to the user when they are done.

Good point. I removed the line to the left of the first circle as well for symmetry and while I'm not completely satisfied with how it looks now (since it does not extend to the edges of the pages any more and it is not easy to fix that with CSS keeping the coloured line), I think it is better than before.

> Regarding the "Done" phrasing, I must disagree, though – not only is pretix versatile enough to be used without admission tickets – it is also possible to finish a booking and only receive the actual ticket weeks or months later, so that a "Tickets" point would be actively misleading. pretix could, of course, try to figure out if there will be a ticket waiting for download, but in my eyes, this would be needlessly complex and (due to plugins, addons, etc) error-prone.

Unfortunately, @rixx got a good point there. I'm also not completely happy with the word "Done" but "Tickets" might be worse.

raphaelm:
Hmm, what do you think about renaming the step "Confirm order" to "Review order" and then changing the last step to "Order confirmed"?

domke:
> what do you think about renaming the step "Confirm order" to "Review order" and then changing the last step to "Order confirmed"?

I like it. This would better indicate that the order is not yet confirmed in the "Confirm order" phase.

turegjorup:
> ...what do you think about renaming the step "Confirm order" to "Review order" and then changing the last step to "Order confirmed"?

I like "Review order" and "Order confirmed". Think they better indicate when action is required on part of the user. As an alternative I had considered "Receipt" as the label for the final last step but I think this is just as good if not better.

> ...and while I'm not completely satisfied with how it looks now (since it does not extend to the edges of the pages any more and it is not easy to fix that with CSS keeping the coloured line), I think it is better than before.

I think this will work for now. However you could look at the examples at [semantic-ui.com](https://semantic-ui.com/elements/step.html) I posted above. Some of them might work better edge-to-edge.

raphaelm:
> I think this will work for now. However you could look at the examples at semantic-ui.com I posted above. Some of them might work better edge-to-edge.

I like them, but this approach takes way too much space which is problematic both on mobile and with an unclear, possibly higher number of steps (due to plugins adding steps)

PR #682, opened 2017-11-13
Use rel="noopener" with target="_blank".

rixx:
Required reading: https://mathiasbynens.github.io/rel-noopener/


raphaelm:
I'm a bit disappointed that your branch name contained no emoji, but I'll merge it nevertheless. Thanks!

PR #685, opened 2017-11-19
fix "typo" Referrer-Options -> Referrer-Options

ari-s:
this header was probably meant - at least it's the one that's actually used in the wild.


raphaelm:
Thanks! :)

PR #688, opened 2017-11-22
Implement automatic spell-check for docs

### LINKED ISSUE #663: Question: Add automatic spell checking?
The last three commits I made were born by running aspell on the .po-files. I quite enjoyed this and would maybe like to work on this some more - If the changes are merged, I'll definitely run the same "tests" on the informal german translation.
Apart from this, would it be an idea worth pursuing to add automatic spell checking to the project?

koebi:
fixes #663

The only thing unclear to me so far is how the output of `make spelling`
should be checked - is there a possibility to check for a file (i.e.
`_build/spelling/output.txt`) to be empty, and report a failed build
otherwise?


raphaelm:
> The only thing unclear to me so far is how the output of make spelling
> should be checked - is there a possibility to check for a file (i.e.
> _build/spelling/output.txt) to be empty, and report a failed build
> otherwise?

If the spellcheck command returns a non-zero exit code, travis should do that automatically.

Regarding the current build issues → see our jabber conversation :)

raphaelm:
> If the spellcheck command returns a non-zero exit code, travis should do that automatically.

Apparently, it doesn't :( Django itself performs a `[ -s _build/spelling/output.txt ]` command to do return non-zero if the file is non-empty, this would indeed be the simplest option.

doc/spelling_wordlist.txt
```diff
@@ -0,0 +1,118 @@
...
+backend
+backends
+banktransfer
+boolean
+booleans
+cancelled
+casted
+checkbox
+checkin
+checkins
```
raphaelm: We could discuss (not in the scope of this PR) if we write ``checkin`` or ``check-in`` consistently, I think it is inconsistent right now. Do you happen to have an opinion on this?

koebi: As of now, I don't. I'll have a look into how inconsistenly this is currently used, and report back, maybe also with an opinion ;)

PR #690, opened 2017-11-23
Remove middleware deprecation mixin

rixx:


raphaelm:
Hi! :) Sorry to say this, but I don't think inline-vendoring the mixin in multiple places makes it more readable or maintainable. Instead, we should rewrite all middlewares to new-style. I created an issue for this (#691) in case you or someone else wants to take this on, for all or even just for one or two of them. :)

PR #701, opened 2017-12-07
Pin bleach version due to mozilla/bleach#337

rixx:


raphaelm:
Thanks a lot! Although bleach already seems to be out, so we better pin belach>=2.1.2?
https://github.com/mozilla/bleach/releases/tag/v2.1.2

rixx:
The bleach people are quick, this is not necessary any more.

PR #711, opened 2017-12-12
standardize use of check-in

koebi:
The main problem here is, that for readability, variables used in code should be spelled without dash, e.g. the variable `checkin_count` should not be spelled `check-in_count`. When documenting these variables, pyenchant seems to tokenize them on the `_`, and the `checkin`-part gets spell-checked and marked as wrong. 

To overcome this, we implement a custom filter, accepting words that start with `checkin_`. The problem here is, that then the `count`-part of `checkin_count` doesn't get spell-checked. 

I'm currently looking into solving this, but my inability to non-yakshave makes this hard ;)

### DIFF
--- a/doc/api/resources/checkinlists.rst
+++ b/doc/api/resources/checkinlists.rst
@@ -213,7 +213,7 @@ Endpoints
 
 .. http:delete:: /api/v1/organizers/(organizer)/events/(event)/checkinlist/(id)/
 
-   Delete a check-in list. Note that this also deletes the information on all checkins performed via this list.
+   Delete a check-in list. Note that this also deletes the information on all check-ins performed via this list.
 
    **Example request**:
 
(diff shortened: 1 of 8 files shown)


koebi:
The latest commit fixes the aforementioned problem.
What remains, is, that in [the api documentation](https://github.com/pretix/pretix/blob/master/doc/api/resources/orders.rst), "checkins" might occurr both as the name of a variable, which shouldn't be changed, and a misspelling of "check-ins", in which case sphinxcontrib-spelling should notify us.

Therefore, it is not suitable to add "checkins" to the spelling-directive for this file. Instead, a way to mark single words as correctly spelled should be found (which I haven't been able to do upto now). I opened an [issue in sphinxcontrib-spelling](https://github.com/sphinx-contrib/spelling/issues/9) for this, let's see how it turns out :)

koebi:
…I have no clue why I closed that…

koebi:
I (hope that I) have fixed the building issue for now. As for the imperfect solution of the spelling of "checkins" here, I will transfer this to an issue, so this PR is mergeable from my side :)

raphaelm:
Thanks! :)

PR #715, opened 2017-12-13
[WIP] Fix #707 -- Extend spell-checking to .po files

### LINKED ISSUE #707: Expand spell-check to translation files
In [this commit](https://github.com/pretix/pretix/commit/9ab39904e851cbb228eaa53960ce330803b4f0ef), automatic spell-checking for the documentation is implemented. This spell-checking should now be expanded to the translation files in the [locale-directory](https://github.com/pretix/pretix/tree/master/src/pretix/locale).

So far, I have not found a convenient way to simply pass the .po-files to the spinx-contrib-spellchecker. 
One could now either look to modify its source code to be able to do this, or look into other means of spell-checking, e.g. wrapping aspell or maybe [pospelcheck](https://github.com/holygeek/pospelchek) or something else :)

koebi:
This is a first proof of concept for a script spell-checking pretix'
translation files automatically.

It is a wrapper around the polib-library handling .po-files and the
pyenchant-lib for spell-checking.


koebi:
While this is nowhere near finished, I'd kindof like a review on the current progress, since my python is a bit rusty and the code is probably rather ugly right now, and I could use some help on beautification ;) Maybe @cherti could provide that?

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,84 @@
...
+    print('Build directory already exists.')
+
+for c in checks:
+    for entry in c.po:
+        if entry.obsolete:
+            continue
+
+        en_ckr.set_text(entry.msgid)
+        for err in en_ckr:
+            print("ERROR:", c.popath, ":", entry.linenum, ":", err.word)
```
cherti: Although this perfectly works, `"ERROR: {}: {}: {}".format(c.popath, entry.lineenum, err.word)` might be better readable for this number of strings printed, better indicating the structure of the message and its changable parts.

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,84 @@
...
+            output.write("ERROR:" + c.popath + ":" + str(entry.linenum) + ":" + err.word + "\n")
+
+
+        c.checker.set_text(entry.msgstr)
+        for err in c.checker:
+            print("ERROR:", c.popath, ":", entry.linenum, ":", err.word)
```
cherti: See above

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,84 @@
...
+            c.output.write("ERROR:" + c.popath + ":" + str(entry.linenum) + ":" + err.word + "\n")
```
cherti: Same here, `"ERROR: {}: {}: {}\n".format(…)`. In this case the format option also has the advantage of taking care of converting the arguments to string, which saves you the explicit cast to `str` for `entry.linenum`.

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,84 @@
...
+
+print("Spell-checking done. You can find the outputs in \"_build/<lang>_output\".")
```
cherti: Use single quotes `'` for the outer quotes and you can drop the escaping of the double quotes inside of the string. Python recognizes the different types of quotes and as a result does not consider inner quotes to be string delimiters for the Python-level.

    print('Hello, this is a demo of "double quotes"')

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,84 @@
...
+        if f.endswith(".po"):
+            checks.append(Check(os.path.join(root, f)))
+
+en_dict = ec.DictWithPWL("en_US", pwl='./ignore_en.txt')
+en_ckr = SpellChecker(en_dict, chunkers=[HTMLChunker], filters=[PythonFormatFilter])
+output = open('_build/en_output.txt', 'w')
+
+try:
+    print('Creating build directory…')
+    os.mkdir('_build')
```
cherti: You might want to move `_build` into a variable `build_dir` so that in case it gets used more than once it does not need refactoring for changing the name.

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,84 @@
...
+except FileExistsError:
+    print('Build directory already exists.')
+
+for c in checks:
+    for entry in c.po:
+        if entry.obsolete:
+            continue
```
cherti: I personally like to abbreviate these things as `if entry.obsolete: continue` in one line to keep things compact, but this is certainly something to be cross-checked with the pretix-style guilde (which in doubt is @raphaelm) and just a minor nitpick. Code itself is totally fine.

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,84 @@
...
+
+        en_ckr.set_text(entry.msgid)
+        for err in en_ckr:
+            print("ERROR:", c.popath, ":", entry.linenum, ":", err.word)
+            output.write("ERROR:" + c.popath + ":" + str(entry.linenum) + ":" + err.word + "\n")
+
+
```
cherti: Double empty line, is this intentional? (There are uses for double empty lines, but this seems accidential to me.)

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,84 @@
...
+        self.get_ignorefile(lang)
+        check_dict = ec.DictWithPWL(checklang, pwl=self.ignore)
+        self.checker = SpellChecker(check_dict, chunkers=[HTMLChunker], filters=[PythonFormatFilter])
+        self.output = open('_build/' + lang + '_output.txt', 'w')
+
+    def get_ignorefile(self, lang):
+        for f in ignores:
+            if "ignore" in f and lang in f:
+                self.ignore = f
+
```
cherti: I think some styleguides for Python suggest closing classes with two empty lines to make the Class-close more distinct as Python does not have an explicit closing bracket. Seems reasonable enough to me to follow that.

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,84 @@
...
+        checklang = languages[lang]
+        self.get_ignorefile(lang)
+        check_dict = ec.DictWithPWL(checklang, pwl=self.ignore)
+        self.checker = SpellChecker(check_dict, chunkers=[HTMLChunker], filters=[PythonFormatFilter])
+        self.output = open('_build/' + lang + '_output.txt', 'w')
+
+    def get_ignorefile(self, lang):
+        for f in ignores:
+            if "ignore" in f and lang in f:
+                self.ignore = f
```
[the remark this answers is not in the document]
koebi: There should only be one ignore-file for each language.

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,98 @@
...
+}
+
+# These wordlists in the IGNORES_DIR only need to contain uncapitalized words,
+# the corresponding capitalized words are automatically recognized.
+LOCALES_DIR = os.path.abspath('../../pretix/locale')
+IGNORES_DIR = os.getcwd()
+BUILD_DIR = os.path.abspath('./_build')
+
+try:
+    print('Creating build directory at' + BUILD_DIR + '…')
```
cherti: for proper insertion of whitespace repace `+` by `,`, otherwise you end up printing something like "…at/foo/bar/builddir".
The `…` can be ommitted, I' say, shortening the print statement.

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,98 @@
...
+    os.mkdir(BUILD_DIR)
+except FileExistsError:
+    print("File or directory" + BUILD_DIR + "already exists, deleting…")
+    rmtree(BUILD_DIR)
```
cherti: This should still be able to fail due to the build-directory not being deletable (because it is owned by another user, for example). One could take care of that exception, or just assume that this case never happens...
However, if you do this in a filesystem where you are only allowed to add, but not to remove, this might crash nevertheless…
Depends on how many special cases you want to cover.

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,98 @@
...
+# the corresponding capitalized words are automatically recognized.
+LOCALES_DIR = os.path.abspath('../../pretix/locale')
+IGNORES_DIR = os.getcwd()
+BUILD_DIR = os.path.abspath('./_build')
+
+try:
+    print('Creating build directory at' + BUILD_DIR + '…')
+    os.mkdir(BUILD_DIR)
+except FileExistsError:
+    print("File or directory" + BUILD_DIR + "already exists, deleting…")
```
cherti: See above, `s/+/,`

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,98 @@
...
+        if name in files:
+            self.output_file = open(os.path.join(BUILD_DIR, name), 'a')
+        else:
+            self.output_file = open(os.path.join(BUILD_DIR, name), 'w')
+
+    def get_ignorefile(self, lang):
+        for f in os.listdir(IGNORES_DIR):
+            if lang in f:
+                self.ignore = f
+                return
```
cherti: `return  # as there should not be more language files`

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,98 @@
...
+checks = []
+
+for root, dirs, files in os.walk(LOCALES_DIR):
+    for f in files:
+        if f.endswith(".po"):
+            checks.append(Check(os.path.join(root, f)))
+
+en_dict = ec.DictWithPWL("en_US", pwl='./ignore_en.txt')
+en_ckr = SpellChecker(en_dict, chunkers=[HTMLChunker], filters=[PythonFormatFilter])
+output_file = open('_build/en_output.txt', 'w')
```
cherti: `BUILD_DIR` missing

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,98 @@
...
+            print("ERROR: {}:{}: {}".format(path, entry.linenum, err.word))
+            output_file.write("ERROR: {}:{}: {}\n".format(path, entry.linenum, err.word))
+
+        c.checker.set_text(entry.msgstr)
+        for err in c.checker:
+            path = os.path.relpath(c.popath, start=LOCALES_DIR)
+            print("ERROR: {}:{}: {}".format(path, entry.linenum, err.word))
+            c.output_file.write("ERROR: {}:{}: {}\n".format(path, entry.linenum, err.word))
+
+print("Spell-checking done. You can find the outputs in '_build/<lang>_output'.")
```
cherti: `BUILD_DIR`

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,98 @@
...
+            self.output_file = open(os.path.join(BUILD_DIR, name), 'w')
+
+    def get_ignorefile(self, lang):
+        for f in os.listdir(IGNORES_DIR):
+            if lang in f:
+                self.ignore = f
+                return
+
+
+# checks contains tuples of po-files and corresponding checkers
```
cherti: Aehm, according to 5 lines down, this is not correct, is it?

src/tests/locale/test_spelling.py
koebi: I personally don't think that we need to cover too many special cases, I'd cover them if they appear. For the moment, checking for an existing directory should be enough.
However, I'll leave this to @raphaelm for decision ;)

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,125 @@
...
+output_file = open(BUILD_DIR + '/en_output.txt', 'w')
+
+for c in checks:
+    for entry in c.po:
+        if entry.obsolete: continue
+
+        en_ckr.set_text(entry.msgid)
+        for err in en_ckr:
+            path = os.path.relpath(c.popath, start=LOCALES_DIR)
+            print("ERROR: {}:{}: {}".format(path, entry.linenum, err.word))
```
cherti: import sys
    print("Because it's an error message, it should go to stderr instead of stdout", file=sys.stderr)

src/tests/locale/test_spelling.py
```diff
@@ -0,0 +1,125 @@
...
+            output_file.write("ERROR: {}:{}: {}\n".format(path, entry.linenum, err.word))
+
+        c.checker.set_text(entry.msgstr)
+        for err in c.checker:
+            path = os.path.relpath(c.popath, start=LOCALES_DIR)
+            print("ERROR: {}:{}: {}".format(path, entry.linenum, err.word))
```
cherti: And how about a function

def errmsg(path, linenum, word):
    return "ERROR: {}: {}: {}".format(path, linenum, word)

and then just

    print(errmsg(path, entry.linenum, err.word)

?
This would have the advantage to have the errormessage consistent in one place at the beginning, having it a) consistent over the entire text and more importantly b) better adustable without larger refactoring.

raphaelm:
Hi @koebi! I totally lost count on the progress here after 34c3, sorry for that. Are we still going for the separate package? Let me know if there is anything I can help with.

koebi:
Hey, 
yes, we are :) Currently, I'm kinda blocked timewise by my last two exams, but after Thursday, I'm free to do whatever I want again :) It should move forward from there on, I'll keep you updated and ask if I stumble upon problems :)

koebi:
Hey,
I think I have the spellchecking working now. I'll spend some time fixing all the mistakes it has found, and then open another pull request and close this one. Expect it within a week :))

koebi:
The new pull request is #896 and this one will be closed.

PR #723, opened 2017-12-19
remove alert since mail-typo-checking is disabled

koebi:
[This commit](https://github.com/pretix/pretix/commit/a62fbd54d4a1eba5cf243a449a20ac4610d478c4) removes the typo-checking for e-mail-adresses. Therefore, this box isn't needed either. Just running the shop from the current master leads to 
pretix_typo_check_alert

This might be because of my inability to work with pretix correctly, but I think this is the right thing to do ;)


raphaelm:
Oops, thanks!

Issue #730, opened 2017-12-31
Add Date, Time and Datetime Question Types


felixrindt:
I'd like to add time related question types to ask for things like birthday, time of arrival or some other event related point in time.

felixrindt:
PR #732 solved this.

PR #732, opened 2017-12-31
Fix #732 -- Add date and time question types

felixrindt:
This adds Date, Time and Datetime question types for #730 
I already implemented parts on the python side. Questions get displayed and validated.

screenshot from 2017-12-31 12-52-54

I'm still missing the nice bootstrap date/time pickers from pretix/control, but I couldn't figure out how to add them yet. I'd love some hints on how to do that.

Also, where are the answers validate? Since the DateTimePicker produces two outputs, it needs special handling for validation and answer display in pretix/control.


src/pretix/base/models/orders.py
```diff
@@ -499,7 +499,7 @@ def __str__(self):
         elif self.question.type == Question.TYPE_FILE:
             return str(_("<file>"))
         else:
-            return self.answer
+            return str(self.answer)
```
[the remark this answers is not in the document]
felixrindt: I wasn't really sure about that. The docstring says it's a string, but it's defined as `answer = models.TextField()`. That confused me, but I read up on how Field classes work. The `TextField` does always deliver a string, so this really doesn't do anything.

felixrindt:
This is the current status:
progress

Date and time work on their own in the presale. All date and time fields are showing the correct bootstrap pickers with correct formatting. There are two problems I'm noticing at the moment with the splitdatetime.

As you can see, the split datetime is not in a row, because the parent div is missing the `splitdatetimerow` class. In pretix/control, this class is applied in the template:
`{% bootstrap_field form.date_from layout="control" horizontal_field_class="col-md-9 splitdatetimerow" %}`. In `pretix/presale/templates/pretixpresale/event/checkout_quesitons.html` the forms are rendered in a loop with the `{% bootstrap_form %}` tag. I don't really know how to apply the class to only the splitdatetimeview widget. It would probably need to happen somewhere in the template rendering python code?

The other thing is: even if the question is marked as 'not required', I always get the validation error as seen in the screenshot, no matter if there was valid input or not.

src/pretix/base/models/orders.py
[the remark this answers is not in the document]
felixrindt: Reverted in de38710

src/pretix/presale/forms/checkout.py
```diff
@@ -299,6 +300,27 @@ def __init__(self, *args, **kwargs):
...
+                )
+            elif q.type == Question.TYPE_TIME:
+                field = forms.TimeField(
+                    label=q.question, required=q.required,
+                    help_text=q.help_text,
+                    initial=initial.answer if initial else None,
+                    widget=TimePickerWidget(),
+                )
+            elif q.type == Question.TYPE_DATETIME:
+                field = forms.DateTimeField(
```
raphaelm: > The other thing is: even if the question is marked as 'not required', I always get the validation error as seen in the screenshot, no matter if there was valid input or not.

You need to use ``forms.SplitDateTimeField`` here :)

src/pretix/presale/forms/checkout.py
```diff
@@ -299,6 +300,27 @@ def __init__(self, *args, **kwargs):
                     initial=initial.file if initial else None,
                     widget=UploadedFileWidget(position=pos, event=event, answer=initial)
                 )
+            elif q.type == Question.TYPE_DATE:
+                field = forms.DateField(
+                    label=q.question, required=q.required,
+                    help_text=q.help_text,
+                    initial=initial.answer if initial else None,
```
raphaelm: ``initial=initial.answer if initial else None`` probably needs to be ``initial=dateutil.parser.parse(initial.answer).astimezone(tz).date() if initial and initial.answer else None`` with ``tz = pytz.timezone(event.settings.timezone)`` further above.

src/pretix/presale/forms/checkout.py
```diff
@@ -299,6 +300,27 @@ def __init__(self, *args, **kwargs):
...
+                    widget=DatePickerWidget(),
+                )
+            elif q.type == Question.TYPE_TIME:
+                field = forms.TimeField(
+                    label=q.question, required=q.required,
+                    help_text=q.help_text,
+                    initial=initial.answer if initial else None,
```
raphaelm: ``initial=initial.answer if initial else None`` probably needs to be ``initial=dateutil.parser.parse(initial.answer).astimezone(tz).time() if initial and initial.answer else None``

src/pretix/presale/forms/checkout.py
```diff
@@ -299,6 +300,27 @@ def __init__(self, *args, **kwargs):
...
+                    widget=TimePickerWidget(),
+                )
+            elif q.type == Question.TYPE_DATETIME:
+                field = forms.DateTimeField(
+                    label=q.question, required=q.required,
+                    help_text=q.help_text,
+                    initial=initial.answer if initial else None,
```
raphaelm: ``initial=initial.answer if initial else None`` probably needs to be ``initial=dateutil.parser.parse(initial.answer).astimezone(tz) if initial and initial.answer else None``

src/pretix/control/forms/__init__.py
```diff
@@ -131,3 +130,37 @@ def __init__(self, attrs=None, date_format=None, time_format=None):
...
+        attrs = attrs or {}
+        if 'placeholder' in attrs:
+            del attrs['placeholder']
+        date_attrs = dict(attrs)
+        date_attrs.setdefault('class', 'form-control')
+        date_attrs['class'] += ' datepickerfield'
+
+        df = date_format or get_format('DATE_INPUT_FORMATS')[0]
+        date_attrs['placeholder'] = now().replace(
+            year=2000, month=1, day=1, hour=0, minute=0, second=0, microsecond=0
```
felixrindt: What about using `month=12 ,day=31` in these placeholders to avoid confusion between month and day?

raphaelm: > What about using month=12 ,day=31 in these placeholders to avoid confusion between month and day?

Absolutely, great idea!

felixrindt: Maybe we could also clearly show that it's 24hrs (no am/pm), but I think also using `23:59:59` would be visually distracting. Using something like `18:00:00` seems a bit arbitrary. What do you think?

raphaelm: I think 18 is fine, the year 2000 is just as arbitrary ;)

felixrindt:
I applied your patch and fixed up the questions. I didn't notice any errors yet. 

I did not apply timezone handling to date questions, as changing the timezone would change the date the user put in and that's not what we want.

The only thing bugging me is that the input format includes seconds and the `TIME_FORMAT` and `SHORT_DATETIME_FORMAT` don't include seconds. I think including the seconds in a format would make the most sense. What's the way to go here? Is there already a format with seconds?

raphaelm:
> I did not apply timezone handling to date questions, as changing the timezone would change the date the user put in and that's not what we want.

Agreed. It would be probably great to have some unit tests around this time format and timezone handling because it is too easy to get subtly wrong to test it manually.

> The only thing bugging me is that the input format includes seconds while TIME_FORMAT and SHORT_DATETIME_FORMAT don't include seconds. I think including the seconds in a format would make the most sense. What's the way to go here? Is there already a format with seconds?

Good question. I do not have a good idea right now. I'm not sure seconds are really relevant here (dropping them would work for me), but we should either consistently have them or not have them, you're right there. We could roll our own formats for that, but that makes it harder to add new languages. Do you see an elegant way to drop the seconds from the input formats?

felixrindt:
There you go! Since we don't want to add a new format, we can only use the existing ones which don't have seconds. Since the input formats are in a list, we needed a way to access the other elements of that list, since the default element at position 0 always included seconds.

felixrindt:
I ran through the codestyle checks and added a test. Also, GitHub said there was a merge conflict with another test being added at the same position. Let me know if there's anything else to do. 😄

src/pretix/base/models/items.py
```diff
@@ -544,6 +544,9 @@ class Question(LoggedModel):
     * a boolean (``TYPE_BOOLEAN``)
     * a multiple choice option (``TYPE_CHOICE`` and ``TYPE_CHOICE_MULTIPLE``)
     * a file upload (``TYPE_FILE``))
+    * a date (``TYPE_DATE``))
```
raphaelm: duplicate )) here and below :)

doc/api/resources/questions.rst
```diff
@@ -23,6 +23,9 @@ type                                  string                     The expected ty
                                                                  * ``C`` – choice from a list
                                                                  * ``M`` – multiple choice from a list
                                                                  * ``F`` – file upload
+                                                                 * ``D`` – date
+                                                                 * ``H`` – time
+                                                                 * ``W`` – date and time
```
raphaelm: Would you mind adding the following below the table?

src/pretix/presale/forms/checkout.py
```diff
@@ -295,7 +301,28 @@ def __init__(self, *args, **kwargs):
...
+                    help_text=q.help_text,
+                    initial=dateutil.parser.parse(initial.answer).astimezone(tz).time() if initial and initial.answer else None,
+                    widget=TimePickerWidget(time_format=get_format('TIME_INPUT_FORMATS')[2]),
+                )
+            elif q.type == Question.TYPE_DATETIME:
+                field = forms.SplitDateTimeField(
+                    label=q.question, required=q.required,
+                    help_text=q.help_text,
+                    initial=dateutil.parser.parse(initial.answer).astimezone(tz) if initial and initial.answer else None,
+                    widget=SplitDateTimePickerWidget(time_format=get_format('TIME_INPUT_FORMATS')[2]),
```
raphaelm: I've gone through all of Django's supported locales and this will break if we ever add:
* Esperanto
* Hungarian

felixrindt: Ok. sorry for not thinking about that. Will read up on that.

src/pretix/base/models/orders.py
```diff
@@ -498,6 +500,21 @@ def __str__(self):
...
+                d = d.astimezone(tz)
+            return date_format(d, "SHORT_DATETIME_FORMAT")
+        elif self.question.type == Question.TYPE_DATE:
+            d = dateutil.parser.parse(self.answer)
+            return date_format(d, "SHORT_DATE_FORMAT")
+        elif self.question.type == Question.TYPE_TIME:
+            d = dateutil.parser.parse(self.answer)
+            if self.orderposition:
+                tz = pytz.timezone(self.orderposition.order.event.settings.timezone)
+                d = d.astimezone(tz)
```
raphaelm: Currently, this fails for me with the exception ``ValueError: astimezone() cannot be applied to a naive datetime`` as the TYPE_TIME values are, in contrast to the TYPE_DATETIME values, saved to the database without timezone information (just as ``18:30``, not ``18:30+00:10``). Intuitively, we would either need to

* modify TimeField or the questions form in a way that they store the timezone in the ``answer`` field with a timezone
* ensure that the ``answer`` field always contains UTC (see also my comment on the test below) and then state this here as ``d = pytz.UTC.localize(d).astimezone(tz)``

However, we have a more fundamental problem at hand here: We don't know for what day the input is! Suppose we have an event that goes from 2018-10-26 to 2018-10-28 and the organizer defines two time input questions with the texts "When do you arrive?" and "When do you leave?". The event timezone might be Europe/Berlin, but in the first case, the organizer probably wants UTC+2 and in the second case, UTC+1, since daylight saving time switches in the night of 2018-10-28. We can't ever now and asking the organizer is probably too much trouble.

So sorry for the confusion, I think we need to take all the timezone handling out again for TYPE_TIME and only keep it for TYPE_DATETIME (where it actually makes a lot of sense).

felixrindt: I agree that this is a problem. How are all the datetimes defined in pretix/control represented in the database? Also, I'm wondering what timezone is used with a datetime in the one hour that is repeated when daylight savings time switches in October.

src/tests/presale/test_checkout.py
```diff
@@ -73,6 +74,52 @@ def test_empty_cart(self):
...
+        o1 = QuestionAnswer.objects.get(question=q1)
+        o2 = QuestionAnswer.objects.get(question=q2)
+        o3 = QuestionAnswer.objects.get(question=q3)
+        order = Order.objects.create(event=self.event, status=Order.STATUS_PAID,
+                                     expires=now() + timedelta(days=3),
+                                     total=4)
+        op = OrderPosition.objects.create(order=order, item=self.ticket, price=42)
+        o1.cartposition, o2.cartposition, o3.cartposition = None, None, None
+        o1.orderposition, o2.orderposition, o3.orderposition = op, op, op
+        # timezone differences can differ, we only test that the time has changed
```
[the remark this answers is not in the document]
felixrindt: Regarding what you commented above we can indeed check for an exact timezone difference with datetime answers, because we know the date. Since we are dropping timezones from time answers, that other time assertion will removed.

src/pretix/presale/forms/checkout.py
felixrindt: I created a `get_format_without_seconds` function in `pretix/control/utils/i18n.py` (and also one for javascript formats) for that and will probably add a test. Maybe this should be moved to `pretix/base`, since it's in `pretix/presale` now?

raphaelm: Yes, you can move it, if you want, it makes sense.

src/pretix/base/models/orders.py
raphaelm: **datetimes** are represented always as UTC in the database. On some database (I think postgres) this is explicitly stored with +00:00, on others (mysql) it is just assumed that it is always UTC.
For datetimes, that is no issue.

For **times** without a date, we don't have that up to now, and as I suggested I'd say we store them without a timezone and don't handle them at all and just output them as they are inputted by the user.

felixrindt:
Ambigous or non-existent times are actually already handled by the forms.

screenshot from 2018-01-08 18-38-31

src/pretix/base/models/orders.py
felixrindt: > **datetimes** are represented always as UTC in the database.

I submitted a datetime question answer to a test event with Europe/Berlin as timezone via a local `./manage.py` sqlite instance. `str(QuestionAnswer.answer)` was `2018-10-14 14:00:00+02:00`, which is not UTC. I'm confused about that.

felixrindt:
Do I need to worry about the 2 failed tests? 

> 2 failed
>         - tests/base/test_mail.py:38 test_send_mail_with_event_sender
>         - tests/base/test_mail.py:63 test_send_mail_with_default_sender

(`Dummy <pretix@localhost>` != `pretix@localhost`)

raphaelm:
> Do I need to worry about the 2 failed tests?

No, that is my fault, I broke them yesterday, sorry for that. I'm travelling until friday, so it'll take a few days until I can look at this in detail.

felixrindt:
There was no formatting on the dates in the control question overview. 

screenshot-2018-1-14 question when will you arrive pretix de

I fixed it, but I'm sure there's a more elegant solution using django's or python's inbuilt abstraction functions to get the answer strings via a `str(QuestionAnswer)` call.

screenshot-2018-1-14 question when will you arrive pretix de

raphaelm:
I gave it another round of testing and the result looks pretty good to me, including your most recent change. Do you agree that this is merge-ready or is there anything open from your side?

felixrindt:
I'm glad you like it. I also didn't notice any other problems. I'd consider this merge-ready then. 😃

raphaelm:
Merged, thanks! \*high-five\*

felixrindt:
Thanks a lot for your support. I really had a lot of fun working on this and learned many different things.

PR #733, opened 2017-12-31
Fix #731 -- Display negative payment method fees correctly

### LINKED ISSUE #731: Payment fee formatting in payment selection form
When the user is in checkout and selects a payment method, if a payment fee is negative, e.g. -4.00, it is dislayed with a prefixed + sign. It should just be a minus. 

Also I think that the '+ 0.00 EUR' looks very bulky and numbery. Maybe the font for these fees could be a little smaller?

screenshot from 2017-12-31 12-10-19

felixrindt:
This was the nicest version I could think of. 😄 

This is how it looks like:
screenshot from 2017-12-31 19-50-30

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/checkout_payment.html
+++ b/src/pretix/presale/templates/pretixpresale/event/checkout_payment.html
@@ -12,7 +12,7 @@
                         <div class="panel-heading">
                             <h4 class="panel-title">
                                 {% if show_fees %}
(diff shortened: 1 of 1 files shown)


raphaelm:
Thanks! ✨

felixrindt:
Glad someone else is doing this tonight 💥 🎉

raphaelm:
<3 Happy new year! 🎆

PR #736, opened 2018-01-04
Update development documentation, update demo URL

m0llux:
When using the test data generation script, the demo conference no longer has the URL ``/bigevents/2018``, but is now moved to ``/bigevents/2019`` . This confused me in the beginning.


raphaelm:
Perfect, thansk! :)

PR #744, opened 2018-01-15
Fix #674 -- Assigning bank transactions with a dash in the event slug

### LINKED ISSUE #674: Order code splitted at wrong character during bank transfer assignment
Just ran into an issue while manually resolving an unresolved transaction. When I enter the order code of the transaction (just the 5 characters) into the input field, a dropdown opens up to reveal the order. This includes the entire string with the name of the event, the full transfer description that should have been used by the payer. When I click it, this entire description is filled into the input field. When I click confirm however, pretix does not recognize this string. Instead it just wants the 5 characters.

So either clicking the recommendation in the dropdown should just insert the 5 characters instead of the entire transfer string, or the input field should accept both. I guess the second option would be the nicer one.

Sorry I didn't take any screenshots, by the time I had solved my issue, it was already too late ;)

JindalMohit:


raphaelm:
Thank you very much! :)

PR #752, opened 2018-01-18
Fix #751 -- calculate payment fees in OrderChangeManager

### LINKED ISSUE #751: order_free email send with non-free payment provider
When only buying free products, pretix skips the payment checkout step and uses the `FreePaymentProvider` (identifier `free`). It then sends the email `event.settings.mail_text_order_free`. So far so good.

What if someone (read "I") is using a payment provider with `-100%` payment fee?  
(e.g. if one is a member of a club, they get in for free)  
When the customer has a cart with value > 0, you can select a payment provider. If that payment provider reduces your order total to 0, inconsistencies show:

- On the order confirmation page, the customer sees the text defined by the chosen payment provider
- The email template used is `event.settings.mail_text_order_free`, stating something else than the potential payment provider wants the user to know (e.g. "please confirm their club membership")

What I would expect, is that the email is also of type `event.settings.mail_text_order_placed` in which the payment provider can insert information, just like every normal payment mail. 

To do that, I think only this line should be changed to check the payment method, not order total:
https://github.com/pretix/pretix/blob/d622f38e1da6a98935c2c8720685fdee9cde85f1/src/pretix/base/services/orders.py#L516

Also, are there any problems with negative product prices and order totals?

felixrindt:
Is there a way to make this 'more robust'?

### DIFF
--- a/src/pretix/base/services/orders.py
+++ b/src/pretix/base/services/orders.py
@@ -824,7 +824,7 @@ class OrderChangeManager:
             raise OrderError(self.error_messages['paid_price_change'])
 
     def _check_paid_to_free(self):
-        if self.order.total == 0:
+        if self.order.total == 0 and self._totaldiff < 0:
             try:
                 mark_order_paid(
                     self.order, 'free', send_mail=False, count_waitinglist=False,
@@ -1015,6 +1015,14 @@ class OrderChangeManager:
         self.order.total += sum([f.value for f in self.order.fees.all()])
         self.order.save()
 
+    def _payment_fee_diff(self):
+        prov = self._get_payment_provider()
+        if prov:
+            old_fee = prov.calculate_fee(self.order.total)
+            new_total = self.order.total + self._totaldiff
+            if new_total != 0:
+                self._totaldiff += prov.calculate_fee(new_total) - old_fee
+
     def _reissue_invoice(self):
         i = self.order.invoices.filter(is_cancellation=False).last()
         if i and self._invoice_dirty:
@@ -1065,6 +1073,7 @@ class OrderChangeManager:
         if not self._operations:
             # Do nothing
             return
+        self._payment_fee_diff()
         with transaction.atomic():
             with self.order.event.lock():
                 if self.order.status not in (Order.STATUS_PENDING, Order.STATUS_PAID):
--- a/src/pretix/base/services/orders.py
+++ b/src/pretix/base/services/orders.py
@@ -513,7 +513,7 @@ def _perform_order(event: str, payment_provider: str, position_ids: List[str],
             invoice = generate_invoice(order, trigger_pdf=not event.settings.invoice_email_attachment)
             # send_mail will trigger PDF generation later
 
-    if order.total == Decimal('0.00'):
+    if order.payment_provider == 'free':
         email_template = event.settings.mail_text_order_free
         log_entry = 'pretix.event.order.email.order_free'
     else:
--- a/src/pretix/base/services/orders.py
+++ b/src/pretix/base/services/orders.py
@@ -825,8 +825,10 @@ class OrderChangeManager:
             raise OrderError(self.error_messages['paid_price_change'])
 
     def _check_paid_to_free(self):
-        if self.order.total == 0 and (self._totaldiff < 0 or self.split_order and self.split_order.total > 0):
+        if self.order.total == 0 and (self._totaldiff < 0 or (self.split_order and self.split_order.total > 0)):
             # if the order becomes free, mark it paid using the 'free' provider
+            # this could happen if positions have been made cheaper or removed (_totaldiff < 0)
+            # or positions got split off to a new order (split_order with positive total)
             try:
                 mark_order_paid(
                     self.order, 'free', send_mail=False, count_waitinglist=False,
--- a/src/pretix/base/services/orders.py
+++ b/src/pretix/base/services/orders.py
@@ -1018,10 +1018,12 @@ class OrderChangeManager:
     def _payment_fee_diff(self):
         prov = self._get_payment_provider()
         if prov:
-            old_fee = prov.calculate_fee(self.order.total)
-            new_total = self.order.total + self._totaldiff
+            old_total = sum([p.price for p in self.order.positions.all()])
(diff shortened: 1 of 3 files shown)


felixrindt:
Closing this for now, since I'm not really happy with what I was going to do with it. (see #751)

raphaelm:
Huh, wow, I was just convinced that this is actually a good idea by your e-mail. I'll reply to your e-mail later or tomorrow.

felixrindt:
Thanks! I just sent you another mail. Generally, this change might be a good idea, but I think there's more to be discussed here 🤔

felixrindt:
I noticed while doing tests with `py.test -n 4` using `pytest-xdist` as described in the docs, 2 tests were failing which run alone or without `-n 4` didn't fail.

src/pretix/base/services/orders.py
```diff
@@ -824,7 +824,7 @@ def _check_paid_price_change(self):
             raise OrderError(self.error_messages['paid_price_change'])
 
     def _check_paid_to_free(self):
-        if self.order.total == 0:
+        if self.order.total == 0 and self._totaldiff < 0:
```
felixrindt: It the order was free before too, don't do anything.

src/pretix/base/services/orders.py
```diff
@@ -1065,6 +1073,7 @@ def commit(self):
         if not self._operations:
             # Do nothing
             return
+        self._payment_fee_diff()
```
felixrindt: This happens before the checks to incorporate a change in payment fees into `self._totaldiff`. I did not touch `self._recalculate_total_and_payment_fee()`, because it expects the change to be already applied to the order, since it is after `self._perform_operations()`.

src/pretix/base/services/orders.py
```diff
@@ -1015,6 +1015,14 @@ def _recalculate_total_and_payment_fee(self):
         self.order.total += sum([f.value for f in self.order.fees.all()])
         self.order.save()
 
+    def _payment_fee_diff(self):
+        prov = self._get_payment_provider()
+        if prov:
+            old_fee = prov.calculate_fee(self.order.total)
+            new_total = self.order.total + self._totaldiff
+            if new_total != 0:
```
felixrindt: This if statement is also found here:
https://github.com/pretix/pretix/blob/fc7bafe3d98b499f9ae2b2fe4f17e79552b62e4c/src/pretix/base/services/orders.py#L1002
I think this should have the same behaviour, although I don't see a reason for why this check exists.

src/pretix/base/services/orders.py
```diff
@@ -1015,6 +1015,16 @@ def _recalculate_total_and_payment_fee(self):
         self.order.total += sum([f.value for f in self.order.fees.all()])
         self.order.save()
 
+    def _payment_fee_diff(self):
+        prov = self._get_payment_provider()
+        if prov:
+            old_total = sum([p.price for p in self.order.positions.all()])
+            new_total = old_total + self._totaldiff
+            old_fee = self.order.fees.get(fee_type=OrderFee.FEE_TYPE_PAYMENT).value
```
felixrindt: We need to get the fee from the model, as some settings could have changed with the payment provider or the fees could depend on some economical factor.

felixrindt:
I'm having a test with tax rules fail. Will look for the cause later 🛏

felixrindt:
When running the tests, this line

https://github.com/pretix/pretix/blob/56ca2305bdcefae044b4c0d0678f90bc000cb75d/src/tests/base/test_orders.py#L985

fails, because `self._check_paid_price_change()` raises `OrderError(self.error_messages['paid_price_change'])`.

I'm not sure, but I think the test uses the order change manager in faulty way. In the test, there are two calls to `ocm.commit()`, meaning that between commits, variables like `_totaldiff` don't get resetted. https://github.com/pretix/pretix/blob/56ca2305bdcefae044b4c0d0678f90bc000cb75d/src/pretix/base/services/orders.py#L683
Simply creating a new OrderChangeManager after the first commit resolves this. Am I thinking correct?

felixrindt:
There is a print statement in the test. Is that intended?

https://github.com/pretix/pretix/blob/56ca2305bdcefae044b4c0d0678f90bc000cb75d/src/tests/base/test_orders.py#L914

raphaelm:
> There is a print statement in the test. Is that intended?

No, print statements in pretix are never intended.

src/pretix/base/services/orders.py
```diff
@@ -1015,6 +1015,22 @@ def _recalculate_total_and_payment_fee(self):
         self.order.total += sum([f.value for f in self.order.fees.all()])
         self.order.save()
 
+    def _payment_fee_diff(self):
+        prov = self._get_payment_provider()
+        if self.order.status != Order.STATUS_PAID and prov:
+            #payment fees of paid orders do not change
+            try:
+                old_fee = self.order.fees.get(fee_type=OrderFee.FEE_TYPE_PAYMENT).value
+            except OrderFee.MultipleObjectsReturned:
```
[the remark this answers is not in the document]
felixrindt: Haha ok that's much better :+1:

felixrindt:
> it wasn't a problem until now.

There were actually 7 places where the ChangeManager is used twice. I'm surprised that it never came to attention before. 

> print statements in pretix are never intended.

I just searched on github and most print statements are found in tests, but I found two in the stripe plugin which looked like they should not be there.

src/pretix/base/services/orders.py
```diff
@@ -824,7 +825,8 @@ def _check_paid_price_change(self):
             raise OrderError(self.error_messages['paid_price_change'])
 
     def _check_paid_to_free(self):
-        if self.order.total == 0:
+        if self.order.total == 0 and (self._totaldiff < 0 or self.split_order and self.split_order.total > 0):
```
raphaelm: 1. Do you know the precedence rules of ``and`` and ``or`` without looking them up? I don't, so please add brackets here to make clear what you mean ;)
2. I just had to think about the logic of that line for at least 10 minutes. I got it now and I think it is correct, but would you mind adding an explanatory comment for the next one reading it? 😉

raphaelm:
Thanks! :)

PR #753, opened 2018-01-18
Fix #735 -- Display of event series on public organizer page

### LINKED ISSUE #735: listing of event with subevents on organizer page irritating
When creating an event with subevents you need to provide a `date_from` date. Since there are multiple dates, this can be used for the timespan the workshops take place in. For example, an art club does 4 painting workshops every February, it could set the event date from Feb 1st to Feb 28th and have the individual workshops as subevents with the actual start times. 

On the organizer page, events can be publicly listed. The issue is, that if you look at that page after the event `date_from` date, this event does not get listed as a future event, even though there might be subevents still in the future you want to sell tickets for. 

Similarly, this page does not list the event in "past events" when loading the page before `date_to`, even though there might be subevents in the past.

In this screenshot, the conference does not have subevents, but the workshop has multiple subevents in February.

screenshot from 2018-01-03 18-08-13

To fix this issue, the conditions when to show events with subevents in the future and past events lists need to be changed. Also there should be a way for a customer to see that there are multiple dates available.

JindalMohit:


src/pretix/presale/views/organizer.py
```diff
@@ -99,14 +99,19 @@ def get(self, request, *args, **kwargs):
...
+            )
+            qs = qs.annotate(
+                min_to=Min('subevents__date_to')
+            ).filter(min_to__lte=now())
             order = '-date_from'
         else:
             query &= Q(Q(date_from__gte=now()) | Q(date_to__gte=now()))
             order = 'date_from'
-        qs = Event.objects.filter(
-            Q(organizer=self.request.organizer) & query
```
felixrindt: The organizer condition is duplicated and could be moved up to the initial definition of `query`.

src/pretix/presale/views/organizer.py
```diff
@@ -99,14 +99,19 @@ def get(self, request, *args, **kwargs):
     def get_queryset(self):
         query = Q(is_public=True) & Q(live=True)
         if "old" in self.request.GET:
-            query &= Q(Q(date_from__lte=now()) & Q(date_to__lte=now()))
+            qs = Event.objects.filter(
+                Q(organizer=self.request.organizer) & query
+            )
+            qs = qs.annotate(
+                min_to=Min('subevents__date_to')
+            ).filter(min_to__lte=now())
```
felixrindt: You filter out normal events from the past here.

felixrindt:
I think not using the `date_from` and `date_too` attributes would be better suited for subevents in the upcoming list aswell, as mentioned at #735. In this case, you'd need to add a `max_to` annotation similar to the one you already added.

Also, what can be done about the date range annotation? This is not that trivial. Here's my idea:

In the case of upcoming events
- The range start could show the `date_from` of the first subevent that has not ended yet.
- The range end could show the latest `date_to` date. 

Simlarly for old events
- The range start could show the oldest `date_from` 
- The range end could show latest `date_to` from subevents that have ended.

This way, ongoing subevents are accounted for in the upcoming list, like it is done for normal events. In all cases, ordering should happen using the ranges start dates. 

I also imagine some hint to the customer that this entry actually represents multiple events and not one long one. What do you think?

src/pretix/presale/views/organizer.py
felixrindt: You can run `py.test tests/presale/test_organizer_page.py` to test your changes. You should also add new tests for the behavior you are working on.

JindalMohit:
I have updated query set for upcoming list.
For events with sub-events we can do something like:
screenshot from 2018-01-20 17-20-10
And I think we should keep the date range as same as it will give customer the idea for the duration of the event, how much event has passed and how much event is remaining. What do you think ?

felixrindt:
> For events with sub-events we can do something like

I don't like how that looks, but possibly we don't even need something like this, as the event name is often in a pluralized form (like workshop**s**), hinting that there are multiple subevents. 

> And I think we should keep the date range as same as it will give customer the idea for the duration of the event, how much event has passed and how much event is remaining. What do you think ?

At the moment, the `date_from` and `date_to` attributes are shown, which don't hava a well defined meaning for subevents as far as I know. Not changing this would enable the user to define what is shown (which is good), but as you add or remove workshops or time moves on, the dates need to be manually changed. I don't think these quite fundamental attributes should be changed only to have a date range displayed nicely, as they more represent the timeframe in which the subevent group exists. The date column should be about the time the (sub)events actually take place.

If the intent is to show the range of upcoming (sub)events (which is the view's headline, so I think it is), the approach I outlined in my last comment would be better. In my opinion, this is the most useful, as people looking for events to attend shouldn't really care about events in the past, right? In other words, subevents that lie in the past shouldn't really play a role in the date range of *upcoming* events.

A third option would be to show the overall minimal `date_from` and maximum `date_to`, but I think that's the least useful.

src/pretix/presale/views/organizer.py
```diff
@@ -97,16 +97,24 @@ def get(self, request, *args, **kwargs):
...
     def get_queryset(self):
-        query = Q(is_public=True) & Q(live=True)
+        query = Q(is_public=True) & Q(live=True) & Q(organizer=self.request.organizer)
+        qs = Event.objects.filter(query)
         if "old" in self.request.GET:
-            query &= Q(Q(date_from__lte=now()) & Q(date_to__lte=now()))
+            qs = qs.annotate(
+                min_to=Min('subevents__date_to')
+            ).filter(
+                Q(Q(has_subevents=False) & Q(date_from__lte=now()) & Q(date_to__lte=now())) |
```
felixrindt: I like this, but I think we can work with the assumption, that `date_from` is never after `date_to`, meaning the check `date_from__lte+now()` could be removed, as `date_to` has to also be <= `now()`.

src/pretix/presale/views/organizer.py
```diff
@@ -97,16 +97,24 @@ def get(self, request, *args, **kwargs):
...
+                Q(Q(has_subevents=True) & Q(min_to__lte=now()))
+            )
             order = '-date_from'
         else:
-            query &= Q(Q(date_from__gte=now()) | Q(date_to__gte=now()))
+            qs = qs.annotate(
+                max_to=Max('subevents__date_to')
+            ).filter(
+                Q(Q(has_subevents=False) & Q(Q(date_from__gte=now()) | Q(date_to__gte=now()))) |
```
felixrindt: Similar to above, the `date_from__gte=now()` check isn't really needed, as `date_to` should be `gte` anyway.

JindalMohit:
Okay I will change the date range display as you suggested earlier.
Can you tell me where to make changes?
According to me, the HTML code for organizer page calls the function:
[https://github.com/pretix/pretix/blob/master/src/pretix/base/models/event.py#L101](url)
for date range display. Changing directly this fuction will have effect on some other places also where we do not want the change.

felixrindt:
I did not take a deep look, but I guess you could overwrite the function in `Event` (checking if it has subevents) or you could create a seperate function just for the in public list, because there are two different usages of the function at

https://github.com/pretix/pretix/blob/56ca2305bdcefae044b4c0d0678f90bc000cb75d/src/pretix/base/models/event.py#L101

By the way: Clicking on the source code links in your comments leads to a 404 for me.

JindalMohit:
The organizer page after the changes(workshop is an event-series):

screenshot from 2018-01-21 16-40-10

screenshot from 2018-01-21 16-40-24

JindalMohit:
@raphaelm I have made the changes suggested by you to handle the case where date_to is NULL.

Regarding displaying of date range, as Felix Rindt suggested in earlier comments, I have calculated date_range_from and date_range_to to take care of sub-events.

Felix Rindt suggestions:

screenshot from 2018-01-22 01-21-31

src/pretix/base/models/event.py
```diff
@@ -109,6 +109,17 @@ def get_date_range_display(self, tz=None) -> str:
...
+    def get_date_range_display_for_organizer_page(self, tz=None) -> str:
+        """
+        Returns a formatted string containing the start date and the end date
+        of the event with respect to the current locale and to the ``show_times`` and
+        ``show_date_to`` settings taking consideration of sub-events.
+        """
+        if not self.has_subevents:
+            return self.get_date_range_display(tz)
+        tz = tz or pytz.timezone(self.settings.timezone)
+        return daterange(self.date_range_from.astimezone(tz), self.date_range_to.astimezone(tz))
```
felixrindt: Where is `data_range_from` and `date_range_to` defined?

JindalMohit: They are defined in the query set in this file : src/pretix/presale/views/organizer.py

raphaelm: Yes, it'd be great if you either document that here with a comment, or find a different way to do this than with a model method as this is very specific to one single view.

raphaelm:
> Regarding displaying of date range, as Felix Rindt suggested in earlier comments, I have calculated date_range_from and date_range_to to take care of sub-events.

Sorry, I missed Felix' comment, but I think I disagree, sorry 😉  I'm in favour of always showing the complete range of the event series, it makes for a simpler query and is less confusing if you're looking for something. Also, your ``date_range_to`` will be NULL if the date_to values are not set. Better use and ``max_fromto`` (which solves the NULL problem) here :)
(Also, ``min_from`` == ``date_range_from``, we do not need both.)

JindalMohit:
@raphaelm, @felixrindt  Thanks for your support. 
So finally, we want date range to be complete date range of an event-series in both upcoming and past event list ? Please tell me what to do finally.
Thanks.

felixrindt:
>  I'm in favour of always showing the complete range of the event series.

To be clear, do you mean the `date_from` and `date_to` dates configured or `min_from` and `max_fromto`?

raphaelm:
> So finally, we want date range to be complete date range of an event-series in both upcoming and past event list ? Please tell me what to do finally.

Yes, I think that would be the most consistent way to display it. Like in the list of the backend, we could display in some way that it is an event series, but that doesn't need to be a part of this PR.

> To be clear, do you mean the date_from and date_to dates configured or min_from and max_fromto?

min_from and max_fromto. date_from/date_to of a series is basically pointless and only there because it would be effort to get rid of the fields ;)

JindalMohit:
I have made the changes.

src/pretix/presale/views/organizer.py
```diff
@@ -4,7 +4,8 @@
 
 import pytz
 from django.conf import settings
-from django.db.models import Exists, OuterRef, Q
+from django.db.models import Exists, OuterRef, Q, Min, Max, Case, When
```
felixrindt: `Case` and `When` imports are not used. Please remove them.

felixrindt: Also, flake8 is complaining about some indentation further down.

JindalMohit:
I have corrected the errors. All the checks are also passed. Following is the view of organiser page for the following event series:
dates
past
upcoming

PR #756, opened 2018-01-19
Refs #654 -- API: Writable event endpoints

turegjorup:
Starting to implement write operations for the event endpoint. Very much the early stages, but a few questions have come up already.

Event object structure:
For plugins I have opted for the following implementation where the Boolean value is true for enabled plugins
```
  "plugins": {
    "pretix.plugins.banktransfer": true,
    "pretix.plugins.stripe": false,
    "pretix.plugins.paypal": false,
    "pretix.plugins.ticketoutputpdf": true,
    "pretix.plugins.sendmail": false,
    "pretix.plugins.statistics": true,
    "pretix.plugins.reports": true,
    "pretix.plugins.pretixdroid": true
  }
```
I also considered having to keys, a `plugin_enabled` and a `plugin_disabled` key but I'm thinking this will be easier for an API-consumer to work with. However, any feedback on this is welcome.

I'm planning to follow a similar structure for the various sections and tabs in the interface. I.e. implement a `display_settings` key with the relevant options. Again, feedback and alternative suggestions are welcome. 

Plugin settings:
This also brings me to the question of how to handle the different options the plugins make available. I have read the section on "REST API viewsets" under Plugin Development in the documentation. And I can see that the bank transfer plugin adds an endpoint to the `orga_router` but no other plugins seems to implement anything in regard to the API. Do you have any thoughts on what the convention should be in regard to URL structure? I'm thinking something along the lines of recommending that `/api/v1/organizers/(organizer)/events/(event)/(pluginname)/settings/` is always implemented and matches the settings that are available in the interface.


src/pretix/base/models/event.py
```diff
@@ -544,6 +546,59 @@ def get_users_with_permission(self, permission):
             Q(is_superuser=True) | Q(twp=True)
         )
 
+    @staticmethod
+    def has_paid_things(event):
```
raphaelm: Is there any reason this method is static? ``event.has_paid_things()`` looks cleaner to me than ``Event.has_paid_things(event)``.

src/pretix/base/models/event.py
```diff
@@ -544,6 +546,59 @@ def get_users_with_permission(self, permission):
...
+        item = apps.get_model('pretixbase', 'Item')
```
raphaelm: You can just use a local import, i.e. ``from . import Item`` inside this method, you just can't have it in the file header to avoid the circular import. Using ``apps.get_model`` breaks tab completion in IDEs etc.

src/pretix/base/models/event.py
```diff
@@ -544,6 +546,59 @@ def get_users_with_permission(self, permission):
...
+        item_variation = apps.get_model('pretixbase', 'ItemVariation')
+        return item.objects.filter(event=event, default_price__gt=0).exists()\
+            or item_variation.objects.filter(item__event=event, default_price__gt=0).exists()
+
+    @staticmethod
+    def has_payment_provider(event):
```
raphaelm: Same here, I see no reason to have this as a static method, this could be a ``@property`` just as well.

src/pretix/base/models/event.py
```diff
@@ -544,6 +546,59 @@ def get_users_with_permission(self, permission):
...
+        result = False
+        for provider in event.get_payment_providers().values():
+            if provider.is_enabled and provider.identifier != 'free':
+                result = True
+                break
+        return result
+
+    @staticmethod
```
raphaelm: Same here, can be a normal method.

src/pretix/base/models/event.py
```diff
@@ -544,6 +546,59 @@ def get_users_with_permission(self, permission):
...
+    def clean_payment_methods(event):
+        if Event.has_paid_things(event) and not Event.has_payment_provider(event):
+            raise ValidationError(_('You have configured at least one paid product but have not enabled any payment '
+                                    'methods.'))
+
+    @staticmethod
```
raphaelm: Same here, can be a normal method.

src/pretix/base/models/event.py
```diff
@@ -544,6 +546,59 @@ def get_users_with_permission(self, permission):
...
+        if event is None or not event.quotas.exists():
+                raise ValidationError(_('You need to configure at least one quota to sell anything.'))
+
+    @staticmethod
+    def clean_live(event):
+        Event.clean_payment_methods(event)
+        Event.clean_quotas(event)
+
+    @staticmethod
+    def clean_slug(organizer, event, slug):
```
raphaelm: Just to make the distinction clear: This and the following two methods are rightfully static methods, since they do not depend on an ``Event`` object at all and can be used on serializer or form level even before the ``Event`` object is constructed!

turegjorup:
Back from other projects 😃 

Have changed the format for plugins, and fixed static methods, etc.

As for your plan on doing this in 3 steps, that sounds like the way to go. I haven't looked in detail at the how the settings store is implemented, so still unsure of the scope of the task. But I like the idea of limiting the scope of each PR. So, let's focus this PR as you describe in 1.

This leaves the issue of the "copy settings+products from" feature. While I understand the convenience of this feature it's also one of the things I find problematic, both in the current UI but also in a RESTful API. 

To clarify I don't think it scales very well in the UI as it's currently implemented. As the number of events grows the list of events that you can copy from becomes quickly becomes too big to be useful. I would have preferred the use of "event templates", so that you could define named templates that define settings + products (or alternatively you could promote events to template status). This would allow for a more useful UI, and potentially make the API design easier because you could reference the template. I also understand that it would take one or more totally separate PR's to get this done so it's not realistic in the scope of this PR(s)

As for the API design, the problem I see is how to best implement an "action" for a RESTful API. I've looked around for suggestions on how to design this. But haven't found any solution I really like.

Working with the current "copy settings+products from" feature I see a couple of different ways forward on this:

1. Finish this PR without the copy feature. This would limit the usefulness of the API until PR's "2" and "3" are complete. But it would allow us to bypass the question of how to integrate this in the API.
2. Implement this with a query parameter, e.g. `POST /api/v1/organizers/bigevents/events/?copy-from-event=big-event` This keeps the action separate from the event resource but using url parameters in a post request seems counter intuitive. 
3. Introduce a virtual property on events like `{"settings-plugin-template": "big-event"}`. This would work but I'm not really a fan given that it's not a proper field and would only be relevant for POST requests. 

Of the tree I prefer the 2. But feedback and other ideas are welcome 🤔

raphaelm:
> To clarify I don't think it scales very well in the UI as it's currently implemented. As the number of events grows the list of events that you can copy from becomes quickly becomes too big to be useful.

Agreed, I recently noticed that too. The simple solution for this is to replace the group of radio boxes by an autocomplete-equipped text field as we started to do in basically all other places where you can select a (sub)event.

> I would have preferred the use of "event templates", so that you could define named templates that define settings + products (or alternatively you could promote events to template status). This would allow for a more useful UI, and potentially make the API design easier because you could reference the template. I also understand that it would take one or more totally separate PR's to get this done so it's not realistic in the scope of this PR(s)

Interesting idea. I've spent some thought on another thing event templates could be, namely presets ("Concert", "Theater", "Conference") to make it easier for new pretix users to start.

I see one problem with your idea: The "copy" feature is useful for two groups of users:

* People who run e.g. an annual conference (like myself). Every year, I just start with copying last year to save time. It would be uncomfortable to create a template first, and then if things change, change both my current event and the template, to have the change next year.

* Power users who run hundreds of events. Templates would make perfectly sense here. The power users I know already do exactly this, they just have one or more events called "TEMPLATE EVENT". The only problem here is discoverability in the list, as you said, that would be solved by both a "this is a template" flag and an autocompleted form field. I'm not sure which one creates more/less confusion here.

> Finish this PR without the copy feature. This would limit the usefulness of the API until PR's "2" and "3" are complete. But it would allow us to bypass the question of how to integrate this in the API.

Would be okay for me. I'm not committed on having this in the API at all, for now, I don't really care. My thinking was that with this feature in the API as well as an event that you use as a template event, the API might already be useful for your use case with this PR already, as you can already create useful events via the API.

> Of the tree I prefer the 2. But feedback and other ideas are welcome

I agree that 3 is not nice. I think the most RESTful approach would be to have a separate endpoint for this. I can imagine two designs from the top of my head:

a)

1. ``POST /api/v1/organizers/bigevents/events/``
2. ``POST /api/v1/organizers/bigevents/events/newevent/copy`` with the ID of the old event as an argument

This looks clean to me, but the downside is that the second endpoint needs to either ensure that the event has not yet been filled with things or handle conflicts, which doesn't sound good.

b)

To create an event with copying, you don't do a ``POST /api/v1/organizers/bigevents/events/``, but instead a ``POST /api/v1/organizers/bigevents/events/oldevent/clone`` with the new event resource in the body.

I think I like (b) most. But I'd also be fine with not implementing it or another design.

turegjorup:
> I think I like (b) most. But I'd also be fine with not implementing it or another design.

Have implemented (b) 😃 

This means this PR is almost complete. One issue that remains is test coverage for permissions. The way the `event_permission_urls` are structured in `test_permissions.py `means that I can't just follow the same template for testing event permissions given that the tests are built to test event "sub" url like `categories` and ìtems`.

I'm unsure if the best approach is to change the tests and `event_permission_urls` and test methods to be able to test on events. That is change all `resp = getattr(token_client, urlset[0])('/api/v1/organizers/{}/events/{}/{}'.format(…` so that it's not hardcoded to sub-resources. Or alternatively to setup new tests for events.

I'm guessing both can work, but any input is appreciated 🤔

turegjorup:
So I gave the permissions issue a shot. Not sure if it's the best option but DRF's auth/permission model is still a learning exercise for me. Depending on your feedback, I think this is ready to be merged 😃

raphaelm:
Thank you very much! Parts of this are tough to review but I'll try to get around in the next days.

src/pretix/api/serializers/event.py
```diff
@@ -15,14 +17,73 @@ def to_representation(self, value):
...
+        return data
+
+    def validate_slug(self, value):
+        Event.clean_slug(self.context['request'].organizer, self.instance, value)
+        return value
+
+    def validate_live(self, value):
+        if value:
+            if self.instance is None:
+                raise ValidationError(_('Events cannot be create as \'live\'. Quotas and payment must be added to the '
```
raphaelm: create → created

src/pretix/api/serializers/event.py
```diff
@@ -15,14 +17,73 @@ def to_representation(self, value):
...
+        full_data = self.to_internal_value(self.to_representation(self.instance)) if self.instance else {}
+        full_data.update(data)
+
+        Event.clean_dates(data.get('date_from'), data.get('date_to'))
+        Event.clean_presale(data.get('presale_start'), data.get('presale_end'))
+
+        return data
+
+    def validate_slug(self, value):
+        Event.clean_slug(self.context['request'].organizer, self.instance, value)
```
raphaelm: We also need to make sure that the slug cannot be changed once it has been set.

src/pretix/api/views/event.py
```diff
@@ -10,23 +13,89 @@
...
+    write_permission = 'can_create_events'
+
+    def perform_create(self, serializer):
+        serializer.save(organizer=self.request.organizer)
+        plugins = serializer.instance.plugins
+
+        event = Event.objects.filter(slug=self.kwargs['event'], organizer=self.request.organizer.pk).first()
+        serializer.instance.copy_data_from(event)
+
+        serializer.instance.plugins = plugins
```
raphaelm: Hmm, do you think there would be a good idea to allow to specify that one wants to copy this as well? I'm not sure about a nice design here.

src/pretix/api/urls.py
```diff
@@ -14,6 +14,7 @@
 
 event_router = routers.DefaultRouter()
 event_router.register(r'subevents', event.SubEventViewSet)
+event_router.register(r'clone', event.CloneEventViewSet)
```
raphaelm: Just for the record, if I had implemented this, I'd probably used a [detail_route](http://www.django-rest-framework.org/api-guide/viewsets/#marking-extra-actions-for-routing) on the existing viewset instead of a new viewset, but I don't see any downsides to your approach, so I guess we keep it.

src/pretix/api/views/event.py
```diff
@@ -10,23 +13,89 @@
...
+                instance.organizer.log_action(
+                    'pretix.event.deleted', user=self.request.user,
+                    data={
+                        'event_id': instance.pk,
+                        'name': str(instance.name),
+                        'logentries': list(instance.logentry_set.values_list('pk', flat=True))
+                    }
+                )
+                instance.items.all().delete()
+                instance.subevents.all().delete()
```
raphaelm: It is very likely that these two lines need to be extended by more lines in the future, so let's better create a model method instead of having it duplicated between pretix.control and here.

src/pretix/api/auth/permission.py
```diff
@@ -59,6 +59,22 @@ def has_permission(self, request, view):
...
 
+class EventCRUDPermission(EventPermission):
+
+    def has_permission(self, request, view):
+        if not super(EventCRUDPermission, self).has_permission(request, view):
+            return False
+        elif view.action == 'create' and 'can_create_events' not in request.orgapermset:
+            return False
+        elif view.action == 'destroy' and 'can_create_events' not in request.eventpermset:
+            return False
```
raphaelm: Currently, this requires ``can_change_event_settings`` permission, not ``can_create_event`` in the backend. I agree that neither choice is perfect, but we should be consistent :)

doc/api/resources/events.rst
```diff
@@ -27,12 +27,20 @@ location                              multi-lingual string       The event locat
...
+Note that events with orders cannot be deleted to ensure data integrity.
+
 .. versionchanged:: 1.7
 
    The ``meta_data`` field has been added.
 
+.. versionchanged:: 1.14
+
+   The ``plugins`` field has been added.
+   The write operations POST, PATCH, AND PUT have been added.
```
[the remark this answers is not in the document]
turegjorup: ...and I could have sworn that i put it in the documentation :-/

doc/api/resources/events.rst
```diff
@@ -118,11 +132,239 @@ Endpoints
...
+   :statuscode 201: no error
+   :statuscode 400: The event could not be created due to invalid submitted data.
    :statuscode 401: Authentication failure
-   :statuscode 403: The requested organizer/event does not exist **or** you have no permission to view it.
+   :statuscode 403: The requested organizer does not exist **or** you have no permission to create this resource.
+
+
+.. http:post:: /api/v1/organizers/(organizer)/events/(event)/clone/
+
+   Creates a new event with properties as set in the request body and settings and items copied from the exiting event.
```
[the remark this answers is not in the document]
turegjorup: Been giving this some thought. No good ideas unfortunately. As long as we nest most endpoints under `organizer` I think it will be close to impossible to come up with a good design.

src/pretix/api/serializers/event.py
```diff
@@ -15,14 +17,73 @@ def to_representation(self, value):
...
+
+        plugins = {plugin_name for plugin_name in data}
+
+        return {
+            'plugins': ",".join(plugins)
+        }
+
+
 class EventSerializer(I18nAwareModelSerializer):
-    meta_data = MetaDataField(source='*')
```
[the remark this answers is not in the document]
turegjorup: My bad, removed it initially to avoid having to implement `MetaDataField.to_internal_value()`.

src/pretix/api/serializers/event.py
turegjorup: Implemented in `Event.clean_slug` and covered by `test_events.py:L275`

src/pretix/api/views/event.py
turegjorup: I tried to consider what should determine this. The post body or the events that is cloned. I chose let the post body decide because I consider this the most explicit request. In the documentation I deliberately only wrote that "settings and items" are copied.

src/pretix/api/urls.py
turegjorup: Just gave the detail_route a quick test. There are some side effects regarding validation and responses that made it a bit tricky. So leaving as is for now.

src/pretix/api/views/event.py
```diff
@@ -10,23 +13,89 @@
...
+
 
 class SubEventFilter(FilterSet):
     class Meta:
         model = SubEvent
         fields = ['active']
 
 
-class SubEventViewSet(viewsets.ReadOnlyModelViewSet):
+class SubEventViewSet(viewsets.ModelViewSet):
```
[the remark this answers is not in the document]
turegjorup: Yes. My bad. Fixed.

src/pretix/api/views/event.py
turegjorup: Done. Maybe not the best name, ideas are welcome 😄

src/pretix/api/auth/permission.py
turegjorup: Changed to     
```
permission = 'can_change_event_settings'
write_permission = 'can_change_event_settings'
```

src/pretix/base/models/event.py
```diff
@@ -557,9 +573,43 @@ def get_users_with_permission(self, permission):
...
+            raise ValidationError(_('You have configured at least one paid product but have not enabled any payment '
+                                    'methods.'))
+
+    def clean_quotas(self):
+        if not self.quotas.exists():
+            raise ValidationError(_('You need to configure at least one quota to sell anything.'))
+
+    def clean_live(self):
+        self.clean_payment_methods()
+        self.clean_quotas()
```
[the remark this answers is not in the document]
turegjorup: I have tried to address this, but unsure if it's the best approach. To follow the principle of handling validation in the model I have moved the 'issues' property from the `EventLive`view to the `Event`model and renamed it to `live_issues` to give it context.

src/pretix/api/serializers/event.py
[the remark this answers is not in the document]
turegjorup: I tried to implement `to_internal_value()` but couldn't figure out a clean way of doing it. When I returned a dict it couldn't be applied as value. Any idea as to how to get around this?

turegjorup:
Jumped the gun on the completeness, so have re-added [WIP] label. I have tried to address the various issues, but would like your feedback on the `live_issues` and `meta_data` parts especially.

Also both Mikkel and I are suddenly seeing a lot of failing tests. This happens on both master and on my feature branch. I can see on Travis that the tests pass on your commits, so we are  unsure what the issue is:
```
Results (840.13s):
    1963 passed
       1 xpassed
      36 failed
         - tests/base/test_models.py:134 QuotaTestCase.test_reserved
         - tests/base/test_models.py:388 QuotaTestCase.test_subevent_isolation
         - tests/base/test_models.py:185 QuotaTestCase.test_unlimited
         - tests/base/test_models.py:198 QuotaTestCase.test_voucher_product
         - tests/base/test_models.py:224 QuotaTestCase.test_voucher_quota
         - tests/base/test_models.py:278 QuotaTestCase.test_voucher_quota_expired
         - tests/base/test_models.py:270 QuotaTestCase.test_voucher_quota_expiring_soon
         - tests/base/test_models.py:237 QuotaTestCase.test_voucher_quota_multiuse
         - tests/base/test_models.py:211 QuotaTestCase.test_voucher_variation
         - tests/base/test_models.py:326 QuotaTestCase.test_waitinglist_item_active
         - tests/base/test_models.py:336 QuotaTestCase.test_waitinglist_variation_active
         - tests/base/test_models.py:346 QuotaTestCase.test_waitinglist_variation_fulfilled
         - tests/presale/test_checkout.py:1143 CheckoutTestCase.test_confirm_expired_partial
         - tests/presale/test_checkout.py:1114 CheckoutTestCase.test_subevent_confirm_expired_partial
         - tests/presale/test_event.py:71 ItemDisplayTest.test_link_rewrite
         - tests/presale/test_event.py:212 ItemDisplayTest.test_subevent_net_prices
         - tests/presale/test_event.py:193 ItemDisplayTest.test_subevent_prices
         - tests/presale/test_event.py:179 ItemDisplayTest.test_subevents
         - tests/presale/test_event.py:99 ItemDisplayTest.test_timely_available
         - tests/presale/test_event.py:91 ItemDisplayTest.test_without_category
         - tests/presale/test_event.py:328 VoucherRedeemItemDisplayTest.test_hide_without_voucher_item
         - tests/presale/test_event.py:310 VoucherRedeemItemDisplayTest.test_in_quota
         - tests/presale/test_event.py:368 VoucherRedeemItemDisplayTest.test_sold_out_blocking
         - tests/presale/test_event.py:376 VoucherRedeemItemDisplayTest.test_sold_out_ignore
         - tests/presale/test_event.py:315 VoucherRedeemItemDisplayTest.test_specific_item
         - tests/presale/test_event.py:463 VoucherRedeemItemDisplayTest.test_subevent_net_prices
         - tests/presale/test_event.py:489 VoucherRedeemItemDisplayTest.test_subevent_prices
         - tests/presale/test_event.py:337 VoucherRedeemItemDisplayTest.test_variations_all
         - tests/presale/test_event.py:394 VoucherRedeemItemDisplayTest.test_variations_sold_out_blocking
         - tests/presale/test_event.py:349 VoucherRedeemItemDisplayTest.test_variations_specific
         - tests/presale/test_event.py:406 VoucherRedeemItemDisplayTest.test_voucher_price
         - tests/presale/test_event.py:414 VoucherRedeemItemDisplayTest.test_voucher_price_percentage
         - tests/presale/test_event.py:422 VoucherRedeemItemDisplayTest.test_voucher_price_variations
         - tests/presale/test_event.py:579 WaitingListTest.test_display_link
         - tests/presale/test_event.py:766 DeadlineTest.test_in_time
         - tests/presale/test_event.py:748 DeadlineTest.test_not_set
```

doc/api/resources/events.rst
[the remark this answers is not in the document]
raphaelm: Alright, let's keep it this way.

src/pretix/api/serializers/event.py
raphaelm: Thanks!

src/pretix/api/views/event.py
raphaelm: I agree that it makes sense to honor the most explicit request. "Settings" is a vague term to many users, though, so we might make this *even more* explicit in the documentation.

src/pretix/api/urls.py
raphaelm: Alright, good to know!

raphaelm:
Thank you very much!

> Jumped the gun on the completeness, so have re-added [WIP] label. I have tried to address the various issues, but would like your feedback on the live_issues and meta_data parts especially.

I'll comment on them inline.

As an additional question: Why did you remove the EventCRUDPermission thing again? It wasn't all that wrong except for the one edge case. Do I see it correctly that I can now create events without the can_create_events permission? This shouldn't be possible.

> Also both Mikkel and I are suddenly seeing a lot of failing tests. This happens on both master and on my feature branch. I can see on Travis that the tests pass on your commits, so we are unsure what the issue is:

Weird, I haven't seem them before and they appear to be unrelated to this PR. Could you maybe open a new issue with (at least one of) the stacktraces you see?

src/pretix/api/serializers/event.py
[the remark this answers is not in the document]
raphaelm: Since this makes use of many-to-many relations, it seems to be impossible to implement inside the ``Field`` class, it behaves more like the nested resources we have elsewere. I took the liberty of pushing the most elegant version I could come up with (which isn't much) to this PR, feel free to disregard or change: https://github.com/pretix/pretix/pull/756/commits/319621677bcb6e501fb0860ac58ba55d021c1709

src/pretix/api/auth/permission.py
turegjorup: And changed back again 😄

turegjorup:
Thank you for your commit on 'meta_data'. Saved me a bit of time.

> As an additional question: Why did you remove the EventCRUDPermission thing again?

I misunderstod your comment on EventCRUDPermission to mean that both `create`and `destroy` required ´can_change_event_settings` - maybe because of the way Github shows inline comments. I have now re-added `EventCRUDPermission` and change the permission needed to delete events.  

The API related tests pass but I'm still seeing the other failures. I'll just wait and see how they perform on Travis.

raphaelm:
Wonderful, thanks for your work. It looks all good to me currently,  but I'll try to find time for another in-depth review before I merge this, just to be sure we don't get the permission things wrong ;)

turegjorup:
Glad I kept the WIP label this time 😄, a few questions before I try to address this:

> **FIX:** In the UI, it is impossible to create an event that you will not have access to later. [...]. In this scenario, the UI creates a new team.

This is a though (impossible?) one. Both design wise and because the API's permission model have no concept of users.

I can understand why you do this in the UI even though I don't completely agree with the design. While you never want the user to end up confused or at a dead end in the UI, I would properly have chosen to honor the team/permissions setup for the organization. Then on create event given the user a confirmation along the lines of "The event have been created but you do not have permission to view it...". To me, silently/automatically creating a new team on event create is not expected behavior. Because of this, and in line with trusting API users to make educated decisions, I would recommend against this. 

But a whole other problem is that the permissions model for the API have no concept of users. The API keys are linked to teams, not to users. As I see it, this makes it impossible to do the same in a meaningful way in the API. Even if we did copy the behavior and created a new team, how would this benefit the API consumer? The consumer would have no way to know about the new team, and would have no access to it because it would have no API key. We _could_ auto generate a key but I can't think of either a clean design for transmitting it back to the API consumer nor a way of doing it without severe security implications.

> **UNCLEAR:** When creating an event with `has_subevents=True`, the UI automatically creates a first subevent.

This is quite similar to the problem we a had with `items`, `has_variations`, `variations` were we ended up supporting nested variations on `CREATE`, but not on `UPDATE`. For consistency we should consider a similar design here. However this would require us to add a `subevents`property to the `events` endpoint. Given that events are larger objects and that the list of sub-events are potentially longer than for variations this might not be desirable from a performance viewpoint. 

As for the other options I'm not sure which i prefer. Allowing `has_subevents=True` while having no sub-events seems inconsistent. Automatically creating one is not expected behavior and we opted against this design for `variations`. But both could properly be documented.

Feedback and thoughts greatly appreciated 😃

raphaelm:
> This is a though (impossible?) one. Both design wise and because the API's permission model have no concept of users.

Right. I totally forgot that API tokens are not users.

> I can understand why you do this in the UI even though I don't completely agree with the design. While you never want the user to end up confused or at a dead end in the UI, I would properly have chosen to honor the team/permissions setup for the organization. Then on create event given the user a confirmation along the lines of "The event have been created but you do not have permission to view it...". To me, silently/automatically creating a new team on event create is not expected behavior. Because of this, and in line with trusting API users to make educated decisions, I would recommend against this.

I see your point. I also think it is acceptable if the UI differs from the API here, as long as we make it explicit in the documentation.

> But a whole other problem is that the permissions model for the API have no concept of users. The API keys are linked to teams, not to users. As I see it, this makes it impossible to do the same in a meaningful way in the API. Even if we did copy the behavior and created a new team, how would this benefit the API consumer? The consumer would have no way to know about the new team, and would have no access to it because it would have no API key. We could auto generate a key but I can't think of either a clean design for transmitting it back to the API consumer nor a way of doing it without severe security implications.

You are absolutely correct, I totally missed the complexity here. Ignore this, I'd say :)

> This is quite similar to the problem we a had with items, has_variations, variations were we ended up supporting nested variations on CREATE, but not on UPDATE. For consistency we should consider a similar design here. However this would require us to add a subeventsproperty to the events endpoint. Given that events are larger objects and that the list of sub-events are potentially longer than for variations this might not be desirable from a performance viewpoint.
> As for the other options I'm not sure which i prefer. Allowing has_subevents=True while having no sub-events seems inconsistent. Automatically creating one is not expected behavior and we opted against this design for variations. But both could properly be documented.

Since there is no logical reason to forbid having a series with no dates, I believe I should fix the crashes in the UI instead.

src/pretix/api/views/event.py
```diff
@@ -10,15 +13,84 @@
...
+        serializer.save(organizer=self.request.organizer)
+        plugins = serializer.instance.plugins
+        is_public = serializer.instance.is_public
+
+        event = Event.objects.filter(slug=self.kwargs['event'], organizer=self.request.organizer.pk).first()
+        serializer.instance.copy_data_from(event)
+
+        if plugins:
+            serializer.instance.plugins = plugins
+        if is_public:
```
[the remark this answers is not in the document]
turegjorup: Fixed

src/pretix/base/models/event.py
```diff
@@ -556,9 +590,51 @@ def get_users_with_permission(self, permission):
...
+
+        for receiver, response in self.live_issues:
```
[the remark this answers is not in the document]
turegjorup: Correct. Fixed

turegjorup:
Have fixed some of the issues you spotted:
- I have tried to expand on the documentation regarding permissions. For event endpoint I have noted for each endpoint which permission is required for access. If you like this approach we should it expand it to the other endpoints? 
- `has_subevents` can no longer be changed.
- Added logging of live activated/deactivated

A few are still needs to be addressed:
- For `subevents=True` I'm hoping you will fix the crashes in the UI if the list is empty?
- I have looked at the plugins enable/disable issue but not started an implementation. I believe I found a bug in the current UI logic that needs to be addressed. If you have an event with paid products and no plugins activated you cannot set it to 'live' because it has no payment options. However if you have a live event with paid products you can disable all plugins without error, in effect removing all payment options. Would like your thoughts on this before looking more at this.

doc/api/fundamentals.rst
```diff
@@ -44,6 +44,25 @@ like the following:
           adding OAuth2 support in the future for user-level authentication. If you want
           to use session authentication, be sure to comply with Django's `CSRF policies`_.
 
+Permissions
+-----------
+
+The API follows Pretix team based permissions model. Each organizer can have several teams
```
raphaelm: ``Pretix`` → ``pretix'``

turegjorup:
> Enabling a plugin is more than just changing Event.plugins. We also (a) have restricted plugins that can only be enabled by sysadmins, (b) have an optional installed method on plugins that should be called (eg to initialize database tables), and (c) log this action separately. It probably makes sense to refactor this into a new Event.enable_plugin() method and call it from both the view and EventSerializer.

Refactored the way plugins are enabled and disabled to address this:

(a)
This is still only accounted for in the UI. The concept of "sysadmins" is linked to users, not to teams as far as I can tell. I think this is fine given the different permission model for the API and in line with API users being expected to make educated decisions. If we need to address this we would have to introduce the "Is site admin" or "Has site admin rights" on the team level as far as I can tell.

(b)
I have added `enable_plugin`, `disable_plugin` and `set_active_plugins` to `Event` and refactored the form view and API serializer to use these.

(c)
Logging for plugins now handled in `EventViewSet`

Depending on your feedback on (a), and a final review I think this is now feature complete 😃

raphaelm:
> This is still only accounted for in the UI. The concept of "sysadmins" is linked to users, not to teams as far as I can tell. I think this is fine given the different permission model for the API and in line with API users being expected to make educated decisions. If we need to address this we would have to introduce the "Is site admin" or "Has site admin rights" on the team level as far as I can tell.

I see your point, this really maps to the API badly. However, I want to make the tradeoff differently and pushed 0bf0dda5 to disallow enabling these plugins via the API completely. Restricted plugins are a rarely-used thing that we use when we absolutely need to roll out something on our hosted infrastructure that is specific to a single or small group of customers. Usually, we avoid such situations but when they occur, we don't want other customers to enable these plugins without talking to us.

I also took the freedom to push 13040feb, removing unused code that seems to be a copy-paste relict.

I did these two minimal changes myself because I now want to press the "Confirm squash and merge [x] wait for successful checks" button :wink: 
Thank you so much for your continued work on this tremendous effort. There are not many better feelings than merging PRs with perfect test and docs coverage, high-quality code and fruitful discussions on the way. :sparkles:

turegjorup:
> However, I want to make the tradeoff differently and pushed 0bf0dda to disallow enabling these plugins via the API completely.

I can easily accept this. Was unsure exactly what the use case was, this makes good sense. 

> I also took the freedom to push 13040fe, removing unused code that seems to be a copy-paste relict.

Ups. Thank you 😊 

> I did these two minimal changes myself because I now want to press the "Confirm squash and merge [x] wait for successful checks" button 

Please do 🚀

PR #759, opened 2018-01-24
Fix register exporter signal name in doc

felixrindt:
The signal is defined at 
https://github.com/pretix/pretix/blob/353dce789d5dfe6c0ebf6e98257999325b6958f1/src/pretix/base/signals.py#L143
and ends with an s.


raphaelm:
Thanks!

PR #761, opened 2018-01-25
Fix #740 -- Date picker: Fix line height for decade span

### LINKED ISSUE #740: rendering issue with decade selection in date picker
While trying out the datepicker in the event settings view I saw this.
screenshot from 2018-01-10 11-26-57

As you can see, the year ranges are seperated by a big vertical gap. What I expected to see was this

screenshot from 2018-01-10 11-36-37 

from https://eonasdan.github.io/bootstrap-datetimepicker/.
I noticed this with a fresh installation of the pretix master branch using firefox-dev 58.

LeonardoChirivi:
This seems to be fixing the problem.
#740

### DIFF
--- a/src/pretix/static/datetimepicker/_bootstrap-datetimepicker.scss
+++ b/src/pretix/static/datetimepicker/_bootstrap-datetimepicker.scss
@@ -300,7 +300,7 @@ $bs-datetimepicker-text-shadow: 0 -1px 0 rgba(0, 0, 0, 0.25) !default;
                 display: inline-block;
                 width: 54px;
                 height: 54px;
-                line-height: 1.8em;
+                line-height: 54px;
                 margin: 2px 1.5px;
                 cursor: pointer;
                 border-radius: $bs-datetimepicker-border-radius;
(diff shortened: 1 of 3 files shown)


raphaelm:
Hi @LeonardoChirivi, thank you very much!

This ``src/pretix/static/datetimepicker/_bootstrap-datetimepicker.scss`` file, however, is vendored from the bootstrap datetimepicker library, so I fear we'll probably accidentally override your change if we ever update that library. Would it make more sense to override this in a file of our own, whenever we include ``_bootstrap-datetimepicker.scss``, what do you think?

LeonardoChirivi:
Good day @raphaelm 
I can either make a pull request to [bootstrap-datetimepicker](https://github.com/Eonasdan/bootstrap-datetimepicker/blob/master/src/sass/_bootstrap-datetimepicker.scss) repository, and as soon they accept it we can keep use the library, or make our own css file that overrides the span.
Whatever works for you.

raphaelm:
> I can either make a pull request to bootstrap-datetimepicker repository, and as soon they accept it we can keep use the library, or make our own css file that overrides the span.

Since it doesn't look broken on https://eonasdan.github.io/bootstrap-datetimepicker/ I'm not even sure it's their fault, maybe their style just conflicts with ours somewhere?

> Whatever works for you.

Given the activity on the bootstrap-datetimepicker repo – maybe both? ;)

raphaelm:
I now moved this to another file myself. If you want, do create a PR to bootstrap-datetimepicker. Thank you! :sparkles:

PR #770, opened 2018-02-17
Use config.getboolean to get metrics enabled value

timfreund:
Given the following configuration:

[metrics]
enabled=False

Using config.get results in a METRICS_ENABLED value that always
evaluates to True. This PR switches to config.getboolean so that metrics
can be disabled without deleting the configuration values.


raphaelm:
Top, thanks! :)

PR #776, opened 2018-02-21
Refs #754 -- check item tax_rule is not none

### LINKED ISSUE #754: Error when adding item with no tax rule and custom price to order
I have an event series with a single product, for which I did not configure any tax rule. When trying to add a product to an order with a custom price, I get an Error. 

https://github.com/pretix/pretix/blob/fc7bafe3d98b499f9ae2b2fe4f17e79552b62e4c/src/pretix/base/services/orders.py#L777

`item.tax_rule` is `None`, so the function call fails.

felixrindt:
While this is a simple fix, it at least makes it possible again to add products without tax rules.

### DIFF
--- a/src/pretix/base/services/orders.py
+++ b/src/pretix/base/services/orders.py
@@ -812,7 +812,7 @@ class OrderChangeManager:
         if price is None:
             price = get_price(item, variation, subevent=subevent, invoice_address=self._invoice_address)
         else:
-            if item.tax_rule.tax_applicable(self._invoice_address):
+            if item.tax_rule and item.tax_rule.tax_applicable(self._invoice_address):
                 price = item.tax(price, base_price_is='gross')
             else:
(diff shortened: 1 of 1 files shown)


raphaelm:
I'll merge this for now, but I'm pretty sure that we'll disallow tax_rule=None very soon (and come up with a migration that creates a tax rule of 0% if required).

felixrindt:
Just for understanding completely: What are your reasons for deprecating tax_rule=None?

felixrindt:
And if you want to deprecate it "very soon" why should we merge this at all?

raphaelm:
> Just for understanding completely: What are your reasons for deprecating tax_rule=None?

Two reasons:
* There are two ways to express the same thing (no tax rule and tax rule with 0%) which creates confusion both in the code and the UI
* My observation is that nobody ever accidentally puts in 0%, but quite a lot of people accidentally set no tax rule.

Issue #757, opened 2018-01-22
TypeError when redeeming voucher at subevent overview


felixrindt:
**What I did** 🔧 
I have an event series called `testseries` under the organizer `bigevents`. Naturally, the event can be seen at `http://localhost:8000/bigevents/testseries/`. 

I have a voucher for the event with subevents `testseries` which is not tied to a specific subevent. I paste the voucher code in the field on `http://localhost:8000/bigevents/testseries/`.

**What happened** 💥 
I get a type error stating that a subevent must be provided. ([Stack Trace](http://dpaste.com/2PZW8K0))

**What I expected** 🤔 
I expected the website to report a userfriendly error that I should select a subevent before trying to redeem the voucher.

Using a voucher works fine under an URL tied to a specific subevent, e.g. `http://localhost:8000/bigevents/testseries/3/`. 👍

felixrindt:
I just saw that the service Django is uploading the stacktraces to expires them after a day, so here's the trace again:

> Environment:
> 
> Request Method: GET
Request URL: http://localhost:8000/bigevents/testseries/redeem?voucher=NZL45DWD342PZ8MH&subevent=
> 
> Django Version: 1.11
Python Version: 3.6.4
Installed Applications:
['django.contrib.auth',
 'django.contrib.contenttypes',
 'django.contrib.sessions',
 'django.contrib.messages',
 'django.contrib.staticfiles',
 'pretix.base',
 'pretix.control',
 'pretix.presale',
 'pretix.multidomain',
 'pretix.api',
 'rest_framework',
 'django_filters',
 'compressor',
 'bootstrap3',
 'djangoformsetjs',
 'pretix.plugins.banktransfer',
 'pretix.plugins.stripe',
 'pretix.plugins.paypal',
 'pretix.plugins.ticketoutputpdf',
 'pretix.plugins.sendmail',
 'pretix.plugins.statistics',
 'pretix.plugins.reports',
 'pretix.plugins.checkinlists',
 'pretix.plugins.pretixdroid',
 'easy_thumbnails',
 'django_markup',
 'django_otp',
 'django_otp.plugins.otp_totp',
 'django_otp.plugins.otp_static',
 'statici18n',
 'django_countries',
 'pretix_cashpayment',
 'pretix_drkmsppayment',
 'pretix_bgukpayment',
 'pretix_telephone',
 'debug_toolbar.apps.DebugToolbarConfig']
Installed Middleware:
[...]

PR #777, opened 2018-02-21
Refs #757: show voucher input for subevents only if subevent is selected

felixrindt:
This reworks the condition on when the voucher input field should be shown in the presale event page.

Now, not only do vouchers need to exists, but if subevents exists, one must be selected for the input field to show up.

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/index.html
+++ b/src/pretix/presale/templates/pretixpresale/event/index.html
@@ -442,7 +442,7 @@
             </form>
         {% endif %}
     {% endif %}
-    {% if vouchers_exist and not event.has_subevents or vouchers_exist and subevent %}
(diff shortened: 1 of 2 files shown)


src/pretix/presale/templates/pretixpresale/event/index.html
```diff
@@ -442,7 +442,7 @@ <h3>{{ tup.0.name }}</h3>
             </form>
         {% endif %}
     {% endif %}
-    {% if vouchers_exist %}
+    {% if vouchers_exist and not event.has_subevents or vouchers_exist and subevent %}
```
felixrindt: This is complicated logic, so I think I'm going to rename `vouchers_exist` to `show_vouchers` and add the logic to the python.

PR #778, opened 2018-02-21
Add "button" operation mode of the widget

raphaelm:


raphaelm:
I'm not happy with the technical design of the items="" parameter but I don't have a better idea right now.

PR #781, opened 2018-02-22
Fix #779: add form field for unpaid option of checkin lists in subevent detail view

### LINKED ISSUE #779: Check-in list include unpaid option in subevent detail view
All properties of a check-in list linked to a subevent can be altered in the subevents detail view. The option `include unpaid` is new, but can't be edited there.

This also has the effect that when duplicating a subevent, the attribute value `include unpaid` of the check-in list isn't copied over to the form.

felixrindt:
This was easier than I thought it would be, although I had to guess a bit...

screenshot from 2018-02-22 13-47-00

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/subevents/detail.html
+++ b/src/pretix/control/templates/pretixcontrol/subevents/detail.html
@@ -178,9 +178,9 @@
                                         </h4>
                                     </div>
                                     <div class="panel-body form-horizontal">
+                                        {% bootstrap_field cl_formset.empty_form.include_pending layout="control" %}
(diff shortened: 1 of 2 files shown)


felixrindt:
Are you ok with the order? Maybe it should be at the top and also have headings like the checkin list change view?

raphaelm:
I don't think headings are required here, but I think it would be better above the product list, since if the product list gets very long, it's easy to miss it this way. Otherwise: perfect!

felixrindt:
Seems good now :)

raphaelm:
You forgot the empty form, sorry ;)

felixrindt:
argh! Sorry ^^

PR #786, opened 2018-02-23
Refs  #774: make question options sortable

felixrindt:
I already added a position field to the question options. I never wrote a migration before, please check it.

Next, my guess is that we use Javascript to let the user reorder the options, just as adding options works. 
The position would be a hidden field in the form. Is there any place in pretix where I can see how to do this?


raphaelm:
I'll look into this in more detail in the next days, but as for your question: The library we use for formsets supports ordering and you can already see the python + js side code for it in the variations view of a product :)

raphaelm:
Hey :) Did you have the chance to have another look at this already? The migration looks good so far.

I have a customer who is waiting for this feature, so if you think you won't have time for another look in the next week, just let me know and I'll finish it up :wink:

felixrindt:
I had my last exam yesterday :laughing: Will continue to work on this now :)

raphaelm:
No worries! :D I hope it went well!

felixrindt:
Ok. Sorting seems to work now. I tested it a bit (changing, reordering, adding, deleting)...
The only problem is, that in my browser, when scrolled to the bottom of the page, the page jumps when starting the reorder animation.

raphaelm:
This doesn't seem to be a problem with your code specifically, but with all places where we use sorting. I'll look into it.

raphaelm:
Yep, your code works perfectly well. I've changed https://github.com/pretix/django-formset-js/commit/c608d7eba461f3ff1c3ae49d5e680e207c67857f and released django-formset-js-improved and then rebased this pull request and updated the dependency so I can merge it right now before I release pretix 1.13 :) Thank you very much! \*high-five\*

felixrindt:
@raphaelm Have you seen my latest comment on this PR?

raphaelm:
> @raphaelm Have you seen my latest comment on this PR?

Uhm, which ones?

felixrindt:
Sorry. [This one](https://github.com/pretix/pretix/pull/786/files/4d8636a2f9a711a95f65c4f882560b996c927199#r172029696). There's some inconsistency with the position fields...

raphaelm:
Can you paste the comment here? I can't see anything except the diff if I click your link.

felixrindt:
Oh I'm sorry. I thought everyone could see that.

In the second commit, I changed the position field of **Question** to `PositiveInteger`, as I saw it [here](https://github.com/felixrindt/pretix/blob/4d8636a2f9a711a95f65c4f882560b996c927199/src/pretix/base/models/items.py#L444). What I *intended* to to was to change the position field in **QuestionOption** to `PositiveInteger`, because in the first commit, I was using a simple Integer field. `PositiveIntegerField` is also used in the Migration, so there are two inconsistancies here:

* `QuestionOption` doesn't use `PositiveInteger`, even though the migration adds it.
* Some Models use a simple Integer, others PositiveInteger for their position field.

raphaelm:
Okay, thanks for noting, I'll probably fix this with the next release! It shouldn't do any damage.

raphaelm:
(Also, postgres is the only database where it actually makes a difference IIRC)

felixrindt:
> (Also, postgres is the only database where it actually makes a difference IIRC) 

Ok. I was wondering why django nor any of my setups where complaining...

> Okay, thanks for noting, I'll probably fix this with the next release! It shouldn't do any damage.

Great. Sadly we rushed this PR a bit and so didn't notice this in review. Again, sorry for the back and forth here :grin:

raphaelm:
> QuestionOption doesn't use PositiveInteger, even though the migration adds it.

Sure? The migration looks consistent to me. It's not very consistent between models, but it seems to be free of contradictions. Compare

```
        migrations.AddField(
            model_name='questionoption',
            name='position',
            field=models.IntegerField(default=0),
        ),
```
to
```
class QuestionOption:
…
    position = models.IntegerField(default=0)
```
and
```
        migrations.AlterField(
            model_name='question',
            name='position',
            field=models.PositiveIntegerField(default=0, verbose_name='Position'),
        ),
```
to
```
class Question:
...
    position = models.PositiveIntegerField(
        default=0,
        verbose_name=_("Position")
    )
```

felixrindt:
You're right. the migrations are consistent. I guess the fact that some models use the positive field and some don't confused me.

PR #788, opened 2018-02-26
Improve voucher filter

rixx:
I had no real clue what "valid" was supposed to mean.

Bonus: Instead of "unredeemed" one could provide "redeemable", which would need to add an annotation to the filter queryset.

### DIFF
--- a/src/pretix/base/models/vouchers.py
+++ b/src/pretix/base/models/vouchers.py
@@ -42,7 +42,7 @@ class Voucher(LoggedModel):
     :param max_usages: The number of times this voucher can be redeemed
     :type max_usages: int
     :param redeemed: The number of times this voucher already has been redeemed
-    :type redeemed: bool
+    :type redeemed: int
     :param valid_until: The expiration date of this voucher (optional)
     :type valid_until: datetime
     :param block_quota: If set to true, this voucher will reserve quota for its holder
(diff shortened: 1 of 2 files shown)


raphaelm:
Hmm, "valid" does not filter "unredeemed" but "Unredeemed & not expired", so we better name it that way? What do you think?

rixx:
In that case, "redeemable" would be more like it, but not entirely true. Maybe we'll just change the filter to show all vouchers that are still redeemable?

raphaelm:
> In that case, "redeemable" would be more like it, but not entirely true. Maybe we'll just change the filter to show all vouchers that are still redeemable?

Hmm, what do you mean by "not entirely true", i.e. what difference would you have in mind?
 I can imagine quite a number of things involving "redeemability", involcing current carts, item availability, presale times (relevant when subevents are in use), quota, … I think this would get to complicated, if we want to make this "entirely true".

rixx:
Well, yeah, that would be too much, but currently you filter for `redeemed=0`, which means a voucher that can be redeemed 10 times and has been redeemed once won't show up, right? That seems easier to fix and like a thing that would probably be expected of "valid" aswell.

raphaelm:
> Well, yeah, that would be too much, but currently you filter for redeemed=0, which means a voucher that can be redeemed 10 times and has been redeemed once won't show up, right? 

Ouch! This really is a bug, then, not only a wording problem!

rixx:
Well, I guess this PR won't be needed after all then.

PR #792, opened 2018-03-03
Improve voucher redemption filter

raphaelm:
@rixx Would love to hear your opinion if you think this makes it clearer


raphaelm:
Thanks! :)

PR #796, opened 2018-03-05
[WIP] Add Dutch translation

mhvis:
As the title suggests. Now 21% complete, and also includes a grammar correction in the English version (this could also be made into a separate commit/pull request if desired).


raphaelm:
Wonderful! As more and more requests for translations are incoming we're planning to publish an official translation workflow and move this from GitHub to a dedicated collaborative translation platform, but we'll just port it there then, so this shouldn't be a reason for you to wait (but might be a reason for us to wait with merging for a bit) :wink:

src/pretix/base/settings.py
```diff
@@ -321,7 +321,7 @@
         'default': LazyI18nString.from_gettext(ugettext_noop("""Hello,
 
 we did not yet receive a payment for your order for {event}.
-Please keep in mind that if we only guarantee your order if we receive
+Please keep in mind that we only guarantee your order if we receive
```
raphaelm: Thanks, will fix on master was well

mhvis: No problem of course, it is a very nice and cool piece of software that you created! So thank you for that :smiley:. I can remove this grammar correction from this pull request or make it into a new one if you desire.

I will continue with the Dutch translation, also because we are planning to use it for our next event (we are using the software for events in our student music association ESMG Quadrivium). When at least all front-end texts have been translated I'll update this pull request.

raphaelm:
Hi @mhvis!

I'm going close this PR for the simple reason that we today launched a new workflow for translating pretix. You can read more about it in [our blogpost](https://pretix.eu/about/de/blog/20180308-translation-process/). The blogpost also explains why we prefer this process over separate Pull Requests.

Don't worry, your work is still much appreciated and remains just as useful. You can just take the .po(s) file from this pull request and upload it [here](https://translate.pretix.eu/projects/pretix/pretix/nl/#upload) and [here (djangojs)](https://translate.pretix.eu/projects/pretix/pretix-js/nl/#upload) after you signed up on our translation platform. I can do this for you, if you want, but did not do so in order to not take credit for the translation submission :wink: 

You can either do this now and continue translating on the platform itself (and others might join you!), or you can upload one now and once later, or you can just upload it after you're done. It's up to you, although I think uploading now would be great to make the progress visible and encourage others to join translating.

Best
Raphael

mhvis:
Great! That's a lot sooner than I was expecting. I will read the blog post and upload the .po files to the platform and continue to work there.

Thanks again for all the work on this great software!

raphaelm:
> Great! That's a lot sooner than I was expecting.

Me too :P Thank you for the kind words!

mhvis:
Hi! I have a small remark on my experience with Weblate so far. It works great, but I have a minor problem which is that saving of translations is slow. That is of course a bit unpleasant because of the large number of strings. I understand that it is probably hard/expensive to fix and I'll just circumvent it using export/import, but just letting you know :smiley:

raphaelm:
> Hi! I have a small remark on my experience with Weblate so far. It works great, but I have a minor problem which is that saving of translations is slow. That is of course a bit unpleasant because of the large number of strings. I understand that it is probably hard/expensive to fix and I'll just circumvent it using export/import, but just letting you know smiley

Thanks for the feedback. I think it is for the fact that it is actually making the git commits synchronously in the background and I'm not sure there is much we could do about it :(

PR #808, opened 2018-03-07
Fix #798 -- Show availability for addon products

### LINKED ISSUE #798: Display remaining quota of add-on products
Currently if you enable "Show number of tickets left" on an event that only applies to products, not add-on products. In my opinion if there is an quota presend for add-on products it should also be displayed (if the quota is exceeded that is already displayed).

Janfred:
This is a very basic approach to address the request of Issue #798 .  
Probably this is not really what you want to see here, since it really looks awkward. Please don't blame me for it :relaxed: 

It's currently showing at https://pretix.460.kif.rocks/sandbox/kif-sandbox/

This is a sandbox shop, so feel free to put something in the cart to see the addon page.

### DIFF
--- a/src/pretix/presale/forms/checkout.py
+++ b/src/pretix/presale/forms/checkout.py
@@ -152,6 +152,9 @@ class AddOnsForm(forms.Form):
             n += ' – {}'.format(_('SOLD OUT'))
         elif avail[0] < 100:
             n += ' – {}'.format(_('Currently unavailable'))
+        else:
+            if avail[1] is not None and event.settings.show_quota_left:
+                n += ' – {}'.format(_('%(num)s currently available') % {'num': avail[1]})
 
         return n
 


src/pretix/presale/forms/checkout.py
```diff
@@ -152,6 +152,9 @@ def _label(self, event, item_or_variation, avail, override_price=None):
             n += ' – {}'.format(_('SOLD OUT'))
         elif avail[0] < 100:
             n += ' – {}'.format(_('Currently unavailable'))
+        else:
+            if avail[1] is not None:
+                n += ' - {}: {}'.format(_('Current availability'), avail[1])
```
dennisschuerholz: I would recommend to use the same dash as in the other cases (`–` not `-` ("–" not "-"..) monospace killing the differece so probably overlooked), looks weird if it's displayed one the same page

dennisschuerholz: Also I personally feel more comfortable if we embed the message in a `small`-tag like it's done for "full-grown" articles - that would again mean we should also modify the other two cases..

raphaelm: Yeah, it should be ``–``. I'm less sure about the ``<small>`` tag but I admit it looks nice – we'd need to be consistent here, though, as you noted.

src/pretix/presale/forms/checkout.py
```diff
@@ -152,6 +152,9 @@ def _label(self, event, item_or_variation, avail, override_price=None):
             n += ' – {}'.format(_('SOLD OUT'))
         elif avail[0] < 100:
             n += ' – {}'.format(_('Currently unavailable'))
+        else:
+            if avail[1] is not None:
```
raphaelm: Replace by:
```
if avail[1] is not None and event.settings.show_quota_left:
```
and this is already fixing #798 completely :)

src/pretix/presale/forms/checkout.py
raphaelm: Also, maybe we choose the wording to be
```
n += ' – {}'.format(_('%(num)s currently available') % {'num': avail[1]})
```
It looks a bit worse in code, but it (a) leaves the order of text and number up to the translator and (b) re-uses the exact same translation string as the other place where we do this, leading to a more consistant language and less work for translators.

raphaelm:
Thanks for the pull request! 

> Probably this is not really what you want to see here, since it really looks awkward. 

Not at all! This is only two very minor changes away from the "correct" solution! :)

src/pretix/presale/forms/checkout.py
Janfred: Sorry, I totally forgot about this setting.

raphaelm:
Thanks! Sorry to be annoying, but I'm still in favour of

```
n += ' – {}'.format(_('%(num)s currently available') % {'num': avail[1]})
```

Yes, it looks unclean, but it will give the string "%(num)s currently available" to translation, which is *exactly* the same string as generated by the ``blocktrans`` statement in ``src/pretix/presale/templates/pretixpresale/event/fragment_quota_left.html`` by Django's translation engine, leading to consistant translations.

raphaelm:
Thanks!

PR #818, opened 2018-03-16
Refs #654 -- API: Writable category endpoints

turegjorup:
Added writable endpoints for categories. I think it's feature complete but any feedback is welcome. 

One note on removing categories from items on delete. I have copied the code from the control view:
```
for item in instance.items.all():
  item.category = None
  item.save()
``` 
I considered having this as a method to avoid duplication but couldn't decide where such a method would logically be placed.


raphaelm:
Well, this is easier to review than the events one. Will be merged as soon as Travis is done with it!

PR #838, opened 2018-03-25
Add order expired notification

felixrindt:
As there are already notifcations for order placed, paid, cancelled and refunded, why not add the only missing order expired notifcation type?


raphaelm:
Good catch, thanks!

PR #849, opened 2018-03-30
Fix #357 -- Implement go to for vouchers

### LINKED ISSUE #357: Search orders/go to voucher
- When I enter an order code in the input box on the order list, and the order exists, I'll see the detail view of the order.
- When I enter a partial order code, I'll get no matches.
- When I enter a voucher code in the input box on the voucher list, and the voucher exists, I'll see a list view containing that one voucher.
- When I enter a partial voucher code, I'll get all vouchers matching that code.

I'd like to see consistent behavior here; at the very least going to a voucher if the entered code corresponds to a voucher exactly. Substring search for order codes would be nice, too, unless you have security concerns there?

mohkami:
In regards to the issue #357 (Search orders/go to voucher):
It seems that partial search works for Orders.
And this PR is about adding Go input to the Vouchers page.
But I also have a couple questions:
1_ I haven't used Djnango translation systems before and my question is about that. The text I added ("There is no voucher with the given voucher code."), even though is passed to the _ugettext_lazy_ function does not appear in the _.po_ file. How is that being added?
2_ As you can see in the images below, the filter section of the Vouchers page is getting cluttered and I'm wondering what you think should be done here.

Order list:
screen shot 2018-03-30 at 2 34 38 pm

Vouchers list:
screen shot 2018-03-30 at 2 34 48 pm

This is my first PR to this repo, and I think my code conforms to the guidelines. But please let me know if there is something I'm missing.

Thank you

### DIFF
--- a/src/pretix/control/views/vouchers.py
+++ b/src/pretix/control/views/vouchers.py
@@ -230,10 +230,10 @@ class VoucherGo(EventPermissionRequiredMixin, View):
     permission = 'can_view_vouchers'
 
     def get_voucher(self, code):
-        return Voucher.objects.get(code=code, event=self.request.event)
+        return Voucher.objects.get(code__iexact=code, event=self.request.event)
 
     def get(self, request, *args, **kwargs):
-        code = request.GET.get("code", "").upper().strip()
+        code = request.GET.get("code", "").strip()
         try:
             voucher = self.get_voucher(code)
             return redirect('control:event.voucher', event=request.event.slug, organizer=request.event.organizer.slug,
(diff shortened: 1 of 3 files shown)


src/pretix/control/views/vouchers.py
```diff
@@ -225,6 +226,23 @@ def post(self, request, *args, **kwargs):
...
 
 
+class VoucherGo(EventPermissionRequiredMixin, View):
+    permission = 'can_view_vouchers'
+
+    def get_voucher(self, code):
+        return Voucher.objects.get(code=code, event=self.request.event)
+
+    def get(self, request, *args, **kwargs):
+        code = request.GET.get("code", "").upper().strip()
```
raphaelm: Instead of calling ``upper()`` here, please remove this and use ``Voucher.objects.get(code__iexact=code, …)`` above, this is more failsave as vouchers aren't strongly guaranteed to be uppercase.

PR #858, opened 2018-04-05
Update manual_smallscale.rst

loceur:
File "/var/pretix/venv/lib/python3.4/site-packages/hierarkey/models.py", line 6, in <module>
    from typing import Any, Callable, Optional
ImportError: No module named 'typing'

Added typing to the venv pip install list, as it seems to be required.


loceur:
Doh!  I missed that!  And I agree.  Ignore my pull request.

PR #862, opened 2018-04-09
Fix #861 -- Add event name to admin notification email

### LINKED ISSUE #861: Add event name to admin notification
When recieving an admin notification, e.g. when an order is placed, only the order number is displayed in the mail. It would be great to also see the event, right now you have to look it up in the "Detail" button url.

I thought this could be an easy task by adding `n.add_attribute(_('Event'), order.event.name)` to `src/pretix/base/notifications.py`, but unfortunately I was not successful.

Thanks for your work!

jfwiebe:
This adds the event name to admin notifications. Without this, it is not possible to distinguish between different event, if you have notifications enabled for multiple accounts (exept from the url).

### DIFF
--- a/src/pretix/base/notifications.py
+++ b/src/pretix/base/notifications.py
@@ -174,6 +174,7 @@ class ParametrizedOrderNotificationType(NotificationType):
             title=self._title.format(order=order, event=logentry.event),
             url=order_url
         )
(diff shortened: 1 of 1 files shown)


raphaelm:
Thanks! :)

PR #883, opened 2018-04-24
Fix #881 -- dont redirect tel scheme

### LINKED ISSUE #881: rich_text: tel schema
When providing email links using the mailto schema in richtext description like `[Email us!](mailto:mail@weandus.com)`, resulting in [Email us!](mailto:mail@weandus.com), Pretix creates the correct `<a>` tag. However, users also use their mobile phone. It would be awesome to also be able to use the `tel` schema like `[Call us!](tel:+1-202-555-0102)`. At the moment, pretix just creates an `<a>` tag without an `href`, so the Telephone app is not opened.

felixrindt:
closes #881


raphaelm:
Great!

PR #893, opened 2018-05-03
fix placeholder for event deletion password

luto:
probably due to copypasta the event deletion form says "new password". This changes it to "Your password" to be in line with the data shredder pw confirmation field.


raphaelm:
Oops! Thanks :)

Issue #828, opened 2018-03-20
Order thank you text misleading


felixrindt:
Looking at this string:
https://github.com/pretix/pretix/blob/1ef3f83e468b4e6205eb43c3b7fbe915f2ff2237/src/pretix/presale/templates/pretixpresale/event/order.html#L26

What if the organizer is not using the ticket feature or not allowing changeing order details? Maybe something like the following better generalizes to every pretix event.

| old | new |
| --- | --- |
| `Please bookmark or save the link to this exact page if you want to download your ticket or  change your details later. We also sent you an email containing the link to the address you specified.` | `Please bookmark or save the link to this exact page if you want to access your order later. We also sent you an email containing the link to the address you specified.` |

raphaelm:
Agreed, your test is better :)

PR #897, opened 2018-05-07
Presale: change email field help text

felixrindt:
Just like #828, I'd like to change the email field description in the presale to be more general.


raphaelm:
Great!

PR #906, opened 2018-05-11
Add setting to configure subevent ordering on frontpage

### LINKED ISSUE #898: Configurable subevent ordering
At the moment, subevent ordering mainly depends on their start date. This is not always the most intuitive, e.g. when there are events in two different locations at various times and people mainly decide which event to attend based on the events location. If the subevents where grouped by in this case location, it would be easier to find the desired subevent in the list on the presale page.

We could

- make the subevent positions configurable in a per subevent basis (like in many other places in pretix, where paginated item list items can be reordered), or
- provide functionality to chose an ordering method (by event start, event name, location->name, ...)

What do you think?

felixrindt:
Fixes #898.

### DIFF
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -528,12 +528,13 @@ class Event(EventMixin, LoggedModel):
 
     @property
     def active_future_subevents(self):
+        orderfields = self.settings.get('frontpage_subevent_ordering', default=['-date_from', 'name'], as_type=list)
         return self.subevents.filter(
             Q(active=True) & (
                 Q(Q(date_to__isnull=True) & Q(date_from__gte=now()))
                 | Q(date_to__gte=now())
             )
-        ).order_by('date_from', 'name')
+        ).order_by(*orderfields)
 
     @property
     def meta_data(self):
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -550,7 +550,7 @@ class Event(EventMixin, LoggedModel):
                 Q(Q(date_to__isnull=True) & Q(date_from__gte=now()))
                 | Q(date_to__gte=now())
             )
-        )
+        )  # order_by doesn't make sense with I18nField
         return sorted(subevs, key=attrgetter(*orderfields))
 
     @property
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -2,6 +2,7 @@ import string
 import uuid
 from collections import OrderedDict
 from datetime import datetime, time
+from operator import attrgetter
 
 import pytz
 from django.conf import settings
@@ -528,6 +529,15 @@ class Event(EventMixin, LoggedModel):
 
     @property
     def active_future_subevents(self):
+        return self.subevents.filter(
+            Q(active=True) & (
+                Q(Q(date_to__isnull=True) & Q(date_from__gte=now()))
+                | Q(date_to__gte=now())
+            )
+        ).order_by('date_from', 'name')
+
+    @property
+    def subevent_list_subevents(self):
         ordering = self.settings.get('frontpage_subevent_ordering', default='date_ascending', as_type=str)
         orderfields = {
             'date_ascending': ('date_from', 'name'),
@@ -535,12 +545,13 @@ class Event(EventMixin, LoggedModel):
             'name_ascending': ('name', 'date_from'),
             'name_descending': ('-name', 'date_from'),
         }[ordering]
-        return self.subevents.filter(
+        subevs = self.subevents.filter(
             Q(active=True) & (
                 Q(Q(date_to__isnull=True) & Q(date_from__gte=now()))
                 | Q(date_to__gte=now())
             )
-        ).order_by(*orderfields)
+        )
+        return sorted(subevs, key=attrgetter(*orderfields))
 
     @property
     def meta_data(self):
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -528,7 +528,13 @@ class Event(EventMixin, LoggedModel):
 
     @property
     def active_future_subevents(self):
-        orderfields = self.settings.get('frontpage_subevent_ordering', default=['-date_from', 'name'], as_type=list)
+        ordering = self.settings.get('frontpage_subevent_ordering', default='date_ascending', as_type=str)
+        orderfields = {
+            'date_ascending': ('date_from', 'name'),
+            'date_descending': ('-date_from', 'name'),
+            'name_ascending': ('name', 'date_from'),
+            'name_descending': ('-name', 'date_from'),
+        }[ordering]
         return self.subevents.filter(
             Q(active=True) & (
                 Q(Q(date_to__isnull=True) & Q(date_from__gte=now()))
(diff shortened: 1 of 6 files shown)


felixrindt:
Here's the problem I'm having with this:
I want the setting to be a list of subevent field names so I can pass it directly to the `order-by` function. Deserialization works fine using the `as_type` keyword arg of `settings.get`, but the SettingsView is not serializing a list value using json, but a call to `str`, which then breaks unserialization because of `'` and `"` getting swapped.

That's why I'm delcaring it with the type `str` and then deserializing it using the `as_type` arg. This is hacky and therefore not a viable solution yet.

src/pretix/control/forms/event.py
```diff
@@ -854,6 +854,16 @@ class DisplaySettingsForm(SettingsForm):
         label=_("Show variations of a product expanded by default"),
         required=False
     )
+    frontpage_subevent_ordering = forms.ChoiceField(
+            label=_('Subevent ordering'),
+            choices=[
+                ('[ "date_from", "name"]', _('Event start time')),
+                ('["-date_from", "name"]', _('Event start time (descending)')),
+                ('[ "name", "date_from"]', _('Name')),
+                ('["-name", "date_from"]', _('Name (descending)')),
```
felixrindt: Would love to use an actual list instead of the json representation of a list here, but serialization is not using json atm.

felixrindt:
Here's a screenshot of a szenario where I'd say this makes sense:

### sorted by date
screenshot_2018-05-11 presale climbing workshop before

### vs sorted by name
screenshot_2018-05-11 presale climbing workshop

raphaelm:
>  I want the setting to be a list of subevent field names so I can pass it directly to the order-by function. Deserialization works fine using the as_type keyword arg of settings.get, but the SettingsView is not serializing a list value using json, but a call to str, which then breaks unserialization because of ' and " getting swapped.

We could fix the settings view, but I'd still consider it a hack. I believe, storing actual order_by values into the database here is a level too "low", I'd rather store string constants that represent the user's intentions. When selecting "Name", the user's intention is to sort by name, the secondary sorting by date is an implementation detail. I'd just store "name" as a string into the settings store and then map it to the full parameter when it is used. What do you think`

felixrindt:
If that's what you think is more appropriate here, let's do it like that.

src/pretix/base/models/event.py
```diff
@@ -528,12 +528,19 @@ def active_subevents(self):
 
     @property
     def active_future_subevents(self):
+        ordering = self.settings.get('frontpage_subevent_ordering', default='date_ascending', as_type=str)
+        orderfields = {
+            'date_ascending': ('date_from', 'name'),
+            'date_descending': ('-date_from', 'name'),
+            'name_ascending': ('name', 'date_from'),
+            'name_descending': ('-name', 'date_from'),
+        }[ordering]
```
felixrindt: Using this inline dictionary now. There's still two places one would have to edit to add a new sorting method.

felixrindt:
Using an inline dictionary now. There's still two places one would have to edit to add a new sorting method.

src/pretix/control/forms/event.py
```diff
@@ -854,6 +854,15 @@ class DisplaySettingsForm(SettingsForm):
         label=_("Show variations of a product expanded by default"),
         required=False
     )
+    frontpage_subevent_ordering = forms.ChoiceField(
+        label=_('Subevent ordering'),
```
raphaelm: Please replace by
```
pgettext('subevent', 'Date ordering')
```
since we don't expose the word "subevent" to end users :)

src/pretix/control/forms/event.py
```diff
@@ -854,6 +854,15 @@ class DisplaySettingsForm(SettingsForm):
...
+        choices=[
+            ('date_ascending', _('Event start time')),
+            ('date_descending', _('Event start time (descending)')),
+            ('name_ascending', _('Name')),
+            ('name_descending', _('Name (descending)')),
+        ],
+    )
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
```
raphaelm: Can we drop out the field if the event doesn't use subevents? e.g. something like
```
if not self.event.has_subevents:
    del self.fields['frontpage_subevent_ordering']
```
here in the ``__init__`` method after the super call

src/pretix/control/templates/pretixcontrol/event/display.html
```diff
@@ -11,6 +11,7 @@
             {% bootstrap_field form.logo_image layout="control" %}
             {% bootstrap_field form.frontpage_text layout="control" %}
             {% bootstrap_field form.show_variations_expanded layout="control" %}
+            {% bootstrap_field form.frontpage_subevent_ordering layout="control" %}
```
raphaelm: If we show the field only conditionally, we need to wrap this
```
{% if form.frontpage_subevent_ordering %}
    {% bootstrap_field form.frontpage_subevent_ordering layout="control" %}
{% endif %}
```

src/pretix/control/forms/event.py
felixrindt: Of course!

src/pretix/control/forms/event.py
felixrindt: `kwargs['event']` raises a KeyError for me, although I see `self.event = kwargs.pop('event')` used throughout the file. `self.obj` gives me the event though.

raphaelm:
Ah, I completely overlooked something here: Does ordering by a I18nField actually make any sense?

felixrindt:
Good catch... While sorting in python should work via [the LazyI18nString[(https://github.com/raphaelm/django-i18nfield/blob/4fbac3188de37a70706f3578c05983bc37a0c8dd/i18nfield/strings.py#L87), the database ordering is done via the actual field value, which is the json string. That of course can have the wrong language listed first, which might provide a different ordering than expected, a cause of not having atomic database values...

A fix to this would be to explicitly sort this "in python"?!

felixrindt:
We now sort in python, which evaluates the LazyI18nString before comparing it. CI seems to be broken?!

felixrindt:
@raphaelm Anything else I should do here?

felixrindt:
Ok. Hope you had fun at DjangoCon :)

I'm not sure sorting those entries is a big deal, as the `name` field needs to be evaluated anyway and sorting hundrets of entries is a quick operation, especially using the `attrgetter`. I did some testing on a local instance with 2272 subevents with german and english name.
It takes around 0.28 seconds to get through the `subevent_list_subevents()` method.
I first evaluate the query set using `list()` and then sort it using `sorted()` while measuring the time between commands.
Evaluation takes around 0.280 seconds while the actual sorting is just 0.003 seconds.
As you can see, evaluating the query set (=constructing actual objects from the database) is most of the work. This is currently done through the iteration over the queryset in the list template. [Django Reference](https://docs.djangoproject.com/en/2.0/ref/models/querysets/)
The time added by sorting in python is negligible in my opinion.

raphaelm:
Of course you are absolutely correct. Let me put my argument in a different way: With large numbers of subevents, we *should* be using pagination anyways, we just never implemented it due to laziness; implementing pagination efficiently will now be impossible once we start to sort in Python not possible when sorting with Python; therefore sorting in Python is "bad".

However, it is probably still useful and in cases with many events the calendar view is more usable anyways, so I'm think I'm going to merge it. I hope to get around reviewing it in detail in the next 2 days.

felixrindt:
Thank you :smile:

PR #910, opened 2018-05-14
Fix #908 -- Fix display of ticket download message

### LINKED ISSUE #908: Ticket download buttons/message should only be shown when ticket output providers are anabled
When ticket download is enabled for an event and no ticket output providers are enabled, the order details page will show a message like

```
You will be able to download your tickets here starting on .
```

This message should only be shown when ticket output providers are enabled.

(In the example above, the “Ticket download date” is not set which makes the message look weird).

rimi-itk:
Hide download tickets message when no download date is specified.


raphaelm:
Great, thanks!

Issue #847, opened 2018-03-29
Limit number of characters in one-line-answers (attendee registration)


Problemkerze:
We want to print the company name of the attendee onto our tickets. Sometimes the company name is quite long and the company name won't fit onto the ticket any more.

Thus, it would be great to be able to limit the number of characters in one (or multiline) answers.

raphaelm:
If we do this, I suggest we do it for both one and multiline answers :)

felixrindt:
I think question answer validation could be also done for other types. Think min/max date/time for date/time questions or min/max for number input.

raphaelm:
This has been implemented a while ago

PR #916, opened 2018-05-18
Fix bug in date/time question stats

felixrindt:
As you can see, the table and graph data is messed up. Aggregation is not correct and links don't point to the correct filtered list:

screenshot from 2018-05-18 14-14-57

I streamlined the code and it's showing correct results now. Also, I changed the ordering for date/time questions.

screenshot from 2018-05-18 14-29-22

### DIFF
--- a/src/pretix/control/views/item.py
+++ b/src/pretix/control/views/item.py
@@ -429,12 +429,12 @@ class QuestionView(EventPermissionRequiredMixin, QuestionMixin, ChartContainingV
         elif self.object.type in (Question.TYPE_TIME, Question.TYPE_DATE, Question.TYPE_DATETIME):
             qs = qs.order_by('answer')
             model_cache = {a.answer: a for a in qs}
-            qs = qs.values('answer').annotate(count=Count('id'))
+            qs = qs.values('answer').annotate(count=Count('id')).order_by('answer')
             for a in qs:
                 a['alink'] = a['answer']
                 a['answer'] = str(model_cache[a['answer']])
         else:
-            qs = qs.values('answer').annotate(count=Count('id')).order_by('-count')
+            qs = qs.order_by('answer').values('answer').annotate(count=Count('id')).order_by('-count')
 
             if self.object.type == Question.TYPE_BOOLEAN:
                 for a in qs:
--- a/src/pretix/control/views/item.py
+++ b/src/pretix/control/views/item.py
@@ -427,14 +427,9 @@ class QuestionView(EventPermissionRequiredMixin, QuestionMixin, ChartContainingV
                 a['answer'] = str(a['options__answer'])
                 del a['options__answer']
         elif self.object.type in (Question.TYPE_TIME, Question.TYPE_DATE, Question.TYPE_DATETIME):
-            qs = qs.order_by('answer')
-            qs_model = qs
-            qs = qs.values('answer').annotate(count=Count('id')).order_by('-count')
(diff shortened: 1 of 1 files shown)


felixrindt:
Yes I know that birthday is in the future ^^ See my comment in #847.

raphaelm:
There's a reason this was more complex. Once you set your pretix to a different language, you need to re-format the datetimes before you can display them, so they are displayed with the correct locale. Additionally, once you're dealing with date**times**, you need to do timezone calculations since we always store UTC in the database. I believe this fix breaks both of these cases. These are all things that the database and Django do automatically for DateTimeFields, but here we're silly and need to do it ourselves unfortunately.

So yes, there's a bug, but this isn't the proper fix yet. :wink: Sorry!

src/pretix/control/views/item.py
```diff
@@ -427,14 +427,9 @@ def get_answer_statistics(self):
                 a['answer'] = str(a['options__answer'])
                 del a['options__answer']
         elif self.object.type in (Question.TYPE_TIME, Question.TYPE_DATE, Question.TYPE_DATETIME):
-            qs = qs.order_by('answer')
-            qs_model = qs
-            qs = qs.values('answer').annotate(count=Count('id')).order_by('-count')
-            for a, a_model in zip(qs, qs_model):
-                a['alink'] = a['answer']
-                a['answer'] = str(a_model)
```
felixrindt: So this `str` is for localization. Sorry for not thinking about that.

raphaelm:
> So this str is for localization. Sorry for not thinking about that.

Yes :( And the weird zipping with objects is for getting all data for that/reusing the existing code for it. Could you check if that zip() step goes all wrong or if the data is wrong before already?

felixrindt:
I guess the problem is in the reordering.

felixrindt:
`a['answer'] = str(qs_model.filter(answer=a['answer']).first())` does work, but it's a query for every answer...

raphaelm:
> a['answer'] = str(qs_model.filter(answer=a['answer']).first()) does work, but it's a query for every answer...

Then maybe use a cache like
```
model_cache = {
  a.answer: a for a in qs_model
}
```

felixrindt:
ok

src/pretix/control/views/item.py
```diff
@@ -428,13 +428,13 @@ def get_answer_statistics(self):
...
-            for a, a_model in zip(qs, qs_model):
+            model_cache = {a.answer: a for a in qs}
+            qs = qs.values('answer').annotate(count=Count('id'))
+            for a in qs:
                 a['alink'] = a['answer']
-                a['answer'] = str(a_model)
+                a['answer'] = str(model_cache[a['answer']])
         else:
-            qs = qs.order_by('answer').values('answer').annotate(count=Count('id')).order_by('-count')
+            qs = qs.values('answer').annotate(count=Count('id')).order_by('-count')
```
raphaelm: Are you sure this change is necessary? Normally, order is relevant when dealing with GROUP BY clauses. Otherwise, the PR looks fine to me now. Probably we should just put ``.order_by()`` at the end of the queries, also for the datetimes. See also: https://docs.djangoproject.com/en/2.0/topics/db/aggregation/#interaction-with-default-ordering-or-order-by

felixrindt: I don't think GROUP BY cares about the order, but we wouldn't get consistent ordering based just on the count.

PR #923, opened 2018-05-26
[WIP] Default value in questions

### LINKED ISSUE #794: Default value in questions
Questions should allow setting a default value, which is automatically filled.

Our special use case is to ask attendees for their arrival and departure date which should default to the start and end date of the event.

johan12345:
working to fix #794 

To do list:
- [x] Model field and validation for default value
- [x] Input in question settings to set default value
    - [x] Make the input type adjust depending on the question type -> I think we need JavaScript for this
- [x] Show default values in checkout questions form
- [ ] implement support for Choice and multiple choice question types
- [x] implement support for file question type (if that makes sense) or hide the default value field in this case

pretix


johan12345:
So this seems to be working now for all question types except the Choice/MultipleChoice types.

To replace the input for the default answer depending on the selected question type, I needed to use some JavaScript, which seems kind of hacky, but I don't really see another way.

For implementing this for the Choice/MultipleChoice types, we would need to add a "selected by default" field to the `QuestionOption` class, right?

src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -6,6 +6,44 @@ function question_page_toggle_view() {
 
     show = $("#id_type").val() == "B" && $("#id_required").prop("checked");
     $(".alert-required-boolean").toggle(show);
+
+    let input = $('#id_default_value');
```
raphaelm: Could you wrap this in a function of if statement that only executes on the question edit page? Mostly to make it obvious to future developers what this is used for.

src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -6,6 +6,44 @@ function question_page_toggle_view() {
...
+        case 'D':
+            input.replaceWith('<input type="date" class="form-control" value="' + value + '" ' + common_attrs + '>');
+            $('.form-group:has(#id_default_value)').show();
+            break;
+        case 'H':
+            input.replaceWith('<input type="time" class="form-control" value="' + value + '" ' + common_attrs + '>');
+            $('.form-group:has(#id_default_value)').show();
+            break;
+        case 'W':
+            input.replaceWith('<input type="datetime-local" class="form-control" value="' + value + '" ' + common_attrs + '>');
```
raphaelm: We're not using type=date, type=time or type=datetime-local anywhere in pretix, as their UX is horrible in nearly all Desktop browsers. Could you instead try to use the date/time pickers we use everywhere already? :)

src/pretix/base/forms/questions.py
```diff
@@ -88,21 +91,21 @@ def __init__(self, *args, **kwargs):
                 field = forms.DecimalField(
                     label=q.question, required=q.required,
                     help_text=q.help_text,
-                    initial=initial.answer if initial else None,
+                    initial=initial.answer if initial else default if default else None,
```
raphaelm: Nested ``if`` statements of this kind are very hard to read, could you add in some ``()`` to make the order of execution clear?

src/pretix/static/pretixcontrol/js/ui/main.js
johan12345: As far as I can see, the `question_page_toggle_view` function is already only executed on the question edit page, right? Of course, I can still extract this piece in a function that is then executed from `question_page_toggle_view`.

raphaelm:
It is never that easy when datetimes are concerned ;) The way you implemented it, the database contents depend on the active locale of the user editing the form, and if someone enters a default date in German and later wants to edit the event in English, the form won't work.

I added some more hacks to normalize the database contents to ISO formats (and in case of datetimes, to UTC) and rebased this against current master properly. Since this makes it non-trivial, it would be great to have some very basic tests. Do you want to write them or should I create them?

johan12345:
Oh, right, that makes sense - thanks!
I can have a look at adding some tests. Probably just of the backend logic, not the JavaScript stuff for switching out the form fields, right?

raphaelm:
Just to save you the scrolling through travis when you revisit this, it's failing because

```
./pretix/control/forms/item.py:1:1: F401 'copy' imported but unused
```

PR #929, opened 2018-05-29
Update Dutch translation for October

jochemvankessel:
Dutch translation for October should be Oktober


raphaelm:
Weblate creates a PR automatically (after some time), so this will be merged in #930 :)

jochemvankessel:
@raphaelm Great, thx!

PR #950, opened 2018-06-16
Strip [] in mail subject prefix

rixx:


raphaelm:
Yeah, that makes sense :D

PR #980, opened 2018-08-03
Widget: set default number to 1 if there is only one product

johan12345:
I think this makes sense as it is already done in the normal purchase flow outside the widget (since 8f2adf0a).
Was requested by internal support ticket # 2336058.

### DIFF
--- a/src/pretix/presale/views/widget.py
+++ b/src/pretix/presale/views/widget.py
@@ -195,7 +195,7 @@ class WidgetAPIProductList(View):
                     } for item in g
                 ]
             })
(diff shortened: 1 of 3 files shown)


raphaelm:
I added one commit to fix the failing test cases, and another to prevent this from happening if there is only one item, but with variations (we don't want to default to buy one of each variation, that doesn't make sense). My local clone of pretix is currently messed up due to the Django 2.1 migration, so I haven't tested it locally, do you think the changes make sense?

johan12345:
Thanks! I tested it locally and your change works.

PR #990, opened 2018-08-15
Show provider-specific icons for TicketOutput

pc-coholic:
pretix-passbook has already led the way by not only setting download_button_text, but also download_button_icon.

Here is the missing link, that also enables pretix to actually display a set icon. If none is set, it defaults back to fa-download.

While we're at it, we could use the opportunity to also change the PDF-output's plugin to `fa-file-pdf`.


raphaelm:
Could you rebase this against current master so the diff is readable? :) Thanks!

PR #994, opened 2018-08-16
Rename module async to tasks

### LINKED ISSUE #993: python3.7 async keyword breaks module name
When trying to import the file `pretix/src/pretix/base/services/async.py`, a syntax error is thrown when using python 3.7 (which is supported as per [Python 3.4 or newer](https://docs.pretix.eu/en/latest/development/setup.html))

      File "/home/felix/projects/pretix/src/pretix/base/services/invoices.py", line 24
        from pretix.base.services.async import TransactionAwareTask
                                      ^
    SyntaxError: invalid syntax

I guess changing the module name should fix it.

felixrindt:
Fixes #993


src/pretix/presale/views/tasks.py
```diff
@@ -0,0 +1,3 @@
+from pretix.base.views.tasks import *  # noqa
+
+# file kept for backwards compatibility
```
felixrindt: We could remove this, as it must get renamed at least.

felixrindt:
Should be good now. Somewhat glad we got this done for 2.0 :)

raphaelm:
Thanks! :)

PR #1000, opened 2018-08-30
Prevent email enumeration

bockstaller:
Here is my attempt to prevent user enumeration. 
I've made the following changes:

**Application:**
- replaces success and failure messages in the form with two (with/without redis) information messages 
- adds logging for attempted password resets of unknown users
- adds logging for failing emails

**Tests:**
- test_unknown asserts a redirect instead of a ok
- adds test_email_reset_twice_redis to assert the correct logging of a twice reset email 
- adds a FakeRedis class similiar to the one implemented in test_metrics.py. I could refactor them into the testutils folder if prefered. 

Please excuse the commit mess. I am currently fighting with my tooling.


bockstaller:
Seems like I have tripped over the conditional requirement django-redis.
Digging further into the issue unveiled, or at least I think so, that the requirements-files and the setup.py aren't in sync. At the moment setup.py requires django-redis as a required dependency:
 [https://github.com/pretix/pretix/blob/35ddf6790e226aabc52441b1b06d42bcecfa7f6d/src/setup.py#L124](url)
This isn't the case in the .travis.sh file. It installs only the requirements/dev.txt and not requirements/redis.txt

I am not sure what is the best way to proceed from here.

@raphaelm. Given your current inactivity:
I hope you enjoy your time off  👍

bockstaller:
Seems like I have tripped over the conditional requirement django-redis.
Digging further into the issue unveiled, or at least I think so, that the requirements-files and the setup.py aren't in sync. At the moment setup.py requires django-redis as a required dependency:
https://github.com/pretix/pretix/blob/35ddf6790e226aabc52441b1b06d42bcecfa7f6d/src/setup.py#L124
This isn't the case in the .travis.sh file. It installs only the requirements/dev.txt and not requirements/redis.txt

I am not sure what is the best way to proceed from here.

@raphaelm. Given your current inactivity:
I hope you enjoy your time off 👍

raphaelm:
Great, this should fix the problem, at least except for timing attacks, which are probably still possible but really hard to prevent here.

> Seems like I have tripped over the conditional requirement django-redis.
> Digging further into the issue unveiled, or at least I think so, that the requirements-files and the setup.py aren't in sync. At the moment setup.py requires django-redis as a required dependency:

Good catch, that seems to be a bug. Redis itself is an optional dependency, so historically we treated django-redis the same way. However, since apparently we install it for everyone through setup.py, we can also just go with it and include it everywhere. I will deal with it after this PR.

I did a few cosmetic changes to get this going asap and will merge is as soon as travis has passed.

PR #1028, opened 2018-09-20
Fix outside of body script element

chrko:
```
</body>
<script src="/static/pretixbase/js/errors.4db07680b709.js"></script>
</html>
```
The script tag is after the body tag. This merge request switches the corresponding lines.


raphaelm:
Thanks! :)

PR #1029, opened 2018-09-20
[WIP] Allow to transfer ticket to new owner

raphaelm:
- [ ] Require email confirmation from original owner?
- [ ] Write tests


raphaelm:
I don't think we'll finish this any time soon, if at all. This can be done in a plugin and https://github.com/rixx/pretix-swap achieves a similar thing.

PR #1046, opened 2018-10-13
Improve accessibility

FlaviaBastos:
Proposed changes to improve accessibility.


src/pretix/presale/templates/pretixpresale/base.html
```diff
@@ -33,14 +33,14 @@
...
-    <meta name="viewport" content="width=device-width, initial-scale=1, maximum-scale=1.0, user-scalable=0">
+    <meta name="viewport" content="width=device-width, initial-scale=1, user-scalable=0">
     <link rel="icon" href="{% static "pretixbase/img/favicon.ico" %}">
     {% block custom_header %}{% endblock %}
 </head>
 <body class="nojs" data-locale="{{ request.LANGUAGE_CODE }}" data-now="{% now "U.u" %}" data-datetimeformat="{{ js_datetime_format }}" data-timeformat="{{ js_time_format }}" data-dateformat="{{ js_date_format }}" data-datetimelocale="{{ js_locale }}">
 {% block above %}
 {% endblock %}
-<div class="container">
+<div class="container" role="main">
```
rixx: Is this correct? As I understand it, `role="main"` should contain only the actual main content of the document, allowing users to skip repetetive content that appears on every page. For pretix, this would include at least the event header, I'd say.

FlaviaBastos: Hmm... :thinking: I don't know much about this so I'll do some research and get back to you. :)

FlaviaBastos: Sorry about the delay. 
Yeah, you are right. However, I tried to add this to a more relevant part (say, the form) but that cannot be done that way. So I ended up deciding to remove this attribute altogether. 
Now, sad story: my branch was stale and I messed up refreshing it :disappointed: 

So here's the real question: considering I never discussed this proposal with any of you before, do you think it's valuable to add? If so, I will see if I can figure out a way of recovering from my mess or might end up closing this PR and sending a new cleaner one. On the other hand, if you think it's not the time, I will be super ok with that too (and we can close this PR).

raphaelm:
Hi Flavia,

> So here's the real question: considering I never discussed this proposal with any of you before, do you think it's valuable to add? If so, I will see if I can figure out a way of recovering from my mess or might end up closing this PR and sending a new cleaner one. On the other hand, if you think it's not the time, I will be super ok with that too (and we can close this PR).

I'm sooo sorry that I didn't react to this PR for so long. You, and this PR, certainly deserve better. :heart:  However, the PR happened a days before I left the country for two weeks and somehow it totally got lost on my agenda. I'll do better next time!

Here's what I think of the changes in detail:

* The ``meta name=viewport`` change is *probably* good, but before I merge it I'd want to test it on real Android and iOS devices to see if it works as expected or if it makes the site behave weird. If you want, you can create a new PR just for that, or I'll just do it myself when I get around.
* I'm not really in favor of the color changes (that now somehow got lost from the PR). We hired a designer a while back to create a matching set of colors, and I'm hesistant of changing them without him involved even if you are right that the red color doesn't have a great contrast. On the other hand, we use it for text in only one or two places, so even though it isn't great for accessibility, I don't think it's a major hassle either.
* For the ``role`` attribute, as Tobi said and you noticed, it's probably non-trivial to put it in the right place, since there's no element that's really right. The best would probably be to create a new element around places like ``src/pretix/presale/templates/pretixpresale/event/base.html`` line 75-76, but we'd need to check if that has any layouting impact.

FlaviaBastos:
Hey, @raphaelm! 
No worries! As I mentioned, I opened this without consulting you so really, it was my bad. Sorry! I realize now that this added some extra work for you and @rixx ... :(
All your arguments make total sense and thank you so much for taking the time to explain that to me. That's very considerate of you!
I will keep an eye on the open issues and grab something from there next time. :)

PR #1048, opened 2018-10-16
Added markdown formatting to event location

MyIgel:


src/pretix/presale/templates/pretixpresale/event/index.html
```diff
@@ -130,7 +130,7 @@ <h2 class="subevent-head">{{ subevent.name }}</h2>
                     <div class="info-row">
                         <span class="fa fa-map-marker fa-fw"></span>
                         <p>
-                            {{ ev.location|linebreaksbr }}
+                            {{ ev.location|rich_text|linebreaksbr }}
```
raphaelm: ``|rich_text`` and ``|linebreaksbr`` should not be combined, since it creates duplicate line breaks. e.g.

```
Line 1
Line 2

Paragraph 2
```

would be converted to
```
Line 1<br>Line 2</p><br><p>Paragraph 2</p>
```
instead of
```
Line 1<br>Line 2</p><p>Paragraph 2</p>
```

Usually, we just use ``|rich_text``. However, this would be backwards-incompatible, since a usual location specifier like

```
Congress Center Hamburg
12345 Hamburg
```

would turn into

```
Congress Center Hamburg 12345 Hamburg
```

with only ``rich_text``, which is also not intended.

We *could* change ``rich_text`` to keep line breaks intact, which would likely be a good idea (but also be backwards incompatible everywhere).

I'm not sure how to resolve this best.

That said, I'm undecided whether adding markdown support here is a good idea at all. I see why it's useful, but the location field is also used in places where it might be distracting (eg. in event lists) or even a problem (e.g. in the passbook module we use it to autosuggest geolocations).

MyIgel: Hm, that is a point, a "strip formatting" would be useful there
A usecase would be to provide a link to a map or similar in the address field

raphaelm: Hmm. Maybe it would be more useful to add a geolocation field to core (like the passbook plugin already has one) and then create plugins that add a nice inline GMaps/OSM widget :)

MyIgel: That would be a better option :D
(And more expandable)

raphaelm:
Let's close this in favor of #1062 then :) Thanks!

PR #1056, opened 2018-10-29
Allow empty plugin responses

rixx:
While plugin developers are supposed to return an empty dictionary, it's
conceivable that they might just put in a `return` if their field is not
needed, and pretix being generous about this would be cool.


raphaelm:
Thanks!

PR #1063, opened 2018-11-03
Removed old configuration for events settings

oocf:
This part is no longer needed, and for the users could be complicated to have it, maybe we can delete this part.


raphaelm:
Hmm, you're right, by now no one will be looking for it there because they know how the system was *before*. However, the team/organizer settings are still hard to discover. We are working on that(tm), but until the navigation structure is new, I'm not sure if this might still be useful to keep around. Are there other opinions on this?

rixx:
You could consider leaving just the "permissions" heading there, but linking it directly to the organizer settings, at least for now.

PR #1064, opened 2018-11-04
Add language attribute to html element

rixx:


rixx:
Ah, I forgot about pretix' non-standard language codes. Yes, you want standardised language attributes, and choose the shortest attribute applicable (favoring the "de" over the subtag+region subtag combination "de-DE". At the same time, the region subtag in "pt-BR" or in "nl-BE" is required and cannot be left out.) 
As pretix currently seems to use shortened versions wherever possible, it seems to me the only thing we need to strip is the custom "informal" subtag.

src/pretix/control/forms/filter.py
```diff
@@ -739,7 +739,6 @@ class VoucherFilterForm(FilterForm):
         label=_('Search voucher'),
         widget=forms.TextInput(attrs={
             'placeholder': _('Search voucher'),
-            'autofocus': 'autofocus'
```
raphaelm: Did you remove this on purpose?

rixx: I did, as the commit message states: [control] Remove second autofocus element on voucher list

raphaelm: Sorry, I didn't look at the commit message.

rixx:
Sorry, I don't think I'll find the time to finish this any time soon.

Issue #1058, opened 2018-10-30
Pending order email for banktransfer needs more translatable field


mdehollander:
Currently, the confirmation email with information on the pending payment when using the bank transfer option contains English sentences. It would be nice to have them as translatable strings.

The template [order_pending.txt](https://github.com/pretix/pretix/blob/master/src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/email/order_pending.txt) contains the sentence `Please transfer the full amount to the following bank account.` and the words `Reference` and `Amount`. 
https://github.com/pretix/pretix/blob/df2d8925ed74c7195f48f1d4bca9b0748e041e5c/src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/email/order_pending.txt#L2

It would be nice translations can be made for these things using translate.pretix.eu
 

Also the email subject says "Your order" 
https://github.com/pretix/pretix/blob/df2d8925ed74c7195f48f1d4bca9b0748e041e5c/src/pretix/base/services/orders.py#L631

rixx:
All of those strings are translatable – in the order_pending.txt you can see that the "blocktrans" marker marks the whole block as translateable, whereas the subject is enclosed by a `_()`, which also marks translateable strings. I can also see that they are present in the translation files which are generated by translate.pretix.eu, e.g. the first example is translated here: https://github.com/pretix/pretix/blob/master/src/pretix/locale/nl/LC_MESSAGES/django.po#L13068

mdehollander:
English, Dutch and French (that's why i am running my own instance(2.2.0.dev0), with Dutch as default language. I did not try this in the pretix hosted version.

raphaelm:
I assume in that case, that your user's language is English (what language is shown in the order details in the top box?) but you only have a Dutch mail template in Setting>Email

mdehollander:
The order locale is Nederlands and there is text for all templates in all languages. I did change the text of the template, not sure of this caused a problem. Could it be a problem that I am running the 2.2.0.dev0 version and not the last stable release?

raphaelm:
> Could it be a problem that I am running the 2.2.0.dev0 version and not the last stable release?

Might be, if we changed that string in source and did not yet push the current sources to the translate portal, we only do that once in a while.

raphaelm:
Thanks for the heads-up that French support is good enough that people self-host just because of that :P I'm deploying French on the SaaS platform now.

PR #1068, opened 2018-11-05
Implement #623 docker-compose

agowa338:
A working docker compose setup.

I also would recommend separating the docker-compose from the pretix repository into pretix/pretix-docker-compose for example, so this repository pretix/pretix can be included as git submodule and would not require copying files from other folders before build.

Implements: #623


mdehollander:
I tested this installation procedure on a cloud instance at hetzner and it works smoothly.

However, I get mixed languages in the notification emails when I enable more languages. This does not happen on the pretix.eu hosted version, so I think it has something to do with the configuration in this docker installation. More info here: https://github.com/pretix/pretix/issues/1058. Does anyone of you tried this as well and experiences the same?

agowa338:
@mdehollander Does this happen with the current version in the branch, too? I've not changed anything that would impact that. I've not changed anything in the current pretix codebase.

raphaelm:
I agree that this should be a separate repository.

I'm not sure if it's wise to have it under the pretix organization, since we're not capable of supporting it properly (since we don't user docker compose ourselves), but maybe we can add it nevertheless. I created an empty repository which you can do a PR against :) https://github.com/pretix/pretix-docker-compose

raphaelm:
Ah, also: If it isn't too much trouble, could you consider replacing mariadb with postgresql? It just is the better database ;) :elephant:

agowa338:
@raphaelm Can you please add a README.md to the new repository? Without it Github does not allow me to fork, nor to create a pull request.
Also I use mariadb, because I use it all the time and I don't know why postgresql should be better. Is there anything I missed futurewise, why postgresql should be prefered?

raphaelm:
> @raphaelm Can you please add a README.md to the new repository? Without it Github does not allow me to fork, nor to create a pull request.

Done!

> Also I use mariadb, because I use it all the time and I don't know why postgresql should be better. Is there anything I missed futurewise, why postgresql should be prefered?

It's a long, and obviously opinionated story. Historically, PostgreSQL has a reputation of favoring correctness (transaction safety!) over speed, where MySQL often made the trade-off the other way. PostgreSQL handles Unicode (Emojis!) a lot better and today, PostgreSQL tends to be faster (not in all cases, but as far as we see for pretix) and supports many advanced features better (like JSON data types, that we started using in pretix 2.2, but also a lot more stuff like geospatial support, functional indexes, …). At the same time, PostgreSQL has arguably better community support and is faster to add more features and improvements. Also, lots of new features are very different on MariaDB vs MySQL (like JSON fields) which makes it hard to support both of them. pretix will continue support MariaDB/MySQL going forward and we have no specific plans on stopping that, but we do see the possibility of adding features that will only work on PostgreSQL. Therefore, if we'd start giving people something like docker-compose that just installs pretix without much manual effort, I'd suggest we go for the recommended database option directly :)

Issue #1071, opened 2018-11-06
Payments box in order detail view too narrow


felixrindt:
When viewing orders in the control interface, the box showing payments is narrow on my screen.
![image](https://user-images.githubusercontent.com/5970076/48058305-5e77b100-e1b6-11e8-9740-21904251e75b.png)

Scrolling in this small box is cumbersome, so I'd recommend not having this half the width of the content area like this:

![image](https://user-images.githubusercontent.com/5970076/48058404-a1398900-e1b6-11e8-815f-d9db69b8539c.png)

(The content column is 1040 pixel wide on my screen.)

PR #1072, opened 2018-11-06
make payments and invoice address full width panels

felixrindt:
Solve #1071.

I'm not happy about how the invoice address panel is really wide now.

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/order/index.html
+++ b/src/pretix/control/templates/pretixcontrol/order/index.html
@@ -536,8 +536,10 @@
                         </div>
                     {% endif %}
                 </div>
+            </div>
+            <div class="row">
                 {% if request.event.settings.invoice_address_asked or order.invoice_address %}
-                    <div class="col-md-12">
+                    <div class="col-md-6">
                         <div class="panel panel-default">
                             <div class="panel-heading">
                                 <div class="pull-right">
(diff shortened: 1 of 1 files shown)


felixrindt:
Here are some screenshots:  
**Before**
image


**After**
image

Maybe we could make use of bootstrap different screen size classes, to only split e.g. for content widths above 1600 (or whatever constants there are in bootstrap). On a simulated 1080p screen, only refunds get a scrollbar in this situation:

image

Here not using half-width columns has the worse look:
image

I'm gonna update this PR with different `col` classes.

felixrindt:
I looked up the classes, the biggest  `col` class we have is `col-lg-xx`, which considers my screen to be large, so it's not big enough.

raphaelm:
Can we maybe put the invoice information next to the "Internal comment" box below that? That one also doesn't really need full width.

src/pretix/control/templates/pretixcontrol/order/index.html
```diff
@@ -388,7 +388,7 @@ <h3 class="panel-title">
             </div>
             {% eventsignal event "pretix.control.signals.order_info" order=order request=request %}
             <div class="row payments">
```
felixrindt: Is this `row` still needed now? What does the `payments` class do?

felixrindt:
Snuggly fits next to the comment panel, yes.

src/pretix/control/templates/pretixcontrol/order/index.html
raphaelm: The .payments class themes the ``dt`` element, see pretix/static/pretixbase/scss/_theme.scss

raphaelm: Since the subelements still use ``col``, even though it's ``col-md-12``, we should keep the ``row`` as well, although  we could drop both, probably.

raphaelm:
Thanks a lot!

PR #1084, opened 2018-11-11
Fix #369 -- Connect with PayPal

### LINKED ISSUE #369: Automate or document obtaining PayPal API credentials

pc-coholic:
New feature to offer something similar to Stripe Connect.

As of now, the eMail-Address of the payee is transmitted instead of the MerchantID. For a later iteration, it might be interesting to use that one instead as it should be less volatile than an email-address. However - for the time being - I was unable to get the MerchantID using the REST API.

### DIFF
--- a/src/pretix/plugins/paypal/payment.py
+++ b/src/pretix/plugins/paypal/payment.py
@@ -4,17 +4,22 @@ import urllib.parse
 from collections import OrderedDict
 
 import paypalrestsdk
+from paypalrestsdk.openid_connect import Tokeninfo
 from django import forms
 from django.contrib import messages
 from django.core import signing
 from django.http import HttpRequest
 from django.template.loader import get_template
+from django.urls import reverse
+from django.utils.crypto import get_random_string
+from django.utils.http import urlquote
 from django.utils.translation import ugettext as __, ugettext_lazy as _
 
 from pretix.base.decimal import round_decimal
-from pretix.base.models import OrderPayment, OrderRefund, Quota
+from pretix.base.models import Event, OrderPayment, OrderRefund, Quota
 from pretix.base.payment import BasePaymentProvider, PaymentException
 from pretix.base.services.mail import SendMailException
+from pretix.base.settings import SettingsSandbox
 from pretix.helpers.urls import build_absolute_uri as build_global_uri
 from pretix.multidomain.urlreverse import build_absolute_uri
 from pretix.plugins.paypal.models import ReferencedPayPalObject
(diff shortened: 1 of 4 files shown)


src/pretix/plugins/paypal/payment.py
```diff
@@ -56,24 +68,77 @@ def settings_form_fields(self):
...
+            if not self.settings.connect_user_id:
+                return (
+                    "<p>{}</p>"
+                    "<a href='{}' class='btn btn-primary btn-lg'>{}</a>"
+                ).format(
+                    _('To accept payments via PayPal, you will need an account at PayPal. By clicking on the '
+                      'following button, you can either create a new PayPal account connect pretix to an existing '
+                      'one.'),
+                    self.get_connect_url(request),
+                    _('Connect with <i class="fa fa-paypal"></i> PayPal')
```
raphaelm: Can we refactor this into
```
 _('Connect with {icon} PayPal').format(icon='<i class="fa fa-paypal"></i>')
```
to avoid confusion with translators?

src/pretix/plugins/paypal/payment.py
```diff
@@ -56,24 +68,77 @@ def settings_form_fields(self):
...
+            )
 
     def init_api(self):
-        paypalrestsdk.set_config(
-            mode="sandbox" if "sandbox" in self.settings.get('endpoint') else 'live',
-            client_id=self.settings.get('client_id'),
-            client_secret=self.settings.get('secret'))
+        if self.settings.connect_client_id:
+            paypalrestsdk.set_config(
+                mode="sandbox" if "sandbox" in self.settings.get('endpoint') else 'live',
```
raphaelm: Where is this ``endpoint`` setting set? As far as I can see it, it's neither part of the global settings nor of the event-level settings when OpenID Connect is in use?

src/pretix/plugins/paypal/payment.py
```diff
@@ -115,7 +184,13 @@ def checkout_prepare(self, request, cart):
                         "currency": request.event.currency,
                         "total": self.format_price(cart['total'])
                     },
-                    "description": __('Event tickets for {event}').format(event=request.event.name)
+                    "description": __('Event tickets for {event}').format(event=request.event.name),
+                    "payee": {
+                        "email": "" or request.event.settings.payment_paypal_connect_user_id,
```
raphaelm: ``"" or B`` does not make sense, it will always return ``B``. Did you mean ``… or ""``?

src/pretix/plugins/paypal/payment.py
pc-coholic: Yes, definitely. :-)

src/pretix/plugins/paypal/payment.py
pc-coholic: Again, agreed: Will flip those two around.

src/pretix/plugins/paypal/payment.py
pc-coholic: I guess this only worked for me with the sandbox env, because I had PayPal manually set up before making my changes. Welp...

But I am not 100% sure how we should handle the whole thing. So the following paragraphs might be a little bit confusing.

| Global Setting | Customer can choose to log into | If Customer logged into | Then Customer can choose to do transactions in |
|----------------|---------------------------------|-------------------------|------------------------------------------------|
| Sandbox        | Sandbox only                    | Sandbox                 | Sandbox only                                   |
| Production     | Sandbox or Production           | Sandbox                 | Sandbox only                                   |
| Production     | Sandbox or Production           | Production              | Production only                                |

Given this is more complex than Stripe, I am not 100% sure where we should offer the env-selection.

One issue seems to be, that the accounts in production/sandbox are not the same. So even if a customer logged into production, and then switched to sandbox, things would break.

So here is an idea: 
- Have a env-switch in  the global settings.
- If global == production, show env-switch to customer before pairing with PayPal.
- If global == sandbox, customer-env is forcibly set to sandbox before pairing with PayPal.
- After Pairing, event-level env cannot be changed. In order to switch env, customer has to un- and relink PayPal.

What are your thoughts on that? Too over-engineered?

PR #1094, opened 2018-11-19
Dockerfile improvements for layer caching and commands order

aruanoguate:
Some changes to the Dockerfile to provide the below benefits:

- Better readability of the packages installed on the apt-get instruction
  - Only one package name by line 
  - The packages are alphabetically ordered
- Better use of Docker cache during image build by installing from PIP before creating the layer with the source code (the PIP installations are less probable to change than the rest of the source code, so this layer will not be recreated with all the source code changes)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,26 @@
 FROM python:3.6
 
 RUN apt-get update && \
-    apt-get install -y git libxml2-dev libxslt1-dev python-dev python-virtualenv locales \
-      libffi-dev build-essential python3-dev zlib1g-dev libssl-dev gettext libpq-dev \
-      default-libmysqlclient-dev libmemcached-dev libjpeg-dev supervisor nginx sudo \
-	  --no-install-recommends && \
+    apt-get install -y --no-install-recommends \
+            build-essential \
+            default-libmysqlclient-dev \
+            gettext \
+            git \
+            libffi-dev \
+            libjpeg-dev \
+            libmemcached-dev \
+            libpq-dev \
+            libssl-dev \
+            libxml2-dev \
+            libxslt1-dev \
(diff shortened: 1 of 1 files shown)


raphaelm:
Hi,

thanks a lot!

> Better readability of the packages installed on the apt-get instruction

I fully agree, that is a nice change!

> Better use of Docker cache during image build by installing from PIP before creating the layer with the source code (the PIP installations are less probable to change than the rest of the source code, so this layer will not be recreated with all the source code changes)

I'm less sure about this one. Using the cache here also means that you will more likely end up with outdated packages in cases where requirements.txt does not contain specific version but just requires the last version of a package. This means that with your change, the docker container is (a) less reproducible since the output depends strongly on the state of your docker cache and (b) might contain outdated/vulnerable software. Since this Dockerfile is commonly used for production, but I don't know people using it for development, I'd personally favor security and stability over build time here. What do you think?

aruanoguate:
Thanks for the feedback. 

I see your point about outdated packages and, if you know that this Dockerfile is commonly used for production, I see two options:

- Get rid at all of this part of the change
- Update the documentation indicating that, in order to build this image for production environments, is recommended to send the `--no-cache=true` flag to the `docker build ` command. 

I ended up with a version like the one proposed in this PR because I was working on some changes that required constant rebuilds of the docker image and I have a very slow internet connection, that's why I felt tedious to re-download the packages from PIP each time I made a change to any file included in the repository, even if the change was unrelated to the requirements files, but probably this is not a common use case for most of the community.

Let me know your thoughts.

aruanoguate:
Hi,

I ended up removing the part related to the caching and only left the part about the readability. 

Let me know if you see any issue with this.

PR #1107, opened 2018-11-24
Fix #1066 -- Change installation tutorial to PostgreSQL

### LINKED ISSUE #1066: Change installation tutorial to show how to use PostgreSQL, not MySQL
PostgreSQL is our preferred database, so we should showcase it in our tutorials. To do this, you should try to install pretix following the tutorial, but with PostgreSQL, and change the tutorial where necessary.

https://docs.pretix.eu/en/latest/admin/installation/index.html

oocf:
This is the recommended database server so this documentation should use that

### DIFF
--- a/doc/admin/installation/docker_smallscale.rst
+++ b/doc/admin/installation/docker_smallscale.rst
@@ -58,12 +58,17 @@ Database
 --------
 
 Next, we need a database and a database user. We can create these with any kind of database managing tool or directly on
-our database's shell, e.g. for MySQL::
+our database's shell, e.g. for POSTGRESQL::
 
-    $ mysql -u root -p
-    mysql> CREATE DATABASE pretix DEFAULT CHARACTER SET utf8mb4 DEFAULT COLLATE utf8mb4_unicode_ci;
-    mysql> GRANT ALL PRIVILEGES ON pretix.* TO pretix@'localhost' IDENTIFIED BY '*********';
-    mysql> FLUSH PRIVILEGES;
+    $ psql postgres -U root
+    postgres=# CREATE ROLE pretixuser WITH LOGIN PASSWORD '*********'; 
+    postgres=# \du
+    postgres=# ALTER ROLE patrick CREATEDB; 
+    postgres=# \du 
+    postgres=# \q
+    $ psql postgres -U pretixuser
+    postgres=# CREATE DATABASE pretix;
+    postgres=# GRANT ALL PRIVILEGES ON DATABASE pretix TO pretixuser;
 
 Replace the asterisks with a password of your own. For MySQL, we will use a unix domain socket to connect to the
 database. For PostgreSQL, be sure to configure the interface binding and your firewall so that the docker container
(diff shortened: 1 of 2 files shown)


doc/admin/installation/docker_smallscale.rst
```diff
@@ -58,12 +58,17 @@ Database
 --------
 
 Next, we need a database and a database user. We can create these with any kind of database managing tool or directly on
-our database's shell, e.g. for MySQL::
-
-    $ mysql -u root -p
-    mysql> CREATE DATABASE pretix DEFAULT CHARACTER SET utf8mb4 DEFAULT COLLATE utf8mb4_unicode_ci;
-    mysql> GRANT ALL PRIVILEGES ON pretix.* TO pretix@'localhost' IDENTIFIED BY '*********';
-    mysql> FLUSH PRIVILEGES;
+our database's shell, e.g. for POSTGRESQL::
```
rixx: Please don't use ALLCAPS here. I think the preferred style is "PostgreSQL".

doc/admin/installation/docker_smallscale.rst
```diff
@@ -58,12 +58,17 @@ Database
...
+
+    $ psql postgres -U root
```
rixx: This seems like a convoluted way of creating a user and database – why are those "\du" lines included? Why do you need to sign out and then use postgres as the pretix user?

Either this:
```SQL
$ psql
postgres=# CREATE DATABASE yourdbname;
postgres=# CREATE USER youruser WITH ENCRYPTED PASSWORD 'yourpass';
postgres=# GRANT ALL PRIVILEGES ON DATABASE yourdbname TO youruser;
```
or even just this
```
sudo -u postgres -i
createuser -W <username>
createdb -O <username> <dbname>
```
should be sufficient, shouldn't it?

doc/admin/installation/docker_smallscale.rst
```diff
@@ -114,13 +119,14 @@ Fill the configuration file ``/etc/pretix/pretix.cfg`` with the following conten
     datadir=/data
 
     [database]
-    ; Replace mysql with postgresql_psycopg2 for PostgreSQL
-    backend=mysql
+    ; Replace postgresql_psycopg2 with mysql for MySQL
+    backend=postgresql_psycopg2
```
raphaelm: I know, the comment is misleading, but ``postgresql_psycopg2`` is actually deprecated by now in Django, so we can just use ``postgresql`` here instead.

doc/admin/installation/manual_smallscale.rst
```diff
@@ -50,12 +50,17 @@ Database
 --------
 
 Having the database server installed, we still need a database and a database user. We can create these with any kind
-of database managing tool or directly on our database's shell, e.g. for MySQL::
-
-    $ mysql -u root -p
-    mysql> CREATE DATABASE pretix DEFAULT CHARACTER SET utf8mb4 DEFAULT COLLATE utf8mb4_unicode_ci;
-    mysql> GRANT ALL PRIVILEGES ON pretix.* TO pretix@'localhost' IDENTIFIED BY '*********';
-    mysql> FLUSH PRIVILEGES;
+of database managing tool or directly on our database's shell, e.g. for POSTGRESQL:
```
raphaelm: This needs to be a double-colon (``::``) at the end of the line such that it is recognized as a code snippet.

raphaelm:
Replaced by 7219575b

PR #1112, opened 2018-12-04
Fix #785 -- Show availability in (sub)event list

### LINKED ISSUE #785: Show ticket availability in (sub)event list
#### Situation
Imagine an event series with many dates of which many are sold out. A potential customer now needs to click through every subevent to find one which is not sold out (in the lists or calendar view).

screenshot from 2018-02-23 18-11-07screenshot from 2018-02-23 18-23-28

That's not only tedious for the customer but might also lower conversion rate.

#### What to do?

This is not an easy question, since it's hard to decide when an event is considered 'sold out'.
I could imagine summing the availability of all products that provide admission for the event. If that sum is zero, the event could be considered sold out, although stuff like merchandise could still be available. Also we wouldn't want customers to not discover the waiting list.

Here's a mockup: screenshot from 2018-02-23 18-22-21
Maybe yellow would be a better color?

What do you think about this? What would be a good way to implement this?

raphaelm:
- [x] subevent list
- [x] event list
- [x] Lots of tests
- [ ] test reseller module
- [ ] get an idea how bad performance is
- [x] decide on (non-)admission products
- [ ] decide if caching is good enough, especially wrt cart expiration

### DIFF
--- a/doc/spelling_wordlist.txt
+++ b/doc/spelling_wordlist.txt
@@ -6,6 +6,7 @@ api
 auditability
 auth
 autobuild
+availabilities
 backend
 backends
 banktransfer
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -209,6 +209,8 @@ class EventMixin:
                 return res[0]
             elif res[0] > best:
                 best = res[0]
+        if not self.active_quotas:
+            return None
         return best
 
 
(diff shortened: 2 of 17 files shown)


raphaelm:
Meh. Turns out, this approach of going via the list of quotas is screwed: This will report back "tickets available" even if there aren't any because all items are blocked by a separate quota. Also, this will always report "tickets available" if there is *any* unlimited quota, which is something you might have just to group tickets together for vouchers.

Going via the list of products would solve this, but require more queries and much more complex code :(

Maybe we can just live with it since this error still fails in the more acceptable direction, but it certainly isn't great.

PR #1120, opened 2018-12-11
Fix #1119 - Proper indent and pluralisation

### LINKED ISSUE #1119: Backend-Signals are not displayed in documentation
I'm not sure why, but when looking at https://docs.pretix.eu/en/latest/development/api/general.html#backend, it seems to me like quite a few signals are not being displayed here...

Comparing to https://github.com/pretix/pretix/blob/master/doc/development/api/general.rst#backend, for example all the `html` and `navbar`-signals are missing...

pc-coholic:
Well - that was a surprisingly easy fix... Just add some indents ;-)

### DIFF
--- a/doc/development/api/general.rst
+++ b/doc/development/api/general.rst
@@ -26,7 +26,7 @@ Frontend
 --------
 
 .. automodule:: pretix.presale.signals
-   :members: html_head, html_footer, footer_links, front_page_top, front_page_bottom, fee_calculation_for_cart, contact_form_fields, question_form_fields, checkout_confirm_messages, checkout_confirm_page_content, checkout_all_optional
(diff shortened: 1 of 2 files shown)


src/pretix/settings.py
```diff
@@ -150,6 +150,7 @@
 EMAIL_USE_TLS = config.getboolean('mail', 'tls', fallback=False)
 EMAIL_USE_SSL = config.getboolean('mail', 'ssl', fallback=False)
 EMAIL_SUBJECT_PREFIX = '[pretix] '
+EMAIL_BACKEND = 'django.core.mail.backends.dummy.EmailBackend'
```
rixx: :mailbox_with_no_mail:

pc-coholic: Öh.... Too much commitment here :-P

raphaelm:
Thanks!

PR #1121, opened 2018-12-11
Add media-src CSP to middleware

pc-coholic:
As of now, pretix does not have any default Content-Security-Policy for `media-src`es.

This PR will take care of this - and help me for a future plugin development :)


raphaelm:
I'm… looking forward to seeing what this is for :D

PR #1142, opened 2019-01-20
Fix #1111 -- Duplicate voucher warning

### LINKED ISSUE #1111: If a voucher was used several times in one order, show order only once
pretix warns you if a voucher was used in an order when viewing the voucher. If the voucher was redeemed several times in the same order, you'll get repeated links to the same order, which is amusing, but not necessary.

maartenberg:
Adds a new method to Voucher that selects all distinct orders containing
a position where the Voucher has been used, and changes the Voucher
detail view to use this method for the warning. Includes regression test.


raphaelm:
Great work, thanks a lot! :heart:

PR #1143, opened 2019-01-20
Updates the dependency versions for flake8

bockstaller:
Following the setup instructions resulted in the following error message:
```
pip3 install -r requirements.txt -r requirements/dev.txt
[...]
flake8 3.6.0 has requirement pyflakes<2.1.0,>=2.0.0, but you'll have pyflakes 1.1.0 which is incompatible.
```

**Fix:**
Pretix doesn't version pin flake8, but it's two requirements pep8 and pyflakes.
The current release flake8 3.6.0 (http://flake8.pycqa.org/en/latest/release-notes/3.6.0.html#new-dependency-information) updates the version requirements for pyflakes and pycodestyle (former pep8).
I've transfered the version range into setup.py and requirements/dev.txt.

I am now able to rerun the dev-setup tutorial without errors.

### DIFF
--- a/src/requirements/dev.txt
+++ b/src/requirements/dev.txt
@@ -1,8 +1,8 @@
 django-debug-toolbar==1.9.1
 sqlparse==0.2.1  # pinned due to difficulties with django-debug-toolbar
 # Testing requirements
-pep8==1.5.7  # exact requirement by flake8 2.4.0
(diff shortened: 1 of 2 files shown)


raphaelm:
Well, apparently those versions do not cause problems any more. Thanks!

Issue #1091, opened 2018-11-16
useless hint about options


j-herden:
If an add-on has only one option I see a hit, that does not make sense for me. I would suggest to hide this hint, if there is only one option available.
See screenshot:
![grafik](https://user-images.githubusercontent.com/2026587/48618497-d786bd80-e999-11e8-90cf-016d2b8e40ce.png)

raphaelm:
Great suggestion!

PR #1147, opened 2019-01-22
Adjust message if there's only one addon

FlaviaBastos:
Relates to #1091 

I'm not sure this is a good approach... Could there be a case for the condition in line 44 or that should be dropped?


FlaviaBastos:
Thank you for the feedback! I will take a look and test later on today.

FlaviaBastos:
I made this change but now the alignment of the form becomes more obvious. I'm sorry I know very little about css and it seems that this is controlled by bootstrap form. Any suggestions on how to improve it?
screenshot from 2019-01-23 05-50-36

raphaelm:
Thanks, looks great (feature-wise). Layout-wise, you're right, but we can't easily fix it since in the case of a add-on product with variations there *is* something to the left of the radiobox, so if we'd move it to the left, it would be unaligned in the case that we have both add-on products with and without variations. I'll give this some thought and see if I can come up with a good solution.

PR #1151, opened 2019-01-27
Fix UI crash in log entry view when `logentry.data` is list of dict

krav:
When e.g. creating vouchers with the `voucher/batch_create` api, `logentry.data` is a list and not a dict.


raphaelm:
Hi, thanks for reporting! I'm surprised this never came up. I'd prefer it to be a dictionary everywhere for consistency and view this as a bug in the API (fixed 38f4d6dd) and enforce it going forward (see 3b85a156).

For your case, since it seems to be a very rare edge case, I'd ask to fix this manually with something like ``UPDATE pretixbase_logentry SET data = CONCAT('{"data": ', CONCAT(data, '}')) WHERE data LIKE '[%]'`` (untested!).

PR #1162, opened 2019-02-10
Enable nl2br plugin for Markdown rendering

ahus1:
The frontpage text is already markdown, and will receive its formatting via the rich_text filter.
When applying the additional linebreaksbr filter, it will add unnecessary blank lines.

I'm using the hosted pretix version. 

Test for frontpage text: 

````
Test 

* test1
* test2
````

Before (screenshot): 

---

before

----

After (screenshot): 

----
after

----

### DIFF
--- a/src/pretix/base/templatetags/rich_text.py
+++ b/src/pretix/base/templatetags/rich_text.py
@@ -84,7 +84,7 @@ def markdown_compile(source):
             source,
             extensions=[
                 'markdown.extensions.sane_lists',
-                # 'markdown.extensions.nl2br',  # TODO: Enable, but check backwards-compatibility issues e.g. with mails
+                'markdown.extensions.nl2br'
             ]
         ),
         tags=ALLOWED_TAGS,
--- a/src/pretix/presale/templates/pretixpresale/event/index.html
+++ b/src/pretix/presale/templates/pretixpresale/event/index.html
@@ -102,7 +102,7 @@
 
     {% if frontpage_text and not cart_namespace %}
         <div>
-            {{ frontpage_text|rich_text|linebreaksbr }}
+            {{ frontpage_text|rich_text }}
         </div>
     {% endif %}
 

### CODE REVIEW

**Comment by ahus1** [comment]
This issue has been introduced in #1049 commit 94be46ffdb2df3601fffde8a3411a5d1119398e7 and was present from pretix v2.2.0

**Comment by raphaelm** [comment]
As I mentioned via email already, I would like to find out *why* we added this before I remove it again, because a distant memory says it was to prevent something else looking odd.

However, I have no idea how it ended up in the commit you referenced, that is really weird, must have slipped into the wrong development branch at the time.

I believe the reasoning was probably that the field was previously unfriendly to people who know nothing about markdown, they'd enter
```
Line1
Line2
```
and end up with
```
Line1 Line2
```
…which is inconsistent with Markdown formatting, yes, but what non-technical users excpect. I agree that this should not affect your case (lists) and should rather be "solved" within the rich_text filter to be used consistently throughout the system. There's a similar problem with email rendering (where we use markdown as well) where even our default templates contain
```
Best regards
Your event team
```
which ends up on one line and looks ugly.

Do you happen to have a good idea how to solve this? Otherwise I need to come up with one on my own ;)

**Comment by raphaelm** [comment]
Heh, I just saw 
https://github.com/pretix/pretix/blob/afd766999c4af13f8b003cf242ad678798ec9405/src/pretix/base/templatetags/rich_text.py#L87

which solves this problem but we were anxious about activating it for the same reason. I'd say we activate this now, remove the template filter and test around if anything breaks. I'm a bit busy atm, but maybe I can dive into this tomorrow or Monday.

**Comment by ahus1** [comment]
I've applied the change and tested your suggestion.

This works for me. I like that this will handle all richtext consistently. It's up to you to determine if this will work for other users as well :-) I'm looking forward to the change, but don't rush!

**Comment by raphaelm** [comment]
Bad news, I found a major incompatibility. In most places on the web, I think we can tolerate the changes, but in email, there's a problem. Our default email templates look like this:

```
Hello,

we successfully received your order for {event} with a total value
of {total_with_currency}. Please complete your payment before {date}.

{payment_info}

You can change your order details and view the status of your order at
{url}

Best regards,
Your {event} team
```

They got a lot of explicit newlines. In the "Best regards," part, this change would improve things, but the paragraphs above would start looking really silly in the HTML emails with lines breaking in very weird places.

Currently, I don't really know how to fix this. Removing the linebreaks from these templates would be the best solution (because with the placeholders, they do not make any sense in the first place), but is hard for two reasons:

* 90%+ of events will, either on purpose or by accident, have these defaults overridden with copies of these values or their own texts that share the same problem. I'm not sure if there's a reasonable way to automatically fix them
* It breaks the translation files which will need to be fixed manually

Unless you or someone else comes up with something more clever, I see two options:

* We make this change conditional in the rich_text filter and enable it everywhere *but* for emails (doesn't solve the Best Regards problem)
* We make this change everywhere but automatically convert all email templates by removing single newlines, except where a heuristical approach believes it should stay (i.e. only at the end, before URLs, …)

raphaelm:
Turns out, email rendering isn't actually using rich_text at the moment, so even though it would nice to fix email rendering, this PR doesn't touch it.

PR #1186, opened 2019-02-20
Filter sensitive keys from log-messages

pc-coholic:


src/pretix/base/models/base.py
```diff
@@ -77,6 +77,13 @@ def log_action(self, action, data=None, user=None, api_token=None, auth=None, sa
 
         logentry = LogEntry(content_object=self, user=user, action_type=action, event=event, **kwargs)
         if isinstance(data, dict):
+            sensitivekeys = ['password', 'smtp_password', 'payment_stripe_secret_key', 'payment_paypal_secret',
+                             'payment_mollie_api_key', 'payment_sofort_api_key', 'payment_wirecard_secret']
```
raphaelm: Maybe instead of the plugin-specific things just ``api_key`` and ``secret``, if we do substring matching anyways?

src/pretix/base/models/base.py
```diff
@@ -77,6 +77,13 @@ def log_action(self, action, data=None, user=None, api_token=None, auth=None, sa
...
+
+            for k, v in data.items():
+                if k in sensitivekeys:
```
raphaelm: I propose a ``and v``, so we log empty strings and null values correctly

PR #1217, opened 2019-03-18
Refs #654 -- Add writable API for subevents

pc-coholic:
- [x] Write operations for subevents
- [x] Tests
- [x] Documentation

### DIFF
--- a/doc/api/resources/subevents.rst
+++ b/doc/api/resources/subevents.rst
@@ -123,16 +123,38 @@ Endpoints
       POST /api/v1/organizers/bigevents/events/sampleconf/subevents/ HTTP/1.1
       Host: pretix.eu
       Accept: application/json, text/javascript
+      Content: application/json
+
+      {
+        "name": {"en": "First Sample Conference"},
+        "active": false,
+        "date_from": "2017-12-27T10:00:00Z",
+        "date_to": null,
+        "date_admission": null,
+        "presale_start": null,
+        "presale_end": null,
+        "location": null,
+        "item_price_overrides": [
+          {
+            "item": 2,
+            "price": "12.00"
+          }
+        ],
+        "variation_price_overrides": [],
+        "meta_data": {}
+      }
+
 
    **Example response**:
 
    .. sourcecode:: http
 
-      HTTP/1.1 200 OK
+      HTTP/1.1 201 Created
       Vary: Accept
       Content-Type: application/json
 
       {
+        "id": 1,
         "name": {"en": "First Sample Conference"},
         "active": false,
         "date_from": "2017-12-27T10:00:00Z",
@@ -210,7 +232,9 @@ Endpoints
 
 .. http:patch:: /api/v1/organizers/(organizer)/events/(event)/subevents/(id)/
 
-   Updates a sub-event, identified by its ID.
+   Updates a sub-event, identified by its ID. You can also use ``PUT`` instead of ``PATCH``. With ``PUT``, you have to
+   provide all fields of the resource, other fields will be reset to default. With ``PATCH``, you only need to provide
+   the fields that you want to change.
 
    Permission required: "Can change event settings"
 
@@ -218,9 +242,20 @@ Endpoints
 
    .. sourcecode:: http
 
-      GET /api/v1/organizers/bigevents/events/sampleconf/subevents/1/ HTTP/1.1
+      PATCH /api/v1/organizers/bigevents/events/sampleconf/subevents/1/ HTTP/1.1
       Host: pretix.eu
       Accept: application/json, text/javascript
+      Content: application/json
+
+      {
+        "name": {"en": "New Subevent Name"},
+        "item_price_overrides": [
+          {
+            "item": 2,
+            "price": "23.42"
+          }
+        ],
+      }
 
    **Example response**:
 
@@ -254,7 +289,7 @@ Endpoints
    :param organizer: The ``slug`` field of a valid organizer
    :param event: The ``slug`` field of the main event
    :param id: The ``id`` field of the sub-event to update
-   :statuscode 201: no error
+   :statuscode 200: no error
    :statuscode 400: The sub-event could not be created due to invalid submitted data.
    :statuscode 401: Authentication failure
    :statuscode 403: The requested organizer/sub-event does not exist **or** you have no permission to update this resource.
--- a/doc/api/resources/subevents.rst
+++ b/doc/api/resources/subevents.rst
@@ -45,6 +45,8 @@ meta_data                             dict                       Values set for 
 
    The ``event`` field has been added, together with filters on the list of dates and an organizer-level list.
 
+.. versionchanged:: 2.6
+   The write operations ``POST``, ``PATCH``, ``PUT``, and ``DELETE`` have been added.
 
 Endpoints
 ---------
@@ -103,10 +105,60 @@ Endpoints
    :query is_past: If set to ``true`` (``false``), only events that are over are (not) returned.
    :query ends_after: If set to a date and time, only events that happen during of after the given time are returned.
    :param organizer: The ``slug`` field of a valid organizer
-   :param event: The ``slug`` field of the event to fetch
+   :param event: The ``slug`` field of the main event
    :statuscode 200: no error
    :statuscode 401: Authentication failure
    :statuscode 403: The requested organizer does not exist **or** you have no permission to view it.
+
+.. http:post:: /api/v1/organizers/(organizer)/events/(event)/subevents/
+
+   Creates a new subevent.
+
+   Permission required: "Can create events"
+
+   **Example request**:
+
+   .. sourcecode:: http
+
+      POST /api/v1/organizers/bigevents/events/sampleconf/subevents/ HTTP/1.1
+      Host: pretix.eu
+      Accept: application/json, text/javascript
+
+   **Example response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Vary: Accept
+      Content-Type: application/json
+
+      {
+        "name": {"en": "First Sample Conference"},
+        "active": false,
+        "date_from": "2017-12-27T10:00:00Z",
+        "date_to": null,
+        "date_admission": null,
+        "presale_start": null,
+        "presale_end": null,
+        "location": null,
+        "item_price_overrides": [
+          {
+            "item": 2,
+            "price": "12.00"
+          }
+        ],
+        "variation_price_overrides": [],
+        "meta_data": {}
+      }
+
+
+   :param organizer: The ``slug`` field of a valid organizer
+   :param event: The ``slug`` field of the main event
+   :statuscode 201: no error
+   :statuscode 400: The sub-event could not be created due to invalid submitted data.
+   :statuscode 401: Authentication failure
+   :statuscode 403: The requested organizer does not exist **or** you have no permission to create this resource.
+
 
 .. http:get:: /api/v1/organizers/(organizer)/events/(event)/subevents/(id)/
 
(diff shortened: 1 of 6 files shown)


src/pretix/api/serializers/event.py
```diff
@@ -202,6 +202,87 @@ class Meta:
...
+
+        for item_price_override_data in item_price_overrides_data:
+            item, created = SubEventItem.objects.get_or_create(subevent=subevent, item=item_price_override_data['item'])
+            item = SubEventItem(id=item.id, subevent=subevent, **item_price_override_data)
+            item.save()
+        for variation_price_override_data in variation_price_overrides_data:
+            variation, created = SubEventItemVariation.objects.get_or_create(subevent=subevent, variation=variation_price_override_data['variation'])
+            variation = SubEventItemVariation(id=variation.id, subevent=subevent, **variation_price_override_data)
+            variation.save()
+
```
raphaelm: Remember deleting ones that have been present before but are no longer :) I recommend to fetch all existing SubEventItem(Variation)s before the loop into a dictionary mapping item IDs to SubEventItem objects, then doing the loop and .pop()-ing all objects from the dictionary that already exists, then deleting all that are still in the dictionary. This way, you can potentially save a database lookups at the same time!

src/pretix/api/views/event.py
```diff
@@ -240,6 +241,42 @@ def get_queryset(self):
...
+        serializer.instance.log_action(
+            'pretix.subevent.added',
+            user=self.request.user,
+            auth=self.request.auth,
+            data=self.request.data
+        )
+
+    def perform_destroy(self, instance):
+        if not instance.allow_delete():
+            raise PermissionDenied('The event can not be deleted as it already contains orders. Please set \'live\''
```
raphaelm: For subevents, this should reference the ``active`` property, not ``live`` status of the event :)

src/pretix/api/views/event.py
```diff
@@ -217,9 +217,10 @@ def is_future_qs(self, queryset, name, value):
             return queryset.exclude(expr)
 
 
-class SubEventViewSet(ConditionalListView, viewsets.ReadOnlyModelViewSet):
+class SubEventViewSet(ConditionalListView, viewsets.ModelViewSet):
     serializer_class = SubEventSerializer
     queryset = ItemCategory.objects.none()
+    permission_classes = (EventCRUDPermission,)
```
raphaelm: You don't need permission_classes (this is special for events as their creation works differently), you can just do:
```
    write_permission = 'can_change_event_settings'
```

src/pretix/api/serializers/event.py
```diff
@@ -202,6 +202,87 @@ class Meta:
                   'presale_start', 'presale_end', 'location', 'event',
                   'item_price_overrides', 'variation_price_overrides', 'meta_data')
 
+    def validate(self, data):
+        data = super().validate(data)
+
+        full_data = self.to_internal_value(self.to_representation(self.instance)) if self.instance else {}
+        full_data.update(data)
```
raphaelm: It's not a security issue in this case, but just to be safe we should validate ``item_price_overrides`` and ``variation_price_overrides`` and check that all IDs are actually items *within the correct event*. You can look at quotas for inspiration.

doc/api/resources/subevents.rst
```diff
@@ -103,11 +105,61 @@ Endpoints
...
+
+   **Example response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Vary: Accept
+      Content-Type: application/json
+
+      {
```
raphaelm: The response will probably contain an ``id``

doc/api/resources/subevents.rst
```diff
@@ -149,13 +201,93 @@ Endpoints
...
+
+   Updates a sub-event, identified by its ID.
+
+   Permission required: "Can change event settings"
+
+   **Example request**:
+
+   .. sourcecode:: http
+
+      GET /api/v1/organizers/bigevents/events/sampleconf/subevents/1/ HTTP/1.1
```
raphaelm: This should be PATCH and the example should have a request body

doc/api/resources/subevents.rst
```diff
@@ -103,11 +105,61 @@ Endpoints
...
+
+   Permission required: "Can create events"
+
+   **Example request**:
+
+   .. sourcecode:: http
+
+      POST /api/v1/organizers/bigevents/events/sampleconf/subevents/ HTTP/1.1
+      Host: pretix.eu
+      Accept: application/json, text/javascript
```
raphaelm: We should add an example body to make clearer how to use this

doc/api/resources/subevents.rst
```diff
@@ -103,11 +105,61 @@ Endpoints
...
+
+   **Example response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
```
raphaelm: POST usually returns ``201 Created`` in our APIs

doc/api/resources/subevents.rst
```diff
@@ -149,13 +201,93 @@ Endpoints
...
+   :param organizer: The ``slug`` field of a valid organizer
+   :param event: The ``slug`` field of the main event
+   :param id: The ``id`` field of the sub-event to fetch
    :statuscode 200: no error
    :statuscode 401: Authentication failure
    :statuscode 403: The requested organizer/event does not exist **or** you have no permission to view it.
 
+.. http:patch:: /api/v1/organizers/(organizer)/events/(event)/subevents/(id)/
+
+   Updates a sub-event, identified by its ID.
```
raphaelm: For completeness, this should have the text

```
You can also use ``PUT`` instead of ``PATCH``. With ``PUT``, you have to provide all fields of
the resource, other fields will be reset to default. With ``PATCH``, you only need to provide the fields that you
want to change.
```

doc/api/resources/subevents.rst
```diff
@@ -149,13 +201,93 @@ Endpoints
...
+          }
+        ],
+        "variation_price_overrides": [],
+        "meta_data": {}
+      }
+
+   :param organizer: The ``slug`` field of a valid organizer
+   :param event: The ``slug`` field of the main event
+   :param id: The ``id`` field of the sub-event to update
+   :statuscode 201: no error
```
raphaelm: 200, not 201

src/pretix/api/serializers/event.py
```diff
@@ -202,6 +202,105 @@ class Meta:
...
+        data = super().validate(data)
+        event = self.context['request'].event
+
+        full_data = self.to_internal_value(self.to_representation(self.instance)) if self.instance else {}
+        full_data.update(data)
+
+        Event.clean_dates(data.get('date_from'), data.get('date_to'))
+        Event.clean_presale(data.get('presale_start'), data.get('presale_end'))
+
+        SubEvent.clean_items(event, [item['item'] for item in full_data.get('subeventitem_set')])
```
raphaelm: I haven't tested it, but I have the feeling that ``validate`` runs before the serializer-level variation, so this might throw a KeyError in case anyone submits a ``subevent_item`` without a ``item`` key, leading to a ``500`` instead of ``400``.

src/pretix/api/serializers/event.py
```diff
@@ -202,6 +202,105 @@ class Meta:
...
+
+        for item_price_override_data in item_price_overrides_data:
+            if item_price_override_data['item'] in existing_item_overrides:
+                id = existing_item_overrides.pop(item_price_override_data['item'])
+                SubEventItem(id=id, subevent=subevent, **item_price_override_data).save()
+            else:
+                SubEventItem(subevent=subevent, **item_price_override_data).save()
+
+        for id in existing_item_overrides.values():
+            SubEventItem.objects.get(id=id).delete()
```
raphaelm: We can do this quicker with ``SubEventItem.objects.filter(id__in=existing_item_overrides.values()).delete()``

src/pretix/api/serializers/event.py
```diff
@@ -202,6 +202,105 @@ class Meta:
...
+        subevent = super().update(instance, validated_data)
+
+        existing_item_overrides = {item.item: item.id for item in SubEventItem.objects.filter(subevent=subevent)}
+
+        for item_price_override_data in item_price_overrides_data:
+            if item_price_override_data['item'] in existing_item_overrides:
+                id = existing_item_overrides.pop(item_price_override_data['item'])
+                SubEventItem(id=id, subevent=subevent, **item_price_override_data).save()
+            else:
+                SubEventItem(subevent=subevent, **item_price_override_data).save()
```
raphaelm: In general, instead of reconstructing the model object manually, I'd recommend storing `{item.item: item}` in the dictionary and using that object directly, since it' less error-prone if we ever extend that model. However, in this case, it shouldn't really matter. If you keep it that we, you can probably even simplify:

```

                id = existing_item_overrides.pop(item_price_override_data['item'], None)
                SubEventItem(id=id, subevent=subevent, **item_price_override_data).save()

```
No need for the if/else :)

src/pretix/api/views/event.py
```diff
@@ -240,6 +241,42 @@ def get_queryset(self):
...
+            'pretix.subevent.added',
+            user=self.request.user,
+            auth=self.request.auth,
+            data=self.request.data
+        )
+
+    def perform_destroy(self, instance):
+        if not instance.allow_delete():
+            raise PermissionDenied('The event can not be deleted as it already contains orders. Please set \'active\''
+                                   ' to false to hide the event and take the shop offline instead.')
```
raphaelm: I think this is still unclear in terms of terminology, how about ``The subevent can not be deleted as it has already been used in orders. Please set \'active\' to false instead to hide it from users.``

doc/api/resources/subevents.rst
pc-coholic: Regression of the events-resource documentation. Fixed it there, too.

src/pretix/api/serializers/event.py
pc-coholic: Do you have a proposal how to handle this?

I could probably just extend it to 
```[item['item'] if 'item' in item else None for item in full_data.get('subeventitem_set')]``` and have `clean_items()` skip those entries... But if there is a better way, I'd like to learn it :)

Comment (pc-coholic) [reply src/pretix/api/serializers/event.py]: Nevermind, I actually worked around this by filtering incomplete objects in a new `validate_item_price_overrides`.

**Comment by pc-coholic** [comment]
Sorry for the mess in the documentation... Too much copy/paste at work :)

File: src/tests/api/test_subevents.py [hunk 15]
```diff
@@ -50,6 +109,242 @@ def test_subevent_list(token_client, organizer, event, subevent):
...
+            "variation_price_overrides": [],
+            "meta_data": {
+                "type": "Workshop"
+            },
+        },
+        format='json'
+    )
+    assert resp.status_code == 201
+    assert not organizer.events.get(slug="dummy").subevents.get(id=resp.data['id']).active
+    assert organizer.events.get(slug="dummy").subevents.get(id=resp.data['id']).meta_values.filter(
```
raphaelm: Just a style thing, you can just use the ``event`` fixture variable, no need to get it from the database. Not a real problem, though ;)

src/tests/api/test_subevents.py
```diff
@@ -50,6 +109,242 @@ def test_subevent_list(token_client, organizer, event, subevent):
...
+            "date_from": "2018-12-27T10:00:00Z",
+            "date_to": "2018-12-28T10:00:00Z",
+        },
+        format='json'
+    )
+    assert resp.status_code == 200
+    event = Event.objects.get(organizer=organizer.pk, slug=event.slug)
+    assert organizer.events.get(slug=event.slug).subevents.get(id=resp.data['id']).meta_values.filter(
+        property__name=meta_prop.name, value="Workshop"
+    ).exists()
```
raphaelm: What exactly is this first part testing? It's setting dates – and then checking for meta information.

src/tests/api/test_subevents.py
```diff
@@ -50,6 +109,242 @@ def test_subevent_list(token_client, organizer, event, subevent):
...
+                    "price": "99.99"
+                }
+            ],
+        },
+        format='json'
+    )
+    assert resp.status_code == 400
+    assert resp.content.decode() == '{"item_price_overrides":[{"item":["Invalid pk \\"123\\" - object does not exist."]}]}'
+
+
```
raphaelm: I know it's annoying, but I'd like to request a few more tests there. I believe the item_price_override and variation_price_override are the most complex and most unique thing in this whole PR and I'd like to think it's worth testing that well. By that I mean that the tests should, for both parameters, cover all three major cases of

* a new price override being added in a PATCH command
* a price override being deleted in a PATCH command
* a price override being changed to a new price in a PATCH command

All of these could go wrong in different ways if we mess the logic up one day and it would be nice to notice that ;)

src/tests/api/test_subevents.py
```diff
@@ -50,6 +109,242 @@ def test_subevent_list(token_client, organizer, event, subevent):
...
+                {
+                    "item": 123,
+                    "price": "99.99"
+                }
+            ],
+        },
+        format='json'
+    )
+    assert resp.status_code == 400
+    assert resp.content.decode() == '{"item_price_overrides":[{"item":["Invalid pk \\"123\\" - object does not exist."]}]}'
```
raphaelm: Could you also add a test for an item that doesn't belong to the event? you can look in ``test_items.py``, there are fixtures ``event2`` and ``item2`` you can copy over for exactly that purpose.

pc-coholic:
It actually does work - test committed here :)

Issue #1170, opened 2019-02-14
check-in list displays wrong email address


j-herden:
When I display a check-in list in the browser, I will not see email address of the participant, but the one from the user who will get the invoice.
However, when I export the same check-in list, I will get the correct email addresses in the exportet data.

I would think that this is a bug and the online visible check-in lists should show the email addresses of the participants too.

abramzog:
Hi I'm new to open source contribution -  would like to get started on this.  Is it OK?

raphaelm:
Sure, go ahead!

PR #1231, opened 2019-03-25
Added email to exported pdf

abramzog:
Related to #1170. Added email column to exported PDF and adjusted the columns widths accordingly.

### DIFF
--- a/src/pretix/plugins/checkinlists/exporters.py
+++ b/src/pretix/plugins/checkinlists/exporters.py
@@ -180,7 +180,7 @@ class PDFCheckinList(ReportlabExportMixin, CheckInListMixin, BaseExporter):
         headlinestyle.fontName = 'OpenSansBd'
         colwidths = [3 * mm, 8 * mm, 8 * mm] + [
             a * (doc.width - 8 * mm)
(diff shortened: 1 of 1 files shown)


raphaelm:
Hi @abramzog! Thank you so much for taking this on! I'm afraid there has been a misunderstanding though, the issue is not about including the email in the PDF. That could be useful as well, but I don't think it's worth it since the PDF has very limited space already and when you're printing lists out, you usually do not need email addresses.

The issue is about the check-in list view that you find at the "Check-in" menu point of an event in the pretix backend. You can find the template in ``pretix/control/templates/pretixcontrol/checkin/index.html``. Here, issue #1170 asks not to display the e-mail address of the *order*, but of the *attendee*, if it exists, e.g. 
```
<td>
{% if e.addon_to and e.addon_to.attendee_email %}
    {{ e.addon_to.attendee_email }}
{% elif e.attendee_email %}
    {{ e.attendee_email }}
{% else %}
    {{ e.order.email }}
{% endif %}
</td>
```
instead of
```
<td>{{ e.order.email }}</td>
```

PR #1250, opened 2019-04-10
Fix #1247 -- Allow team invites to be resent

### LINKED ISSUE #1247: Resend team invite
When inviting users to a team of an Organizer Account it would be useful to be able to resend the email with the invitation in case it got lost somehow. Currently I have to revoke the invitation and send it out again.

Sohalt:
The functionality is working, but it currently looks horrible, because the button gets rendered as a button (without any styling). It should be possible to remove the button styling, but that would probably require custom CSS for that single case. I thought it would be nice to have the resend functionality close to the envelope icon indicating, that an invite is still pending. Alternatively one could embrace the button being a button and style it like the "remove" button on the right and also move it to the right.

### DIFF
--- a/src/pretix/control/logdisplay.py
+++ b/src/pretix/control/logdisplay.py
@@ -389,6 +389,9 @@ def pretixcontrol_logentry_display(sender: Event, logentry: LogEntry, **kwargs):
     if logentry.action_type == 'pretix.team.invite.created':
         return _('{user} has been invited to the team.').format(user=data.get('email'))
 
+    if logentry.action_type == 'pretix.team.invite.resent':
+        return _('Invite for {user} has been resent.').format(user=data.get('email'))
+
     if logentry.action_type == 'pretix.team.invite.deleted':
         return _('The invite for {user} has been revoked.').format(user=data.get('email'))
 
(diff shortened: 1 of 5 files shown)


raphaelm:
Hi @Sohalt, thank you so much for contributing! Just so you know, I'm currently on a conference and can't promise I'll be able to review this properly this week, but I'll try my best.

Sohalt:
Thank you so much for creating pretix!

I don't know, why the tests still fail on the CI, locally they all pass. Maybe you have an idea. But take your time and enjoy Djangocon!

raphaelm:
You need to clear out the "fake email outbox" at the beginning of the test, e.g. ``djmail.outbox = []``. Otherwise, things from previous tests might be in there depending on the order of test execution.

Sohalt:
I expected something like this, but I'm clearing the outbox on both of my tests.
The failing test is one that I did not touch. And in the `setUp` for the failing test the outbox gets cleared.

raphaelm:
Weird, okay, I'll need to look into it deeper then.

Sohalt:
Thanks for the quick response though! As said, take your time!

raphaelm:
I wasn't able to reproduce this locally, either. Since Travis IIRC does not test your branch, but the hypothetical merge result, I took the liberty of rebasing and force-pushing your branch, let's see if that helps.

PR #1252, opened 2019-04-14
[WIP] Banktransfer code prefix (Fix #529)

### LINKED ISSUE #529: Custom prefix for bank transfer references
It defaults to the slug, but why should it need to be the slug? E.g. in case of a long slug, this might be impractical.

tynsh:
Add an optional setting in the banktransfer plugin to change the code
prefix from the default value of the event slug to an arbitary string.

This is a work is progress, so there are no tests yet. I'd just like to have some feedback if my approach makes sense :-)

This implementation does retain codes used in previous orders. I've added a function, that takes an organizer or event and returns a map of valid reference code prefixes to the actual event. This is used during transfer processing and in the review functionality for not matched transfers.

### DIFF
--- a/src/pretix/plugins/banktransfer/payment.py
+++ b/src/pretix/plugins/banktransfer/payment.py
@@ -118,6 +118,16 @@ class BankTransfer(BasePaymentProvider):
     def settings_form_fields(self):
         d = OrderedDict(
             list(super().settings_form_fields.items()) + list(BankTransfer.form_fields().items()) + [
+                ('code_prefix', forms.CharField(
+                    label=_('Bank transfer code prefix.'),
+                    help_text=_('This value is prepended to the unique order code.'),
+                    widget=forms.TextInput(
+                        attrs={
+                            'placeholder': self.event.slug.upper(),
+                        },
+                    ),
+                    required=False
+                )),
                 ('omit_hyphen', forms.BooleanField(
                     label=_('Do not include a hypen in the payment reference.'),
                     help_text=_('This is required in some countries.'),
@@ -212,10 +222,13 @@ class BankTransfer(BasePaymentProvider):
         return template.render(ctx)
 
     def _code(self, order):
+        prefix = self.settings.get('code_prefix', as_type=str)
+        if not prefix:
+            prefix = self.event.slug.upper()
         if self.settings.get('omit_hyphen', as_type=bool):
-            return self.event.slug.upper() + order.code
+            return prefix + order.code
         else:
-            return order.full_code
+            return prefix + '-' + order.code
 
     def shred_payment_info(self, obj):
         if not obj.info_data:
--- a/src/pretix/plugins/banktransfer/tasks.py
+++ b/src/pretix/plugins/banktransfer/tasks.py
@@ -175,6 +175,26 @@ def _get_unknown_transactions(job: BankImportJob, data: list, event: Event=None,
     return transactions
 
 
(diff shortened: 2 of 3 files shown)


src/pretix/plugins/banktransfer/payment.py
```diff
@@ -118,6 +118,16 @@ def test_mode_message(self):
     def settings_form_fields(self):
         d = OrderedDict(
             list(super().settings_form_fields.items()) + list(BankTransfer.form_fields().items()) + [
+                ('code_prefix', forms.CharField(
+                    label=_('Bank transfer code prefix.'),
+                    help_text=_('This value is prepended to the unique order code.'),
```
raphaelm: We should add something like "If not set, the short form of the event name will be used."

src/pretix/plugins/banktransfer/payment.py
```diff
@@ -118,6 +118,16 @@ def test_mode_message(self):
     def settings_form_fields(self):
         d = OrderedDict(
             list(super().settings_form_fields.items()) + list(BankTransfer.form_fields().items()) + [
+                ('code_prefix', forms.CharField(
+                    label=_('Bank transfer code prefix.'),
```
raphaelm: I think the context of bank transfer is clear, let's jsut call it "Reference code prefix" (without a fullstop at the end).

src/pretix/plugins/banktransfer/payment.py
```diff
@@ -118,6 +118,16 @@ def test_mode_message(self):
...
+                    help_text=_('This value is prepended to the unique order code.'),
+                    widget=forms.TextInput(
+                        attrs={
+                            'placeholder': self.event.slug.upper(),
```
raphaelm: :+1:

src/pretix/plugins/banktransfer/tasks.py
```diff
@@ -190,13 +175,32 @@ def _get_unknown_transactions(job: BankImportJob, data: list, event: Event=None,
...
+    elif organizer:
+        events = organizer.events.all()
+    else:
+        return {}
+
+    for event in events:
+        prefix_event_map[event.slug.upper()] = event
+        if 'banktransfer' in event.get_payment_providers():
+            custom_prefix = event.get_payment_providers()['banktransfer'] \
+                .settings.get('code_prefix', as_type=str)
```
raphaelm: `event.get_payment_providers()` is comparatively expensive to call, so let's call it only once per event. We might even not call it at all and just access ``event.settings.get('payment_banktransfer_code_pretix')`` directly.

src/pretix/plugins/banktransfer/tasks.py
```diff
@@ -60,31 +60,16 @@ def notify_incomplete_payment(o: Order):
 @transaction.atomic
 def _handle_transaction(trans: BankTransaction, code: str, event: Event=None, organizer: Organizer=None,
```
raphaelm: Event should now no longer be an optional argument.

src/pretix/plugins/banktransfer/payment.py
```diff
@@ -118,6 +118,16 @@ def test_mode_message(self):
...
+                        },
+                    ),
+                    required=False
```
raphaelm: We might want to add a validator to make sure only letters and characters are used. Everything else will both cause problems with the banks and with the RegEx generation.

raphaelm:
Superseded by #1710

PR #1275, opened 2019-05-08
Stripe SCA

pc-coholic:
- Upgrade to latest Stripe API
- Deprecate Stripe Checkout for CC
- Migrate CC payments to Payment Intents

This is not thoroughly tested yet (as in: not tested beyond manually inserting CC-numbers)

Known issues:
- Payments are not displayed right away as "Paid" to the customers (PaymentIntent is confirmed, status is `confirmed`. Upon next poll of the PaymentIntent, it only has changed to `successful`. This could happen right away, after a few seconds or with a substantial delay).
- No tests done as to how the API-answers have changed with the latest API (webhooks, etc.).

SCA enabled test cards: https://stripe.com/docs/testing#three-ds-cards


vincentzimmer:
Nice work! We need this too :)

src/pretix/plugins/stripe/payment.py
```diff
@@ -618,73 +605,185 @@ def _use_3ds(self, card):
...
+                    description='{event}-{code}'.format(
+                        event=self.event.slug.upper(),
+                        code=payment.order.code
+                    ),
+                    metadata={
+                        'order': str(payment.order.id),
+                        'event': self.event.id,
+                        'code': payment.order.code
+                    },
+                    # TODO: Is this sufficient?
```
raphaelm: Is there a reason you dropped ``statement_descriptor``?

src/pretix/plugins/stripe/views.py
```diff
@@ -467,3 +474,37 @@ def _redirect_to_order(self):
...
+
+        if intent.status == 'requires_action' and intent.next_action.type == 'use_stripe_sdk':
+            r = render(request, 'pretixplugins/stripe/sca.html', {
+                'stripe_settings': StripeSettingsHolder(self.order.event).settings,
+                'payment_intent_client_secret': intent.client_secret,
+            })
+            r._csp_ignore = True
+            return r
+        else:
+            StripeCC._handle_payment_intent(prov, request, self.payment)
```
raphaelm: First we retrieve the payment intent, then we call ``_handle_payment_intent``, which retrieves it from the API again. This feels wasteful, any ideas to optimize it?

src/requirements/production.txt
```diff
@@ -40,7 +40,7 @@ oauthlib==2.1.*
 django-jsonfallback>=2.1.2
 psycopg2-binary
 # Stripe
-stripe==2.0.*
+stripe==2.27.*
```
raphaelm: Since the PR was started, Stripe release ``2.29.*``, I think we can upgrade to that right away.

src/pretix/plugins/stripe/views.py
```diff
@@ -150,22 +150,28 @@ def webhook(request, *args, **kwargs):
...
-        objid = event_json['data']['object']['id']
     else:
         return HttpResponse("Not interested in this data type", status=200)
 
+    if 'payment_intent' in event_json['data']['object'] and event_json['data']['object']['payment_intent']:
+        objid = event_json['data']['object']['payment_intent']
+    elif 'payment_method' in event_json['data']['object'] and event_json['data']['object']['payment_method']:
+        objid = event_json['data']['object']['payment_method']
+    else:
+        objid = event_json['data']['object']['id']
```
raphaelm: Doesn't this break the ``event_json['data']['object']['object'] == "dispute"`` case where we actually want the charge ID, not the dispute ID?

src/pretix/plugins/stripe/views.py
```diff
@@ -467,3 +474,37 @@ def _redirect_to_order(self):
...
         }) + ('?paid=yes' if self.order.status == Order.STATUS_PAID else ''))
+
+
+@method_decorator(xframe_options_exempt, 'dispatch')
+class ScaView(StripeOrderView, View):
+    def get(self, request, *args, **kwargs):
+        prov = self.pprov
+        prov._init_api()
+
+        payment_info = json.loads(self.payment.info)
```
raphaelm: Could we start this view off by checking whether ``self.payment.state`` is already in a "final" state like ``PAYMENT_STATE_CONFIRMED``, ``PAYMENT_STATE_CANCELED``, or ``PAYMENT_STATE_FAILED`` and then redirect directly to the order without anything else? For the confirmed case, this would increase performance, for the other cases, it would prevent irregular behaviour.

PR #1293, opened 2019-05-17
Document Debian dependency python-venv

ligi:
otherwise in the next step I get
```
$ python3 -m venv env
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt-get install python3-venv
```


rixx:
It might be less confusing to make it clear that this package is only needed on Debian (and maybe Ubuntu? but I don't think so). Most other distributions ship venv with Python, since it's part of the standard library.

ligi:
yea - I had the issue on Debian GNU/Linux 9.9 (stretch)

raphaelm:
I added that clarification. Thanks!

PR #1321, opened 2019-06-14
Decrement voucher usage counter when deleting testmode orders

pc-coholic:


src/pretix/base/models/orders.py
```diff
@@ -203,6 +205,11 @@ def gracefully_delete(self, user=None, auth=None):
                 'code': self.code,
             }
         )
+
+        for position in OrderPosition.all.filter(order=self):
```
raphaelm: This ignores the fact that we can't do this for canceled orders and positions, since they have already been subtracted. This will make it possible to use vouchers too many times. Therefore:

* Let's only execute the new block of code at all if ``self.status != Order.STATUS_CANCELED``
* Let's use ``order.positions.all()`` for iteration, which already have been canceled individually.

I think it would be worth it to cover these cases in tests as well.

PR #1334, opened 2019-07-08
Option to notify users when questions have been changed in backend

pc-coholic:


pc-coholic:
Merged by accidental force-push to master :grimacing:

PR #1346, opened 2019-07-22
Autofocus login form

Sohalt:
This automatically focuses the username field of the login form on browsers that support the `autofocus` attribute.


raphaelm:
Thanks, good idea!

PR #1358, opened 2019-08-04
MIMEEncode inline images as CID

pc-coholic:
Quite a few (web-)mail clients do not display the images we base64-encode as data-URLs into our emails - leaving our emails still readable, but missing the graphical elements.

Screenshot from 2019-08-04 14-49-18

By passing the images as `MIMEImage` (so basically still just a base64-encoded version of the image) and replacing the references by the `CID`-notation, we should be able to get most mailclients onboard with displaying the images - even (or despite) not having remote-image loading enabled.

As a side-effect (which we also could just disable by stripping a few lines of code), we can now also include relative or absolute `img src`'es, which will be encoded on the fly.


src/pretix/base/services/mail.py
```diff
@@ -332,3 +340,75 @@ def render_mail(template, context):
...
+
+        msg.mixed_subtype = 'related'
+        for image in cid_images:
+            try:
+                mime_image = convert_image_to_cid(
+                    image['src'], image['cid_id'], verify_ssl)
+                if mime_image:
+                    msg.attach(mime_image)
+            except Exception as e:
+                print("ERROR attaching CID image %s[%s] %s" % (image['cid_id'], image['src'], str(e)))
```
raphaelm: Could you replace this by

```
except:
    logger.exception("ERROR attaching CID image %s[%s]" % (image['cid_id'], image['src']))
```

This will automatically log it to the right place and also add a stacktrace automatically

src/pretix/base/services/mail.py
```diff
@@ -332,3 +340,75 @@ def render_mail(template, context):
...
+            cid_id = "image_%s" % (image_counter)
+            image_counter = image_counter + 1
+            original_image_src = image['src']
+
+            image['src'] = "cid:%s" % (cid_id)
+
+            cid_images.append({
+                'src': original_image_src,
+                'cid_id': cid_id
+            })
```
raphaelm: Hmm, if I think about our sample email design, we use the same image multiple times, e.g. our separator bar. Could we use (at least internally in this method) a dict instead of a list here and build it in a way that the same ``src`` will only lead to one attachment, not multiple?

src/pretix/base/services/mail.py
```diff
@@ -332,3 +340,75 @@ def render_mail(template, context):
...
+    return msg
+
+
+def convert_image_to_cid(image_src, cid_id, verify_ssl=True):
+    try:
+        image_src_split = image_src.split('data:image/png;base64,')
+        if len(image_src_split) == 2:
+            mime_image = MIMEImage(image_src_split[1], _subtype="png", _encoder=encode_noop)
+            mime_image.add_header('Content-Transfer-Encoding', 'base64')
+        else:
```
raphaelm: The ``else`` branch will also be triggered for ``data:`` URLs of other kinds, such das ``data:image/jpeg``. Currently, pretix' default renderers don't use such URLs, but custom renderers might. I'm not saying we need to deal with every image format on the planet here, be we should maybe at least fail gracefully.

That said, I'm not so sure ``.split()`` is the clearest way to detect if a string is present, maybe replace with ``if image_src.startswith(…)`` and then split use something like ``image_src[image_src.index(",")+1:]``

src/pretix/base/services/mail.py
```diff
@@ -332,3 +340,75 @@ def render_mail(template, context):
...
+
+            response = requests.get(image_src, verify=verify_ssl)
+            mime_image = MIMEImage(
+                response.content, _subtype=guess_subtype)
+
+        mime_image.add_header('Content-ID', '<%s>' % cid_id)
+
+        return mime_image
+    except Exception as e:
+        print("ERROR creating mime_image %s[%s] %s" % (cid_id, image_src, str(e)))
```
raphaelm: Same as above :)

src/pretix/base/services/mail.py
```diff
@@ -332,3 +340,75 @@ def render_mail(template, context):
...
+        return None
+
+
+def normalize_image_url(url):
+    if '//' not in url.lower():
+        if settings.STATIC_URL.startswith('http'):
+            url = "%s%s" % (settings.MEDIA_ROOT, url)
+        else:
+            url = "%s%s%s" % (settings.SITE_URL, settings.STATIC_URL, url)
```
raphaelm: I wrote a review comment that this function is fundamentally wrong, just to then realize it is not. Twice. However, there are still some subtleties here and I think we should add a comment here for future readers, since it's really not trivial. Also, there's the following things to consider:

* Please use [urljoin](https://docs.python.org/3/library/urllib.parse.html?highlight=urljoin#urllib.parse.urljoin) to concatenate URLs, to avoid ending up with double-slashes or even broken URLs.
* ``'//' not in url.lower()`` and ``.startswith('http')`` are kinda two ways to do the same thing within two lines, and neither of them is really bullet-prof. I suggest ``'://' not in url`` for both places.
* ``MEDIA_ROOT`` is an internal filesystem path, we need ``MEDIA_URL``
* If this *is* a static file, ``STATIC_URL`` is probably already included in the path.

Here's a proposal:

```
def normalize_image_url(url):
    if '://' not in url:
        """
        If we see a relative URL in an email, we can't know if it is meant to be a media file
        or a static file, so we need to guess. If it is a static file included with the
        ``{% static %}`` template tag (as it should be), then ``STATIC_URL`` is already prepended.
        If ``STATIC_URL`` is absolute, then ``url`` should already be absolute and this
        function should not be triggered. Thus, if we see a relative URL and ``STATIC_URL`` 
        is absolute *or* ``url`` does not start with ``STATIC_URL``, we can be sure this
        is a media file (or a programmer error …).

        Constructing the URL of either a static file or a media file from settings is still
        not clean, since custom storage backends might very well use more complex approaches
        to build those URLs. However, this is good enough as a best-effort approach. Complex
        storage backends (such as cloud storages) will return absolute URLs anyways so this
        function is not needed in that case.
        """
        if '://' not in settings.STATIC_URL and url.startswith(settings.STATIC_URL):
            url = urljoin(settings.SITE_URL, url)
        else:
            url = urljoin(settings.MEDIA_URL, url)
    return url
```

pc-coholic:
> * It's unclear to me what the _actual_ client support looks like. I know Outlook doesn't support base64-encoded images, but I thought most other clients did. [This article](https://sendgrid.com/blog/embedding-images-emails-facts/) states that many web-based clients can't do CID embedding, which I'm not sure if it's accurate. Can you find any data on that?

I was not really able to find any reliable data on this subject.

However - just out of trying it out myself - it seems to me almost like CID seems to be more adopted since it's "the older technology" compared to dataURLs. From a quick test, I was able to confirm that GMail, Yahoo, Web.de and Anonbox do indeed render CID without an issue. (I was also very surprised, that I still had a web.de mailaccount from 199something *and* still remembered the password).

All in all, I'd say that this is a mixed bag - but we'll probably always have a few clients not showing pictures at all.

PR #1364, opened 2019-08-12
Improve order creation through API

raphaelm:


raphaelm:
756a7829ae0bcb6079947b156144d41c8e94c13d fixes an important bug that we should put into the next bugfix release

Issue #1369, opened 2019-08-14
Mass emails to people according to their check-in status 


cbrocas:
Hello,

Pretix gives organizers the capability to send mass emails to attendees according to their tickets types and to their payment status.

But to be able to deal with after event communication, it would be very nice to be able to send emails to people that have done their check-in and those who have not.

It may be especially **useful for organizers of free access conferences** who usually have to deal with some high level of attendees that finally do not attend the event (20-25% style).

**Use case:** sending a post event survey *only* to people who finally came to the event.

This issue can be complementary to #1192 :)

Thanks!

raphaelm:
This sounds useful :)

cbrocas:
Thank Raphaël! Do you think you would integrate it to your roadmap?

raphaelm:
I don't think we'll prioritize it very high, but it's possible :) It also should be pretty straightforward to add in, in case someone wants to make a contribution.

cbrocas:
"pretty straight forward to add in" : not so sure from my point of view ;-) Thanks for everything you do on Pretix.

Issue #1313, opened 2019-06-06
Add email option for specific answers to questions


katrinleinweber:
At [deRSE19](https://www.de-rse.org/de/conf2019/index.html), the organisers used a bunch of questions during the sign-up (e.g. "Are you a de-RSE e.V. member?"). In pretix' admin interface, it is already easy to filter registrations down to the answers to such sign-up questions.

It would additionally be nice to have a "E-mail these participants" button in that view, _or_ provide question & answer filters in the E-mail section. cc @caha42

raphaelm:
https://github.com/rixx/pretix-question-placeholders might solve this

PR #1382, opened 2019-09-05
Sendmail plugin: Allow filtering for check-ins

pajowu:
This PR adds the ability to send mass emails to all attendees who have checked in (with the ability to filter for specific checkin lists). This is sort-of what was asked in #1369 (It is not possible to filter for users who never checked in)

### DIFF
--- a/src/pretix/plugins/sendmail/tasks.py
+++ b/src/pretix/plugins/sendmail/tasks.py
@@ -34,11 +34,12 @@ def send_mails(event: Event, user: int, subject: dict, message: dict, orders: li
                     continue
 
                 if filter_checkins:
-                    checkins = p.checkins.all()
-                    if not_checked_in:
-                        if checkins:
-                            continue
-                    elif not any(c.list_id in checkin_lists for c in checkins):
+                    checkins = list(p.checkins.all())
+                    allowed = (
+                        (not_checked_in and not checkins)
+                        or (any(c.list_id in checkin_lists for c in checkins))
+                    )
+                    if not allowed:
                         continue
 
                 if not p.attendee_email:
--- a/src/pretix/plugins/sendmail/tasks.py
+++ b/src/pretix/plugins/sendmail/tasks.py
@@ -7,10 +7,12 @@ from pretix.base.services.mail import SendMailException, mail
 from pretix.base.services.tasks import ProfiledEventTask
 from pretix.celery_app import app
 
+from . import forms
+
 
 @app.task(base=ProfiledEventTask)
 def send_mails(event: Event, user: int, subject: dict, message: dict, orders: list, items: list,
-               recipients: str, filter_checkins: bool, checkin_lists: list) -> None:
+               recipients: str, checkin_lists: list) -> None:
     failures = []
     user = User.objects.get(pk=user) if user else None
     orders = Order.objects.filter(pk__in=orders, event=event)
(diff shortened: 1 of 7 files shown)


src/pretix/plugins/sendmail/views.py
```diff
@@ -183,6 +199,9 @@ def get_context_data(self, **kwargs):
             log.pdata['items'] = [
                 itemcache[i['id']] for i in log.pdata.get('items', [])
             ]
+            log.pdata['checkin_lists'] = [
+                checkin_list_cache[i['id']] for i in log.pdata.get('checkin_lists', [])
```
raphaelm: You could maybe add a ``if i['id'] in checkin_list_cache`` to prevent crashes if a check-in list got deleted in the meantime

pajowu:
I considered having only one option first, similar to what there is currently for items. I decided against it because that would mean one of three things:
1. **Show all checkin lists, send the eMails to all chosen lists.**
This would mean that orders which were not checked in until now cannot be reached via mass email. This is not an option
2. **Show all checkin lists. Selecting none means sending to all users** (that match the other criteria)
I saw this as too unintuitiv
3. **Show a "not checked in" list**
This is probably the best idea but I decided against it first because it might a bit confusing.

Since 3. also solves the question in the PR better I will take a shot at it and update the PR once it is ready

src/pretix/plugins/sendmail/views.py
pajowu: Will do

raphaelm:
Hmm, I'm not sure I'm a fan of option two. I understand your reasoning and think either option two or the current approach is the best. I'll sleep on it ;)

pajowu:
Oops, I saw your message too late. The last commit adds a (slight hacky IMO) implementation of option 3. Depending on what you deem better can be merged. I *think* you can revert the commit if you want to merge option 2 (I enabled `Allow edits from maintainers.`), but I am also happy to revert it and add the small thing you asked for.

raphaelm:
Well, seeing you did that already, let's keep it that way for now and I'll have a look ;) Maybe I start do like it… Hopefully tomorrow, but I don't know, the day is rather full.

src/pretix/plugins/sendmail/tasks.py
```diff
@@ -36,6 +39,9 @@ def send_mails(event: Event, user: int, subject: dict, message: dict, orders: li
                 if p.item_id not in items and not any(a.item_id in items for a in p.addons.all()):
                     continue
 
+                if forms.MailForm.NOT_CHECKED_IN not in checkin_lists and not any(str(c.list_id) in checkin_lists for c in p.checkins.all()):
+                    continue
```
raphaelm: The logic seems off here.

If I only select "Not checked in", it does send emails to everyone, even those who checked in, doesn't it?

pajowu:
I think reverting to the previous approach might also be a good Idea if more filters are added later (as proposed for example in #1313) to keep the form to a reasonable length.

pajowu:
I changed the input method to a checkbox for filtering at all which expands to a checkbox for filtering not-checked-in orders plus a ajax-based multi-select for filtering checkin lists.

2019-10-20-114113_1595x184_scrot

I also rebased the branch to the current master so it should be merge-able without problems.

raphaelm:
Thank you so much! :heart: 

I was planning for weeks to finish this one up, but didn't get around to. Now I felt bad for making you jump through even more hoops, so I did some more tweaks myself:

* I cleaned up some code that was changed through all the back-and-forth

* I resized the panel in the form so it looks better and does not "break up" the form

* I made a slight change in semantics of the options. With your version, it was *possible* to select both "only send to customers not checked in" *and* a check-in list, but the result of that operation was neither well-defined nor really useful. We could have just prevented that situation by blocking this combination, but I imagine that there is a real use case of "send to all people who are either not checked in at all or just checked in on list X" to exclude everyone who already checked in on a list Y. Therefore, I just removed the "Only" from the form fields and adjusted the logic accordingly.

* I fixed a "bug" in the query that existed before this PR as well, but turned worse through this PR. Previously, if you would filter by "product X" and "checked-in on list Y", the email might have been sent *even if those two properties are true for two different positions on the order, not in combination*. To make things worse, this was handled correctly only when sending only to atteendee email addresses, but the number of matching orders in the success message was off in this case. I fixed this by introducing a subquery for the postions.

PR #1396, opened 2019-09-10
Support for gift cards

raphaelm:
This is a proof of concept for how pretix could support gift cards. It's faaaar away from being usable, but it proves the concept. Hence the name!

- [x] Data model
- [x] Proper UI during checkout
- [x] Proper display of pending orders
- [x] Proper checks of card acceptance
- [x] Cross-organizer acceptance
  - [x] configuration
- [x] Refund support
- [x] Handle cancellation of gift card sales
- [x] Allow to redeem gift cards when changing payment methods
- [x] Backend UI
  - [x] Payment details
  - [x] Gift card list and details
  - [x] Create gift cards
- [x] Allow to sell gift cards
  - [x] Gift card products
  - [x] Prevent users from getting taxation wrong
  - [x] Send gift cards to users nicely
  - [x] Deal with refunds
- [x] Properly prevent race conditions
- [x] Show helpful error when users try to double-spend
- [x] Acceptance settings
- [x] Show helpful error when users confuse gift cards and vouchers
- [x] API for gift cards
- [x] API for redeeming gift cards
- [x] Add lots of tests
- [x] Manually squash migrations
- [x] Write user guide
- [x] Review notes
  - [x] Giftcards can be paid with giftcards - I don't think this should be allowed (or at least the giftcard should only be used for the non-giftcard portion of the purchase)
  - [x] A giftcard can be an (attendance-)ticket and a giftcard at the same time (and as such also shows up in Checkin-Lists)
  - [x] Unless aptly named, there is no way to decern from the order-view in the backend, that a certain order is a giftcard. A order-info-box with a link to the new giftcard might make handling easier.
  - [x] Deletion of Testmode-Orders is not possible.
- [ ] pretix.eu
  - [ ] set new permission in setup
  - [ ] do not charge fee on sale of gift card
  - [ ] Support in pretixPOS
  - [ ] Support in Reseller mode?


src/pretix/presale/checkoutflow.py
```diff
@@ -528,6 +529,40 @@ def provider_forms(self):
...
+                if gc.currency != request.event.currency:
+                    messages.error(self.request, _("This gift card does not support this currency."))
+                    return self.render()
+                if gc.pk in self.cart_session['gift_cards']:
+                    messages.error(self.request, _("This gift card is already used for your payment."))
+                    return self.render()
+                if gc.value <= Decimal("0.00"):
+                    messages.error(self.request, _("All credit on this gift card has been used."))
+                    return self.render()
+                if 'gift_cards' not in self.cart_session:
```
felixrindt: This check needs to be moved up. I'm getting a key error in line 541. :D

felixrindt:
I took a look at what you build already. Looks like a very flexible tool for organizers!
Using fees in the presale and an Order Payment in the backend while actually being a GiftCardTransfer seems a bit hacky :smile: but it seems like most sensible thing to do at this stage.

I think users should be able to deactivate the Gift Card Payment Provider, if they don't want to accept gift cards. Maybe the check for if the organizer has gift cards could be moved to the `is_allowed` instead of the template rendering?

Have a nice day :)

src/pretix/presale/checkoutflow.py
raphaelm: Silly me ;)

raphaelm:
>  Using fees in the presale and an Order Payment in the backend while actually being a GiftCardTransfer seems a bit hacky smile but it seems like most sensible thing to do at this stage.

Yeah, the fee thing is hacky :( The only alternative would be to support multiple partial payments in the checkout process and I'm not really sure that's a road worth going down…

The payment is correct in my eyes. Gift cards are a means of payment and in this case, pretix is both the payment provider and the payment consumer. However, since cross-organizer acceptance of gift cards is possible, it makes sense to have a record on both sides of the equation, thus the transaction object and the payment object.

> I think users should be able to deactivate the Gift Card Payment Provider, if they don't want to accept gift cards. Maybe the check for if the organizer has gift cards could be moved to the is_allowed instead of the template rendering?

Hmm, that would make sense. I'll need to check what breaks if we actually make it configurable like a payment provider.

raphaelm:
@felixrindt I reworked it a little to adress your concerns, I like it much better now, less of a special case.

raphaelm:
> Giftcards can be paid with giftcards - I don't think this should be allowed (or at least the giftcard should only be used for the non-giftcard portion of the purchase)

Hmm. Do you see a specific reason why it shouldn't be allowed? I agree that it doesn't make sense, but it won't be that easy to work around.

> A giftcard can be an (attendance-)ticket and a giftcard at the same time (and as such also shows up in Checkin-Lists)
> Unless aptly named, there is no way to decern from the order-view in the backend, that a certain order is a giftcard. A order-info-box with a link to the new giftcard might make handling easier.

Fixed! Also in the order view in the frontend, thus removing the need to issue tickets.

> Deletion of Testmode-Orders is not possible.

Ouch, I completely missed that. That's gonna be fun to integrate… Thanks!

pc-coholic:
> Hmm. Do you see a specific reason why it shouldn't be allowed? I agree that it doesn't make sense, but it won't be that easy to work around.

I just feel like it's a general consensus in the industry™ not to allow this...

As of now, giftcards do not have an expiration data attached to them (but trust me: Not before long, people will ask for exactly that...). If we would allow to buy new giftcards with existing giftcards, we could extend the validity of a card indefinitely.

I'm sure there are also other reasons to not allow that, but I'm kinda blanking on those right now...


> Ouch, I completely missed that. That's gonna be fun to integrate… Thanks!

Feels like groundhog day, after having found the same issue with our Reseller-module :-P

raphaelm:
> As of now, giftcards do not have an expiration data attached to them (but trust me: Not before long, people will ask for exactly that...). If we would allow to buy new giftcards with existing giftcards, we could extend the validity of a card indefinitely.

Good point. I didn't add expiries for now, since it's not really legally a possibility in Germany… ;)

raphaelm:
Okay, I implemented test mode and I removed the possibility to buy gift cards with gift cards. I chose the easy way of not allowing this combination at all, since it will be near-impossible to get this right otherwise when partially paying orders later.

Gift cards now have a test mode flag on their own. Test gift cards can't be deleted, but the orders can now be deleted. I think it this fine – the only reason orders *can* be deleted in test mode is that test mode was introduced late into the system and third-party code such das plugins or API consumers was not prepared to ignore test mode orders. We do not have this problem for newly added features.

I'd ask you to especially review the changes since your last review, to make sure I didn't miss any operations where we'd need to respect test mode.

PR #1402, opened 2019-09-11
Allow ticket output providers to handle downloads externally

pc-coholic:
In addition to adding the missing ``download_button_icon`` property to the documentation, this PR introduces a few more optional attributes, which should make it easier for having external TicketOutput-Providers.

``preview_allowed`` was already an internal property, which I think might be useful to have. For example, creating previews for Google Play Passes is rather... difficult/annoying, having to push data to a server - especially when the generating plugin does not only operate on the passed ``order`` and ``orderposition``, but also actively tries to query the database (which will probably fail for preview-data).

``is_downloadable`` will prevent attaching the output of the provider to any outgoing emails - useful if there is nothing to actually output as a file but for example only a URL to forward the user to.

``download_action`` replaces the default async POST for generation of the ticket-file with a regular ``<a href>``-link, however enriched with a variety of ``data-``-attributes relating to the order(-position). This way, we can have providers which can just forward the user to another URL or have JavaScript-listeners for ``click``-actions on the buttons - without the current hacky solutions to intercept the async form submissions.

Not gonna lie: This PR will probably mostly profit the Google Pay Passes integration in the future... But I think, it might also be useful for other integrations in the future.


src/pretix/base/ticketoutput.py
```diff
@@ -161,3 +161,47 @@ def download_button_icon(self) -> str:
...
+        """
+        By default, the ``generate()`` method is called for generating a preview in the pretix backend.
+        In case your plugin cannot generate previews for any reason, you can manually disable it here.
+        """
+        return True
+
+    @property
+    def is_downloadable(self) -> bool:
+        """
+        Returns whether or whether the output of this plugin can be downloaded and/or attached to an email.
```
raphaelm: ``whether or whether``? :)

src/pretix/base/ticketoutput.py
```diff
@@ -161,3 +161,47 @@ def download_button_icon(self) -> str:
...
+        The link will contain the following ``data-``-attributes, accessible for further usage from your JavaScript:
+
+        - ``data-organizer``: Organizer Slug
+        - ``data-event``: Event Slug
+        - ``data-order``: Order Code
+        - ``data-secret``: Order Secret
+        - ``data-position``: ID of the concerned OrderPosition
+
+        To facilitate access to these ``data-``-attributes, the button will have a ``class``-attribute of
+        ``btn-identifier`` (where ``identifier`` is the identifier of your ticket download provider).
```
raphaelm: Do you think we should use e.g. ``btn-ticket-itentifier`` to avoid any clashes with bootstrap's class names (like btn-default)?

src/pretix/base/ticketoutput.py
```diff
@@ -161,3 +161,47 @@ def download_button_icon(self) -> str:
...
+        Returns whether or whether the output of this plugin can be downloaded and/or attached to an email.
+        By default, this is setting defaults to ``True`` for backwards compatibility.
+        """
+        return True
+
+    @property
+    def download_action(self) -> str:
+        """
+        By default, the download-button will call an asynchronous task to generate the downloadable ticket file.
+        By overriding this property, the default behavior can be changed - for example a ``href``-target to another
```
raphaelm: Why "for example"? On first read I thought this can do other things than just returning an URL, but in fact it's the only thing it can.

src/pretix/base/ticketoutput.py
```diff
@@ -161,3 +161,47 @@ def download_button_icon(self) -> str:
...
+        In case of using a JavaScript-call, make sure to also include your JavaScript-file on all pages that might
+        display the download button (such as the ``control`` and ``presale`` order views and the ticket output settings.
+
+        The link will contain the following ``data-``-attributes, accessible for further usage from your JavaScript:
+
+        - ``data-organizer``: Organizer Slug
+        - ``data-event``: Event Slug
+        - ``data-order``: Order Code
+        - ``data-secret``: Order Secret
+        - ``data-position``: ID of the concerned OrderPosition
```
raphaelm: For clarity, we should add if we mean the ``id`` or the ``positionid`` attribute.

src/pretix/base/ticketoutput.py
```diff
@@ -161,3 +161,47 @@ def download_button_icon(self) -> str:
...
+    @property
+    def is_downloadable(self) -> bool:
+        """
+        Returns whether or whether the output of this plugin can be downloaded and/or attached to an email.
+        By default, this is setting defaults to ``True`` for backwards compatibility.
+        """
+        return True
+
+    @property
+    def download_action(self) -> str:
```
raphaelm: I can think of usecases where JavaScript is not required and where one would just want to return the URL to an external resource here. In that case, designing this as a ``@property`` will not be very helpful, as that URL will likely depend on the details. A more flexible signature could be

``
def download_action(self, order: Order, position: OrderPosition) -> Optional[str]
``

src/pretix/control/templates/pretixcontrol/event/tickets.html
```diff
@@ -28,9 +28,14 @@ <h1>{% trans "Ticket download" %}</h1>
                 {% for provider in providers %}
                     <div class="panel panel-default ticketoutput-panel">
                         <div class="panel-heading">
-                            <a href="{% url "control:event.settings.tickets.preview" event=request.event.slug organizer=request.organizer.slug output=provider.identifier %}"
-                                    class="btn btn-default btn-sm pull-right {% if not provider.preview_allowed %}disabled{% endif %}"
-                                    target="_blank">
+                            {% if provider.download_action %}
+                                <a href="{{ provider.download_action }}" class="btn btn-default btn-sm pull-right {% if not provider.preview_allowed %}disabled{% endif %}" target="_blank"
```
raphaelm: With the changes proposed above, this would not longer work as easily. However, I'm fine with not having previews for non-downloadable providers.

src/pretix/base/ticketoutput.py
```diff
@@ -161,3 +161,47 @@ def download_button_icon(self) -> str:
...
+        - ``data-position``: ID of the concerned OrderPosition
+
+        To facilitate access to these ``data-``-attributes, the button will have a ``class``-attribute of
+        ``btn-identifier`` (where ``identifier`` is the identifier of your ticket download provider).
+
+        .. note:: Please be aware, that ``download_action`` cannot contain any direct direct ``javascript:xxxx();``
+                  -calls, as this will violate pretix' Content Security Policy of excluding ``self`` as a
+                  ``script-src``. This is a sane default and should not easily be overridden. Please consider using
+                  event-listeners instead.
+        """
```
raphaelm: This docstring should mention what ``return None`` means in this case and that it is the default.

pc-coholic:
As discussed, I've changed a few things around...

- Dropped the `is_downloadable` completely.
- `download_action` is now `download_handled_by_frontend` strictly a bool
- Expanded on the documentation of the different `data-`-attributes that are provided to frontend-handled JS.
- Said `data-`-attributes are note leaking any secrets they shouldn't - but it is now up to the developer to interpret the individual fragments properly.
- Also added a `data-generate-url` which contains the original POST-URL. This should take out the guess-work, if the creation of a proper order/secret-link is required.

doc/development/api/ticketoutput.rst
```diff
@@ -69,3 +69,11 @@ The output class
...
 
    .. autoattribute:: download_button_text
+
+   .. autoattribute:: download_button_icon
+
+   .. autoattribute:: preview_allowed
+
+   .. autoattribute:: is_downloadable
+
+   .. autoattribute:: download_action
```
raphaelm: I think this is out of date now

src/pretix/control/templates/pretixcontrol/event/tickets.html
```diff
@@ -29,7 +29,7 @@ <h1>{% trans "Ticket download" %}</h1>
                     <div class="panel panel-default ticketoutput-panel">
                         <div class="panel-heading">
                             <a href="{% url "control:event.settings.tickets.preview" event=request.event.slug organizer=request.organizer.slug output=provider.identifier %}"
-                                    class="btn btn-default btn-sm pull-right {% if not provider.preview_allowed %}disabled{% endif %}"
+                                    class="btn btn-default btn-sm pull-right {% if not provider.preview_allowed or provider.download_handled_by_frontend %}disabled{% endif %}"
```
raphaelm: Your PR still introduces ``preview_allowed`` (which is fine by me), so that one should be used here.

src/pretix/control/templates/pretixcontrol/order/index.html
```diff
@@ -281,15 +281,21 @@ <h3 class="panel-title">
...
-                                                        method="post" data-asynctask data-asynctask-download
-                                                        class="form-inline helper-display-inline">
-                                                    {% csrf_token %}
-                                                    <button type="submit"
-                                                            class="btn btn-xs btn-default">
+                                                {% if b.download_handled_by_frontend %}
+                                                    <a href="#" class="btn btn-default btn-xs btn-ticket-{{ b.identifier }}" target="_blank" data-organizer="{{ request.event.organizer.slug }}" data-event="{{ event.slug }}" data-order="{{ order.code }}" data-secret="{{ line.secret }}" data-position="{{ line.pk }}" data-generate-url="{% url "control:event.order.download.ticket" code=order.code event=request.event.slug organizer=request.event.organizer.slug position=line.pk output=b.identifier %}">
```
raphaelm: ``line.secret`` is used for ``data-secret`` here. However, the documentation states it is the ``line.web_secret`` in same cases and ``order.secret`` in others.

src/pretix/control/views/event.py
```diff
@@ -822,7 +822,6 @@ def provider_forms(self) -> list:
             provider.settings_content = provider.settings_content_render(self.request)
             provider.form.prepare_fields()
 
-            provider.preview_allowed = True
             for k, v in provider.settings_form_fields.items():
                 if v.required and not self.request.event.settings.get('ticketoutput_%s_%s' % (provider.identifier, k)):
                     provider.preview_allowed = False
```
raphaelm: Hmm. Does this still work as intended? I think setting a property this way is illegal.

pc-coholic:
As discussed:
- Ditching all `data-`-attributes
- Keeping `preview_allowed`
- Adding `javascript_required`
- Adding the possibility of redirects instead of downloads

PR #1409, opened 2019-09-18
Auto-check-in for specific sales channels

pc-coholic:
- [x] Extend Checkin-model
- [x] Show in form
- [x] Show in List-Overview
- [x] Expose in API
- [x] Extend Serializer
- [x] Write API tests
- Actually check-in automatically for each sales-channel
  - [x] Web
  - [x] pretixPOS
  - [x] resellers
- Make auto-checkins visible in
  - [x] checkinlists
  - [x] orders
  - [x] API
  - [x] Exports
- [x] Update API-docs


src/pretix/control/views/checkin.py
```diff
@@ -33,12 +34,18 @@ def get_queryset(self, filter=True):
             m=Max('datetime')
         ).values('m')
 
+        cqs2 = Checkin.objects.filter(
```
pc-coholic: We should try to eliminate this sub-query

src/pretix/plugins/checkinlists/exporters.py
```diff
@@ -93,10 +93,17 @@ def _get_queryset(self, cl, form_data):
         ).order_by().values('position_id').annotate(
             m=Max('datetime')
         ).values('m')
+
+        cqs2 = Checkin.objects.filter(
```
pc-coholic: We should try to also eliminate this sub-query

raphaelm: For any specific reason? Subqueries aren't inherently bad.

We should change it though, see below.

src/pretix/control/views/checkin.py
raphaelm: See below

src/pretix/control/templates/pretixcontrol/order/index.html
```diff
@@ -255,7 +255,16 @@ <h3 class="panel-title">
...
+                                            </svg>
```
raphaelm: :joy: See above

src/pretix/plugins/checkinlists/exporters.py
```diff
@@ -240,7 +252,7 @@ def get_story(self, doc, form_data):
 
             row = [
                 '!!' if op.item.checkin_attention or op.order.checkin_attention else '',
-                CBFlowable(bool(op.last_checked_in)),
+                CBFlowable(checked=bool(op.last_checked_in), auto_checked_in=bool(op.auto_checked_in)),
```
raphaelm: I don't think we *really* need to make this PDF more crowded by this information, I'm not sure in what use case it would be relevant.

src/pretix/base/services/checkin.py
```diff
@@ -155,3 +157,16 @@ def perform_checkin(op: OrderPosition, clist: CheckinList, given_answers: dict,
...
+
+
+@receiver(order_placed, dispatch_uid="autocheckin_order_placed")
+def order_placed(sender, **kwargs):
+    order = kwargs['order']
+    event = sender
+
+    cls = list(event.checkin_lists.filter(auto_checkin_sales_channels__contains=order.sales_channel).prefetch_related(
+        'limit_products'))
+    for op in order.positions.all():
```
raphaelm: One small performance improvement in case this feature is not used at all:
```suggestion
    if not cls:
        return
    for op in order.positions.all():
```

PR #1411, opened 2019-09-19
Code style improvements

felixrindt:
Just some stuff I noticed while looking at #1396 :)

### DIFF
--- a/src/pretix/base/payment.py
+++ b/src/pretix/base/payment.py
@@ -531,7 +531,7 @@ class BasePaymentProvider:
         containing the URL the user will be redirected to. If you are done with your process
         you should return the user to the order's detail page.
 
-        If the payment is completed, you should call ``payment.confirm()``. Please note that ``this`` might
+        If the payment is completed, you should call ``payment.confirm()``. Please note that this might
         raise a ``Quota.QuotaExceededException`` if (and only if) the payment term of this order is over and
         some of the items are sold out. You should use the exception message to display a meaningful error
         to the user.
--- a/src/pretix/plugins/banktransfer/payment.py
+++ b/src/pretix/plugins/banktransfer/payment.py
@@ -102,7 +102,12 @@ class BankTransfer(BasePaymentProvider):
                 label=_('Payment method name'),
                 widget=I18nTextInput,
                 required=False
-            ))
+            )),
+            ('omit_hyphen', forms.BooleanField(
+                label=_('Do not include a hypen in the payment reference.'),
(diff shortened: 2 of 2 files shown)


src/pretix/base/payment.py
```diff
@@ -174,7 +174,7 @@ def settings_form_fields(self) -> dict:
 
         We suggest that you return an ``OrderedDict`` object instead of a dictionary
         and make use of the default implementation. Your implementation could look
-        like this::
```
rixx: The "::" is intentional, as it signals in RST that a code block is about to begin. You can see the output here: https://docs.pretix.eu/en/latest/development/api/payment.html#pretix.base.payment.pretix.base.payment.BasePaymentProvider.BasePaymentProvider.settings_form_fields

src/pretix/plugins/banktransfer/payment.py
```diff
@@ -116,16 +121,7 @@ def test_mode_message(self):
...
-            list(super().settings_form_fields.items()) + list(BankTransfer.form_fields().items()) + [
-                ('omit_hyphen', forms.BooleanField(
-                    label=_('Do not include a hypen in the payment reference.'),
-                    help_text=_('This is required in some countries.'),
-                    required=False
-                )),
-
-            ]
-        )
+        d = OrderedDict(list(super().settings_form_fields.items()) + list(BankTransfer.form_fields().items()))
```
raphaelm: This is also on purpose. This hyphen thing is a very specific option, only really a problem in Finland. With your change, we would start asking it to every user who goes through the quick event setup.

felixrindt:
Haha ok sorry for the interruption then and thanks for explaining ;)

src/pretix/plugins/banktransfer/payment.py
felixrindt: I searched for usages of `BankTransfer.form_fields()`. In https://github.com/pretix/pretix/blob/76e75bef65a9543594ac418e0ebc16f61a7f9a0e/src/pretix/control/forms/event.py#L1495

the fields shown in the quick setup are hardcoded, so I thought that wouldn't be a problem.

src/pretix/base/payment.py
felixrindt: Thanks for explaining. I removed that change.

src/pretix/plugins/banktransfer/payment.py
raphaelm: Agreed

PR #1428, opened 2019-10-04
remove .csv from default filename in List Exporter

felixrindt:
The file type suffix is appended here

https://github.com/pretix/pretix/blob/9fca3188b29df9b79e3adaa3aa20af3101231338/src/pretix/base/exporter.py#L121

so at the moment there are filenames like `export.csv.csv`.


raphaelm:
Oops, thanks!

PR #1433, opened 2019-10-07
Reordering questions with drag'n'drop

### LINKED ISSUE #807: Simplify reordering of questions
The reordering-process is quite annoying right now because it reloads the page every time I move a question up or down. Especially time consuming is it if you have a lot of questions and you want to move one from the very bottom to the top position.

Possible solutions I have in mind:
* make asynchronous requests and move the question within the web page, making complete reloads unnecessary
* add an number input field (maybe displayed only if a button is clicked) for redefining the order in one step for all questions
* drag-and-drop UI (inplace or in a new view)

Sohalt:
Closes #807.

This implements drag'n'drop to reorder questions.

TODO:
- [x] support pagination


src/pretix/control/views/item.py
```diff
@@ -283,6 +283,22 @@ def get_queryset(self):
         return self.request.event.questions.prefetch_related('items')
 
 
+def reorder_questions(request, organizer, event):
```
raphaelm: There's a ``@event_permission_required("can_change_items")`` missing here. Tip: Also add the URL to ``tests/control/test_permissions.py``

Sohalt:
> However, do you have any ideas on how to handle the pagination problem yet?

I currently can think of 3 approaches:

1. disabling drag'n'drop when pagination kicks in
2. allowing drag'n'drop within one page and keeping the up arrow for the first element in the list and the down arrow for the last element, to allow reordering across page boundaries
3. replacing pagination with "infinite scrolling", progressively loading more items and allowing drag'n'drop for all of them.

The first would probably be the easiest. It still makes reordering more pleasant for most use cases and if you have to reorder more than 100 elements you are already in a pretty bad place anyway, no matter if you have drag'n'drop or not.

The second approach might be a little confusing to users, because some items have arrows and others don't, but it would allow drag'n'drop to be used for more use cases. It should still be fairly easy to implement.

The third would probably be best from a UX standpoint, but I fear it is way to complicated to implement.

raphaelm:
Yeah, I think option 3 is a story for another day.

Option 1 would be the simplest, yes, although the up/down arrows are *especially* annoying once you have more items, so it is kinda contra-intuitive to disable it in the edge case where it is most useful.

Option 2 sounds like a decent compromise.

I would be willing to merge option 1, but if you have the time to work on option 2, I think it would be more useful.

Sohalt:
I've started work on option 2.

I'm not familiar with django pagination, so I wanted to ask if there's an easy way for me to validate the provided input, i.e. make sure that the provided `ids` correspond to the actual question ids at the provided `offset`.

raphaelm:
> I'm not familiar with django pagination, so I wanted to ask if there's an easy way for me to validate the provided input, i.e. make sure that the provided ids correspond to the actual question ids at the provided offset.

This should be possible with the [Paginator](https://docs.djangoproject.com/en/2.2/topics/pagination/) object, but I'm not sure that this is even necessary. If we have a list of IDs that are all valid (i.e. of the correct event), we should be able to just order them, regardless of if they really were on the correct page, aren't we? We'd just need to somehow figure out how to keep the order of the other pages untouched.

src/pretix/control/views/item.py
```diff
@@ -283,6 +284,32 @@ def get_queryset(self):
...
+
+    questions = request.event.questions.filter(id__in=ids)
+
+    if questions.count() != len(ids):
+        raise Http404(_("Some of the provided question ids are invalid."))
+
+    positions = questions.values_list('position', flat=True)
+
+    if positions.last() - positions.first() + 1 != questions.count():
+        return HttpResponseBadRequest("ids have to be from a consecutive range")
```
raphaelm: Hmm, I'm not sure that's going to work, ``position`` values can have gaps for very valid reasons. Can we instead check something like:
```
if request.event.questions.filter(position__gte=positions.first(), position__lte=positions.last()).exclude(id__in=ids).exists():
```

src/pretix/control/views/item.py
```diff
@@ -283,6 +284,32 @@ def get_queryset(self):
...
+
+    for pos, id in zip(positions, ids):
+        qt = questions.get(id=id)
```
raphaelm: This creates a new database query on every iteration, could we pre-cache these in line 294?

src/pretix/control/views/item.py
```diff
@@ -283,6 +284,32 @@ def get_queryset(self):
         return self.request.event.questions.prefetch_related('items')
 
 
+@event_permission_required("can_change_items")
+def reorder_questions(request, organizer, event):
```
raphaelm: A ``@transaction.atomic`` would probably also do some good

src/pretix/control/views/item.py
Sohalt: Yep, just found out that `positions` are neither unique nor consecutive. Can you explain what valid reasons there are for this?
I was just working on a migration that would give all questions a consecutive position. If we don't have this, we run into a problem when more questions than fit on a single page have the same position. E.g. in the default case with `position = 0` when you just create more than 100 questions and then try to reorder them, your suggested query would reject the request (and I don't see any way to perform the reordering with sensible semantics).

src/pretix/control/views/item.py
Sohalt: Ah, I'm not super familiar with Django ORM internals and assumed that the manager caches the result from line 294 and executes the query against that. Will fix.

src/pretix/control/views/item.py
raphaelm: Well, first of all, up until now, this wasn't a problem, so we didn't care that much.

When using pretix' interface, gaps can easily occur when you delete a question, because we don't re-calculate all positions in this case. I don't think non-uniqueness should happen through using the interface.

However, it is possible to create non-unique positions through the REST API, since we currently neither validate nor auto-assign positions there. I would very hesitant to add in validation here since it would be a pretty breaking API change.

Therefore, a migration will not fix our problem since these situations can re-occur at any time.

Even if we *would* fix both of these issues, I would be very cautious about relying on uniquness and consecutiveness of the positions, since there is a third reason how it can break: Race conditions when deleting/sorting/adding questions at the same time. I feels like over-engineering to go through all the trouble of preventing these issues just for this feature.

> If we don't have this, we run into a problem when more questions than fit on a single page have the same position. E.g. in the default case with position = 0 when you just create more than 100 questions and then try to reorder them, your suggested query would reject the request (and I don't see any way to perform the reordering with sensible semantics).

Well, I don't think it's impossible. Even with non-uniqueness and non-conscutiveness in the position, the order is still well-defined as ``position ASC, id ASC``. We don't even need to loop over all the questions in the event as the current re-ordering code does.

Untested pseudocode:

```
input_questions = request.event.questions.filter(id__in=ids)
ordered_questions = sorted(input_questions, lambda k: ids.index(k.pk))
original_lowest_score = min((q.position, q.pk) for q in input_questions)
original_highest_score = max((q.position, q.pk) for q in input_questions)

highest_position_on_previous_page = request.event.questions.filter(
    Q(position__lt=original_lowest_score[0])
    | Q(Q(position=original_lowest_score[0]) & Q(pk__lt=original_lowest_score[1]))
).aggregate(m=Max('position'))['m'] or 0

for i, q in enumerate(ordered_questions):
    pos = highest_position_on_previous_page + 1 + i
    if pos != q.position:  # Save unneccessary UPDATE queries
        q.position = pos
        q.save(update_fields['position'])

questions_on_later_pages = request.event.questions.filter(
    Q(position__gt=original_highest_score[0])
    | Q(Q(position=original_highest_score[0]) & Q(pk__gt=original_highest_score[1]))
)
for i, q in enumerate(questions_on_later_pages):
    pos = highest_position_on_previous_page + len(input_questions) + 1 + i
    if pos != q.position:  # Save unneccessary UPDATE queries
        q.position = pos
        q.save(update_fields['position'])
```

Sohalt: > Untested pseudocode:

I guess that should work. I didn't think enough about the ordering being defined by both `position` and `id`.

I think we should definitely document this behavior in more detail in the API docs. It was pretty counter intuitive to me (although I get your reasons now).
It might also be helpful to rename the field to something other than position (e.g. "sortkey" or something like that), since position (at least to me) has a pretty strong meaning (i.e. the first position, the second position, and so on). Renaming fields in a public API however is often not worth the pain. It's probably enough to extensively document the field.

raphaelm: Yes. The correct name would probably be something like "priority".

raphaelm:
Thanks a lot!

Sohalt:
Thanks for writing pretix and for the good code review!

PR #1434, opened 2019-10-08
Replace occurrences of "blacklist" with "banlist"

Sohalt:


src/pretix/base/validators.py
```diff
@@ -6,11 +6,11 @@
 
 class BlacklistValidator:
```
raphaelm: Maybe we should fix this class name and it's usages as well, then.

raphaelm:
Apparently, we'll need to rename all the references to the validators in the migration files.

Alternatively, in case any plugins use those validators (which I highly doubt), we could also add a backwards-compatibility option at the end of the file (e.g. ``BlackListValidator = BanListValidator``)

PR #1461, opened 2019-10-23
Add manual order-matching field for duplicate bankimport transactions

pc-coholic:


src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/transaction_list.html
```diff
@@ -43,7 +43,12 @@
                             </button>
                         </div>
                     {% elif trans.state == 'error' or trans.state == 'already' %}
+                        <input type="text" class="form-control" placeholder="{% trans "Order code" %}">
```
raphaelm: Hmm, this also adds the option in ``trans.state == 'error'``, do we really want that?

pc-coholic: Good point - changed that.

PR #1462, opened 2019-10-23
Add question type: phone number

pc-coholic:
Z#2352979

Thoughts on this?

The dropdown-selection is not as pretty as it could be though (country flags, etc.)...

### DIFF
--- a/doc/api/resources/questions.rst
+++ b/doc/api/resources/questions.rst
@@ -31,6 +31,7 @@ type                                  string                     The expected ty
                                                                  * ``H`` – time
                                                                  * ``W`` – date and time
                                                                  * ``CC`` – country code (ISO 3666-1 alpha-2)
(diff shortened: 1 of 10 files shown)


raphaelm:
- [ ] Check impact on pretixSCAN apps

raphaelm:
Okay, so on Android and Desktop, the field just won't show up on pretixSCAN, which is unexpected behaviour, so we should fix that. From reading the source code, the iOS app might even crash, so we definitely need to check and fix that ;)

raphaelm:
@pc-coholic Since this is in high demand and it's unlikely we get the pretixSCAN change rolled out soon enough, could you just add in a validation that does not allow combining the new type with ``ask_during_checkin`` in both the form and the API? This should also be effective for the "file" type, which also does not work in pretixSCAN at the moment.

PR #1471, opened 2019-10-29
PayPal: Show payment ID on paid invoices

pc-coholic:
This is a PoC for displaying the PaymentID on paid invoices for PayPal.

Technically, this is just calling the `render_invoice_text` also on paid invoices, which we didn't before.

If this change is any good, we should probably also look into implementing similar printouts for Stripe and Mollie.

### DIFF
--- a/src/pretix/base/services/invoices.py
+++ b/src/pretix/base/services/invoices.py
@@ -20,9 +20,7 @@ from django_scopes import scope, scopes_disabled
 from i18nfield.strings import LazyI18nString
 
 from pretix.base.i18n import language
-from pretix.base.models import (
-    Invoice, InvoiceAddress, InvoiceLine, Order, OrderPayment,
-)
+from pretix.base.models import Invoice, InvoiceAddress, InvoiceLine, Order
 from pretix.base.models.tax import EU_CURRENCIES
 from pretix.base.services.tasks import TransactionAwareTask
 from pretix.base.settings import GlobalSettingsObject
@@ -37,9 +35,6 @@ logger = logging.getLogger(__name__)
 @transaction.atomic
 def build_invoice(invoice: Invoice) -> Invoice:
     lp = invoice.order.payments.last()
-    open_payment = None
-    if lp and lp.state not in (OrderPayment.PAYMENT_STATE_CONFIRMED, OrderPayment.PAYMENT_STATE_REFUNDED):
-        open_payment = lp
 
     with language(invoice.locale):
         invoice.invoice_from = invoice.event.settings.get('invoice_address_from')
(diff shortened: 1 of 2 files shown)


src/pretix/plugins/paypal/payment.py
```diff
@@ -531,3 +532,16 @@ def shred_payment_info(self, obj):
...
             le.shredded = True
             le.save(update_fields=['data', 'shredded'])
+
+    def render_invoice_text(self, order: Order, payment: OrderPayment) -> str:
+        if order.status == Order.STATUS_PAID:
+            if payment.info_data.get('id', None):
+                return '{}\r\n{}: {}'.format(
+                    _('The payment for this invoice has already been received.'),
+                    _('PayPal payment ID'),
+                    payment.info_data['id']
```
raphaelm: We should *also* display the sale ID that, if it exists, is stored at `payment.info_data['transactions'][0]['related_resources'][0]['sale']['id']`. The ``id`` value is something that seems to be specific to the REST API, so it does not help when trying to find a transaction in the consumer or merchant portal.

src/pretix/base/services/invoices.py
```diff
@@ -37,9 +35,6 @@
 @transaction.atomic
 def build_invoice(invoice: Invoice) -> Invoice:
     lp = invoice.order.payments.last()
-    open_payment = None
-    if lp and lp.state not in (OrderPayment.PAYMENT_STATE_CONFIRMED, OrderPayment.PAYMENT_STATE_REFUNDED):
-        open_payment = lp
```
raphaelm: Hmhmm, interesting that we previously didn't call this at all for paid orders. However, since the default implementation in BasePaymentProvider and all implementations I just checked already take care of this situation, it should be fine.

src/pretix/plugins/paypal/payment.py
```diff
@@ -531,3 +532,18 @@ def shred_payment_info(self, obj):
...
+
+    def render_invoice_text(self, order: Order, payment: OrderPayment) -> str:
+        if order.status == Order.STATUS_PAID:
+            if payment.info_data.get('id', None):
+                return '{}\r\n{}: {}\r\n{}: {}'.format(
+                    _('The payment for this invoice has already been received.'),
+                    _('PayPal payment ID'),
+                    payment.info_data['id'],
+                    _('PayPal sale ID'),
+                    payment.info_data['transactions'][0]['related_resources'][0]['sale']['id']
```
raphaelm: Sorry to give this another spin. For both things, we need to be able to not crash when these resources do not exist, e.g. because of some weird circumstances an incomplete payment resource is stored. At the very least, we need to catch KeyError and IndexError here :)

PR #1473, opened 2019-10-29
PDF renderer: Properly escape HTML answer fields

thegcat:


raphaelm:
Right, I forgot that one!

PR #1474, opened 2019-10-30
missing dot in docker command

telegnom:
There is a missing dot in the docker command to create a custom build with plugins. Inserted that missing dot.
Should be `docker build . -t mypretix` instead of `docker build -t mypretix`


raphaelm:
Thanks!

PR #1475, opened 2019-10-31
Two more things to escape in the PDF renderer

thegcat:
I'd also think the better place to escape the values of variables would be in the same `get_text_content` method so that everything is localised there. This would require moving all the `.replace("\n", "<br/>\n")` there too, it would in my eyes be an improvement though, as the knowledge about the PDF renderer needing HTML to be escaped and newlines to be converted to `<br/>` would be in `get_text_content` and there only.

I can prepare a corresponding pull request if this is an idea which would work for you.

### DIFF
--- a/src/pretix/base/pdf.py
+++ b/src/pretix/base/pdf.py
@@ -405,9 +405,9 @@ class Renderer:
         if not o['content']:
             return '(error)'
         if o['content'] == 'other':
-            return o['text'].replace("\n", "<br/>\n")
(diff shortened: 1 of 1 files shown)


raphaelm:
I did not escape these to values on purpose, and for the same purpose did not implement the escape in the renderer itself: This would be backwards-incompatible, as I know of people out there using <br/> and the like in their custom content. I'm not yet sure how we want to handle this.

raphaelm:
You are right, we should use a different approach. Superseded by #1556

PR #1481, opened 2019-11-11
Restrict Payment Providers to Sales Channels

pc-coholic:
This adds a Sales-Channels Multi-Select to each Payment Provider, allowing the user to restrict the usage of said Payment Provider to specific sales channels.

Backwards compatibility:
If no setting is saved, the system assumes the `web`-sales channel as enabled, not breaking anything for existing events.

However, upon next editing of the Payment Provider Settings by the user, the field is required, forcing the user to actively select which sales channels the payment provider should be enabled for. (Positive?) side effect: The user cannot configure a payment provider to be enabled but not approved for any sales channel.

Also changes `is_implicit` from a property to a method which gets the `request` passed. Legacy-compatibility should be provided by the changes.

### DIFF
--- a/src/pretix/base/payment.py
+++ b/src/pretix/base/payment.py
@@ -283,8 +283,9 @@ class BasePaymentProvider:
                  label=_('Restrict to specific sales channels'),
                  choices=(
                      (c.identifier, c.verbose_name) for c in get_all_sales_channels().values()
-                     if c.identifier != 'pretixpos'
+                     if c.payment_restrictions_supported
                  ),
+                 initial=['web'],
                  widget=forms.CheckboxSelectMultiple,
                  help_text=_(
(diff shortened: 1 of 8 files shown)


src/pretix/base/payment.py
```diff
@@ -278,8 +278,20 @@ def settings_form_fields(self):
                  required=False,
                  disabled=not self.event.settings.invoice_address_required
              )),
+            ('_restrict_to_sales_channels',
+             forms.MultipleChoiceField(
+                 label=_('Restrict to specific sales channels'),
+                 choices=(
+                     (c.identifier, c.verbose_name) for c in get_all_sales_channels().values()
+                     if c.identifier != 'pretixpos'
```
raphaelm: Hmm, that's a little special-casey. Do you think we should add a flag to sales channels?

PR #1483, opened 2019-11-13
New signal: ticketoutput_override_layout

pc-coholic:
New signal that allows plugins to override the layout that the PDF-ticketoutput is using.

### DIFF
--- a/src/pretix/plugins/ticketoutputpdf/ticketoutput.py
+++ b/src/pretix/plugins/ticketoutputpdf/ticketoutput.py
@@ -79,7 +79,9 @@ class PdfTicketOutput(BaseTicketOutput):
         merger = PdfFileMerger()
         with language(order.locale):
             for op in order.positions_with_tickets:
-                layout = ticketoutput_override_layout.send(order.event, op) or self.layout_map.get(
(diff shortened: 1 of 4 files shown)


src/pretix/plugins/ticketoutputpdf/ticketoutput.py
```diff
@@ -78,7 +79,9 @@ def generate_order(self, order: Order):
         merger = PdfFileMerger()
         with language(order.locale):
             for op in order.positions_with_tickets:
-                layout = self.layout_map.get(
+                layout = ticketoutput_override_layout.send_chained(
+                    order.event, 'layoutoverride', orderposition=op
+                ) or self.layout_map.get(
```
raphaelm: With

```
                layout = ticketoutput_override_layout.send_chained(
                    order.event, 'layout', orderposition=op, layout=self.layout_map.get(
                        …
                    )
                )
```

we could even make use of ``send_chained``, such that the signal receives the originally selected layout in a ``layout`` argument and can factor it into the decision. This will be passed iteratively to all receivers. Of course, we should then document the new parameter.

(This is a nice to have, not a must, we can always add it later.)

Issue #1485, opened 2019-11-14
Assign team limitation to event during creation.


MaicoTimmerman:
With the introduction of pluggable authentication (#775) we are now able to provision our users automatically with teams upon login. We allow any of these users to create new events, however we would like the new events to be limited to these teams only. I could not find any open or closed issues except for #914, which is a somewhat similar issue.

Currently when a user has no team allowed access for all events under an organizer, a new team is provisioned:
```python
 t = Team.objects.create(
                    organizer=event.organizer, name=_('Team {event}').format(event=event.name),
                    can_change_event_settings=True, can_change_items=True,
                    can_view_orders=True, can_change_orders=True, can_view_vouchers=True,
                    can_change_vouchers=True
                )
                t.members.add(self.request.user)
                t.limit_events.add(event)
```

We would like the user to be able to select a team during event creation.

My suggestion would be to add a step during event creation, which checks whether the user is in a team with control rights, similar to the current check in the Event wizard:

```
            has_control_rights = self.request.user.teams.filter(
                organizer=event.organizer, all_events=True, can_change_event_settings=True, can_change_items=True,
                can_change_orders=True, can_change_vouchers=True
            ).exists()
```

If such a team is not present, show an additional final event creation step in which the user can select a team or would like a team to be provisioned.

I'd be happy to make an attempt at such a pull request.

raphaelm:
Hmm, I see the problem here, and I also am sometimes annoyed by the many auto-generated teams, so I think it makes sense to let users optionally select an existing team during event creation instead of a new one being created. It doesn't really fit the data model of teams (in theory, you'd need to select N teams that should gain access to the event), but it's a start.

However, I'm cautious of making the event creation process more complex, as this is the very first thing people see of pretix and it should therefore be clear and fricitonless. Most people do not know what teams are at that point.

Therefore, I'd suggest:

* Not adding an additional step, but instead just a form field to the second step
* Only showing the field in the case where we would currently create a new team, as suggested

Feel free to create a PR! :)

MaicoTimmerman:
@raphaelm I have created a pull request, please let me know what you think.

PR #1487, opened 2019-11-14
Feature/optional team provisioning

MaicoTimmerman:
Implementation for #1485

### DIFF
--- a/src/pretix/control/forms/organizer.py
+++ b/src/pretix/control/forms/organizer.py
@@ -213,12 +213,11 @@ class OrganizerSettingsForm(SettingsForm):
     )
 
     event_team_provisioning = forms.BooleanField(
-        label=_('Allow team provisioning for newly created events.'),
+        label=_('Allow creating a new team during event creation'),
         help_text=_('Users that do not have access to all events under this organizer, must select one of their teams '
                     'to have access to the created event. This setting allows users to create an event-specified team'
-                    'on-the-fly.'),
+                    ' on-the-fly, even when they do not have \"Can change teams and permissions\" permission.'),
         required=False,
-        initial=True,
     )
 
     primary_color = forms.CharField(
--- a/src/pretix/control/forms/organizer.py
+++ b/src/pretix/control/forms/organizer.py
@@ -211,6 +211,16 @@ class OrganizerSettingsForm(SettingsForm):
         widget=I18nTextarea,
         help_text=_('Not displayed anywhere by default, but if you want to, you can use this e.g. in ticket templates.')
     )
+
+    event_team_provisioning = forms.BooleanField(
+        label=_('Allow team provisioning for newly created events.'),
+        help_text=_('Users that do not have access to all events under this organizer, must select one of their teams '
+                    'to have access to the created event. This setting allows users to create an event-specified team'
+                    'on-the-fly.'),
+        required=False,
+        initial=True,
+    )
+
     primary_color = forms.CharField(
         label=_("Primary color"),
         required=False,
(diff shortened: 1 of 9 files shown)


src/pretix/control/views/main.py
```diff
@@ -228,19 +228,19 @@ def done(self, form_list, form_dict, **kwargs):
...
-                    can_view_orders=True, can_change_orders=True, can_view_vouchers=True,
-                    can_change_vouchers=True
-                )
-                t.members.add(self.request.user)
-                t.limit_events.add(event)
+            if not EventWizardBasicsForm.has_control_rights(self.request.user, event.organizer):
+                if basics_data["team"] is not None:
+                    t = basics_data["team"]
+                    t.limit_events.add(event)
+                elif event.organizer.settings.get("event_team_provisioning", "True"):
```
raphaelm: I'm not sure if the logic is the right way here – if the setting is True, the user needs to fill in the new field, right?

src/pretix/control/forms/organizer.py
```diff
@@ -211,6 +211,16 @@ class OrganizerSettingsForm(SettingsForm):
...
         help_text=_('Not displayed anywhere by default, but if you want to, you can use this e.g. in ticket templates.')
     )
+
+    event_team_provisioning = forms.BooleanField(
+        label=_('Allow team provisioning for newly created events.'),
+        help_text=_('Users that do not have access to all events under this organizer, must select one of their teams '
+                    'to have access to the created event. This setting allows users to create an event-specified team'
+                    'on-the-fly.'),
+        required=False,
+        initial=True,
```
raphaelm: initial=True won't do it for existing organizers. if it stays, we need to define the default value in pretix/base/settings.py instead.

MaicoTimmerman:
Thanks for the response! I'll adjust the suggestions you made.

Regarding the setting. In our system users are logging in with OAuth2.0, based on the response of an external API we set the teams of the user. This means that the user will be removed from any provisioned teams that might have been created in the past. To prevent our system from flooding with 
teams without any users, therefore I'd really like to keep the setting.

I agree it can be confusion for end-users to have many settings. I can move the setting, including defaults to pretix/base/settings.py and remove it from the organizer levels settings.

Please let me know what you think.

raphaelm:
> Regarding the setting. In our system users are logging in with OAuth2.0, based on the response of an external API we set the teams of the user. This means that the user will be removed from any provisioned teams that might have been created in the past. To prevent our system from flooding with teams without any users, therefore I'd really like to keep the setting.

I don't quite understand. I'm proposing to add the form field and the new behaviour you proposed in #1485 for everyone, so why do you need the setting to avoid a flooding with teams?

MaicoTimmerman:
I think there might be some confusion in what the setting I created is for. The setting is only there to disallow any team provisioning, by not showing the option in the dropdown. The dropdown will appear always when applicable, the setting only affects whether the "empty" option is shown. When the empty option is chosen, a team is provisioned.

raphaelm:
Yes, in fact I did not understand that at all from reading the code :sweat_smile: Could you try to clarify the description and naming of the setting a little? I think the word "provisioning" doesn't mean a lot to most people in that scenario.

The setting could be something along the lines of "Allow creating a new team during event creation" (while the help text clarifies that this also affects users without can_change_teams), while the empty label could be something like "Create a new team for this event with me as the only member" (or if you can come up with something shorter, feel free)

src/pretix/control/forms/organizer.py
MaicoTimmerman: I've moved the initial defaults to pretix/base/settings.py

src/pretix/control/views/main.py
raphaelm: Since we added a proper default, we should be able to just do
```suggestion
                elif event.organizer.settings.event_team_provisioning:
```

raphaelm:
Hmm, I don't seem to have permission to amend this PR; so if you could just apply that one suggest we're ready to go.

PR #1491, opened 2019-11-17
Remove 'no selection' from addon products bulletList

gilbertf:
In the case exacly one addon product has to be choosen by the user the artifical option "no selection" is not required. By this modification the UI experience is improved.
A
B


raphaelm:
Hi! Thanks for suggesting this and preparing the PR! Unfortunately, it is not correct, or at least not as simple. In theory, the category of add-on products could contain multiple products with variations, in which case only one variation of one of the products may be selected and this will hence not allow the user to correct their choice any more :(

PR #1502, opened 2019-11-23
Send widget_data when opening waiting list

MaicoTimmerman:
Currently the waiting list button of the widget does not send the widget_data to the server. Meaning that after successfully registering in the waitinglist, the cart session will not contain the widget data.

We'd like the widget data to be available, as this is the way we pass meta data of the user to the pretix application. Our plugin will save additional information in the database depending on the widget data in the cart session.


raphaelm:
Thanks! It makes sense, even though it might cause some confusion since the widget data is not actually used for the waiting list request itself in any way.

PR #1504, opened 2019-11-26
Allow to create vouchers for *all* products

raphaelm:
This was requested to allow e.g. a global 20% discount across all (sub)events. Of course, it's not really useful once you have products that require a voucher, but there are enough people using pretix without this constraint.


pc-coholic:
Looks good to me.

However one thing that might cause confusion:
When a voucher is created in an event-series and limited to one specific performance of the event and the voucher is entered on the page of another performance, the shop will send the user to the correct event-page. Correct behavior - but might be confusing if it goes unnoticed by the user.

But I guess, that is not an issue tied specifically to this PR

raphaelm:
That is in fact confusing, but not new with this PR, I created #1505 to track this issue!

PR #1506, opened 2019-11-27
Add "Are you sure?"-dialog to dirty forms

pc-coholic:
Z#144010

### DIFF
--- a/src/pretix/static/pretixcontrol/js/ui/main.js
+++ b/src/pretix/static/pretixcontrol/js/ui/main.js
@@ -744,5 +744,7 @@ $(function () {
 });
 
 $(document).ready(function () {
-    $('.form-horizontal').areYouSure( {'message': gettext('You have unsaved changes!')});
+   $('form[method=post]').filter(function () {
+       return $(this).find("button:not([type=button]), input[type=submit]").length > 0;
+   }).areYouSure( {'message': gettext('You have unsaved changes!')});
 });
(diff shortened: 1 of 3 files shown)


src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -742,3 +742,7 @@ $(function () {
     $("#ajaxerr").on("click", ".ajaxerr-close", ajaxErrDialog.hide);
     moment.locale($("body").attr("data-datetimelocale"));
 });
+
+$(document).ready(function () {
+    $('.form-horizontal').areYouSure( {'message': gettext('You have unsaved changes!')});
```
raphaelm: Hmm, I'm not sure if using ``form-horizontal`` really follows the principle of least surprise and won't come back to bite us in the future. For example, this would not make it trigger for the "Comment" field on the order detail page, which is a field you really easily forget to save.

To exclude all search fields, we could use ``form[method=post]``. This might cause some problems with hidden/AJAX-enabled forms, so maybe we could add a ``.filter`` statement to get only those which have a submit button?

```
$('.form-horizontal').filter(function () { return $(this).find("button:not([type=button]), input[type=submit]").length > 0; })…
```

Haven't tested this yet.

pc-coholic: This actually seems to be working quite nicely! Did a few tests in the frontend and couldn't find any fields, where it misfired.

PR #1510, opened 2019-12-04
Improve email contents

### LINKED ISSUE #1415: Use proper MIME structure
Since #1358, we've been sending emails with a MIME structure like this:

```
multipart/related
   multipart/alternative
      text/plain
      text/html
   image/png (images for HTML content)
   application/pdf (tickets)
   …
```

Unfortunately, this broke attachments on Apple devices entirely: iOS Mail no longer showed any attachments. As a quick fix, in ca0c0f4ae, I changed ``multipart/related`` to ``multipart/mixed``. This has the upside of showing all attachments again, but the downside of showing the PNG images as attachments as well.

The correct structure would look like this:

```
multipart/mixed
   multipart/alternative
      text/plain
      multipart/related
         text/html
         image/png (images for HTML content)
   application/pdf (tickets)
   …
```

I verified that this yields the correct results on iOS. What's left to do is changing our code to generate this structure and verify that it works as intended in at least GMail and Outlook.

raphaelm:
- [x] Fix #1415
- [x] Improve looks in dark mode

Test on

- [x] GMail web
- [x] GMail Android
- [x] Apple Mail on macOS
  - [x] dark mode
- [x] Apple Mail on iOS (passbook!)
- [x] Thunderbird
- [x] K9 Mail
- [x] Outlook 365
- [x] Hotmail web
- [x] GMX web

### DIFF
--- a/src/pretix/base/services/mail.py
+++ b/src/pretix/base/services/mail.py
@@ -14,7 +14,10 @@ import requests
 from bs4 import BeautifulSoup
 from celery import chain
 from django.conf import settings
-from django.core.mail import EmailMultiAlternatives, get_connection
+from django.core.mail import (
+    EmailMultiAlternatives, SafeMIMEMultipart, get_connection,
+)
+from django.core.mail.message import SafeMIMEText
 from django.template.loader import get_template
 from django.utils.translation import pgettext, ugettext as _
 from django_scopes import scope, scopes_disabled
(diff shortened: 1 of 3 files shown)


raphaelm:
Looks decent everywhere in light mode:
https://app.emailonacid.com/app/acidtest/nX2v4x12FKeiTrIzVPfAbEFAReULXZUc4lXWSFz8MFKAf/list

Still looks weird in macOS dark mode, but ew can only fix that by breaking Outlook, and it's already better than before.

PR #1514, opened 2019-12-05
Add itemcount (Items, excluding AddOns) to fragment_cart

pc-coholic:
Z#2354436


src/pretix/presale/views/__init__.py
```diff
@@ -173,6 +173,7 @@ def keyfunc(pos):
             'minutes_left': minutes_left,
             'seconds_left': seconds_left,
             'first_expiry': first_expiry,
+            'itemcount': sum(c.count for c in positions if not c.addon_to and hasattr(c, 'count'))
```
raphaelm: what is the ``not hasattr(c, 'count')`` case? Is this a case we really want to display ``0`` in the cart?

PR #1523, opened 2019-12-13
Fix #1521 -- External authenticated users cannot delete events

### LINKED ISSUE #1521: External authenticated users cannot delete events.
With the release of #775 users are able to authenticate using external password validation. Deleting events currently requires the password of the user. Is there any way we can update this, such that external authenticated users can also delete events? I'd be open to implementing a solution, however would like your opinion on how such feature should work.

MaicoTimmerman:
Adds possibility for reauthenticating backends with authentication_url auth method. Additionally replaces all seperate password checks with recent authentication required.

### DIFF
--- a/src/pretix/control/forms/event.py
+++ b/src/pretix/control/forms/event.py
@@ -1,6 +1,5 @@
 from django import forms
 from django.conf import settings
-from django.contrib.auth.hashers import check_password
 from django.core.exceptions import ValidationError
 from django.core.validators import RegexValidator, validate_email
 from django.db.models import Q
--- a/src/pretix/control/forms/event.py
+++ b/src/pretix/control/forms/event.py
@@ -1436,11 +1436,6 @@ class EventDeleteForm(forms.Form):
         'pw_current_wrong': _("The password you entered was not correct."),
         'slug_wrong': _("The slug you entered was not correct."),
     }
-    user_pw = forms.CharField(
-        max_length=255,
-        label=_("Your password"),
-        widget=forms.PasswordInput()
-    )
     slug = forms.CharField(
         max_length=255,
         label=_("Event slug"),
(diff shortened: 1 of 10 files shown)


src/pretix/control/forms/event.py
```diff
@@ -1436,31 +1436,15 @@ class EventDeleteForm(forms.Form):
         'pw_current_wrong': _("The password you entered was not correct."),
```
raphaelm: This can probably go as well now :)

src/pretix/control/templates/pretixcontrol/user/reauth.html
```diff
@@ -8,7 +8,11 @@
         {% csrf_token %}
         <h3>{% trans "Welcome back!" %}</h3>
         <p>
-            {% trans "We just want to make sure it's really you. Please re-enter your password to continue." %}
+            {% if form.backend.url %}
+                {% blocktrans trimmed with verbose_name=form.backend.verbose_name %}We just want to make sure it's really you. Please re-authenticate with '{{ verbose_name}}'.{% endblocktrans %}
```
raphaelm: Can you use e.g. ``login_provider`` instead of ``verbose_name`` in the translated string? It's just more helpful to the translaters to understand whats going on :)

src/pretix/control/views/shredder.py
```diff
@@ -103,7 +104,7 @@ def post(self, request, *args, **kwargs):
         if constr:
             return self.error(ShredError(self.get_error_url()))
 
-        if not self.request.user.check_password(request.POST.get("password")):
-            return self.error(ShredError(_("The current password you entered was not correct.")))
+        if request.event.slug != request.POST.get("slug"):
+            return self.error(ShredError(_("The current slug you entered was not correct.")))
```
raphaelm: Current slug → event slug? Best take the same error message as with event deletion, this saves translator time :)

PR #1534, opened 2019-12-20
Expose help texts on questions' API

patrick91:
Hi there! We are using the API to show the list of questions and I notice that the help text is missing, this PR adds that and adds a test for it as well :)


raphaelm:
Amazing, good catch! Could you also update the ``questions.rst`` file in the docs (both the table at the top as well as the examples and a change comment)?

patrick91:
@raphaelm done! I guessed the version number, not sure how that works 😊

raphaelm:
We don't change the API in patch releases, so it's going to be 3.5, but I fixed that :)

Issue #1229, opened 2019-03-24
Login is not working after updating to 2.5.0


ZEROF:
HI,

After upgrading to 2.5 I can't login any more, I get this error:

    Verification failed
    We could not verify that this request really was sent from you. For security reasons, we      therefore cannot process it.
    Please go back to the last page, refresh this page and then try again. If the problem persists, please get in touch with us.
    Take a step back 
    I got back to loging page, I try again, and I get same error, over and over.

**This is update output:**

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/pretix/src/pretix/__main__.py", line 9, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 381, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.6/site-packages/django/core/management/__init__.py", line 375, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/base.py", line 316, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/base.py", line 353, in execute
    output = self.handle(*args, **options)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/base.py", line 83, in wrapped
    res = handle_func(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/django/core/management/commands/migrate.py", line 85, in handle
    executor.loader.check_consistent_history(connection)
  File "/usr/local/lib/python3.6/site-packages/django/db/migrations/loader.py", line 297, in check_consistent_history
    connection.alias,
django.db.migrations.exceptions.InconsistentMigrationHistory: Migration pretixbase.0089_auto_20180315_1322 is applied before its dependency pretixapi.0001_initial on database 'default'.

Full output can be found here from docker logs (docker logs -f):

https://paste.lugons.org/?1fa5a1f35726c15e#y7ljyeYpRWssgOilncotLNeYbqRzvu8+F4pSpMhl3HE=

raphaelm:
[...]

PR #1539, opened 2020-01-02
Fix when CSRF cookies on login missing

mistressofjellyfish:
For some reason, I did not get a CSRF token cookie set when visiting the login screen during local development of the google oauth backend mentioned in my other PR. Adding this decorator made sure I could log in as I would finally receive the cookies.

### DIFF
--- a/src/pretix/control/views/auth.py
+++ b/src/pretix/control/views/auth.py
@@ -17,6 +17,7 @@ from django.urls import reverse
 from django.utils.functional import cached_property
 from django.utils.http import is_safe_url
 from django.utils.translation import ugettext_lazy as _
+from django.views.decorators.csrf import ensure_csrf_cookie
 from django.views.generic import TemplateView
 from django_otp import match_token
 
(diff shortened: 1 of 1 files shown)


raphaelm:
Hmm, this certainly does not harm but I must say I'm puzzled why it is necessary in your case.

mistressofjellyfish:
I have no idea why I needed that. I followed the development setup from the docs; developing on macOS 10.14.6.

The purpose of this PR is more to provide a google-able solution if this problem is encountered by someone else, as it took me a good half a day trying to figure out what's going on. This solution certainly doesn't address the underlying problem, but eh - for my purposes, it worked.

mistressofjellyfish:
Digging a little bit through the issues, this could be the same scenario as #1229. It'd be interesting to try and reproduce that.

raphaelm:
If you remove the line again, can you reproduce it? If so, can you check in the developer tools if the cookie is actually not sent or if your browser is just not saving it due to problematic cookie flags?

mistressofjellyfish:
Unfortunately, I can't reproduce it now. However, when I checked, the cookie wasn't set in the response headers shown in the django debug toolbar.

raphaelm:
Closing this for now

PR #1542, opened 2020-01-08
Limit Payment Term Days Field

pajowu:
When no payment term is required, some users set this field to a very high value. This produces an "internal error" when a customer creates an order (`timedeltas` can only store up to 999999999 days, `date(time)s` only store until the year 9999). This PR limits the field to 1 million days which is still more than 2700 years.


raphaelm:
:joy: :see_no_evil:

pc-coholic:
Screenshot from 2020-01-08 17-00-42

PR #1546, opened 2020-01-13
Send signal on checkin

pajowu:
Plugins might need this signal to trigger own processes when a user is checked in. One example might be sending a greeting email or activating features on-site. This PR sends a signal when creating a new checkin.

I have two questions regarding this change:
1. Do you think the signal should also be send if a user is force-checked-in? (https://github.com/pretix/pretix/blob/master/src/pretix/base/services/checkin.py#L152)
2. Does the signal need to be added to the documentation on other places?


raphaelm:
> Do you think the signal should also be send if a user is force-checked-in? (/src/pretix/base/services/checkin.py@master#L152)

No, I think that's fine.

> Does the signal need to be added to the documentation on other places?

No, that's fine, too.

I'm unsure about the name of the signal, maybe ``order_position_checked_in`` or ``checkin_created`` would be more consistent?

pajowu:
Thank you very much for the lightning-fast reply. I renamed the signal to `checkin_created`, which is indeed a much better fitting name.

PR #1561, opened 2020-01-24
Solve some Docker Setup scaling issues

gnomus:
Currently the first Scaling Problem you run into while hosting the Small Scale Docker Setup is the sysctl Setting of maximum parallel Socket Connections which defaults to 128.

This means after 128 parallel connections the nginx starts replying with 502 Errors.

If you increate this Limit the next Problem will be the 768 Worker Connections of nginx.

To solve both of these Issues this Pullrequest sets them both to a Value of 4096.

### DIFF
--- a/deployment/docker/nginx.conf
+++ b/deployment/docker/nginx.conf
@@ -4,7 +4,7 @@ pid /var/run/nginx.pid;
 daemon off;
 
 events {
-	worker_connections 768;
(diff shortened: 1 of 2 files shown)


raphaelm:
:+1: Nice! Not sure how many small-scale pretix setups will survive 4096 connections on the next layer, but let's see how it works out ;)

PR #1563, opened 2020-01-25
Comparison for file extensions should be case insensitive

ahus1:
This takes care for situations when buyers upload files with capital letters in the extension (like .JPG); I suppose this is common on Windows OS.

I found this first when buyers uploaded images to questions that ended with .JPG. I browsed the code and found that this is in orders.py

I looked further and found two other places that looked similar. 

As I currently don't have a working python setup I was unable to test this change locally. I would be grateful if a reviewer could test it. 
If this is not possible, please let me know and I'll setup a VM with a python environment.

### DIFF
--- a/src/pretix/base/models/orders.py
+++ b/src/pretix/base/models/orders.py
@@ -900,7 +900,7 @@ class QuestionAnswer(models.Model):
 
     @property
     def is_image(self):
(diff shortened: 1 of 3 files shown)


ahus1:
(build failed due to a spelling error present in master branch, I plead "not guilty")

raphaelm:
Thanks a lot!

PR #1576, opened 2020-02-10
Add meta_data for items

pc-coholic:
This extends the concept of `MetaProperty` on Organizer-Level -> `MetaValue` on Event-Level to `MetaProperty` on Event-Level to `MetaValue` on Item-Level.

Unfortunately, this will require also an update to all pretixSCAN-apps, to handle the new `itemmeta:`-tag correctly.

I know this has been requested in at least one more ticket, but I can't find it right now...

Ref: Z#2356069


src/pretix/control/views/typeahead.py
```diff
@@ -589,3 +590,37 @@ def meta_values(request):
...
+        raise PermissionDenied()
+
+    defaults = defaults.filter(event__organizer_id=organizer.pk)
+    matches = matches.filter(item__event__organizer_id=organizer.pk)
+    all_access = (
+        (request and request.user.has_active_staff_session(request.session.session_key))
+        or request.user.teams.filter(all_events=True, organizer=organizer).exists()
+    )
+    if not all_access:
+        matches = matches.filter(event__id__in=request.user.teams.values_list('limit_events__id', flat=True))
```
raphaelm: Why not filter defaults as well?

PR #1594, opened 2020-02-28
Add option to hide the price of free products

pajowu:
This PR adds a display setting to hide the price for products if they are free. A few questions that popped up for me:
- Should this hide the price if only some products are free or only if all products of the event are free?
- I currently assumed that if the price is zero, the tax is also 0. Is that a valid assumption or should I update the templates to also check that the tax is also 0?
- The setting uses the term "free product" / "the product is free". I also saw the use of "free price" for the "select what you want"-price. Should I change the name of the setting?

Note: this only removes the price from the checkout process, not from tickets/invoices.

### DIFF
--- a/src/pretix/api/serializers/event.py
+++ b/src/pretix/api/serializers/event.py
@@ -550,6 +550,7 @@ class EventSettingsSerializer(serializers.Serializer):
         'show_variations_expanded',
         'hide_sold_out',
         'meta_noindex',
+        'hide_free_prices',
         'redirect_to_checkout_directly',
         'frontpage_subevent_ordering',
         'frontpage_text',
--- a/src/pretix/base/settings.py
+++ b/src/pretix/base/settings.py
@@ -716,6 +716,16 @@ DEFAULTS = {
             label=_('Ask search engines not to index the ticket shop'),
         )
     },
+    'hide_free_prices': {
+        'default': 'False',
+        'type': bool,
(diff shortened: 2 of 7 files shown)


raphaelm:
> I currently assumed that if the price is zero, the tax is also 0. Is that a valid assumption or should I update the templates to also check that the tax is also 0?

That's fine, taxes (in pretix) are always a percentage, and any percentage of zero is zero.

> The setting uses the term "free product" / "the product is free". I also saw the use of "free price" for the "select what you want"-price. Should I change the name of the setting?

I think that's fine.

---

However, I'm pretty unsure we want this PR in this way at all. It adds a setting and lots of additional cases for a very small effect. pretix is very complex already and we need to avoid both settings as well as additional cases to test for wherever possible.

An alternative might be to get rid of the setting and just hide prices when all products are free by default. However, this still adds lots of if/else cases.

Can you tell me a little more about the idea behind this PR? Did showing "FREE" and "0,00" cause actual user confusion?

pajowu:
FREE/0.0 caused issues when pretix was used for registering for an event but the participation fee was either not fixed when the registration started and/or was collected during the event seperately. Currently this was solved by patching the template files

pajowu:
After discussing this issue again with the event we would use it at we have found a better solution that does not need this PR. Therefore I will close it for now but would be available to fix any things that would be needed if you still want to merge it

PR #1603, opened 2020-03-11
Option to hide date-block on frontpage

pc-coholic:
When enabled, the date-block will be completely removed from the ticketshop start page and the (public) organizer list view.

In the backend, the date-block is always displayed, no matter the setting.


raphaelm:
Hmm, we could also make it affect the event header (if no logo is set), that's basically the same thing. Otherwise, this looks good.

pc-coholic:
Good catch - added!

PR #1605, opened 2020-03-11
Subclass MultipleChoiceField to serialize to list

pajowu:
The rest frameworks `serializers.MultipleChoiceField` `to_internal_value` functions returns a set instance[¹]. hierarkey did not like that which lead to internal server errors. Since the code generally assumes that the value of those fields is a list I subclassed MultipleChoiceField so that it now returns a `list` and added a test for that

[¹]: https://www.django-rest-framework.org/api-guide/fields/#multiplechoicefield

### DIFF
--- a/src/pretix/api/serializers/fields.py
+++ b/src/pretix/api/serializers/fields.py
@@ -1,6 +1,6 @@
 from rest_framework import serializers
 
 
-class MultipleChoiceField(serializers.MultipleChoiceField):
+class ListMultipleChoiceField(serializers.MultipleChoiceField):
     def to_internal_value(self, data):
         return list(super().to_internal_value(data))
--- a/src/pretix/api/serializers/fields.py
+++ b/src/pretix/api/serializers/fields.py
@@ -1,6 +1,29 @@
+from collections import OrderedDict
+
 from rest_framework import serializers
+
+
+def remove_duplicates_from_list(data):
+    return list(OrderedDict.fromkeys(data))
 
 
 class ListMultipleChoiceField(serializers.MultipleChoiceField):
     def to_internal_value(self, data):
-        return list(super().to_internal_value(data))
+        if isinstance(data, str) or not hasattr(data, '__iter__'):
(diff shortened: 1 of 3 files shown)


src/pretix/api/serializers/fields.py
```diff
@@ -0,0 +1,6 @@
+from rest_framework import serializers
+
+
+class MultipleChoiceField(serializers.MultipleChoiceField):
```
raphaelm: To avoid confusion, we should probably give this it's own name, e.g. ``ListMultipleChoiceField``.

That said, do we need to provide an even more custom implementation of ``to_internal_value`` since for locales, it actually matters in which order they are set. We don't expose that in the interface, but it would be weird if saving it through the API would "randomize"  the order.

pajowu: I renamed the class as suggested. I don't quite understand the seconds comment? The field stores the selected options and from what I understand, the order of fields in a MultipleChoiceField are given by the `choices`. I don't know if there is any guarantee about the order in Django itself (I guess it doesnt because the rest framework seems to ignore order). What is the order of the active languages of an event used for?

raphaelm: What I was refering to, is:

``MultipleChoiceField.to_internal_value`` creates a ``set`` using a set comprehension literal:
```
        return {
            super(MultipleChoiceField, self).to_internal_value(item)
            for item in data
        }
```

However, unlike dicts, sets still do not have any guarantee on their ordering. So, if I push ``["en", "de"]`` to the API, I might end up with ´`["de", "en"]`` in the database to day and with ``["en", "de"]`` tomorrow, which might be very confusing.

Therefore, I think we should not use ``list(super().…)`` but instead *copy* the code from ``MultipleChoiceField`` and change the set comprehension to a list comprehension.

pajowu: Oh yes you are right I was only thinking about other uses in the code not about consitency of the data from the API. I fixed that and force-pushed to remove a small mistake

PR #1606, opened 2020-03-11
Send widget_data in voucher redeem request.

MaicoTimmerman:
When the checkout process on mobile starts with redeeming a voucher the widget data is not sent. This change adds a hidden field which sends adds the widget data in de GET request, similar to the pretix-button variant of the widget. See [src/pretix/static/pretixpresale/js/widget/widget.js#1047](https://github.com/pretix/pretix/blob/359c46b8e28c8c5e61bedceb6c1723f5ce313b79/src/pretix/static/pretixpresale/js/widget/widget.js#L1047) for reference.


raphaelm:
Good catch, thanks!

Issue #1586, opened 2020-02-18
[External authentication] Login/Reauth next url using oauth state parameters


MaicoTimmerman:
Currently, when presented with the login or reauth screen the `next` query parameter is populated with the url which the user should be send to after login.

According to [RFC6819, an OAuth 2.0 security review from 2013](https://tools.ietf.org/html/rfc6819#section-5.2.3.3) dynamic redirect URIs should not be used. Due to this, most OAuth providers do not support custom query parameters to be reused over the authentication request. Instead the state parameters is should be used to pass the next parameter back to the application.

The simplest solution would be to also check `request.GET['state']` after all calls to `pretix.base.auth.BaseAuthBackend.request_authenticate`. However that is not really dynamic for methods which use other parameter naming. Do you have any suggestions for a more elegant solution of this problem?

PR #1609, opened 2020-03-12
Make next url authentication backend dependent

MaicoTimmerman:
This is a proposal for a fixing #1586. Pretix ever only checks the 'next' GET parameter, however redirect based authentication methods like OAuth use custom named parameters. In case of OAuth this `state`.

I've added additionally functionality to the authentication backend to select the next URL based on the backend used. The default backend and thus all other backends still uses 'next', however my custom backend will be able to implement it using a different parameter name.

### DIFF
--- a/src/pretix/base/auth.py
+++ b/src/pretix/base/auth.py
@@ -85,6 +85,16 @@ class BaseAuthBackend:
         """
         return
 
+    def next_url(self, request):
+        """
+        This method will be called after a successful login to determine the next URL. Pretix in general uses the
+        ``'next'`` query parameter. However, external authentication methods could use custom attributes with hardcoded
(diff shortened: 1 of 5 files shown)


src/pretix/base/auth.py
```diff
@@ -85,6 +85,16 @@ def authentication_url(self, request):
         """
         return
 
+    def next_url(self, request):
```
raphaelm: Can we name this ``get_next_url``? Feels a little more consistent with Django things

raphaelm:
Ah, also, do you think it would be useful/simple enough to add a test for this? :wink:

MaicoTimmerman:
> Ah, also, do you think it would be useful/simple enough to add a test for this?

The default case, `next` is already converted by the current test set. I've added a test for a redirect based on a custom named query parameter `state`.

PR #1611, opened 2020-03-12
Do not validate event creation form on back button

pajowu:
`django-formtool`s `SessionWizardView`does not store the entered form if "wizard_goto_step" is used[1]. Therefore it is not needed to validate the data, which currently lead to confusion when we tried to create an event but wanted to change things in the previous step.

[1]: https://django-formtools.readthedocs.io/en/latest/wizard.html#formtools.wizard.views.WizardView.render_goto_step


raphaelm:
This looks useful, I didn't know about this attribute!

PR #1620, opened 2020-03-23
Fix sample service config

bhaettasch:
Add missing "\" to sample for service configuration when using docker as deployment strategy


raphaelm:
Uh, thank you!

PR #1621, opened 2020-03-24
Render remaining badges onto new page even if it doesn't fill the page

pajowu:
When the number of badges was not a multiple of the badges per page, the last few badges were not rendered. This PR produces badge pdfs where the last page has some empty slots.


raphaelm:
Ouch! Thanks :DD

PR #1624, opened 2020-03-24
Send checkin list mapping in event_copy_data signal

pajowu:
The signal already sends a map of tax, category, item etc ids to their new items but not for checkin lists.


raphaelm:
Hi! LGTM, but could you add this to the definition and documentation of the signal as well?

pajowu:
Oh yes, just added it. Sorry I forgot about this.

PR #1626, opened 2020-03-25
Allow customers to choose to receive their refund as a gift card

raphaelm:


pc-coholic:
LGTM, but:
When Javascript is deactived/not available, the whole becomes unusable and the element for entering the cancellation fee does not make much sense, either.

Screenshot from 2020-03-25 11-26-35

pc-coholic:
Correction: Without JS, the user can enter a value into the input box, that he wants to give as a cancellation fee - the box is not properly styled to show this...

raphaelm:
Okay, since that's not really part of this PR (but the previous one) I'll fix it separetely

PR #1630, opened 2020-03-26
Show Subevent start time in select2-pickers

pc-coholic:


src/pretix/presale/templates/pretixpresale/event/voucher.html
```diff
@@ -13,6 +13,49 @@ <h2>{% trans "Voucher redemption" %}</h2>
     {% if subevent %}
         <h3>{{ subevent.name }}</h3>
     {% endif %}
+    {% if event.settings.show_dates_on_frontpage %}
+        {% with ev=subevent|default:event %}
```
raphaelm: I think I'd pull this into the ``{% if subevent %}`` and *only* show this block if we're dealing with a subevent. If we're on a non-series event, there's less potential for confusion since the date is in the header and there's also only one event with the name.

PR #1656, opened 2020-04-21
Add expiry dates and individual conditions to gift cards

raphaelm:


pc-coholic:
LGTM :ship: 

Tinest of possible improvements: Adding an Edit-Button the giftcard-overview. But I do see the difficulty of having a differentiation between the "view"-button, that brings to a page which allows to edit the value of the giftcard, and the "edit"-button, that brings you to a page to edit the giftcard-properties... So perhaps the current way is actually pretty good... :thinking:

PR #1666, opened 2020-04-28
Add register_event_fonts signal

pc-coholic:
This adds a copycat signal `register_event_fonts`, that behaves the same as `register_fonts` - except that it scopes to individual events.

It works for server-local PDF generation and webfont usage - however it will not work in our apps. I am not sure, if using a font that is not bundled with the apps will break device-local PDF-generation or not.

I guess, we could add a warning in the PDF-editor regarding this...


raphaelm:
Closing for now since we're not sure it's a good idea

PR #1668, opened 2020-05-05
Refactor quota calculation

raphaelm:
This PR refactors quota computation. Instead of methods on the ``Quota`` method, there's now a dedicated ``QuotaAvailability`` for this task. The difference is that the new code is designed to calculate the capacity for multiple quotas *at once*.

Simply speaking, instead of a ``SELECT COUNT(*) FROM orders WHERE event = x AND item = y``, the code now does a ``SELECT COUNT(*), event, item FROM orders GROUP BY event, item`` and then interprets the result Python-side.

This is intended to **massively** speed up many of our critical code paths that are currently slow for some relevant edge cases, such as:

- The list of (sub)events in the shop
- The list of products in an event on the shop frontpage
- The list of (sub)events in the backend
- The list of quotas in the backend
- The backend dashboard
- Cart operations with multiple items

I very very strongly do not want to have this logic twice, so the old code was deleted. This leads to an (hopefully) insignificant slowdown during operations that only affect one quota, since the code is slightly less inefficient in this case (1-2 additional queries per quota, more python-level code to execute). These should be:

- Placing an order
- Marking expired orders as paid, extending expired orders, reactivating orders, …
- Waiting list operations
- Any loops that do not make active use of the new optimization

:fire: This might easily be the most dangerous PR I can remember, we cannot screw this up. Tests are passing, we need to do an extensive review, but I even think we should take a snapshot of the production DB and make sure all quotas have the same result with the old and the new code.

### DIFF
--- a/src/pretix/base/services/quotas.py
+++ b/src/pretix/base/services/quotas.py
@@ -1,15 +1,300 @@
+import sys
+from collections import Counter, defaultdict
 from datetime import timedelta
 
 from django.conf import settings
-from django.db.models import Max, Q
+from django.db.models import Count, F, Func, Max, Q, Sum
 from django.dispatch import receiver
 from django.utils.timezone import now
 from django_scopes import scopes_disabled
 
-from pretix.base.models import Event, LogEntry
+from pretix.base.models import (
+    CartPosition, Event, LogEntry, Order, OrderPosition, Quota, Voucher,
+    WaitingListEntry,
+)
 from pretix.celery_app import app
 
-from ..signals import periodic_task
+from ..signals import periodic_task, quota_availability
+
+
+class QuotaAvailability:
+    """
+    This special object allows so tompute the availability of quotas and inspect their results.
+    Initialize, add quotas with .queue(quota), compute with .compute() and access results.
+    """
+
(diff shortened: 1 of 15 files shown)


raphaelm:
I ran a quota list export for all production quotas, without actively making use of the new optimizations.

With the old code, it took 16 minutes, wit the new code 14 minutes, so even though that wasn't a fair benchmark (machine had other load), we're likely not causing new issues. Proud to day there are no differences:

```
$ diff -u quotas.csv quotas.refactored.csv | wc -l
0
```

src/pretix/base/services/quotas.py
```diff
@@ -1,15 +1,318 @@
...
+from ..signals import periodic_task, quota_availability
+
+
+class QuotaAvailability:
+    """
+    This special object allows so tompute the availability of quotas and inspect their results.
+    Initialize, add quotas with .queue(quota), compute with .compute() and access results.
+    """
+
+    def __init__(self, count_waitinglist=True, ignore_closed=False, full_results=False, early_out=True):
```
pc-coholic, line 26: A little more documentation would be appreciated.

`count_waitinglist` and `ignore_closed` are self-explanatory to me, `full_results` probably too. `early_out` however didn't ring a bell right away...

PR #1671, opened 2020-05-06
Improve global order filters

pc-coholic:
Basically, copy/paste from the order search.

Perhaps just pay specifically attention if this works as intended (`.using()`-statement in conjunction with the annotation: 

https://github.com/pretix/pretix/blob/4b01c42f3193f8083d106139c56e20798b7bef1b/src/pretix/control/views/search.py#L33-L34

### DIFF
--- a/src/pretix/control/forms/filter.py
+++ b/src/pretix/control/forms/filter.py
@@ -363,6 +363,28 @@ class OrderSearchFilterForm(OrderFilterForm):
         )
     )
 
+    status = forms.ChoiceField(
+        label=_('Order status'),
+        choices=(
+            ('', _('All orders')),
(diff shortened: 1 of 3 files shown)


src/pretix/control/forms/filter.py
```diff
@@ -363,6 +363,28 @@ class OrderSearchFilterForm(OrderFilterForm):
         )
     )
 
+    status = forms.ChoiceField(
```
raphaelm, line 366: Do we really need to copy-paste this? OrderSearchFilterForm and EventOrderFilterForm both depend on OrderFilterForm -- can we do this there instead and only have it once?

raphaelm:
> Perhaps just pay specifically attention if this works as intended (.using()-statement in conjunction with the annotation:

It should :sweat_smile:

PR #1672, opened 2020-05-07
Week calendar and more improvements to subevent calendars

raphaelm:


raphaelm:
> LGTM - but I was surprised the title-omitting was not exposed to the widget.*1
> Buuuut: If we expose it to the widget - where the event name is not necessarily displayed, we should probably make it configurable...

Yeah… So I skipped it, if someone doesn't want the name there they can just hide it with CSS after all :)

PR #1685, opened 2020-05-28
Implement corona-safe seating

raphaelm:
(see upcoming blogpost)


raphaelm:
As discussed on the phone:

> Assuming a seating plan with two rows with three seats each:
Rows at 10;310 and 10;335 (25 distance), Seats in each row at 480;0, 510;0, 570;0.
> Setting the minimal_distance to 25 should disallow the booking of the same seat in the other row. It doesn't however (<= vs <?)

is actually correct

> Being über-correct, we would possibly have to include the actual seat-size into the calculation, as right now we are measuring center-to-center...

is a problem for the future

> Selecting multiple seats which block themselves (say 25.1 distance, selecting two seats behind each other) will only yield one seat in the cart and a generic The seat you selected has already been taken. Please select a different seat. error message.
> Generally speaking: it would be nice to see in the error-message the specific seats that are not available anymore. But in this case here especially, I feel an explanation à la "The seat you selected is not available due to minimal distancing rules" or similar is warranted.

was a bug, is fixed

> Seats blocked by minimal_distance-requirements should be also flagged accordingly and - possibly - be overridable by the "Allow to buy blocked seats on these channels"-setting.

is a problem for the future

> pretix-seating/frontrow:
> Blocked seat view is not functional anymore (TypeError: Cannot read property 'original_price' of undefined).

was an unrelated bug, fixed

PR #1692, opened 2020-06-01
widget: disable buy button without selected products

### LINKED ISSUE #1689: Widget data is not saved in cart session when presale starts without selecting an item.
We've been running into a fairly uncommon bug, which I was unable to locate until today. Since #1606 has been merged, all actions which initiate the presale send the widget data to the server. However we were still seeing cases of the widget data not being available in the cart session for our plugin to use.

Currently when someone initiates the presale, without checking an item using the checkboxes, the cart session including widget data is not saved:
pretix_widget_issue

I inspected the code for the adding item to cart (`pretix.presale.views.cart.CartAdd.post`) and when there are no items present, a simple redirect is returned.

I'd be happy to create a fix for this problem, however I'd like some pointers on how to best fix the problem. Can I simply call the `pretix.base.services.cart.add_items_to_cart` task and add logic to have different return values of the view depending on the number of items?

MaicoTimmerman:
Fixes #1689

### DIFF
--- a/src/pretix/static/pretixpresale/js/widget/widget.js
+++ b/src/pretix/static/pretixpresale/js/widget/widget.js
@@ -190,10 +190,10 @@ Vue.component('availbox', {
         + '</div>'
         + '<div class="pretix-widget-availability-available" v-if="!item.require_voucher && avail[0] === 100">'
         + '<label class="pretix-widget-item-count-single-label" v-if="order_max === 1">'
-        + '<input type="checkbox" value="1" v-bind:name="input_name">'
+        + '<input type="checkbox" v-model="amount_selected" :name="input_name">'
         + '</label>'
         + '<input type="number" class="pretix-widget-item-count-multiple" placeholder="0" min="0"'
-        + '       :value="($root.itemnum == 1 && !item.has_variations) ? 1 : false" v-bind:max="order_max" v-bind:name="input_name"'
+        + '       v-model="amount_selected" :max="order_max" :name="input_name"'
         + '       v-if="order_max !== 1">'
         + '</div>'
         + '</div>'),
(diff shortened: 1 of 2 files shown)


src/pretix/static/pretixpresale/js/widget/widget.js
```diff
@@ -775,6 +809,28 @@ Vue.component('pretix-widget-event-form', {
             } else {
                 this.$root.view = "weeks";
             }
+        },
+        calculate_buy_disabled: function() {
```
raphaelm, line 813: Shouldn't this be a ``method`` instead of a ``computed``?

src/pretix/static/pretixpresale/js/widget/widget.js
```diff
@@ -190,26 +190,48 @@ Vue.component('availbox', {
         + '</div>'
         + '<div class="pretix-widget-availability-available" v-if="!item.require_voucher && avail[0] === 100">'
         + '<label class="pretix-widget-item-count-single-label" v-if="order_max === 1">'
-        + '<input type="checkbox" value="1" v-bind:name="input_name">'
+        + '<input type="checkbox" v-model="amount_selected" :name="input_name">'
```
raphaelm, line 193: Have you tested this does the right thing? I think ``v-model`` on checkboxes refers to whether they are checked or not, not what value they transmit on HTML form submission (which needs to be ``1``).

src/pretix/static/pretixpresale/js/widget/widget.js
MaicoTimmerman: I believe it is :), however the diff folding happens to be unfortunate at this point.

src/pretix/static/pretixpresale/js/widget/widget.js
MaicoTimmerman: I'd have to double check this, according to the vuejs docs v-model for checkboxes and radiobuttons uses the checked property and the change event. So I think we'll have to replace this with something like `:value="amount_selected" @change="e => amount_selected = value"` and do some type conversion in the setter of `amount_selected`. That would also remove the need for the unary `+`-operator to convert booleans to integers in ` calculate_buy_disabled()`

MaicoTimmerman: I've tested and the HTML form submission indeed needs to be an integer. I've updated the MR such that the `value` of the checkboxes is an integer, while using the `checked` property in the vue model.

PR #1696, opened 2020-06-15
Optional PayPal Reference Prefix (Z#2359330)

pc-coholic:


src/pretix/plugins/paypal/payment.py
```diff
@@ -99,8 +99,18 @@ def settings_form_fields(self):
...
+             forms.CharField(
+                 label=_('Reference prefix'),
+                 help_text=_('Any value entered here will be added in front of the regular booking reference containing '
+                             'the order number.'),
+                 required=False,
+             ))
+        ]
+
         d = OrderedDict(
-            fields + list(super().settings_form_fields.items())
```
raphaelm, line 103: I think this is special-casy enough the we can move it to the end of the field list

PR #1709, opened 2020-07-02
Ask only for VAT ID if company is inside EU

ateuber:
To avoid confusion in a reverse charge setting we don't want to ask companies that are not in the European Union for the VAD ID. Maybe it would be better, to make this a setting, but I don't have a good idea how to do it.

### DIFF
--- a/src/pretix/base/forms/questions.py
+++ b/src/pretix/base/forms/questions.py
@@ -580,7 +580,7 @@ class BaseInvoiceAddressForm(forms.ModelForm):
                 'data-display-dependency': '#id_is_business_1',
                 'autocomplete': 'organization',
             }),
-            'vat_id': forms.TextInput(attrs={'data-display-dependency': '#id_is_business_1'}),
+            'vat_id': forms.TextInput(attrs={'data-display-dependency': '#id_is_business_1', 'data-countries-in-eu': ','.join(EU_COUNTRIES)}),
             'internal_reference': forms.TextInput,
         }
         labels = {
@@ -630,6 +630,11 @@ class BaseInvoiceAddressForm(forms.ModelForm):
         )
         self.fields['state'].widget.is_required = True
 
+        # Without JavaScript the VAT ID field is not hidden, so we empty the field if a country outside the EU is selected.
+        if cc and cc not in EU_COUNTRIES and fprefix + 'vat_id' in self.data:
+            self.data = self.data.copy()
+            del self.data[fprefix + 'vat_id']
+
         if not event.settings.invoice_address_required or self.all_optional:
             for k, f in self.fields.items():
                 f.required = False
@@ -644,8 +649,6 @@ class BaseInvoiceAddressForm(forms.ModelForm):
             self.fields['company'].widget.is_required = True
             self.fields['company'].widget.attrs['required'] = 'required'
             del self.fields['company'].widget.attrs['data-display-dependency']
-            if 'vat_id' in self.fields:
-                del self.fields['vat_id'].widget.attrs['data-display-dependency']
 
         self.fields['name_parts'] = NamePartsFormField(
             max_length=255,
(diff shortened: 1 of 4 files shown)


src/pretix/base/forms/questions.py
```diff
@@ -630,6 +630,10 @@ def __init__(self, *args, **kwargs):
         )
         self.fields['state'].widget.is_required = True
 
+        if cc and cc not in EU_COUNTRIES and fprefix + 'vat_id' in self.data:
+            self.data = self.data.copy()
+            del self.data[fprefix + 'vat_id']
```
raphaelm, line 635: Modifying ``self.data`` doesn't feel clean, I think the part in the ``clean()`` should suffice?

src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -314,6 +314,31 @@ var form_handlers = function (el) {
...
+                        } else {
+                            dependent.closest(".form-group").hide();
+                        }
+                        dependent.prop("disabled", false);
+                        dependency_country.closest(".form-group").find("label .fa-spin").remove();
+                    });
+                }
+            };
+        dependency_country.on("change", update);
+        dependency_id_is_business_1.on("change", update);
```
raphaelm, line 339: We probably also need one ``update();`` call to evaluate it after page load (i.e. if something is pre-set)

src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -314,6 +314,31 @@ var form_handlers = function (el) {
...
+        var dependent = $(this),
+            dependency_country = $(this).closest(".panel-body, form").find('select[name$=country]'),
+            dependency_id_is_business_1 = $(this).closest(".panel-body, form").find('input[id$=id_is_business_1]'),
+            update = function (ev) {
+                if (dependency_id_is_business_1.length && !dependency_id_is_business_1.prop("checked")) {
+                    dependent.closest(".form-group").hide();
+                } else {
+                    dependent.prop("disabled", true);
+                    dependency_country.closest(".form-group").find("label").prepend("<span class='fa fa-cog fa-spin'></span> ");
+                    $.getJSON('/js_helpers/is_country_in_eu/?country=' + dependency_country.val(), function (data) {
```
raphaelm, line 327: An AJAX call to get a list of 27 two-letter country codes feels like overkill. Maybe we can just pass it to JS as a ``data-`` attribute to the form field or something?

src/pretix/base/forms/questions.py
ateuber: Not important, but the idea is to remove the input if JavaScript is disabled and the form is reinitialized with a selecte country that is outside the EU, similar to https://github.com/pretix/pretix/commit/b0b0f7474d7999a088444e0f1c0649f18397585d
grafik.

src/pretix/static/pretixcontrol/js/ui/main.js
ateuber:
```suggestion
        dependency_id_is_business_1.on("change", update);
        $(window).on("load", update);
```

src/pretix/static/pretixcontrol/js/ui/main.js
ateuber: You are right, it is now passed as a `data-` attribute.

src/pretix/static/pretixcontrol/js/ui/main.js
raphaelm: No, ``load`` is pretty late. ``load`` only happens after *all* resources on the page have been loaded, which can take a while. You can just call ``update();`` – we're in a ``$(function() { … })`` block and jQuery ensures the DOM is ready.

src/pretix/base/forms/questions.py
raphaelm: Ok, got it. Maybe add a small comment for future readers? :)

ateuber: Thanks, I added a comment.

src/pretix/static/pretixcontrol/js/ui/main.js
ateuber: Thanks for the explanation, changed it.

src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -314,6 +314,24 @@ var form_handlers = function (el) {
         dependency.closest('.form-group').find('input[name=' + dependency.attr("name") + ']').on("dp.change", update);
     });
 
+    $("input[name$=vat_id][data-countries-in-eu]").each(function () {
```
raphaelm, line 317: This selector actually broke a form in the invoice settings, so I refined it a little

PR #1710, opened 2020-07-03
Add optional banktransfer code prefix

Flauschbaellchen:
Add an optional setting in the banktransfer plugin to change the code
prefix from the default value of the event slug to an arbitary string.

This does retain codes used in previous orders.

Work based on PR #1252 from @tynsh

### DIFF
--- a/src/pretix/control/views/event.py
+++ b/src/pretix/control/views/event.py
@@ -1297,6 +1297,11 @@ class QuickSetupView(FormView):
                     form.cleaned_data['payment_banktransfer_%s' % f]
                 )
 
+            self.request.event.settings.set(
+                'payment_banktransfer_code_prefix',
+                form.cleaned_data['payment_banktransfer_code_prefix']
+            )
+
         if form.cleaned_data.get('payment_stripe__enabled', None):
             if 'pretix.plugins.stripe' not in plugins_active:
                 self.request.event.log_action('pretix.event.plugins.enabled', user=self.request.user,
--- a/src/pretix/plugins/banktransfer/payment.py
+++ b/src/pretix/plugins/banktransfer/payment.py
@@ -1,4 +1,5 @@
 import json
+import re
 import textwrap
 from collections import OrderedDict
 
@@ -20,6 +21,7 @@ class BankTransfer(BasePaymentProvider):
     identifier = 'banktransfer'
     verbose_name = _('Bank transfer')
     abort_pending_allowed = True
+    code_prefix_patterns = re.compile("^[a-zA-Z0-9-]+$")
 
     @staticmethod
     def form_fields():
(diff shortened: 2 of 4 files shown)


Flauschbaellchen:
Hi @raphaelm 

This is my PR to continue the work done in #1252.

I haven't added any tests yet and I didn't know how exactly it works with the translation... but maybe you can look into it?

It would be infinitely grateful if we could manage to get it live in some way before August 1st, as that's the date when our prepayment starts and it would be the only missing feature.
Would that be possible, or what release cycles do you have?

src/pretix/control/views/event.py
```diff
@@ -1297,6 +1297,11 @@ def form_valid(self, form):
                     form.cleaned_data['payment_banktransfer_%s' % f]
                 )
 
+            self.request.event.settings.set(
+                'payment_banktransfer_code_prefix',
+                form.cleaned_data['payment_banktransfer_code_prefix']
+            )
```
raphaelm, line 1303: I think you can remove this completely, the prefix will not become part of the quick-setup form.

src/pretix/plugins/banktransfer/payment.py
```diff
@@ -144,6 +157,10 @@ def settings_form_clean(self, cleaned_data):
             if not cleaned_data.get('payment_banktransfer_bank_details'):
                 raise ValidationError(
                     {'payment_banktransfer_bank_details': _('Please enter your bank account details.')})
+
+        if (cleaned_data.get('payment_banktransfer_code_prefix') and not self.code_prefix_patterns.match(cleaned_data.get('payment_banktransfer_code_prefix'))):
+            raise ValidationError(
+                {'payment_banktransfer_code_prefix': _('Prefix can only contains alphanumeric characters and hyphens.')})
```
raphaelm, line 163: Please use Django's built-in validators instead of doing this yourself:

https://docs.djangoproject.com/en/3.0/ref/validators/#built-in-validators

Usage example:
https://github.com/pretix/pretix/blob/27b56b5aeaa972180f27e4c797dc3042666e3706/src/pretix/control/forms/organizer.py#L225

src/pretix/plugins/banktransfer/tasks.py
```diff
@@ -66,31 +66,16 @@ def cancel_old_payments(order):
 @transaction.atomic
 def _handle_transaction(trans: BankTransaction, code: str, event: Event=None, organizer: Organizer=None,
```
raphaelm, line 67: ``event`` no longer is optional, so we should drop the ``=None``, organizer is no longer used so we should drop it

Flauschbaellchen:
Hi @raphaelm,

sorry to let you wait the whole time.
I've improved the code.
Regarding the order code there is another route we can go:

We can change the prefix to be a code-prefix. Thus not changing the event's prefix but the way the code is generated.
E.g. the event with the identifier "my-event" would currently generate codes in the form of "xxxxx" which translates into "my-event-xxxxx".
When setting an additional code-prefix, we'll generate the code in the form of "prefix-xxxxx" which translates into "my-event-prefix-xxxxx".

The advantages would be:
* Changing the prefix halfway would not change the code of existing orders
* The order codes can still be clearly assigned to an event in an human readable form
* The whole mapping functionality (get_prefix_event_map) is obsolete.

Disadvantage:
* The order code is currently limited to 16 characters (`settings.ENTROPY['order_code']` + X length if code is already taken). The field's length needs to be increased and the prefix would need to be limited in length.

However, this would slightly change the initial intention of this PR.
What do you think?

raphaelm:
I don't think we should change how order codes work, no, as it is a concept baked very deep into pretix. Order codes are commonly used with and without their prefix, and I think this would break many unexpected things.

raphaelm:
Making the prefix changeable just adds too much complexity to the importer, it's not worth to do that.

@julia-luna replace this with a PR that allows to put an additional prefix in front of the current bank transfer reference, instead of replacing it.

PR #1717, opened 2020-07-12
Allow selecting the same add-on multiple times

raphaelm:
This PR introduces the option to allow booking the same ticket multiple times as an add-on product, enabling a wide number of use cases where we required nasty workarounds so far. At the same time, it cleans up the add-on selection step which often looks weird.

- [x] Implement
- [x] Fix existing tests
- [x] Free price
- [x] Add new tests
- [ ] Decide if we pre-inform customers since it's a kinda visible change to the checkout process

Before:

Screenshot_2020-07-12 Bestellung fortsetzen Addon-Test

After:

Screenshot_2020-07-12 Bestellung fortsetzen Addon-Test(1)


src/pretix/presale/templates/pretixpresale/event/checkout_addons.html
```diff
@@ -45,15 +48,277 @@ <h4 class="panel-title">
                                             {% plural %}
                                                 You need to choose {{ min_count }} options from this category.
                                             {% endblocktrans %}
-                                        {% elif c.min_count == 0 and c.max_count >= c.form.fields|length %}
+                                        {% elif c.min_count == 0 and c.max_count >= c.items|length and not c.multi_allowed %}
+                                        {% elif c.min_count == 0 %}
+                                            {% blocktrans trimmed with max_count=c.max_count %}
+                                                You can choose up to {{ max_count }} options from this category.
```
pc-coholic, line 54: This is only displayed, when the Multi-AddOn-selection is enabled.

In the "classic" step, it should probably read "You can select up to {{ max(max_count, len(items)) }} options" to not confuse people (for example: Can select 0 to 10 items, but only 3 items are available to choose from).

raphaelm:
Will merge on monday (in case we have any deployments on the weekend)

PR #1719, opened 2020-07-16
Allow customers to change to a different product variation

raphaelm:
- [x] tests


src/pretix/base/models/orders.py
```diff
@@ -468,14 +481,41 @@ def user_cancel_fee(self):
 
     @property
     @scopes_disabled()
-    def user_cancel_allowed(self) -> bool:
+    def user_change_allowed(self) -> bool:
         """
         Returns whether or not this order can be canceled by the user.
```
felixrindt, line 486:
```suggestion
        Returns whether or not this order can be changed by the user.
```

src/pretix/base/services/orders.py
```diff
@@ -1131,7 +1131,7 @@ def notify_user_changed_order(order, user=None, auth=None, invoices=[]):
         try:
             order.send_mail(
                 email_subject, email_template, email_context,
-                'pretix.event.order.email.order_changed', user, auth=auth, invoices=invoices,
+                'pretix.event.order.email.order_changed', user, auth=auth, invoices=invoices, attach_tickets=True
```
felixrindt, line 1134: You dropped a trailing comma.

src/pretix/presale/views/order.py
```diff
@@ -1006,3 +1009,100 @@ def get(self, request, *args, **kwargs):
...
+                p.custom_error = str(e)
+                return False
+        return True
+
+    def post(self, *args, **kwargs):
+        ocm = OrderChangeManager(
+            self.order,
+            user=self.request.user,
+            notify=True,
+            reissue_invoice=True
```
felixrindt, line 1093: No trailing comma. (Don't know if I'd want one tbh)

raphaelm:
> Due to user_change_allowed checking only if a change is allowed (as the name implies), the user can also click the button and end up on a page, where nothing in fact can be changed. Perhaps adding a user_change_possible that checks if there is even anything that could be changed might be an improvement?

That would involve very expensive checks (eg quota availability) every time the order detail page is called (which is probably a lot), I'm not sure it's worth it.

> Items that cannot be changed will still be displayed as drop-downs. Perhaps we either get rid of them completely or display them without the drop-down? (More leaning towards the removal of the drop-down).

Hmm, yeaah, but… does it really make it less confusing? Especially if there's theoretically a different variation but I can't select it, it might be weird to not even see that this is where I'd theoretically select it. Not sure how to communicate these nuances to the user without too much effort. Especially in combination with your previous point, in a situation where I am allowed to change but there isn't actually anything to change to (everything sold out), my feeling would be that a dropdown with one option feels less like "hmm, this must be the wrong page/a bug" than if there's not even a form.

> Add-Ons are not properly distinguishable or attributable to the main item, if one does not know that they are displayed in order. See the two following screenshots. Perhaps we can nest the items a little bit?

Good point. Considering that more options for this form might come in the future (changing subevents, selecting new add-ons entirely), do you have suggestions on how we should nest? Like in the questions form?

pc-coholic:
> Hmm, yeaah, but… does it really make it less confusing? Especially if there's theoretically a different variation but I can't select it, it might be weird to not even see that this is where I'd theoretically select it. Not sure how to communicate these nuances to the user without too much effort. Especially in combination with your previous point, in a situation where I am allowed to change but there isn't actually anything to change to (everything sold out), my feeling would be that a dropdown with one option feels less like "hmm, this must be the wrong page/a bug" than if there's not even a form.

Could we just always display all options - even the unavailable ones - and disable those `<select>`-items?

> Good point. Considering that more options for this form might come in the future (changing subevents, selecting new add-ons entirely), do you have suggestions on how we should nest? Like in the questions form?

Yes, that sounds good to me :)

raphaelm:
> Could we just always display all options - even the unavailable ones - and disable those <select>-items?

Not easily. First, we'd need to carefully check if we're allowed to show them, second django doesn't really do disabled select options :( I opted for disabling the select box if there's nothing to select with a reason in the help text.

> [addon hierarchy]

done

PR #1721, opened 2020-07-20
Absent/Checked Out persons in Checkin lists

pc-coholic:


src/pretix/control/forms/filter.py
```diff
@@ -883,6 +884,10 @@ def filter_qs(self, qs):
                 qs = qs.filter(last_entry__isnull=False).filter(
                     Q(last_exit__isnull=True) | Q(last_exit__lt=F('last_entry'))
                 )
+            elif s == '3':
+                qs = qs.filter(last_entry__isnull=False).filter(
+                    Q(last_exit__isnull=False) | Q(last_exit__lt=F('last_entry'))
```
raphaelm: I think this needs to be
```suggestion
                    Q(last_exit__isnull=False) & Q(last_exit__gte=F('last_entry'))
```

PR #1725, opened 2020-07-22
Create invoice exporter mixin

felixrindt:
I created a mixin to unite Form Fields for both Invoice Exporters. Now they both have the same date and payment provider filter.
Also I moved the Invoice Data Exporter over to `exporters/invoices.py`.

Looks like this now:

image


src/pretix/base/exporters/invoices.py
```diff
@@ -84,43 +133,243 @@ def render(self, form_data: dict, output_file=None):
...
+            ).values('order').annotate(
+                m=GroupConcat('provider', delimiter=',')
+            ).values(
+                'm'
+            ).order_by()
+
+            qs = InvoiceLine.objects.annotate(
+                payment_providers=Subquery(p_providers, output_field=CharField()),
+            ).filter(
+                invoice__in=self.invoices_queryset(form_data)
```
raphaelm, line 318: I was a little worried that using a subquery here might have worse performance characteristics than the join we previously used, considering we might realistically have millions of invoices in a cross-event export of a large client.

However, apparently at least PostgreSQL will be smart about it:
https://www.percona.com/blog/2020/04/16/sql-optimizations-in-postgresql-in-vs-exists-vs-any-all-vs-join/

I'm still a little worried about this on other databases, but since it makes the code much simpler I think we can just give it a go and see if it causes actual problems.

felixrindt: I also thought about caching the invoices from when the invoices sheet is rendered. That way we wouldn't need to have that query twice. But that would need state in the exporter/a change in how the Multi Sheet Exporter works.

PR #1726, opened 2020-07-22
Send canonical mails depending on sales channel

felixrindt:
* [x] tests
* [x] passing tests
* [x] texts and description
* [x] API order creation docs
* [x] user guide?!
* [x] migration
* [x] Setting field with defaults
* [x] conditions
  * [x] order placed
  * [x] ticket download reminder
  * [x] payment paid


src/pretix/base/settings.py
```diff
@@ -1190,6 +1190,14 @@ def country_choice_kwargs():
                         "Defaults to your event name."),
         )
     },
+    'mail_sales_channel_placed_paid': {
+        'default': ['web'],
+        'type': list,
+    },
+    'mail_sales_channel_ticket_reminder': {
+        'default': ['web'],
+        'type': list,
```
felixrindt, line 1199: `List[str]`?

src/pretix/control/templates/pretixcontrol/event/mail.html
```diff
@@ -17,6 +17,8 @@ <h1>{% trans "E-mail settings" %}</h1>
                 {% bootstrap_field form.mail_text_signature layout="control" %}
                 {% bootstrap_field form.mail_bcc layout="control" %}
                 {% bootstrap_field form.mail_attach_ical layout="control" %}
+                {% bootstrap_field form.mail_sales_channel_placed_paid layout="control" %}
+                {% bootstrap_field form.mail_sales_channel_ticket_reminder layout="control" %}
```
felixrindt, line 21: Maybe in another place? It has sensible defaults and most users don't need it/don't even need to understand what sales channels are...

src/pretix/control/forms/event.py
```diff
@@ -758,6 +758,26 @@ class MailSettingsForm(SettingsForm):
...
+    )
+
+    mail_sales_channel_ticket_reminder = forms.MultipleChoiceField(
+        choices=[(ident, sc.verbose_name) for ident, sc in get_all_sales_channels().items()],
+        label=_('Sales Channels for Ticket Emails'),
+        help_text=_('Restrict ticket reminder emails to orders from certain sales channels.'),
+        widget=forms.CheckboxSelectMultiple(
+            attrs={'class': 'scrolling-multiple-choice'}
+        ),
+        required=False,
```
felixrindt, line 778: Check that `required` is what we want here...

src/pretix/base/settings.py
raphaelm: ``list`` is correct, ``List[str]`` only works as a type annotation, while this needs to be an actual class that hierarkey can unserialize to.

src/pretix/control/templates/pretixcontrol/event/mail.html
raphaelm: For the ticket reminder, I'd put it in the same box as the ticket reminder text, because that's also where you activate the reminder in the first place, so the context is pretty clear.

For the placed/paid/free, this doesn't work since it would need to go in 3 boxes, so maybe the place you chose is a good one for that, even though that makes it a little inconsistent.

src/pretix/control/forms/event.py
raphaelm: Hmm, I'd actually consider not allowing people to deselect ``web`` by adding a custom validator that requires ``web`` to always be in the result. There's to much to go wrong if no emails are sent.

pc-coholic:
> * This PR is backwards incompatible, in that it disables emails for reseller sales for all events as soon as the PR is applied. We need to decide if we want to change the default to the current value generally, or if we want to apply the PR as is and add a data migration that sets the setting to the old value for all existing events. @pc-coholic do you have an opinion here?

If we agree that not sending the emails is a sane default, I would agree to run a migration for existing events to keep the current behaviour.

felixrindt:
```
When creating orders through the API, API clients can send a send_mail attribute that currently controls whether emails are sent. We should decide which value has higher precedence – the one set through the API or the one through the settings and document that decision. I'm leaning towards trusting the API value.
```

I agree that the API value should have higher precedence. It is already calling the mail functions directly and not using the business logic methods I touched.

src/tests/base/test_orders.py
```diff
@@ -543,6 +543,13 @@ def test_not_sent_after_reminder_date(self):
         send_download_reminders(sender=self.event)
         assert len(djmail.outbox) == 0
 
+    @classscope(attr='o')
+    def test_not_sent_for_disabled_sales_channel(self):
+        self.event.settings.mail_days_download_reminder = 2
+        self.event.settings.mail_sales_channel_ticket_reminder = []
+        send_download_reminders(sender=self.event)
+        assert len(djmail.outbox) == 0
+
```
felixrindt, line 552: So I thought this test would fit here. I did not find a nice place to test the `placed_paid` option.

src/pretix/control/forms/event.py
```diff
@@ -758,6 +762,27 @@ class MailSettingsForm(SettingsForm):
...
+                    'The online shop must be enabled.'),
+        widget=forms.CheckboxSelectMultiple(
+            attrs={'class': 'scrolling-multiple-choice'}
+        ),
+        validators=[contains_web_channel_validate],
+    )
+
+    mail_sales_channel_download_reminder = forms.MultipleChoiceField(
+        choices=lambda: [(ident, sc.verbose_name) for ident, sc in get_all_sales_channels().items()],
+        label=_('Sales Channels'),
```
raphaelm, line 778:
```suggestion
        label=_('Sales channels'),
```

src/pretix/control/forms/event.py
```diff
@@ -758,6 +762,27 @@ class MailSettingsForm(SettingsForm):
         'mail_attach_ical',
     ]
 
+    mail_sales_channel_placed_paid = forms.MultipleChoiceField(
+        choices=lambda: [(ident, sc.verbose_name) for ident, sc in get_all_sales_channels().items()],
+        label=_('Sales Channels for Checkout Emails'),
```
raphaelm, line 767:
```suggestion
        label=_('Sales channels for checkout emails'),
```

PR #1729, opened 2020-07-24
Project setup things

felixrindt:


src/setup.cfg
```diff
@@ -6,13 +6,11 @@ max-complexity = 11
...
 [isort]
 combine_as_imports = true
-default_section = THIRDPARTY
 include_trailing_comma = true
 known_first_party = pretix
 known_third_party = versions
-known_standard_library = typing,enum,mimetypes
+extra_standard_library = typing,enum,mimetypes
 multi_line_output = 5
-not_skip = __init__.py
```
felixrindt, line 15: I took a look at https://timothycrosley.github.io/isort/docs/upgrade_guides/5.0.0/ for this. With and without these changes my isort complains about the order in 20 or so py files :/

src/pretix/base/management/commands/shell_scoped.py
```diff
@@ -18,6 +18,7 @@ def handle(self, *args, **options):
             cmd = 'shell_plus'
         except ImportError:
             cmd = 'shell'
+            del options['skip_checks']
```
felixrindt, line 21: alternatively raise and expect `django_extensions` to be installed/add to requirements-dev

PR #1732, opened 2020-07-28
Fix #1675 -- Allow '0' as answer to number questions

### LINKED ISSUE #1675: question type number should accept 0
when you make a question with the type number a reqired question it should accept the number 0. Currently it refuses:
Selection_501

felixrindt:
Fix #1675

### DIFF
--- a/src/pretix/presale/checkoutflow.py
+++ b/src/pretix/presale/checkoutflow.py
@@ -570,7 +570,7 @@ class QuestionsStep(QuestionsViewMixin, CartMixin, TemplateFlowStep):
                 )
 
             for q in cp.item.questions_to_ask:
-                if question_is_required(q) and not answ.get(q.id):
+                if question_is_required(q) and q.id not in answ.keys():
                     if warn:
                         messages.warning(request, _('Please fill in answers to all required questions.'))
                     return False
--- a/src/pretix/presale/checkoutflow.py
+++ b/src/pretix/presale/checkoutflow.py
@@ -570,7 +570,7 @@ class QuestionsStep(QuestionsViewMixin, CartMixin, TemplateFlowStep):
                 )
 
             for q in cp.item.questions_to_ask:
-                if question_is_required(q) and q.id not in answ.keys():
+                if question_is_required(q) and q.id not in answ:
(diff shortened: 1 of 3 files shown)


raphaelm:
Have you checked if someone can actually proceed with checkout after this or if ``pretix/presale/checkoutflow.py`` line 572ff will throw you back to the question step?

felixrindt:
> Have you checked if someone can actually proceed with checkout after this or if pretix/presale/checkoutflow.py line 572ff will throw you back to the question step?

I checked and also I changed the test to check it. I was unaware of these lines, so I debugged into it. 
`answ.get(q.id)` gives you a `QuestionAnswer` instance which always evaluates to true or you get `None`.

felixrindt:
Will change line 573 to `if question_is_required(q) and q.id in answ.keys():` then.

src/pretix/presale/checkoutflow.py
```diff
@@ -570,7 +570,7 @@ def question_is_required(q):
                 )
 
             for q in cp.item.questions_to_ask:
-                if question_is_required(q) and not answ.get(q.id):
+                if question_is_required(q) and q.id not in answ.keys():
```
raphaelm, line 573:
```suggestion
                if question_is_required(q) and q.id not in answ:
```

you don't even need a the .keys() for a dict :)

felixrindt: I know, but it has been personal preference to be explicit with dict containment.
I'd say its more natural for a dict to iterate over key/value pairs by default, but that's not the way it is :sweat_smile:

raphaelm: > I know, but it has been personal preference to be explicit with dict containment.

Nevermind then, it certainly does not hurt to be explicit. Feel free to do it as you like ;)

> I'd say its more natural for a dict to iterate over key/value pairs by default, but that's not the way it is sweat_smile

For iterating, I think I agree. For containment checks, I'd only agree because it would be consistent if iteration, ``key in dict`` does feel natural to me (i.e. you can say it out loud and it's somewhat clear what it means).

PR #1733, opened 2020-07-29
Copy answers button for addon products

felixrindt:
I added data attributes to make it easy to get the correct source and target inputs. It works the following now:

* For products, the "copy answer" button copies answers from the first products answers down to the product and addon questions where the button is positioned.
* For addon products, the "copy answer" button copies answers from the addon's main product down to the addon.

It's somewhat sensible in the way that it is hierarchical, but I don't love that they behave differently. What do you think? 

TODO:

* [x] position the button correctly
* [x] ~fix the button shrinking after being clicked~

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/item/include_addons.html
+++ b/src/pretix/control/templates/pretixcontrol/item/include_addons.html
@@ -8,8 +8,7 @@
         workshops as add-ons to the conference ticket. With this configuration, the workshops cannot be bought
         on their own but only in combination with a conference ticket. You can here specify categories of products
         that can be used as add-ons to this product. You can also specify the minimum and maximum number of
-        add-ons of the given category that can or need to be chosen. The user can buy every add-on from the
-        category at most once. If an add-on product has multiple variations, only one of them can be bought.
+        add-ons of the given category that can or need to be chosen.
     {% endblocktrans %}
 </p>
 <div class="formset" data-formset data-formset-prefix="{{ formset.prefix }}">
--- a/src/pretix/presale/templates/pretixpresale/event/checkout_questions.html
+++ b/src/pretix/presale/templates/pretixpresale/event/checkout_questions.html
@@ -105,9 +105,18 @@
                             {% for form in forms %}
                                 {% if form.pos.item != pos.item %}
                                     {# Add-Ons #}
-                                    <legend>+ {{ form.pos.item.name }}{% if form.pos.variation %} – {{ form.pos.variation.value }}{% endif %}</legend>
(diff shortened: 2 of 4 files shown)


src/pretix/control/templates/pretixcontrol/item/include_addons.html
```diff
@@ -8,8 +8,7 @@
         workshops as add-ons to the conference ticket. With this configuration, the workshops cannot be bought
         on their own but only in combination with a conference ticket. You can here specify categories of products
         that can be used as add-ons to this product. You can also specify the minimum and maximum number of
-        add-ons of the given category that can or need to be chosen. The user can buy every add-on from the
-        category at most once. If an add-on product has multiple variations, only one of them can be bought.
+        add-ons of the given category that can or need to be chosen.
     {% endblocktrans %}
```
felixrindt, line 12: For a few days this limitation has been removed. Feel free to update this explanation further.

src/pretix/presale/templates/pretixpresale/event/checkout_questions.html
```diff
@@ -105,9 +105,18 @@ <h4 class="panel-title">
                             {% for form in forms %}
                                 {% if form.pos.item != pos.item %}
                                     {# Add-Ons #}
-                                    <legend>+ {{ form.pos.item.name }}{% if form.pos.variation %} – {{ form.pos.variation.value }}{% endif %}</legend>
+                                    <legend> {# FIXME: I wanted to put it to the right side. For that legend should be d-flex and justify-content-between, but that's Bootstrap 4... #}
```
felixrindt, line 108: I need help with the CSS here :sweat_smile:

src/pretix/control/templates/pretixcontrol/item/include_addons.html
raphaelm: :+1:

src/pretix/presale/templates/pretixpresale/event/checkout_questions.html
raphaelm: Putting the button before the actual text of the legend and replacing ``text-right flip`` with ``pull-right flip`` should be enough!

felixrindt:
How it works now is that:

* the copy-button for addons is only shown if there is a common question between the addon and the main product.
* the copy-button for addons copies answers from the first input it fiends. That might be the main product or another addon.
* the copy-button for main products copies answers from the first product section-wise. B's main questions from A's main questions, B's first addon from A's first addon, and so on.

PR #1734, opened 2020-07-29
Allow to create invoices before bank transfer runs

felixrindt:
using a new hook in the payment provider


src/pretix/presale/views/order.py
```diff
@@ -602,13 +602,18 @@ def get_confirm_url(self, payment):
 
 
 def can_generate_invoice(event, order, ignore_payments=False):
+    open_payment = order.payments.last()
+    if open_payment and open_payment.state in (OrderPayment.PAYMENT_STATE_CONFIRMED, OrderPayment.PAYMENT_STATE_REFUNDED):
+        open_payment = None
```
felixrindt, line 607: I'm unsure about the semantics here. This is taken from the `OrderChangeManager`.
I thought Payments in state `created` and `pending` would be "open". Either way, we need to decide in which payment states invoices should be able to be created for payment providers overwriting this option.

src/pretix/presale/views/order.py
```diff
@@ -602,13 +602,18 @@ def get_confirm_url(self, payment):
 
 
 def can_generate_invoice(event, order, ignore_payments=False):
```
felixrindt, line 604: I changed this method so when changing to bank transfer the customer can request an invoice. It is not automatically generated in this case.

src/pretix/plugins/banktransfer/payment.py
```diff
@@ -21,6 +21,11 @@ class BankTransfer(BasePaymentProvider):
     verbose_name = _('Bank transfer')
     abort_pending_allowed = True
 
+    def create_invoice_immediately(self, order):
+        if super().create_invoice_immediately(order):
+            return True
+        return self.event.settings.get('invoice_generate') == 'paid' and self.settings.get("invoice_immediately", False)
+
     @staticmethod
```
felixrindt, line 29: I just put this somewhere in the class near the top :sweat_smile:

src/pretix/base/payment.py
```diff
@@ -175,6 +175,15 @@ def abort_pending_allowed(self) -> bool:
...
         return False
 
+    def create_invoice_immediately(self, order):
+        """
+        Whether or not to create an invoice right after this payment method has been chosen for ``order``.
+        Invoices send on succeeded payment are dealt with separately.
+        By default, this is True only if the event is configured to generate invoices for all orders immediately,
+        but it might be overwritten for e.g. bank transfer.
+        """
+        return self.event.settings.get('invoice_generate') == 'True'
```
raphaelm, line 185: This is more of a general design question: I'm not sure if I like moving this logic into the payment provider. It creates issues where e.g. a invoice is no longer generated if an order currrently has no open payment (see below) and I imagine it will be hard to maintain and explain to users going forward if plugins to crazy stuff here.

I'd rather have this function to be just ``return False`` by default, and keep the logic in the pretix core, put like this (pseudocode):

```
# in services/orders.py:
if self.event.settings.get('invoice_generate') == 'True' or (self.event.settings.get('invoice_generate') == 'paid' and payment_provider.create_invoice_immediately(order)):
    # generate invoice
```

I'd then also change the settings label of ``invoice_generate = paid`` to something like "Automatically after payment, or during payment if required by payment method"

src/pretix/presale/views/order.py
raphaelm: Hmm, you're right, this logic kinda handles failed/canceled payments wrong. However, for the scope of this PR, I think we can drop this again and automatically create the invoice when the payment method has changed, so we don't need to touch this method at all.

src/pretix/base/payment.py
```diff
@@ -168,13 +168,22 @@ def identifier(self) -> str:
...
-        Whether or not a user can abort a payment in pending start to switch to another
+        Whether or not a user can abort a payment in pending state to switch to another
         payment method. This returns ``False`` by default which is no guarantee that
         aborting a pending payment can never happen, it just hides the frontend button
         to avoid users accidentally committing double payments.
         """
         return False
 
+    @property
+    def requires_invoice_immediately(self):
```
raphaelm, line 179: I think you manually need to reference this new property in the documentation

src/pretix/base/payment.py
```diff
@@ -168,13 +168,22 @@ def identifier(self) -> str:
...
+        """
+        Return whether this payment method requires an invoice to exist for an order, even though the event
+        is configured to only create invoices for paid orders.
```
raphaelm, line 182: As an additional help for plugin authors, we could clarify here that the invoice exists before execute_payment is called

src/pretix/plugins/banktransfer/payment.py
```diff
@@ -20,6 +20,7 @@ class BankTransfer(BasePaymentProvider):
     identifier = 'banktransfer'
     verbose_name = _('Bank transfer')
     abort_pending_allowed = True
+    requires_invoice_immediately = True
```
raphaelm, line 23: If we have a setting for it 80 lines below, we should probably also respect that (and keep the default to ``False`` as it is, many people do not want invoices for unpaid orders)

src/pretix/plugins/banktransfer/payment.py
```diff
@@ -98,6 +99,12 @@ def form_fields():
                 }},
                 required=False
             )),
+            ('invoice_immediately',
+             forms.BooleanField(
+                 label=_('Create an invoice for orders using bank transfer immediately if the event is otherwise'
```
raphaelm, line 104:
```suggestion
                 label=_('Create an invoice for orders using bank transfer immediately if the event is otherwise '
```

felixrindt:
I must have been tired working on this yesterday :sweat_smile:

PR #1735, opened 2020-07-30
Allow to set multiple confirm texts

felixrindt:
This allows users to configure multiple confirmation texts.

image

I still want to

* [x] create Log Entries when the texts are changed, as they are not part of the main form anymore, but are important.
* [x] create a `I18nStringList` to avoid dealing with the data attribute in the middle of form code.
* [x] create and test a migration


src/pretix/base/settings.py
```diff
@@ -37,6 +38,20 @@ def country_choice_kwargs():
     }
 
 
+class LazyI18nStringList(UserList):
+    def __init__(self, init_list=None):
+        super().__init__()
+        if init_list is not None:
+            self.data = [v if isinstance(v, LazyI18nString) else LazyI18nString(v) for v in init_list]
```
felixrindt, line 45: I guess this should technically be

```suggestion
            self.data = [LazyI18nString(copy.deepcopy(v.data)) if isinstance(v, LazyI18nString) else LazyI18nString(v) for v in init_list]
```

Or the [initializer of LazyI18nString](https://github.com/raphaelm/django-i18nfield/blob/109de9f1d3c8621dc09e9d5a2095d652f97ba1bd/i18nfield/strings.py#L14) is updated to work with `v` being `dicts`, `str` or another `LazyI18nString` (so it copies the data over). Then this could just be:

```suggestion
            self.data = [LazyI18nString(v) for v in init_list]
```

raphaelm: I think since this is mostly a wrapper for serialization/unserialization, we should be fine without the copy for now. Or did you see it cause any problems?

src/pretix/control/templates/pretixcontrol/event/settings.html
```diff
@@ -99,10 +103,80 @@ <h1>{% trans "General settings" %}</h1>
                 {% bootstrap_field sform.frontpage_text layout="control" %}
                 {% bootstrap_field sform.presale_has_ended_text layout="control" %}
                 {% bootstrap_field sform.voucher_explanation_text layout="control" %}
-                {% bootstrap_field sform.confirm_text layout="control" %}
                 {% bootstrap_field sform.checkout_email_helptext layout="control" %}
                 {% bootstrap_field sform.banner_text layout="control" %}
                 {% bootstrap_field sform.banner_text_bottom layout="control" %}
+            
+                <div class="form-group">
```
raphaelm, line 110: Can we move this back up, preferrably even just below ``frontpage_text``, since it's arguably the most important thing on this page.

src/pretix/base/settings.py
felixrindt: We are fine for now. ^^
I guess not copying elements, just the initial list is actually how `list` works, so that's fine.

src/pretix/control/templates/pretixcontrol/event/settings.html
felixrindt: Yes we can.

PR #1738, opened 2020-07-30
Add minimum postgres version to docs

pajowu:
Django dropped support for postgres 9.4 in Django 3.0 [1]. Operating pretix with a postgres 9.4 or smaller leads to internal server errors.

[1]: https://code.djangoproject.com/wiki/SupportedDatabaseVersions#PostgreSQL


raphaelm:
While I'm inclined to just put 10 or 11 in there, let's go for this for now ;)

PR #1739, opened 2020-07-31
Fix event settings form considered changed even if unchanged

felixrindt:
As discussed, when not having changed anything in the event settings form, pretix still creates a LogEntry that a change has been made, even though there is code to prevent that.

I found a way to correct that for a use of `RelativeDateTime`, but `has_changed()` is still True for the field `checkout_email_helptext`. That's an I18nText, so maybe @raphaelm can have a look at that?! :innocent:

### DIFF
--- a/src/pretix/base/reldate.py
+++ b/src/pretix/base/reldate.py
@@ -273,6 +273,11 @@ class RelativeDateTimeField(forms.MultiValueField):
                 minutes_before=None
             ))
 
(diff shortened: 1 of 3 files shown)


raphaelm:
I finally fixed the I18nText issue here:
https://github.com/raphaelm/django-i18nfield/commit/fc31f2e33e0e9a1fe2a2fed4504c15bf18867fc0

I'll have another look at finishing this PR now

Issue #1663, opened 2020-04-25
The order of products is not stable between event page, cart, checkout


thegcat:
The event page will show products in the order defined for products in the event settings, the cart at the top of the event page and the checkout pages will not always show the same order. This was confusing to me when I first noticed it :-)

As best I can surmise the products in the cart are sorted in the order the products were created (I can't say if this is by ID or by whatever order the DB happens to return). The products seem to be in the same order they have been added to the cart on the checkout pages (same as before, can't say if it's intentionally sorted by ID or whatever order the DB returns). It is very confusing that products are sorted differently in different parts of the app.

raphaelm:
There's two aspects to consider here that pretix handles differently:

* Cart items that contain any specific sub-items (such as download buttons, answers, add-on products) are currently shown in the order of their cart position ID. I think we should keep it this way (so if I add something to my cart later, it appears at the bottom), but should double-check that products are added to the cart in the order the products show up on the event page if I add multiple products at once.

* Cart items that are not specific and can be grouped (e.g. shown as "2x Regular ticket" instead of 2 lines with "Regular ticket") are apparently currently sorted by their product ID, which does not make sense, and should be based on the product and category order instead, agreed. This is handled in the inline ``keyfunc`` function in ``pretix/presale/views/__init__.py::CartMixin::get_cart`` which is used both for grouping and for sorting (which makes it a bit nasty to work with)

raphaelm:
Fixed by #1473

PR #1743, opened 2020-08-03
Fix #1740 - Do not group gift card positions

### LINKED ISSUE #1740: Order page shows only one gift card code
When ordering multiple gift cards in a single order, only a single gift card code is shown on the order page after payment , see images attached. Is this intended?
Frontend:
image
Backend:
image

felixrindt:
This fixes #1740. Also refs #1663.

### DIFF
--- a/src/pretix/presale/views/__init__.py
+++ b/src/pretix/presale/views/__init__.py
@@ -112,16 +112,28 @@ class CartMixin:
                 or self.request.event.settings.attendee_emails_asked
                 or pos_additional_fields.get(pos.pk)
             )
-            addon_penalty = 1 if pos.addon_to_id else 0
-            if downloads or pos.pk in has_addons or pos.addon_to_id:
-                return i, addon_penalty, pos.pk, 0, 0, 0, 0, (pos.subevent_id or 0), (pos.seat_id or 0)
-            if answers and (has_attendee_data or pos.item.questions.all()):
-                return i, addon_penalty, pos.pk, 0, 0, 0, 0, (pos.subevent_id or 0), (pos.seat_id or 0)
 
+            addon_penalty = 1 if pos.addon_to_id else 0
+
+            if downloads \
+                    or pos.pk in has_addons \
+                    or pos.addon_to_id \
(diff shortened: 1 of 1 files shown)


src/pretix/presale/views/__init__.py
```diff
@@ -112,16 +112,28 @@ def keyfunc(pos):
...
-            if downloads or pos.pk in has_addons or pos.addon_to_id:
-                return i, addon_penalty, pos.pk, 0, 0, 0, 0, (pos.subevent_id or 0), (pos.seat_id or 0)
-            if answers and (has_attendee_data or pos.item.questions.all()):
-                return i, addon_penalty, pos.pk, 0, 0, 0, 0, (pos.subevent_id or 0), (pos.seat_id or 0)
 
+            if downloads \
+                    or pos.pk in has_addons \
+                    or pos.addon_to_id \
+                    or pos.item.issue_giftcard \
+                    or answers and (has_attendee_data or pos.item.questions.exists()):
```
raphaelm, line 122:
```suggestion
                    or (answers and (has_attendee_data or pos.item.questions.exists())):
```
Nobody can remember the precedence of or and and :see_no_evil:

src/pretix/presale/views/__init__.py
```diff
@@ -112,16 +112,28 @@ def keyfunc(pos):
...
+                return (
+                    # standalone positions are grouped by main product position id, addons below them also sorted by position id
+                    i, addon_penalty, pos.pk,
+                    # all other places are only used for positions that can be grouped. We just put zeros.
+                ) + (0, ) * 10
```
raphaelm, line 127: Is it really 10, not 7?

felixrindt: It is now because of `position` being used for ordering three times.

raphaelm: AAh, sorry, I misunderstood the "category_key"/"variation_key" thing, I though you were doing nested tuples instead of concatinating them. Should be fine then!

PR #1745, opened 2020-08-06
Improve algorithm for {name} placeholder

raphaelm:


src/pretix/base/email.py
```diff
@@ -474,11 +495,7 @@ def base_placeholders(sender, **kwargs):
         ),
         SimpleFunctionalMailTextPlaceholder(
             'name', ['position_or_address'],
-            lambda position_or_address: (
-                position_or_address.name
-                if isinstance(position_or_address, InvoiceAddress)
-                else position_or_address.attendee_name
-            ),
+            lambda position_or_address: get_best_name(position_or_address),
```
felixrindt, line 498:
```suggestion
            get_best_name,
```

src/pretix/base/email.py
```diff
@@ -258,9 +258,30 @@ def _placeholder_payment(order, payment):
         return str(payment.payment_provider.order_pending_mail_render(order))
 
 
+def get_best_name(position_or_address, parts=False):
```
felixrindt, line 261: I think having the `position_or_address` context creation input smells, but I guess it's not possible for a placeholder right now to specify that it requires either a position or an address?

src/pretix/base/email.py
```diff
@@ -258,9 +258,30 @@ def _placeholder_payment(order, payment):
...
+    """
+    Return the best name we got for either an invoice address or an order position, falling back to the respective other
+    """
+    from pretix.base.models import InvoiceAddress, OrderPosition
+    if isinstance(position_or_address, InvoiceAddress):
+        if position_or_address.name:
+            return position_or_address.name_parts if parts else position_or_address.name
+        position_or_address = position_or_address.order.positions.exclude(attendee_name_cached="").exclude(attendee_name_cached__isnull=True).first()
```
felixrindt, line 269: what if such a position does not exist?

raphaelm: Then it won't be an instance of OrderPosition (since None is not an instance of anything) and we'll fall back to the empty string/dict, since we haven't got anything

PR #1750, opened 2020-08-12
Add eMail-Template for free approval orders

pc-coholic:
Z#2361977 requires a dedicated approval email for free orders, as they cannot repurpose the existing "Free Order" eMail-template.

I don't expect any issues merging this for a function POV.

However, in order to not annoy our hosted customers from Germany, we should probably deploy it only once the German translations for this have been merged, too.

Casus Knacksus @raphaelm: Is this something, we can merge and deploy without writing a migration, which duplicates the existing `mail_text_order_free` into the new `mail_text_order_approved_free` for existing events in order to conserve the current mail templates?

### DIFF
--- a/src/pretix/control/forms/event.py
+++ b/src/pretix/control/forms/event.py
@@ -929,6 +929,13 @@ class MailSettingsForm(SettingsForm):
         required=False,
         widget=I18nTextarea,
         help_text=_("This will only be sent out for non-free orders. Free orders will receive the free order "
+                    "template from below instead."),
+    )
+    mail_text_order_approved_free = I18nFormField(
+        label=_("Approved free order"),
+        required=False,
+        widget=I18nTextarea,
+        help_text=_("This will only be sent out for free orders. Non-free orders will receive the free order "
                     "template from above instead."),
     )
     mail_text_order_denied = I18nFormField(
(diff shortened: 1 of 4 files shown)


src/pretix/control/forms/event.py
```diff
@@ -929,6 +929,13 @@ class MailSettingsForm(SettingsForm):
         required=False,
         widget=I18nTextarea,
         help_text=_("This will only be sent out for non-free orders. Free orders will receive the free order "
+                    "template from below instead."),
+    )
+    mail_text_order_approved_free = I18nFormField(
+        label=_("Approved free order"),
+        required=False,
+        widget=I18nTextarea,
+        help_text=_("This will only be sent out for free orders. Non-free orders will receive the free order "
```
felixrindt, line 938:
```suggestion
        help_text=_("This will only be sent out for free orders. Non-free orders will receive the non-free order "
```

pc-coholic:
@felixrindt So it has been decided, that we need to copy over the `mail_text_order_free` into the new `mail_text_order_approved_free`... Since you have recently gained experience in writing Settings-Migrations: Feel up to do that? :-)

pc-coholic:
For most, the new behaviour will probably all right, as the new approval text is basically the same as the old one.

However, if you are running your own pretix instance and want to preserve the same confirmation text as before, run this in your pretix-shell:
`for s in Event_SettingsStore.objects.filter(key='mail_text_order_free'): 
     s.id = None 
     s.key = 'mail_text_order_approved_free' 
     s.save() 
`

PR #1753, opened 2020-08-18
Allow to reduce the interval of some cronjobs

raphaelm:


raphaelm:
This will especially also be helpful for some cronjobs in plugins of ours (campaigns, coronashredder, pretixeu.sales, …)

felixrindt (approved):

PR #1760, opened 2020-08-28
Payment term in minutes

felixrindt:


felixrindt:
Failing tests are from master if I spotted that correctly.

felixrindt:
I found it hard to come up with a sensible UI while keeping code simple and expressive. Developing a new widget felt strange in the way that we are not only changing on piece of data, but behavior is different.

-------

image

-------

image

src/pretix/base/settings.py
```diff
@@ -464,7 +480,48 @@ def unserialize(cls, s):
...
+    },
+    'payment_term_minutes': {
+        'default': '30',
+        'type': int,
+        'form_class': forms.IntegerField,
+        'serializer_class': serializers.IntegerField,
+        'form_kwargs': dict(
+            label=_('Payment term in minutes'),
+            help_text=_("The number of minutes after placing an order the user has to pay to preserve their reservation. Only use this for real-time "
+                        "payment methods."),
```
raphaelm, line 500: Maybe: "Only use this if you exclusively offer real-time payment methods."? Otherwise sounds as if you could do it per-payment-method.

src/pretix/base/settings.py
```diff
@@ -2015,7 +2071,7 @@ def __delitem__(self, key: str) -> None:
     def __delattr__(self, key: str) -> None:
         del self._event.settings[self._convert_key(key)]
 
-    def get(self, key: str, default: Any=None, as_type: type=str):
```
raphaelm, line 2018: same here

src/pretix/presale/templates/pretixpresale/event/order.html
```diff
@@ -112,51 +122,52 @@ <h3 class="panel-title">
                 <ul class="list-group">
                     {% for r in refunds %}
                         <li class="list-group-item">
-                            {% if r.state == "created" or r.state == "transit" %}
+                        {% if r.state == "created" or r.state == "transit" %}
```
raphaelm, line 125: What happened here? I don't think I understand the  diff github shows :D

src/pretix/base/models/orders.py
```diff
@@ -545,7 +551,7 @@ def user_cancel_allowed(self) -> bool:
             return self.event.settings.cancel_allow_user_paid
         return False
 
-    def propose_auto_refunds(self, amount: Decimal, payments: list=None):
```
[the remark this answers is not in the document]
felixrindt: image

src/pretix/base/email.py
```diff
@@ -319,6 +319,10 @@ def base_placeholders(sender, **kwargs):
             lambda event: LazyDate(now() + timedelta(days=15))
             # TODO: This used to be "date" in some placeholders, add a migration!
         ),
+        SimpleFunctionalMailTextPlaceholder(
+            'expire_time', ['event', 'order'], lambda event, order: LazyTime(order.expires.astimezone(event.timezone)),
+            lambda event: LazyTime(now() + timedelta(minutes=30))
```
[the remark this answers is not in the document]
felixrindt: Same goes for the order detail page, I agree.

src/pretix/base/email.py
```diff
@@ -315,9 +315,8 @@ def base_placeholders(sender, **kwargs):
             lambda event: LazyCurrencyNumber(Decimal('42.23'), event.currency)
         ),
         SimpleFunctionalMailTextPlaceholder(
-            'expire_date', ['event', 'order'], lambda event, order: LazyDate(order.expires.astimezone(event.timezone)),
+            'expire_date', ['event', 'order'], lambda event, order: LazyExpiresDate(order.expires.astimezone(event.timezone)),
             lambda event: LazyDate(now() + timedelta(days=15))
-            # TODO: This used to be "date" in some placeholders, add a migration!
```
felixrindt, line 320: This was added last year.

raphaelm: oops

raphaelm (approved):

src/pretix/base/settings.py
```diff
@@ -464,7 +480,49 @@ def unserialize(cls, s):
             label=_('Only end payment terms on weekdays'),
             help_text=_("If this is activated and the payment term of any order ends on a Saturday or Sunday, it will be "
                         "moved to the next Monday instead. This is required in some countries by civil law. This will "
-                        "not effect the last date of payments configured above."),
+                        "not effect the last date of payments configured below."),
+            widget=forms.CheckboxInput(
+                attrs={
+                    'data-display-dependency': '#id_payment_term_mode_0',
+                    'data-required-if': '#id_payment_term_mode_0'
```
inwwin, line 487: @felixrindt I'm curious. Why do we need to force events that have `payment_term_mode` set to `days` to only end payment terms on weekdays? What if this is not required by the law in the country where the event is located? Is this a feature or a bug?

felixrindt: This indeed seems like an oversight. We only need it to be displayed, but not checked, when `days` are selected. Glad you noticed it, thanks! I created #1862 to fix it.

PR #1762, opened 2020-08-31
Improvements for bank transfer importing

felixrindt:
* Get IBAN and BIC from bank transfer imports and display them in the control interface by storing them in the payment.
* better filter form for filtering bank transfer statements.
* bank transfers are refundable if iban and bic was captured.
* bank transfer refunds can be exported in batches, also as a sepa transfer xml (if currency is EUR)


felixrindt:
You can now also export refunds.

image

src/pretix/plugins/banktransfer/models.py
```diff
@@ -80,3 +86,37 @@ def shred_private_data(self):
...
+
+    @cached_property
+    def currency(self):
+        if self.event:
+            return self.event.currency
+        return self.organizer.events.first().currency
+
+    @cached_property
+    def rows_data(self):
+        return json.loads(self.rows)
```
raphaelm, line 114: A ``@cached_property`` on a model can get really confusing, if you change ``rows``, ``rows_data`` won't change. I think it's fine for the entity_slug and currency since they should never change, but I'd avoid it for the other ones, even if that causes a few more CPU cycles

src/pretix/plugins/banktransfer/payment.py
```diff
@@ -234,3 +235,21 @@ def shred_payment_info(self, obj):
...
+        return self.payment_refund_supported(payment)
+
+    def execute_refund(self, refund: OrderRefund):
+        """
+        We just keep a created refund object. It will be marked as done using the control view
+        for bank transfer refunds.
+        """
+        if refund.payment:
+            refund.info_data = {key: value for key, value in refund.payment.info_data.items() if key in {"payer", "iban", "bic"}}
+            refund.save(update_fields=["info"])
```
raphaelm, line 252: … and else? We should better throw an exception, or is there a valid use case of the else branch?

src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/transaction_list.html
```diff
@@ -4,7 +4,14 @@
...
     {% csrf_token %}
-    <table class="table table-condensed transaction-list" data-url="{% if request.event %}{% url "plugins:banktransfer:import.action" event=request.event.slug organizer=request.organizer.slug %}{% else %}{% url "plugins:banktransfer:import.action" organizer=request.organizer.slug %}{% endif %}">
+    <table class="table table-condensed transaction-list" data-url="
+
+
+
+
+
+
+            {% if request.event %}{% url "plugins:banktransfer:import.action" event=request.event.slug organizer=request.organizer.slug %}{% else %}{% url "plugins:banktransfer:import.action" organizer=request.organizer.slug %}{% endif %}">
```
raphaelm, line 14:
```suggestion
    <table class="table table-condensed transaction-list" data-url="{% if request.event %}{% url "plugins:banktransfer:import.action" event=request.event.slug organizer=request.organizer.slug %}{% else %}{% url "plugins:banktransfer:import.action" organizer=request.organizer.slug %}{% endif %}">
```

src/pretix/plugins/banktransfer/views.py
```diff
@@ -274,6 +291,33 @@ def get_context_data(self, **kwargs):
         return ctx
 
 
+class BankTransactionFilterForm(forms.Form):
+    search_text = forms.CharField(required=False, widget=forms.TextInput(attrs={'class': "form-control", "placeholder": _("Search text")}))
+    amount_min = forms.DecimalField(required=False, widget=forms.NumberInput(attrs={'class': "form-control", "placeholder": _("min"), "size": 8}))
+    amount_max = forms.DecimalField(required=False, widget=forms.NumberInput(attrs={'class': "form-control", "placeholder": _("max"), "size": 8}))
```
raphaelm, line 297: This will work badly with locales like German where you want to enter numbers with comma. Fix: pass ``localize=True`` to the field and don't use NumberInput but a normal TextInput instead

src/pretix/plugins/banktransfer/views.py
```diff
@@ -543,3 +584,223 @@ def order_qs(self):
...
+
+    @transaction.atomic()
+    def post(self, request, *args, **kwargs):
+        unite_transactions = request.POST.get("unite_transactions", False)
+        valid_refunds: Set[OrderRefund] = set()
+        for refund in self.get_unexported().select_related('order', 'order__event'):
+            if not refund.info_data:
+                # Should not happen
+                messages.warning(request,
+                                 _("We could not find bank account information for the refund {}. It was marked as failed.")
```
raphaelm, line 643: In translatables, please always use a named parameter so the translator has a hint what will be inserted there.

src/pretix/plugins/banktransfer/views.py
```diff
@@ -543,3 +584,223 @@ def order_qs(self):
...
+        )
+
+
+class SepaXMLExportForm(forms.Form):
+    account_holder = forms.CharField(label=_("Account holder"))
+    iban = IBANFormField(label="IBAN")
+    bic = BICFormField(label="BIC")
+
+    def set_initial_from_event(self, event):
+        ...
```
raphaelm, line 757: ;)

src/pretix/plugins/banktransfer/urls.py
```diff
@@ -21,6 +29,15 @@
...
+    url(r'^control/event/(?P<organizer>[^/]+)/(?P<event>[^/]+)/banktransfer/refunds/',
+        views.EventRefundExportListView.as_view(),
+        name='refunds.list'),
+    url(r'^control/event/(?P<organizer>[^/]+)/(?P<event>[^/]+)/banktransfer/export/(?P<id>\d+)/$',
+        views.EventDownloadRefundExportView.as_view(),
+        name='refunds.download'),
+    url(r'^control/event/(?P<organizer>[^/]+)/(?P<event>[^/]+)/banktransfer/sepa-export/(?P<id>\d+)/$',
+        views.EventSepaXMLExportView.as_view(),
+        name='refunds.sepa'),
 ]
```
raphaelm, line 42: Can you add all these to the various array in ``test_permissions.py`` to make sure they are restricted correctly?

raphaelm (changes requested):
Have read all the code, but not yet played with it.

General stuff:

* Tests on master are fixed now, so a rebase of this should fix the CI as well. I'm really curious to see the test coverage on this one, it might give me an idea if the tests are enough or if we're missing something important.

* We really, really want to prevent invalid IBANs and BICs. If you end up with an invalid IBAN or BIC in your data set, you will be a support case, because either ``sepaxml`` just fails with an XML validation error, or you get an XML file and your bank will not accept it. We should therefore add a step that validates and normalizes (i.e. uppercase, spaces removed) IBANs and BICs. I'm not sure whether we should do so during import and just ignore invalid IBANs (which might lead to missing data), or whether we should to so when creating the refund / checking if refunds are possible, which leaves more room for the data to be corrected inside pretix. Either way, we should prevent creation of a a refund with invalid data. ``django-localflavor`` has validation logic for both things.

src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/transaction_list.html
felixrindt: I don't know why github shows this. Looks fine in my editor...

PR #1763, opened 2020-09-02
Fix #1133 -- Validate email addresses for valid DNS names

### LINKED ISSUE #1133: Validate email addresses earlier
Sometimes users simply mistype their email domains :( I would suggest to do MX, A, and AAAA lookups on the domain of the email address, and if none of that exists then flag the email as suspicious. Either entirely reject it or offer the user to override this, but least force her/him to look again.

raphaelm:
TODO

- [ ] is this a good idea? does this have bad performance impact?

### DIFF
--- a/src/pretix/presale/forms/checkout.py
+++ b/src/pretix/presale/forms/checkout.py
@@ -1,6 +1,6 @@
 from itertools import chain
 
-import dns
+import dns.resolver
 from django import forms
 from django.core.cache import cache
 from django.core.exceptions import ValidationError
--- a/src/pretix/presale/forms/checkout.py
+++ b/src/pretix/presale/forms/checkout.py
@@ -1,6 +1,8 @@
 from itertools import chain
 
+import dns
 from django import forms
+from django.core.cache import cache
 from django.core.exceptions import ValidationError
 from django.utils.encoding import force_str
 from django.utils.translation import gettext_lazy as _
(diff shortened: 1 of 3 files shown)


src/pretix/presale/forms/checkout.py
```diff
@@ -18,12 +20,52 @@
...
+                    return value
+            except (dns.resolver.NoAnswer, dns.resolver.NXDOMAIN, dns.resolver.YXDOMAIN):
+                # These errors represent a non-existing domain or entry, we want to continue with the
+                # other types of records and if all fail, we'll report an error.
+                continue
+            except:
+                # Timeout, IO errors, etc.: We don't know what's happening! Let's not stop the user
+                # from buying and just not validate this. Let's cache this result for a short while
+                # to prevent issues during peak load
+                cache.set(f"mail_domain_exists_{domain}", "false", 60)
```
Techwolf12, line 51: You pass the current one, but set the cache value to `false` here.
This would mean, that if the DNS server is having a brief outage, a domain like `gmail.com` could be blocked for 60 seconds. I think you'd want to use `true` in the cache here.

Techwolf12:
Regardless of performance impact, which should be minimal but can quickly add up to a large install where people have a lot of custom domains during a peak load.

If this is implemented, I'd personally like to see it as a configurable option.

raphaelm:
In the 4 years since creating this PR, we haven't really missed it, so we probably don't really need it.

PR #1767, opened 2020-09-11
Copy answers from matching product

felixrindt:


src/pretix/base/views/mixins.py
```diff
@@ -53,7 +54,31 @@ def forms(self):
...
+                            if other_form.pos.addon_to_id is None and form.pos.item.id == other_form.pos.item.id
+                        ),
+                        (   # match a position with questions in common
+                            other_form.pos.id for other_form in formlist
+                            if other_form.pos.addon_to_id is None
+                            and set(form.pos.item.questions.values_list("id", flat=True)) & set(other_form.pos.item.questions.values_list("id", flat=True))
+                        )
+                    ),
+                    None  # didn't find a position to copy answers from
+                )
```
raphaelm, line 80: Oof, wow.

This has a level of complexity that I am inclined to just say the complexity isn't worth the feature. Especially, since there's still a lot missing, e.g. support for the "system questions" like name and address for admission products. (Also for ``n`` positions, this does ``2*n`` SQL queries to the questions table, which isn't really great.)

@felixrindt @pc-coholic What do you think? Keep working on this or drop it?

raphaelm (changes requested):

src/pretix/base/views/mixins.py
felixrindt: Couldn't both issues be fixed by matching answer inputs to copy from client-side by looking at the question input suffixes? (could opt for copying from only one item, or multiple)

I think I blindly implemented this in python so I could avoid rendering the button if nothing would happen, but that could be moved to JS as well.

raphaelm: Yes, I think they would, and I think a JS-only implementation would be better in this case.

(I'm not sure if this is worth spending more hours on it, though. Maybe one or two, but no more.)

raphaelm:
Closing as wontfix for now

PR #1774, opened 2020-09-16
Mail service: Allow to attach arbitrary cached files

felixrindt:
When building a plugin, people might want to use the already existing mail infrastructure to send mails.

Currently, it is not possible to send attachments using pretix' mail system. This PR tries to solve that by allowing callers to attach cached files, which might need to be created solely for attaching a file to a mail. This is necessary, because pretix uses distributed celery workers, so we cannot just give plain file/content (also mail_send task's arguments are JSON-serialized).

### DIFF
--- a/src/pretix/base/services/mail.py
+++ b/src/pretix/base/services/mail.py
@@ -7,7 +7,7 @@ import ssl
 import warnings
 from email.mime.image import MIMEImage
 from email.utils import formataddr
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List, Sequence, Union
 from urllib.parse import urljoin, urlparse
 
 import cssutils
@@ -27,7 +27,7 @@ from i18nfield.strings import LazyI18nString
 from pretix.base.email import ClassicMailRenderer
 from pretix.base.i18n import language
 from pretix.base.models import (
-    Event, Invoice, InvoiceAddress, Order, OrderPosition, User,
+    CachedFile, Event, Invoice, InvoiceAddress, Order, OrderPosition, User,
 )
 from pretix.base.services.invoices import invoice_pdf_task
 from pretix.base.services.tasks import TransactionAwareTask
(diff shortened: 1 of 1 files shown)


src/pretix/base/services/mail.py
```diff
@@ -52,10 +52,11 @@ class SendMailException(Exception):
...
 
-def mail(email: str, subject: str, template: Union[str, LazyI18nString],
-         context: Dict[str, Any]=None, event: Event=None, locale: str=None,
-         order: Order=None, position: OrderPosition=None, headers: dict=None, sender: str=None,
-         invoices: list=None, attach_tickets=False, auto_email=True, user=None, attach_ical=False):
+def mail(email: Union[str, Sequence[str]], subject: str, template: Union[str, LazyI18nString],
+         context: Dict[str, Any] = None, event: Event = None, locale: str = None,
+         order: Order = None, position: OrderPosition = None, headers: dict = None, sender: str = None,
+         invoices: Sequence = None, attach_tickets=False, auto_email=True, user=None, attach_ical=False,
+         attach_cached_files: Sequence = None):
```
felixrindt, line 59: Again, I autoformatted here, but pep 8 seems to recommend spaces around the `=` with type annotations.

felixrindt, line 59: Downside here is that these are *a lot* of kwargs. But I don't think I can change the logic for tickets/invoices at this place.

Would be more clean to have a general-purpose `mail` and a `mail_order` that can do all the other stuff, but to delegate, invoices would need to be copied to a cached file, which we probably don't want for this often-called task.

src/pretix/base/services/mail.py
```diff
@@ -349,6 +353,19 @@ def mail_send_task(self, *args, to: List[str], subject: str, body: str, html: st
...
                         pass
 
+        if attach_cached_files:
+            for cf in CachedFile.objects.filter(id__in=attach_cached_files):
+                if cf.file:
+                    try:
+                        email.attach(
+                            cf.filename,
+                            cf.file.file.read(),
+                            mimetype=None  # TODO: can we get that from cachedfile?
```
raphaelm, line 363: I think there is a ``cf.type`` ;)

raphaelm (commented):
Looks good apart from the mime type thing. mime types are important.

src/pretix/base/services/mail.py
felixrindt: Of course there is! They are important... :see_no_evil:

raphaelm (approved):

PR #1779, opened 2020-09-22
Add name scheme with salutation

felixrindt:


src/pretix/base/email.py
```diff
@@ -241,6 +242,8 @@ def get_email_context(**kwargs):
             kwargs['invoice_address'] = kwargs['order'].invoice_address
         except InvoiceAddress.DoesNotExist:
             kwargs['invoice_address'] = InvoiceAddress()
+        finally:
+            kwargs.setdefault("position_or_address", kwargs['invoice_address'])
```
felixrindt, line 246: The changes in this file makes the already existing `{name_*}` placeholders available always when there's an order, because we can always use the orders invoice address as a fallback for lookup (although it might be set to an empty one).

src/pretix/base/forms/questions.py
```diff
@@ -73,6 +73,8 @@ def __init__(self, scheme: dict, field: forms.Field, attrs=None, titles: list=No
             a['data-fname'] = fname
             if fname == 'title' and self.titles:
                 widgets.append(Select(attrs=a, choices=[('', '')] + [(d, d) for d in self.titles[1]]))
+            elif fname == 'salutation':
+                widgets.append(Select(attrs=a, choices=[('', '---')] + [(s, s) for s in PERSON_NAME_SALUTATIONS]))
```
felixrindt, line 77: I put the `---` to indicate that a choice *must* be made. The title doesn't have it, as it can be left empty.

I just noticed that validation should therefore be different then.

src/pretix/base/forms/questions.py
```diff
@@ -162,12 +164,18 @@ def __init__(self, *args, **kwargs):
...
                 )
-                field.part_name = fname
-                fields.append(field)
+
+            elif fname == 'salutation':
+                d = dict(defaults)
+                d.pop('max_length', None)
+                field = forms.ChoiceField(
+                    **d,
+                    choices=[('', '---')] + [(s, s) for s in PERSON_NAME_SALUTATIONS]
```
felixrindt, line 173: Is there a war to not duplicate the choices list creation?

src/pretix/base/settings.py
```diff
@@ -1836,17 +1836,24 @@ def unserialize(cls, s):
...
+
+PERSON_NAME_SALUTATIONS = [
+    pgettext_lazy("person_name_salutation", "Ms"),
+    pgettext_lazy("person_name_salutation", "Mr"),
+]
+
 PERSON_NAME_SCHEMES = OrderedDict([
     ('given_family', {
         'fields': (
+            # field_name, label, width-weight
```
felixrindt, line 1856: I spent a few minutes finding out what the number was :sweat_smile:

src/pretix/base/forms/questions.py
felixrindt: Ah nevermind there is `REQUIRED_NAME_PARTS` :+1:

src/pretix/base/forms/questions.py
raphaelm: Not sure if worth it

src/pretix/base/settings.py
raphaelm: Feel free to improve the comment :D

src/pretix/base/settings.py
```diff
@@ -2001,6 +2008,24 @@ def unserialize(cls, s):
             '_scheme': 'full_transcription',
         },
     }),
+    ('title_salutation_given_family', {
```
raphaelm, line 2011: Maybe we should call this salutation_title_given_family if that's the order? ;)

raphaelm (approved):

src/pretix/base/settings.py
felixrindt: woopsy yes!

src/pretix/base/settings.py
felixrindt: well the comment wasn't there before :sweat_smile:

src/pretix/base/settings.py
```diff
@@ -2001,6 +2008,24 @@ def unserialize(cls, s):
...
+        'fields': (
+            ('salutation', pgettext_lazy('person_name', 'Salutation'), 1),
+            ('title', pgettext_lazy('person_name', 'Title'), 1),
+            ('given_name', _('Given name'), 2),
+            ('family_name', _('Family name'), 2),
+        ),
+        'concatenation': lambda d: ' '.join(
+            str(p) for p in (d.get(key, '') for key in ["salutation", "title", "given_name", "family_name"]) if p
```
felixrindt, line 2019: Just thought about not having the salutation here, as it's probably only used in Mail templates? @raphaelm

raphaelm: Thought about this as well and I think it's a good idea.

felixrindt: :+1:

PR #1788, opened 2020-09-28
Support required-if and display-dependency for more elements

felixrindt:
I'd like this for a plugin.
Want me to change it for presale js as well?


raphaelm:
Yeah, would be easier to reason about if it were consistent, I guess

Issue #1758, opened 2020-08-25
Shredding personal data without exporting to personal computers


MaicoTimmerman:
Currently, it is required to download all the data before confirming the shredding the users data. To me this seems counterproductive towards protecting the data of the user. Shredding the data is to protect the consumer's privacy, by removing data that is no longer needed such as answers to (sensitive) questions. By requiring to download the data we are moving personal data from the server onto personal computers. These copies of the data are more prone to be part of a data breach as they can be accidentally shared, stolen or lost. Human mistakes are the most common in data breaches.

I'd like to be able to make the data download optional and confirm the deletion only by confirming the event's short name.

PR #1801, opened 2020-10-09
Shredder: Only force download for tax-relevant data

MaicoTimmerman:
Using this PR, I'd like to start the discussion on a potential solution to #1758. At our association we process a lot of personal data in Pretix, which is done by many volunteers. We try to take good care of personal data of our members, however when executing the data shred process all this personal data has to be downloaded to personal computers, resulting in a loss of control.

I've currently left the download file generating in, depending on how we'd like this to work I can adjust and expand the implementation.

Please let me know what you think.

### DIFF
--- a/src/pretix/base/services/shredder.py
+++ b/src/pretix/base/services/shredder.py
@@ -73,16 +73,19 @@ def shred(event: Event, fileid: str, confirm_code: str) -> None:
         indexdata = json.loads(zipfile.read('index.json').decode())
     if indexdata['organizer'] != event.organizer.slug or indexdata['event'] != event.slug:
         raise ShredError(_("This file is from a different event."))
-    if indexdata['confirm_code'] != confirm_code:
-        raise ShredError(_("The confirm code you entered was incorrect."))
-    if event.logentry_set.filter(datetime__gte=parse(indexdata['time'])):
-        raise ShredError(_("Something happened in your event after the export, please try again."))
-
+    shredders = []
     for s in indexdata['shredders']:
         shredder = known_shredders.get(s)
         if not shredder:
             continue
+        shredders.append(shredder)
+    if any(shredder.require_download_confirmation for shredder in shredders):
+        if indexdata['confirm_code'] != confirm_code:
+            raise ShredError(_("The confirm code you entered was incorrect."))
+    if event.logentry_set.filter(datetime__gte=parse(indexdata['time'])):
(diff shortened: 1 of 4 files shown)


MaicoTimmerman:
I think the risk of permanent deletion while the data is needed for e.g. tax audits will always be present, no matter how many verification steps are imposed.

The current safeguards present are the downloaded code, together with a download of the data to be shredded and there are alerts present for the Invoice addresses, Invoices and Payment information shredders. To improve these safeguards I'm thinking:
1. Replace the current instant solution with the ability to schedule a shred. This way the user will be able to cancel the shredding of personal data and we are able to notify administrators that data is to be shredded. 
2. Disable certain system shredders on a system/organizer level or limit certain shredders to administrative users only.

Secondly, I'm proposing to leave the download safeguard available, however my primary goal is to make it optional with some form of configuration, such that we can disable it for our users.

raphaelm:
Yeah, I agree. We've proabably overdone it a bit with the safeguards.

> Replace the current instant solution with the ability to schedule a shred. This way the user will be able to cancel the shredding of personal data and we are able to notify administrators that data is to be shredded.

That would be an amazing solution! Doesn't need to be in the scope of this PR, but would you be interested in working on that?

> Disable certain system shredders on a system/organizer level or limit certain shredders to administrative users only.

Hmm, yeah, although our permission system isn't really fine-grained enough to make this straightforward.

> Secondly, I'm proposing to leave the download safeguard available, however my primary goal is to make it optional with some form of configuration, such that we can disable it for our users.

I'd prefer less configuration. Maybe we can make the download safeguard dependent on whether invoice and payment data is affected? That would probably help your use case as well, wouldn't it?

MaicoTimmerman:
> That would be an amazing solution! Doesn't need to be in the scope of this PR, but would you be interested in working on that?
I'm limited on time, so I'd first focus on issues that directly affect my organisation. I'll update #1758 when this PR has been merged, so we can start the implementation discussion.

> I'd prefer less configuration. Maybe we can make the download safeguard dependent on whether invoice and payment data is affected? That would probably help your use case as well, wouldn't it?

I've updated the PR by adding a new property to the base shredder, defining whether the data should be downloaded and confirmed using the `CONFIRM_CODE.txt`. I've left the download button as optional for the other shredders, as otherwise it would render the `generate_files` function irrelevant for all shredders that do not have `require_download_confirmation`.

raphaelm:
I like this solution!

PR #1807, opened 2020-10-13
Docker: use the healthcheck endpoint

fridtjof:
This adds a `HEALTHCHECK` (https://docs.docker.com/engine/reference/builder/#healthcheck) directive to the Dockerfile. It uses the /healthcheck/ endpoint documented here: https://docs.pretix.eu/en/latest/admin/maintainance.html#uptime-monitoring

For an example of another project using this, check out the `synapse` Matrix homeserver: https://github.com/matrix-org/synapse/blob/develop/docker/Dockerfile#L81

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -70,5 +70,7 @@ RUN chmod +x /usr/local/bin/pretix && \
 USER pretixuser
 VOLUME ["/etc/pretix", "/data"]
 EXPOSE 80
+HEALTHCHECK --interval=1m --timeout=2m \
+  CMD curl -fSs http://localhost/healthcheck/ || exit 1
 ENTRYPOINT ["pretix"]
 CMD ["all"]


raphaelm:
Nice idea, I have a few concerns.,

* ``/healthcheck/`` only returns a valid response if both the SQL database and redis work properly. ``/healthckeck/`` failing does not mean the docker container is faulty, it's actually more likely an external component is failing. If docker itself, or some orchestration tool, would then restart the docker container due to the "unhealthy" state, this might even be harmful e.g. in a high-load situation.

* Have you tested this with a production-like config? I have a hunch that it will always fail because localhost is not the correct hostname of the pretix installation as soon as an url is set in pretix.cfg

fridtjof:
I just tested on a production instance with a URL set, and `/healthcheck/` returns an empty 200. Looks okay to me.

I did not know about your first point though, that might indeed be troublesome. For my use case, it would just affect the output of `docker ps` and the like, but as you said, who knows what other tools might do with it?

If you want to keep the current `/healthcheck/` behavior, that's perfectly fine, and we can close the PR in that case. Otherwise, I am sorry I won't have time to work on that for now, as it goes beyond something I could quickly whip up while being sure enough it won't break without testing.

raphaelm:
Yes, the current ``/healthcheck/`` behaviour needs to stay, people rely on it for monitoring and we at pretix.eu even rely on it for load balancing. We don't only want to take a pretix server out of rotation if the webserver hangs (which I haven't heard of once in all my career), but also if it is no longer able to talk to the database.

If anything, we'd need to add another endpoint like ``/healthcheck/webserver/``.

I'm closing this for now, since its unclear what use it has, might actively be harmful, and requires further design. If someone (possibly with docker tooling experience) wants to work on this in the future, I'm happy to talk about it again.

PR #1809, opened 2020-10-15
Add pluggable ticket secret generators

raphaelm:
- [x] trigger re-generation whenever item, variation or subevent is changed
- [x] expose keys to api
- [x] revocation list api
- [x] resolve 404 in online scanning
- [x] libpretixsync integration, incl revokelist  https://github.com/pretix/libpretixsync/pull/7
  - [x] pretixscan android https://github.com/pretix/pretixscan-android/pull/22
  - [ ] pretixscan desktop
  - [ ] make sure ios works online (charset is accepted)
- [x] documentation
- [ ] support for device-specific keys? → later
- [x] fix qr-code-overlay
- [x] revoke on cancellation (how to deal with reactivate)
- [x] tests


src/pretix/base/models/orders.py
```diff
@@ -2326,6 +2332,12 @@ class CancellationRequest(models.Model):
     refund_as_giftcard = models.BooleanField(default=False)
 
 
+class TicketSecretBlacklist(models.Model):
```
felixrindt, line 2335: You probably want to call these Denylists? :)

raphaelm: Argh, thanks, yes. I planned "revocation list" which is even preciser than blacklist, its already in the commit in other places, but I forgot the model. Thanks!

src/pretix/api/views/checkin.py
```diff
@@ -283,13 +283,23 @@ def redeem(self, *args, **kwargs):
...
+                }, user=self.request.user, auth=self.request.auth)
+                raise Http404()
+
+            op = revoked_matches[0].position
+            op.order.log_action('pretix.event.checkin.revoked', data={
                 'datetime': dt,
                 'type': type,
                 'list': self.checkinlist.pk,
                 'barcode': self.kwargs['pk']
             }, user=self.request.user, auth=self.request.auth)
```
pc-coholic, line 302: - [ ] `pretix.event.checkin.revoked` logaction does not exist yet
- [ ] Perhaps I'm not in the right mindspace - but should an online-scan of a revoked ticket be declined with "unknown" or actually with "revoked"? (I can see the both as valid usecases)
- [ ] Displaying the whole ticket-secret from the logentry on the dashboard was already a risky business before (due to space constraints) - but with super-long base64-encoded strings that won't wrap, it's almost guaranteed to break the layout

pc-coholic (changes requested):
LGTM so far; tested against pretixSCAN-android main branch with success

PR #1814, opened 2020-10-19
Sendmail plugin: Allow to attach a file to emails

Sohalt:
This allows to attach a file to emails send out by the sendmail plugin.

Current limitations:
- only allows for 1 attachment
- the attachment is the same for all languages


raphaelm:
Thanks a lot! :) This was mostly correct, but not quite (it didn't actually work for important cases, attachments had no file name). Since we're running a little low on time to review stuff here at the moment, I took the liberty of just finishing it up quickly ;)

Sohalt:
Thanks!

PR #1820, opened 2020-10-21
Update pytest version and related packages

wiffbi:
Updated requirements to fix pytest-django to version 3.* as latest version 4.0.0 requires pytest==5.4, which currently breaks tests.


raphaelm:
Thanks! When we discussed this previously, I understood that the only problem on pytest 5.4 was a "real" failure that has now been fixed. Did I understand that wrongly?

wiffbi:
Sorry, for the missunderstanding: pytest 5.4 fails 1 Test and skipped 3. Also several RemovedInDjango40Warnings were emitted. I will look into this and use this branch to update.

wiffbi:
Sorry for the confusion, all tests (except 3 skipped, 1 xfailed) pass with pytest 5.4. Updated requirements and setup.py to always use latest versions of pytest and pytest-django.

raphaelm:
On CI, there seems to be a failure in pytest-rerunfailures, maybe we need to upgrade that package as well?

PR #1824, opened 2020-10-23
Allow plugins to disable and pre-fill questions and contact form fields

pc-coholic:


src/pretix/presale/checkoutflow.py
```diff
@@ -412,6 +413,9 @@ class QuestionsStep(QuestionsViewMixin, CartMixin, TemplateFlowStep):
     identifier = "questions"
     template_name = "pretixpresale/event/checkout_questions.html"
     label = pgettext_lazy('checkoutflow', 'Your information')
+    contact_overrides = False
+    contact_overrides_initial = {}
+    contact_overrides_disabled = {}
```
raphaelm, line 418: This is generally a bad idea. The way properties work in python, this will be **the same dict** with shared content across all requests handled on this Django thread, so state can leak between requests. These 3 variables should be initialized in ``__init__``

src/pretix/presale/checkoutflow.py
```diff
@@ -423,6 +427,19 @@ def all_optional(self):
...
         return False
 
+    def get_contact_overrides(self):
+        overrides = contact_form_fields_overrides.send_chained(
+            self.request.event,
+            'contact_form_fields_overrides',
+            request=self.request
+        )
+
+        self.contact_overrides = True
```
raphaelm, line 437: Can we name this ``contact_overrides_computed`` or something like that to make clear it doesn't say whether there are overrides, but only whether they've been handled already?

src/pretix/presale/signals.py
```diff
@@ -180,6 +180,20 @@
 argument will contain the request object.
 """
 
+contact_form_fields_overrides = EventPluginSignal(
+    providing_args=["request"]
+)
+"""
+This signals allows you to override fields of the contact form that is presented during checkout
```
raphaelm, line 187:
```suggestion
This signal allows you to override fields of the contact form that is presented during checkout
```

src/pretix/presale/signals.py
```diff
@@ -196,6 +210,22 @@
 As with all plugin signals, the ``sender`` keyword argument will contain the event.
 """
 
+question_form_fields_overrides = EventPluginSignal(
+    providing_args=["position", "request"]
+)
+"""
+This signals allows you to override fields of the questions form that is presented during checkout
```
raphaelm, line 217:
```suggestion
This signal allows you to override fields of the questions form that is presented during checkout
```

src/pretix/presale/signals.py
```diff
@@ -196,6 +210,22 @@
...
+and by default only asks for the questions configured in the backend. You are supposed to return a
+dictionary of dictionaries with globally unique keys. The value-dictionary should contain one or
+more of the following keys: ``initial``, ``disabled``. The key of the dictionary should not be the
+question's form field name (``question_n``) but rather the questions ``identifier``.
+
+The ``position`` keyword argument will contain a ``CartPosition`` object.
```
raphaelm, line 223:
```suggestion
The ``position`` keyword argument will contain a ``CartPosition`` or ``OrderPosition`` object.
```

It's always a cart for now, but let's document it this way, it will give us more options going forward.

raphaelm (commented):
Thanks! :+1: 

* Does the saving of questions still work, even though this does not include the patch for the form that we discussed?

* Tests complain about the two signals not referenced in general.rst in the docs

* isort complains

raphaelm (approved):
LGTM now, @pc-coholic if you could run it once in your setup just to be safe I haven't broken anything with my changes, that'd be great :)

PR #1839, opened 2020-11-02
Fix #1780 -- Trigger exports through API

### LINKED ISSUE #1780: Question: Will it be feasible to trigger order exports via the api?
We're using the pretix api extensively to manage events, and we're currently exposing order lists to user without access to pretix using data from the api. This works great and now we're looking into also exposing answers to order questions for these users.

We can get all the data, i.e. questions and answers, from the api, but we then have to render the data in a way that makes it usable for the users. This gives rise to the following question:

Will it be feasible to make it possible to trigger (and download the result of) an order export, e.g. “Check-in list (PDF)”, through the pretix api?

We appreciate that the exports have a lot of individual options and that we have to pass all these options along when triggering an export, but before getting into the details on this, we'd like to know if it is at all feasible to extend the api to be able to do something like this.

Best regards,
Mikkel

raphaelm:
- [x] Test handling of (model|multiple) choice fields
- [x] Implement organizer-level exports
- [x] Add tests
- [x] Add documentation

### DIFF
--- a/doc/api/resources/exporters.rst
+++ b/doc/api/resources/exporters.rst
@@ -1,0 +1,215 @@
+.. spelling:: checkin
+
+Data exporters
+==============
+
+pretix and it's plugins include a number of data exporters that allow you to bulk download various data from pretix in
+different formats. This page shows you how to use these exporters through the API.
+
+.. versionchanged:: 3.13
+
+   This feature has been added to the API.
+
+.. warning::
+
+   While we consider the methods listed on this page to be a stable API, the availability and specific input field
+   requirements of individual exporters is **not considered a stable API**. Specific exporters and their input parameters
+   may change at any time without warning.
+
+Listing available exporters
+---------------------------
+
+.. http:get:: /api/v1/organizers/(organizer)/events/(event)/exporters/
+
+   Returns a list of all exporters available for a given event. You will receive a list of export methods as well as their
+   supported input fields. Note that the exact type and validation requirements of the input fields are not given in the
+   response, and you might need to look into the pretix web interface to figure out the exact input required.
+
+   **Example request**:
+
+   .. sourcecode:: http
+
+      GET /api/v1/organizers/bigevents/events/sampleconf/exporters/ HTTP/1.1
+      Host: pretix.eu
+      Accept: application/json, text/javascript
+
+   **Example response**:
+
+   .. sourcecode:: http
+
+      HTTP/1.1 200 OK
+      Vary: Accept
+      Content-Type: application/json
+
+      {
+        "count": 1,
+        "next": null,
+        "previous": null,
+        "results": [
+          {
+            "identifier": "orderlist",
+            "verbose_name": "Order data",
+            "input_parameters": [
+              {
+                "name": "_format",
+                "required": true,
+                "choices": [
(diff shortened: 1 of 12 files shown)


src/pretix/api/views/exporters.py
```diff
@@ -0,0 +1,150 @@
...
+        cf = get_object_or_404(CachedFile, id=kwargs['cfid'])
+        if cf.file:
+            resp = ChunkBasedFileResponse(cf.file.file, content_type=cf.type)
+            resp['Content-Disposition'] = 'attachment; filename="{}"'.format(cf.filename)
+            return resp
+        elif not settings.HAS_CELERY:
+            return Response(
+                {'status': 'failed', 'message': 'Unknown file ID or export failed'},
+                status=status.HTTP_410_GONE
+            )
```
pc-coholic, line 57: I'm not 100% sure if the error-message is 100% accurate for non-celery systems...

doc/api/resources/exporters.rst
```diff
@@ -0,0 +1,215 @@
...
+        "download": "https://pretix.eu/api/v1/organizers/bigevents/events/sampleconf/orderlist/download/29891ede-196f-4942-9e26-d055a36e98b8/3f279f13-c198-4137-b49b-9b360ce9fcce/"
+      }
+
+   :param organizer: The ``slug`` field of the organizer to fetch
+   :param event: The ``slug`` field of the event to fetch
+   :param identifier: The ``identifier`` field of the exporter to run
+   :statuscode 202: no error
+   :statuscode 400: Invalid input options
+   :statuscode 401: Authentication failure
+   :statuscode 403: The requested organizer/event does not exist **or** you have no permission to view this resource.
```
pc-coholic, line 194: I think I managed once to get a 302 if the export was still cached?

pc-coholic (changes requested):
Looks generally okay.

Issues I ran into (non-celery setup):
- Triggering a large export (all tickets/all invoices) that will run past the Connection Timeout leave the user with no download URL.
- `emailaddrs` export fails without optional `status`-parameter.
- `emailaddrs` export also fails with `"status": ["p"]`

doc/api/resources/exporters.rst
raphaelm: I don't see how anything in the code could trigger a 302 :thinking:

src/pretix/api/views/exporters.py
raphaelm: It is not. I think it's okay, there's not that much we can do on non-celery systems

raphaelm:
>Triggering a large export (all tickets/all invoices) that will run past the Connection Timeout leave the user with no download URL.

Yep, well, that's expected on a non-celery setup, I'd consider this wontfix

> emailaddrs export fails without optional status-parameter.

Then we should make that required (but not really scope of this PR)

> emailaddrs export also fails with "status": ["p"]

That's weird, do you recall what the failure was?

pc-coholic:
> > emailaddrs export also fails with "status": ["p"]
> 
> That's weird, do you recall what the failure was?

```
EncodeError at /api/v1/organizers/bigevents/exporters/mailaddrs/run/
Object of type set is not JSON serializable
```

PR #1844, opened 2020-11-04
Add seating statistics in shared reports

julia-luna:
Added total seats, sold seats, and reserved seats to the existing free seat counter.

image


pc-coholic (changes requested):
LGTM :tada: 

One tiny request:
- [x] Move the duplicate functions into the `EventMixin`

raphaelm (approved):

PR #1858, opened 2020-11-26
Add min/max validation for date, datetime, and number questions

raphaelm:


pc-coholic (changes requested):
- [ ] Datetime error messages are not localized: Violating the Min- or Max-Value of 01.11.2020 01:00 will be displayed as 01.11.2020 00:00 in the validation-error message

Apart from that: LGTM!

PR #1859, opened 2020-11-26
Move front page text above date selection

raphaelm:
Has been requested multiple times and probably also makes more sense


pc-coholic (approved):
Just one tiny change where the frontpage_text did not get displayed on events that had a logo set.

PR #1864, opened 2020-12-01
Added export of WaitingList for single and mutliple events with filter for voucher-status

wiffbi:
I copied the basic queryset for the WaitingList from `pretix/control/views/waitinglist.py:113`. I am not sure whether I really need to `prefetch_related`. It worked without, so it should probably be removed.

### DIFF
--- a/src/pretix/base/exporters/waitinglist.py
+++ b/src/pretix/base/exporters/waitinglist.py
@@ -1,0 +1,155 @@
+
+from collections import OrderedDict
+
+from django import forms
+from django.db.models import F, Q
+from django.dispatch import receiver
+from django.utils.timezone import now
+from django.utils.translation import gettext_lazy as _, pgettext_lazy
+
+from pretix.base.models.waitinglist import WaitingListEntry
+
+from ..exporter import ListExporter
+from ..signals import (
+    register_data_exporters, register_multievent_data_exporters,
+)
+
+
+class WaitingListExporter(ListExporter):
+    identifier = 'waitinglist'
+    verbose_name = _('Waiting list')
+
+    # map selected status to label and queryset-filter
+    status_filters = [
+        (
+            '',
+            _('All entries'),
+            lambda qs: qs
+        ),
+        (
+            'awaiting-voucher',
+            _('Waiting for a voucher'),
+            lambda qs: qs.filter(voucher__isnull=True)
+        ),
+        (
+            'voucher-assigned',
+            _('Voucher assigned'),
+            lambda qs: qs.filter(voucher__isnull=False)
+        ),
+        (
+            'awaiting-redemption',
+            _('Waiting for redemption'),
+            lambda qs: qs.filter(
+                voucher__isnull=False,
+                voucher__redeemed__lt=F('voucher__max_usages'),
+            ).filter(Q(voucher__valid_until__isnull=True) | Q(voucher__valid_until__gt=now()))
+        ),
+        (
+            'voucher-redeemed',
+            _('Voucher redeemed'),
+            lambda qs: qs.filter(
+                voucher__isnull=False,
+                voucher__redeemed__gte=F('voucher__max_usages'),
+            )
+        ),
+        (
+            'voucher-expired',
+            _('Voucher expired'),
+            lambda qs: qs.filter(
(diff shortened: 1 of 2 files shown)


src/pretix/base/exporters/waitinglist.py
```diff
@@ -0,0 +1,155 @@
+
```
raphaelm, line 1:
```suggestion
```

src/pretix/base/exporters/waitinglist.py
```diff
@@ -0,0 +1,155 @@
...
+    status_filters = [
+        (
+            '',
+            _('All entries'),
+            lambda qs: qs
+        ),
+        (
+            'awaiting-voucher',
+            _('Waiting for a voucher'),
+            lambda qs: qs.filter(voucher__isnull=True)
```
raphaelm, line 32: Fine for now, but as a suggestion for future similar things, you could use [Q objects](https://docs.djangoproject.com/en/3.1/ref/models/querysets/#django.db.models.Q) instead of lambdas, e.g. ``Q(voucher__isnull=True)``

src/pretix/base/exporters/waitinglist.py
```diff
@@ -0,0 +1,155 @@
...
+                voucher__valid_until__isnull=False,
+                voucher__valid_until__lte=now()
+            )
+        ),
+    ]
+
+    def iterate_list(self, form_data):
+        # create dicts for easier access by key, which is passed by form_data[status]
+        status_labels = dict([(k, v) for (k, v, c) in self.status_filters])
+        queryset_mutators = dict([(k, c) for (k, v, c) in self.status_filters])
```
raphaelm, line 70:
```suggestion
        status_labels = {k: v for k, v, c in self.status_filters}
        queryset_mutators = {k: c for k, v, c in self.status_filters}
```

src/pretix/base/exporters/waitinglist.py
```diff
@@ -0,0 +1,155 @@
...
+
+        entries = WaitingListEntry.objects.filter(
+            event__in=self.events,
+        ).select_related('item', 'variation', 'voucher').prefetch_related(
```
raphaelm, line 74:
```suggestion
        ).select_related('item', 'variation', 'voucher', 'subevent').prefetch_related(
```

src/pretix/base/exporters/waitinglist.py
```diff
@@ -0,0 +1,155 @@
...
+            else:
+                status_label = status_labels['awaiting-voucher']
+            row = [
+                entry.created.isoformat(),
+                entry.email,
+                str(entry.item) if entry.item else "",
+                str(entry.variation) if entry.variation else "",
+                entry.event.slug,
+                entry.event.name,
+                str(entry.subevent) if entry.subevent else "",
```
raphaelm, line 117: This will give a value of the form ``Subevent name – December 1st, 2020``. Maybe it's nicer for further processing to split it into two columns, one with the name and one with the date? I think we did so in some other exports already (order data?).

raphaelm (commented):
Overall looks really good, just some very minor remarks.

src/pretix/base/exporters/waitinglist.py
wiffbi: In this case, as it always applies .filter() on the queryset, Q objects are an option. But I originally considered using .exclude() on 'voucher-assigned' (don’t know why it felt semantically better to me). Furthermore using the queryset and lambdas makes it possible to add further manipulations to the queryset beyond just a filter – e.g. sort differently, group, etc. Currently not needed, but I think it is more flexible this way.

src/pretix/base/exporters/waitinglist.py
wiffbi: I will split it.

src/pretix/base/exporters/waitinglist.py
raphaelm: Understood :) ``exclude()`` is still possible with Q objects (``~Q(excluded…)``), but ``annotate`` e.g. would not be, correct. Let's keep it like this for now :)

raphaelm (approved):

PR #1867, opened 2020-12-02
Add option to limit events to specific sales channels

pc-coholic:
c.f.: pretix-resellers/!7

Not sure this is the best way to go about this with regards to limiting access to the events on the `web`-channel - let me know what you think...


src/pretix/api/serializers/event.py
```diff
@@ -124,7 +124,8 @@ class Meta:
         fields = ('name', 'slug', 'live', 'testmode', 'currency', 'date_from',
                   'date_to', 'date_admission', 'is_public', 'presale_start',
                   'presale_end', 'location', 'geo_lat', 'geo_lon', 'has_subevents', 'meta_data', 'seating_plan',
-                  'plugins', 'seat_category_mapping', 'timezone', 'item_meta_properties', 'valid_keys')
+                  'plugins', 'seat_category_mapping', 'timezone', 'item_meta_properties', 'valid_keys',
+                  'sales_channels')
```
raphaelm, line 128: Can you adjust the tests and docs accordingly? :)

src/pretix/base/models/event.py
```diff
@@ -409,7 +410,12 @@ class Event(EventMixin, LoggedModel):
     )
     seating_plan = models.ForeignKey('SeatingPlan', on_delete=models.PROTECT, null=True, blank=True,
                                      related_name='events')
-
+    sales_channels = MultiStringField(
+        verbose_name=_('Restrict to specific sales channels'),
+        help_text=_('Only sell tickets for this event on the following sales channels.'),
+        default=['web'],
+        blank=True,
```
raphaelm, line 417: Hmm, do we really want to allow "empty"? It adds another case for the support case "my event does not show up"…

src/pretix/presale/views/event.py
```diff
@@ -348,6 +348,9 @@ def get(self, request, *args, **kwargs):
             r._csp_ignore = True
             return r
 
+        if 'web' not in request.event.sales_channels:
+            raise Http404
```
raphaelm, line 352: * Can you raise Http404 with a meaningful message? That should make debugging a lot easier.

* Also, doesn't this break the reseller module, which also accesses this page?

* To avoid cheating, we should probably also restrict ``CartAdd``

src/pretix/presale/views/event.py
```diff
@@ -348,6 +348,9 @@ def get(self, request, *args, **kwargs):
             r._csp_ignore = True
             return r
 
+        if 'web' not in request.event.sales_channels:
```
raphaelm, line 351: Maybe better use ``request.sales_channel.identifier`` in case a middleware modifies that?

src/pretix/presale/views/organizer.py
```diff
@@ -101,6 +101,7 @@ class EventListMixin:
     def _get_event_queryset(self):
         query = Q(is_public=True) & Q(live=True)
         qs = self.request.organizer.events.using(settings.DATABASE_REPLICA).filter(query)
+        qs = qs.filter(sales_channels__contains='web')
```
raphaelm, line 104: Same here, better take it from the request? (if that is set on organizer level)

raphaelm (commented):
Besides the minor points in the in-line comments, I believe the widget API view also needs similar adjustments.

raphaelm (approved):

raphaelm (approved):

PR #1875, opened 2020-12-14
Add region setting to supplement localization

raphaelm:


pc-coholic (commented):
LGTM, but I feel like the decimal symbol in monetary values is not yet localized according to the set region: An english ticketshop with region Germany will still produce "€23.42" instead of "€23,42".

raphaelm:
Yep, for monetary values our template tag currently hardcodes that only the language is used, not the region. This apparently was intentional, even though I don't remember why, so I didn't want to touch it. Also, there was no consensus in the team if 300 AUD with `en_AU` should be `300 A$` or `300 $`, so I think we'll keep it as-is until some actual users care for it to be changed?

PR #1880, opened 2020-12-16
Added combined radiob uttons for <name>_asked and <name>_required fields

wiffbi:
This is a UI/UX variant for the event settings. For each combination of _asked and _required fields, one ChoiceField is created and set. The more I look at it, I am torn and not so sure, whether this is the better UX – for me, the settings are easier to grasp with the table of checkboxes. But setting it to „Do not ask“ is more clear and understandable with the <select> than with unticking both boxes. Also one select is easier in the front-end as no logic is needed to keep the checkboxes in sync.

Currently the combined ChoiceField uses the label and help_text of its _asked-field. I am not sure, whether this is the best solution.

### DIFF
--- a/src/pretix/control/forms/event.py
+++ b/src/pretix/control/forms/event.py
@@ -501,16 +501,15 @@ class EventSettingsForm(SettingsForm):
         data = super().clean()
         settings_dict = self.event.settings.freeze()
         settings_dict.update(data)
-        validate_event_settings(self.event, data)
 
         # set all dependants of virtual_keys and
         # delete all virtual_fields to prevent them from being saved
         for virtual_key in self.virtual_keys:
             if virtual_key not in data:
                 continue
-            base_key = virtual_key[:-15]
-            asked_key = base_key + "_asked"
-            required_key = base_key + "_required"
+            base_key = virtual_key.rsplit('_', 2)[0]
+            asked_key = base_key + '_asked'
+            required_key = base_key + '_required'
 
             if data[virtual_key] == 'optional':
                 data[asked_key] = True
@@ -527,9 +526,7 @@ class EventSettingsForm(SettingsForm):
             # hierarkey.forms cannot handle non-existent keys in cleaned_data => do not delete, but set to None
             data[virtual_key] = None
 
-        # TODO: do we need to re-validate?
-        # if len(self.virtual_keys):
-        #     validate_event_settings(self.event, data)
+        validate_event_settings(self.event, data)
         return data
 
     def __init__(self, *args, **kwargs):
@@ -558,10 +555,11 @@ class EventSettingsForm(SettingsForm):
 
         # create "virtual" fields for better UX when editing <name>_asked and <name>_required fields
         self.virtual_keys = []
-        for asked_key in [key for key in self.fields.keys() if key.endswith('_asked') and key[:-5] + 'required' in self.fields]:
-            required_key = asked_key[:-5] + 'required'
+        for asked_key in [key for key in self.fields.keys() if key.endswith('_asked')]:
+            required_key = asked_key.rsplit('_', 1)[0] + '_required'
             virtual_key = asked_key + '_required'
-            if virtual_key in self.fields:
+            if required_key not in self.fields or virtual_key in self.fields:
+                # either no matching required key or
                 # there already is a field with virtual_key defined manually, so do not overwrite
                 continue
 
@@ -582,9 +580,11 @@ class EventSettingsForm(SettingsForm):
             self.virtual_keys.append(virtual_key)
 
             if self.initial[required_key]:
-                self.initial[virtual_key] = ['required']
+                self.initial[virtual_key] = 'required'
             elif self.initial[asked_key]:
-                self.initial[virtual_key] = ['optional']
+                self.initial[virtual_key] = 'optional'
+            else:
+                self.initial[virtual_key] = 'do_not_ask'
 
 
 class CancelSettingsForm(SettingsForm):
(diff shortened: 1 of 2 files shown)


wiffbi:
I think, instead of a <select> using three radio buttons (ideally in one row) would be the best of both worlds: 1. you get the overview like with the checkboxes, 2. you make the option „Do not ask“ explicit (other than with the checkboxes) and 3. it is only one click (one less than with <select> as well as when making it required with the checkboxes).

If you agree, I will look into making it 3 radio buttons.

src/pretix/control/forms/event.py
```diff
@@ -502,6 +502,34 @@ def clean(self):
...
+            elif data[virtual_key] == 'do_not_ask':
+                data[asked_key] = False
+                data[required_key] = False
+
+            # hierarkey.forms cannot handle non-existent keys in cleaned_data => do not delete, but set to None
+            data[virtual_key] = None
+
+        # TODO: do we need to re-validate?
+        # if len(self.virtual_keys):
+        #     validate_event_settings(self.event, data)
```
raphaelm, line 532: Since validate_event_settings can't know anything about "virtual" keys, I think we should **only** validate at this point.

src/pretix/control/forms/event.py
```diff
@@ -528,6 +556,36 @@ def __init__(self, *args, **kwargs):
             (a, {"title": a, "data": v}) for a, v in get_fonts().items()
         ]
 
+        # create "virtual" fields for better UX when editing <name>_asked and <name>_required fields
+        self.virtual_keys = []
+        for asked_key in [key for key in self.fields.keys() if key.endswith('_asked') and key[:-5] + 'required' in self.fields]:
+            required_key = asked_key[:-5] + 'required'
```
raphaelm, line 562: This is probably a matter of taste, but I'd find ``asked_key.rsplit('_', 1)[0]`` way more readable than ``asked_key[:-5]``. It's longer, but it expresses my *intent* and I don't need to think about why there is a 5 and count characters. What do you think?

src/pretix/control/forms/event.py
```diff
@@ -528,6 +556,36 @@ def __init__(self, *args, **kwargs):
...
+                    # default key needs a value other than '' because with '' it would also overwrite even if combi-field is not transmitted
+                    ('do_not_ask', _('Do not ask')),
+                    ('optional', _('Ask, but do not require input')),
+                    ('required', _('Ask and require input'))
+                ]
+            )
+            self.virtual_keys.append(virtual_key)
+
+            if self.initial[required_key]:
+                self.initial[virtual_key] = ['required']
```
raphaelm, line 585: I haven't tested it, but it seems weird to me that this is a list and not just ``'required'``, have you checked if the string on its own also works?

src/pretix/control/forms/event.py
```diff
@@ -502,6 +502,34 @@ def clean(self):
         settings_dict = self.event.settings.freeze()
         settings_dict.update(data)
         validate_event_settings(self.event, data)
+
+        # set all dependants of virtual_keys and
+        # delete all virtual_fields to prevent them from being saved
+        for virtual_key in self.virtual_keys:
+            if virtual_key not in data:
+                continue
+            base_key = virtual_key[:-15]
```
raphaelm, line 511: see below on my opinion this ;)

raphaelm:
I'm not sure why, but this seems to break all tests for this page. Probably ``extract_form_fields`` can't deal with the hidden fields properly?

src/pretix/control/forms/event.py
wiffbi: virtual_keys is just a list of the keys to access the fields, which are _asked_required and mapped to the _asked and _required. So when the form is submitted, before I loop over all fields listed in virtual_keys, I wanted to have them validated? Probably is not needed as I explicitly check for the allowed values do_not_ask|optional|required, so I will call validate_event_settings after having applied the changes from the virtual_keys-fields.

src/pretix/control/forms/event.py
wiffbi: I wasn’t to happy about that magic number either, so totally for removing it! I thought about adding `suffix = '_asked'` and use `asked_key[:-len(suffix)].` but I wasn’t convinced. Your suggestion totally matches the intent of „divide string in two parts at the last _ and use the first part". It might be me, but I had to think a second about the second paramter to resplit. What about `asked_key[:asked_key.rindex('_')]` ? Either way is fine.

src/pretix/control/forms/event.py
wiffbi: Hm, must have been late. It did not seem to work with just a String – against what I expected, so I changed to a list. But I just checked again and it works with string as well. Sorry!

src/pretix/control/forms/event.py
wiffbi: I will update

wiffbi:
> I'm not sure why, but this seems to break all tests for this page. Probably `extract_form_fields` can't deal with the hidden fields properly?

I will look into this.

src/pretix/control/forms/event.py
raphaelm: You're probably missing a ``-`` in your approach, but other than that, I don't really care. Your choice :=

src/pretix/control/forms/event.py
raphaelm: no worries :)

src/pretix/control/forms/event.py
raphaelm: `validate_event_settings` is a pretix-custom function that only exists so we don't need to write the same logic here and in the clean() method of the API. That function only handles edge cases, and will not know how to validate a `_asked_required` key since these only exist in the context of this form, so yeah, validating after sounds better.

raphaelm (approved):

PR #1881, opened 2020-12-17
Fix #1879 -- Do not add a tab’s hash/id to location.hash if it is inside a .panel

### LINKED ISSUE #1879: Email preview broken after reload
Just a small frontend bug, which I found as follows:

1. Go to an event's email settings
2. Edit one of them
3. Check the preview to see how it looks
4. Satisfied and happy, click "save" (Page gets loaded again)
5. Click back to the email text you edited: It's still open on the preview tab, and the preview tab is completely empty

Nothing big, but I first thought that pretix had somehow discarded the whole email text, so maybe worth fixing.

*Browser:* Firefox 85 (current).

wiffbi:


src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -614,7 +614,10 @@ $(function () {
         $('.nav-tabs a[href="#' + url.split('#')[1] + '"]').tab('show');
     }
     $('a[data-toggle="tab"]').on('click', function (e) {
-        window.location.hash = e.target.hash;
+        if (!$(this).closest(".panel").length) {
```
raphaelm, line 617: Do we really want to check for ``.panel`` or rather for ``.tab-content``? The location.hash thing is bound to cause problems whenever tabs are nested, regardless of whether there are panels involved.

wiffbi: Yes, that makes total sense!

raphaelm, line 617:
```suggestion
        if (!$(this).closest(".tab-content").length) {
```

PR #1889, opened 2020-12-22
Show "live issues" when event is already live

MaicoTimmerman:
Using the `event_live_issues` we can prevent an event from going live with all kinds of checks. However once it is set live, these issues are no longer showing. When changes are made to settings of the event, new live issues could arise. I'd like to show these live issues on the live page, so users can see them:
scrot-2020-12-23T00:44:26+01:00

In our plugin we have added a custom `event_dashboard_widgets` based on the `shop_state_widget`, I've not included the widget in this MR, as I wasn't sure on whether you'd want to have it for all events. Let me know if you'd like for me to add the widget as well:
scrot-2020-12-23T00:35:11+01:00

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/event/live.html
+++ b/src/pretix/control/templates/pretixcontrol/event/live.html
@@ -12,6 +12,18 @@
                 <p>
                     {% trans "Your shop is currently live. If you take it down, it will only be visible to you and your team." %}
                 </p>
+                {% if issues|length > 0 %}
+                    <div class="alert alert-warning">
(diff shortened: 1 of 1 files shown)


raphaelm (approved):
This looks good to me.

For the widget, I'm less sure, I fear it's confusing for first time users with too many "status" things on the dashboard. @wiffbi Do you have an opinion here?

raphaelm:
Let's go with this for now and discuss the other, slightly unrelated, issue separately

PR #1893, opened 2020-12-23
Add convenience redirect …/event/(org)/ => …/organizer/(org)/

luto:
Calling https://pretix.eu/control/event/uberspace/ (obtained by stripping the event name off) currently yields a 404. This implementes a redirect to https://pretix.eu/control/organizer/uberspace/ instead.

### DIFF
--- a/src/pretix/control/urls.py
+++ b/src/pretix/control/urls.py
@@ -1,4 +1,5 @@
 from django.conf.urls import include, url
+from django.views.generic.base import RedirectView
 
 from pretix.control.views import (
     auth, checkin, dashboards, event, geo, global_settings, item, main, oauth,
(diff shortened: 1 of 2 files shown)


src/pretix/control/urls.py
```diff
@@ -303,4 +304,5 @@
         url(r'^checkinlists/(?P<list>\d+)/delete$', checkin.CheckinListDelete.as_view(),
             name='event.orders.checkinlists.delete'),
     ])),
+    url(r'^event/(?P<organizer>[^/]+)', RedirectView.as_view(pattern_name='organizer')),
```
rixx, line 307: This doesn't seem like optimal behaviour here: regexes can be notoriously long-running, and additionally `(?P<organizer>[^/]+)` catches more than you'd like! What you are probably looking for is this:

```suggestion
    path('event/uberspace', RedirectView.as_view('/control/organizer/uberspace/')),
```

rixx (commented):
Optimization notes

src/pretix/control/urls.py
rixx: (SCNR, happy holidays everybody 🎅)

PR #1898, opened 2021-01-05
Add option to export multiple choice answers in orderlists grouped

bhaettasch:
Currently, the export of orderlines creates one column per possible answer of a mulitple choice question containing Yes/No.

This pull request introduces a new optional checkbox field to choose grouping these answers in a single column of the order line instead. It will use the question title as column header (instead of question title followed by the answer like it is done in the existing export) and all selected answers for each row as comma-separated list (each comma is followed by a blank as this export format is mainly intended for manual parsing/reading).

The feature is introduced as code change instead of a separate plugin to prevent large amounts of code duplications.

Additionally, one API test was slightly adjusted to support the new option.

### DIFF
--- a/src/pretix/base/exporters/orderlist.py
+++ b/src/pretix/base/exporters/orderlist.py
@@ -59,6 +59,12 @@ class OrderListExporter(MultiSheetListExporter):
                      initial=False,
                      required=False
                  )),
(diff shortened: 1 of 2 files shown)


raphaelm:
Thank you!

PR #1909, opened 2021-01-17
Add redis sentinel support

### LINKED ISSUE #1905: Support redis sentinel
Redis Sentinel provides high availability for Redis \[[redis docs](https://redis.io/topics/sentinel)\].
django-sentinel does not yet support redis sentinel \[[open issue](https://github.com/jazzband/django-redis/issues/426)\]
Therefore, afaik pretix does not support redis sentinel.

It would be great if this support could be added.
Maybe https://pypi.org/project/django-sentinel/ helps with that.

neumantm:
Both for caching and for celery

Fixes #1905

### DIFF
--- a/doc/admin/config.rst
+++ b/doc/admin/config.rst
@@ -319,6 +319,8 @@ to speed up various operations::
     The ``sentinels`` variable should be a json serialized list of sentinels,
     each being a list with the two elements hostname and port.
     You cannot provide a password within the location when using sentinels.
+    Note that the configuration format requires you to either place the entire
+    value on one line or make sure all values are indented by at least one space.
 
 ``password``
     If your redis setup doesn't require a password or you already specified it in the location you can omit this option.
--- a/doc/admin/config.rst
+++ b/doc/admin/config.rst
@@ -297,6 +297,12 @@ to speed up various operations::
     [redis]
     location=redis://127.0.0.1:6379/1
     sessions=false
+    sentinels=[
+            ["sentinel_host_1", 26379],
+            ["sentinel_host_2", 26379],
+            ["sentinel_host_3", 26379]
+        ]
+    password=password
 
 ``location``
     The location of redis, as a URL of the form ``redis://[:password]@localhost:6379/0``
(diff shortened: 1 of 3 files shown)


neumantm:
As discussed in #1905 I understand if you don't want to merge this because of the additional dependency.
In that case you could maybe leave this open until https://github.com/jazzband/django-redis/pull/460 is merged.
Then I (or someone else) can rework the PR so it works without the additional dependency.

raphaelm:
@cla-bot recheck

raphaelm:
Hi, thanks for updating the PR with docs! I think we should add this, sorry for the slow responses. However, I see the django-redis PR you mentioned has since been merged, do you think we should move to the no-new-dependency approach directly?

neumantm:
I don't think the code style / isort failure is caused by this PR as it doesn't even touch `src/pretix/presale/views/cart.py`

raphaelm:
Yeah, I think that's unrelated. I'll fix it on master.

@cla-bot check

neumantm:
> Hi, thanks for updating the PR with docs! I think we should add this, sorry for the slow responses. However, I see the django-redis PR you mentioned has since been merged, do you think we should move to the no-new-dependency approach directly?

Yes, you're right. Didn't notice until now.
Will check and make changes accordingly.

neumantm:
There is no release containing that PR yet, so we can't just use the requirements.txt
Therefore I'd suggest waiting for a release.
I couldn't find any information on their release schedule though.

Whether you want to keep this PR open until then or merge it now and have a new PR when there is a release is your decision.
I'm not really in a hurry. I'm currently using a Docker Image based on my branch anyway.

raphaelm:
Let's wait for that release, as we're not in a hurry here :)

For future reference, this is the commit we're waiting for:
https://github.com/jazzband/django-redis/commit/e2f6bcc9db1e4710ede6ad74fd91c0dc74884839

neumantm:
I've now changed the PR to use the new version django-redis instead of a separate library.
I've also successfully deployed a test instance of this using a redis sentinel cluster.
From my point of view, this can now be reviewed and merged.
The remaining CI failure does not seem to be related to this PR.

doc/admin/config.rst
```diff
@@ -297,6 +297,12 @@ to speed up various operations::
     [redis]
     location=redis://127.0.0.1:6379/1
     sessions=false
+    sentinels=[
+        ["sentinel_host_1", 26379],
+        ["sentinel_host_2", 26379],
+        ["sentinel_host_3", 26379]
+    ]
```
raphaelm, line 304: Does this actually work like this? https://docs.python.org/3/library/configparser.html states:

>  Values can also span multiple lines, as long as they are indented deeper than the first line of the value

Wouldn't this fail for the closing ] in the example?

neumantm: You are absolutely right, missed that. I will fix the docs.

neumantm:
I don't think the CI failure is caused by this PR.
Is there anything still preventing this to be merged?

raphaelm (approved):

raphaelm:
Nope, sorry!

PR #1910, opened 2021-01-18
Updated docs on how to setup Google Analytics (Universal or GA4) for cross-domain tracking when using pretix-widget

wiffbi:


doc/user/events/widget.rst
```diff
@@ -304,8 +304,57 @@ Hosted or pretix Enterprise are active, you can pass the following fields:
...
+        gtag('config', 'G-XXXXXXXX', {
+            'linker': {
+                'domains': ['pretix.eu'],// if you host pretix on your own domain, replace accordingly
+                'decorate_forms': true
+            }
+        });
+    </script>
+
+  Please note, that pretix-widget currently does not fully support cross-domain tracking with GA4 yet. The following code
+  is future proof and as soon as pretix will support cross-domain tracking with GA4, it will just work. If you use GA4 and
```
raphaelm, line 326: Is it, though? This would not yet be compatible with our idea of allowing the embedding website to pass a ``decorate_url`` function to the widget

wiffbi: The embedding website does not pass a `decorate_url` function to the widget. Our idea was to define a `decorate_url` inside widget.js that is called whenever widget.js opens a URL. If GA4 is used on the embedding website, then that `decorate_url` in widget.js notices and runs the URL through the GA4-linker/decorator automatically – so no need to manually add a `decorate_url`.

Using the init-properties for `linker` with `'domains'` gives us cross-domain tracking for normal links (such as „Waiting list“) for free. Using `'decorate_forms'` as a default at least gives us the freedom in the future to refactor widget.js to use „normal“ form submits and get cross-domain tracking in GA4 supported out of the box without using a `decorate_url` with the _hack_/undocumented feature described here: https://www.andrewdixon.co.uk/2019/09/22/manually-decorate-a-link-with-google-analytics-cross-domain-measure-query-parameter-gtag-js/

doc/user/events/widget.rst
```diff
@@ -304,8 +304,51 @@ Hosted or pretix Enterprise are active, you can pass the following fields:
 * If you use the campaigns plugin, you can pass a campaign ID as a value to ``data-campaign``. This way, all orders
   made through this widget will be counted towards this campaign.
 
-* If you use the tracking plugin, you can pass a Google Analytics User ID to enable cross-domain tracking. This will
-  require you to dynamically load the widget, like this::
+* If you use the tracking plugin, you can enable cross-domain tracking. To do so, you need to dynamically build
```
raphaelm, line 307: "dynamically build" sounds like you need to compile something, maybe stick with "load" or use "initialize"?

wiffbi: I guess „initialize“ is the word. „load“ for me somehow made it unclear, whether I still need the html-code from the top of the page. What about „you need to initialize the pretix-widget manually“?

raphaelm: :+1:

PR #1918, opened 2021-01-30
Subevents: Bulk editor

raphaelm:
- [x] Name
- [x] Active
- [x] Location + geo coordinates
  - [x] geo handling
- [x] frontpage text
- [x] show in lists
- [x] time and date of event and presale
- [x] quota
- [x] prices
- [x] checkin lists
- [x] meta
- [x] tests
- [x] "select all that match search"

Things to consider later, but deliberately not in this PR so the complexity stays manageable.

- fields added by plugins (such as seating)
- retrofitting relative time changes ("presale 1hr before" → "presale 2hr before")

Screenshot_2021-02-10 Change multiple dates pretix eu


wiffbi:
When I bulk-edit quotas, there is an issue when adding multiple new quotas for sub-events where orders already exist and the quota is exactly the number of orders. My setup is a follows:

3 sub-events selected for edit. For these 3 sub-events there is an order with one ticket for each sub-event. I delete all quotas for these sub-events and save. Then I bulk-edit these sub-events again and add multiple quotas: one for tickets with „Total capacity“ of 1 (so each sub-event is basically booked out) and another quota for another product (in my case an add-on, but does not matter). If I try and save, I get this error:

`UNIQUE constraint failed: pretixbase_quota_items.quota_id, pretixbase_quota_items.item_id`

When I change the total capacity of the tickets-quota to 2, then adding multiple quotas at once works.

Adding quotas step by step always works: add one quota for tickets with total capacity of 1 („sweet spot“ for booked out), then save, then bulk-edit again and add another quota for another product, then it works. So it must have something to do with adding multiple quotas at once and hitting the „sweet spot“ (a.k.a. booked out) with total capacity for already booked products.

raphaelm:
I think the relation with the quota size was coincidential, the error was triggered whenever you added at least two quotas that included the same product. I pushed a fix, can you verify as well?

wiffbi:
In my tests, I have created the quotas for the different products. But that works now too.

raphaelm:
> In my tests, I have created the quotas for the different products. But that works now too.

yep you're right. it was triggered in any case of adding multiple quotas. my bad!

raphaelm:
Okay, I think this is done. @pc-coholic so you want to also have a quick look of you can think of any problems? otherwise I think we're good to go. :)

wiffbi:
Probably a non-issue as this control is hidden visually, but shouldn’t https://github.com/pretix/pretix/blob/master/src/pretix/control/templates/pretixcontrol/subevents/index.html#L91-L100 be inside an if-statement `{% if "can_change_event_settings" in request.eventpermset %}`?

raphaelm:
I don't quite see what you refer to, line 90 does have such a check?

src/pretix/control/templates/pretixcontrol/subevents/index.html
```diff
@@ -67,21 +73,31 @@ <h1>{% trans "Dates" context "subevent" %}</h1>
...
+                    <tr class="table-select-all warning hidden">
+                        <td>
+                            <input type="checkbox" name="__ALL" id="__all">
+                        </td>
+                        <td colspan="5">
+                            <label for="__all">
+                                {% trans "Select all results on other pages as well" %}
+                            </label>
+                        </td>
+                    </tr>
```
wiffbi, line 100: Sorry, I linked to the file in master, but meant this. I know it is hidden visually, but shouldn’t it be hidden from HTML-source as well, `{% if "can_change_event_settings" in request.eventpermset %}`

raphaelm: yup makes sense

PR #1919, opened 2021-02-01
PDF layout: Allow to show photos from questions

raphaelm:
also needs support in
- [x] pretixpos
- [x] libpretixsync scanning online
- [x] libpretixsync scanning offline
- [x] pretixscan android
  - [x] take
  - [x] print
- [ ] pretixscan desktop
  - [ ] take
  - [ ] print
- [ ] pretixscan proxy support?
- [x] pretixprint

### DIFF
--- a/doc/api/resources/orders.rst
+++ b/doc/api/resources/orders.rst
@@ -220,7 +220,7 @@ downloads                             list of objects            List of ticket 
 └ url                                 string                     Download URL
 answers                               list of objects            Answers to user-defined questions
 ├ question                            integer                    Internal ID of the answered question
-├ answer                              string                     Text representation of the answer
+├ answer                              string                     Text representation of the answer (URL if answer is a file)
 ├ question_identifier                 string                     The question's ``identifier`` field
 ├ options                             list of integers           Internal IDs of selected option(s)s (only for choice types)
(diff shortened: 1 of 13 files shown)


pc-coholic (commented):
LGTM - but didn't test yet against apps.

Tiny room for improvement:
- Image-Preview in PDF-editor shows 'question_{identifier}' as a placeholder. I would love to have the actual question name there - perhaps output the internal ID in addition?
- I am aware that there is a very specific use case for this feature and that our apps will take care of the following... But there will be people asking for the pictures already during signup - and the upload-form will just accept any file. Not looking for a full-blown picture editor - but perhaps an advanced option to limit the upload to image files could be an idea that doesn't involve having a dedicated "image-upload"-question type.

raphaelm:
>Image-Preview in PDF-editor shows 'question_{identifier}' as a placeholder. I would love to have the actual question name there - perhaps output the internal ID in addition?

Yeah :( I'll leave that as an exercise for later, I think :D

> I am aware that there is a very specific use case for this feature and that our apps will take care of the following... But there will be people asking for the pictures already during signup - and the upload-form will just accept any file. Not looking for a full-blown picture editor - but perhaps an advanced option to limit the upload to image files could be an idea that doesn't involve having a dedicated "image-upload"-question type.

Yep, I already have that on my mind as well! But I think it warrants its own PR.

PR #1922, opened 2021-02-05
Add more features to custom meta properties

raphaelm:
- allow to restrict to org-admins
- allow to restrict allowed values
- allow to set properties as required to release event


pc-coholic (commented):
- Converting an existing property to a list-selectable one will:
  - Still output the old value on the API
  - Not cause a `live_issue`
  - Be cleared if saving the settings form without setting a new, approved value
- Same as before, but setting the default-value to one that is not on the approved list:
  - Will display the default value in the UI
  - Will not output this default value on the API
  - Actually: Setting a default value will not output anything on the API unless it is != the default value. But I guess that is by design and reflects existing behaviour.
- Manually editing the DropDown to include an arbitrary value will result in a validation error with corresponding "look below"-message. However, the expected `'Meta data property \'{name}\' does not allow value \'{value}\'` is not displayed inline.

raphaelm:
>Converting an existing property to a list-selectable one will:
> Still output the old value on the API
> Not cause a live_issue
> Be cleared if saving the settings form without setting a new, approved value

I think this is acceptable.

> Same as before, but setting the default-value to one that is not on the approved list:
> Will display the default value in the UI

This will be fixed.

> Will not output this default value on the API
> Actually: Setting a default value will not output anything on the API unless it is != the default value. But I guess that is by design and reflects existing behaviour.

This is intentional

> Manually editing the DropDown to include an arbitrary value will result in a validation error with corresponding "look below"-message. However, the expected 'Meta data property \'{name}\' does not allow value \'{value}\' is not displayed inline.

Will be fixed

PR #1923, opened 2021-02-11
[a11y] Add text "required" to label of required inputs

wiffbi:
As the * is added via CSS-content, which is not visible to screen readers, and also is mainly a visual indicator, that when read aloud is distracting, I added the text „required“ to the label of required inputs in presale-checkout.

See here for reasoning: https://www.a11y-101.com/development/required

Copied and adapted from src/pretix/control/forms/renderers.py

### DIFF
--- a/src/pretix/presale/forms/renderers.py
+++ b/src/pretix/presale/forms/renderers.py
@@ -1,5 +1,38 @@
 from bootstrap3.renderers import FieldRenderer
+from bootstrap3.text import text_value
 from bootstrap3.utils import add_css_class
+from django.forms import CheckboxInput
+from django.forms.utils import flatatt
+from django.utils.html import format_html
+from django.utils.safestring import mark_safe
+from django.utils.translation import pgettext
+
+
+def render_label(content, label_for=None, label_class=None, label_title='', optional=False):
+    """
+    Render a label with content
+    """
+    attrs = {}
+    if label_for:
+        attrs['for'] = label_for
+    if label_class:
+        attrs['class'] = label_class
+    if label_title:
+        attrs['title'] = label_title
+
+    if text_value(content) == '&#160;':
+        # Empty label, e.g. checkbox
+        attrs.setdefault('class', '')
+        attrs['class'] += ' label-empty'
+
+    builder = '<{tag}{attrs}>{content}{opt}</{tag}>'
+    return format_html(
+        builder,
+        tag='label',
+        attrs=mark_safe(flatatt(attrs)) if attrs else '',
+        opt=mark_safe(' <i class="sr-only">{}</i>'.format(pgettext('form', 'required'))) if not optional else '',
+        content=text_value(content),
+    )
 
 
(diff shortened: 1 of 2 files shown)


src/pretix/presale/forms/renderers.py
```diff
@@ -1,5 +1,38 @@
...
+        # Empty label, e.g. checkbox
+        attrs.setdefault('class', '')
+        attrs['class'] += ' label-empty'
+
+    builder = '<{tag}{attrs}>{content}{opt}</{tag}>'
+    return format_html(
+        builder,
+        tag='label',
+        attrs=mark_safe(flatatt(attrs)) if attrs else '',
+        opt=mark_safe('<i class="sr-only"> {}</i>'.format(pgettext('form', 'required'))) if not optional else '',
```
raphaelm, line 33: Would a screen reader read "ZIP code (required)" differently than "ZIP code required"? When reading, it makes a difference and the () would be more appropriate – but when listening it's probably to fast to hear an actual difference even if the screen reader would put in a longer break?

src/pretix/presale/forms/renderers.py
```diff
@@ -1,5 +1,38 @@
...
+        attrs['for'] = label_for
+    if label_class:
+        attrs['class'] = label_class
+    if label_title:
+        attrs['title'] = label_title
+
+    if text_value(content) == '&#160;':
+        # Empty label, e.g. checkbox
+        attrs.setdefault('class', '')
+        attrs['class'] += ' label-empty'
```
raphaelm, line 26: The checkbox case is probably one we need to handle differently, a label of just "required" probably doesn't make a lot of sense?

src/pretix/presale/forms/renderers.py
wiffbi: AFAIK the „()“ will be read out as „parenthesis“, which would sound rather odd. That’s why it uses an i-element, which should change the voice, so it is clear, that „required“ is an instruction and not part of the label.

raphaelm: Perfect, disrecard my comment then

src/pretix/presale/forms/renderers.py
wiffbi: No, but an empty label is wrong anyways. Will look into this. Do you have any examples where I can find such a checkbox with an empty label?

raphaelm: The checkbox will have another, non empty label.

You can probably construct this in the frontend by just adding a question of type Yes/No.

wiffbi: Ah, I see – I had a Yes/no-Question, but it was not required. I will look into this.

wiffbi: Fixed. I added "required" to the  checkbox-label (when inside `put_inside_label`) as well as removed it from the overall-label for checkboxes, when it is empty (which it usually is)

PR #1931, opened 2021-02-18
add bulk selection by click and drag over table rows

wiffbi:
The implementation is currently a bit fiddly as I assigned the trigger to start the drag-selection on the td:first-child, which contains the checkbox and is not that much bigger than the checkbox itself. So basically one has to click the checkbox and start dragging up or down. Once you drag, you can move freely as long as you stay over the table it selects the next records/rows. We probably should extend the hit-area or at least make the td:first-child change the cursor to something meaningful/different – maybe `cursor: cell;`. Or even better put a `<label>` around each checkbox, which brings the opportunity for aria-label as well as a bigger hit-area without JavaScript!

In Firefox, if you click a checkbox, it changes state once you release it. I did not change this behavior, which makes drag-selection a bit weird in the first moment. But IMHO not so bad, that we should change the standard-behavior of the checkbox.


raphaelm:
I like it. Unfortunately, the changes in main.js do conflict with the mreged changes from the bulk-edit branch more than expected. Sorry about that, could you look at this?

wiffbi:
> I like it. Unfortunately, the changes in main.js do conflict with the mreged changes from the bulk-edit branch more than expected. Sorry about that, could you look at this?

Fixed. The .warning class to highlight selected rows interferes with the .table-select-all row, so I removed .warning for selected rows. Result is actually not that bad – as in not that visually cluttered.

PR #1985, opened 2021-03-01
Web-based check-in interface

raphaelm:
- [x] list selection
- [x] entry vs exit
- [x] search
  - [ ] search by question value
  - [x] filter by custom logic
- [x] scan without search
- [x] perform checkin
- [x] implement unpaid dialog
- [x] implement questions dialog
- [x] require attention
- [ ] ~play sounds~
- [x] show stats / "currently inside"
- [x] Add new permission
- [x] link to interface in UI


pc-coholic (changes requested):
- [ ] File Upload questions display only a String Input field and cannot be skipped.
- [ ] Search Input Field does not take focus on idle. Might be useful, since people will keep changing windows.
- [ ] Search input will not clear after pressing enter. Scanning multiple tickets in quick succession will only append them one after another, not checking in the subsequent ones. Perhaps displaying the scanned/confirmed (== pressed Enter) value separately?
- [ ] There is no way to return to the CheckIn overview-page, unless waiting for the timeout, actively emptying the searchbar or reloading the page. Add a button to clear the input and go home?
- [ ] I would have expected the screen to be navigable through key inputs... Typing anything -> Searchbar (or input modal if open). Up/Down to highlight a result. Enter to check it in/Submit the typed data.

raphaelm:
> File Upload questions display only a String Input field and cannot be skipped.

Unsupported on purpose for now, but I can make if ail more loudly.

> Search Input Field does not take focus on idle. Might be useful, since people will keep changing windows.

Have you noticed it will auto-focus as soon as you perform a keystroke anywhere in the tab? :) I hope that's good enough.

> Search input will not clear after pressing enter. Scanning multiple tickets in quick succession will only append them one after another, not checking in the subsequent ones. Perhaps displaying the scanned/confirmed (== pressed Enter) value separately?

Clearing has the issue that you can't modify your search query to refine it, but you need to start typing from scratch. If you start typing somewhere without any focus ("scan the next code") it should override the current content of the field-

> There is no way to return to the CheckIn overview-page, unless waiting for the timeout, actively emptying the searchbar or reloading the page. Add a button to clear the input and go home?

Good point.

>  I would have expected the screen to be navigable through key inputs... Typing anything -> Searchbar (or input modal if open). Up/Down to highlight a result. Enter to check it in/Submit the typed data.

Most of that should already work, I'll look more closely again.

pc-coholic:
> > Search Input Field does not take focus on idle. Might be useful, since people will keep changing windows.
> 
> Have you noticed it will auto-focus as soon as you perform a keystroke anywhere in the tab? :) I hope that's good enough.

I did indeed not...

> > Search input will not clear after pressing enter. Scanning multiple tickets in quick succession will only append them one after another, not checking in the subsequent ones. Perhaps displaying the scanned/confirmed (== pressed Enter) value separately?
> 
> Clearing has the issue that you can't modify your search query to refine it, but you need to start typing from scratch. If you start typing somewhere without any focus ("scan the next code") it should override the current content of the field-

I just tried scanning two barcodes back to back and ran into the issue of it getting appended, sorry.

> > I would have expected the screen to be navigable through key inputs... Typing anything -> Searchbar (or input modal if open). Up/Down to highlight a result. Enter to check it in/Submit the typed data.
> 
> Most of that should already work, I'll look more closely again.

It is actually working - I think my test might have gone somehow wrong, because I had the devtools open.

PR #1987, opened 2021-03-02
Waiting list: Add name and phone number

wiffbi:
Add name and phone number fields to waitinglist. Whether name and/or phone number is asked or even required can be configured through pretix-control under event settings => waitinglist.

Data-shredder is updated to shred these fields.

Add columns „name“ and „phone“ number to all waitinglist exports as well as filtered downloads.


wiffbi:
TODO: `PhoneNumberField` or to be more precise `WrappedPhoneNumberPrefixWidget` does not pick up the initial value provided in `WaitingListForm`.

doc/api/resources/waitinglist.rst
```diff
@@ -13,7 +13,9 @@ Field                                 Type                       Description
 ===================================== ========================== =======================================================
 id                                    integer                    Internal ID of the waiting list entry
 created                               datetime                   Creation date of the waiting list entry
+name                                  string                     Name of the user on the waiting list
```
raphaelm, line 16: * Probably add "or ``null``" both here and for the phone field so API users know these fields are nullable
* This is missing name_parts (see orders.rst for an example)

src/pretix/base/shredder.py
```diff
@@ -203,7 +203,7 @@ def shred_data(self):
 class WaitingListShredder(BaseDataShredder):
     verbose_name = _('Waiting list')
     identifier = 'waiting_list'
-    description = _('This will remove all email addresses from the waiting list.')
+    description = _('This will remove all names, email addresses and phone numbers from the waiting list.')
```
raphaelm, line 206:
```suggestion
    description = _('This will remove all names, email addresses, and phone numbers from the waiting list.')
```

raphaelm (approved):
You can just run ``isort -rc pretix`` to automatically sort the imports in the migration correctly.

Apart from that, this looks done and good to me!

PR #1990, opened 2021-03-02
Manual payment: Add amount placeholder

inwwin:
When I use manual payment, I would like to show texts to my customers with the pending amount rather than the total of the entire order, e.g., "Please transfer {amount} to account XXX", which was not previously possible because pretix only provided placeholder for the total of the order.

This would be useful for customers who redeem gift card for partial payment to the order, and then choose manual payment to pay for the rest of the order.

However, I have one failed test [located at this line](https://github.com/pretix/pretix/blob/master/src/tests/control/test_mail_settings_preview.py#L317) I would like you guys to take a look at, because I'm not sure why it failed.

### DIFF
--- a/src/pretix/base/payment.py
+++ b/src/pretix/base/payment.py
@@ -970,21 +970,17 @@ class ManualPayment(BasePaymentProvider):
                     label=_('Payment process description in order confirmation emails'),
                     help_text=_('This text will be included for the {payment_info} placeholder in order confirmation '
                                 'mails. It should instruct the user on how to proceed with the payment. You can use '
-                                'the placeholders {order}, {total}, {amount}, {currency}, {total_with_currency} '
-                                'and {amount_with_currency}.'),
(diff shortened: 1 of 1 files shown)


raphaelm (commented):
I agree with the goal of this PR. I worry that it will be utterly confusing to people to have both a ``total`` and ``amount`` placeholder, because no normal person could guess what the difference is.

I think there's barely no value for the ``{total}`` placeholder, I'd even consider it a bug that it existed. My proposal for a "soft deprecation" would be to remove the ``{total}`` placeholder from both the help text as well as the validator (but keep it in format_map so we don't break existing configuration silently). @pc-coholic any thoughts on this?

PS: Your test only fails locally, I assume, 'cause you never compiled the translation files (``make localecompile`` or just ``make``)

raphaelm (approved):

raphaelm:
Thank you very much!

PR #1997, opened 2021-03-16
Add force-reactivate checkbox to order

pc-coholic:
The form used is not exactly meant to be used here - but it works and has exactly the same functionality... So in order to avoid code duplication: here it is :-)

### DIFF
--- a/src/pretix/control/forms/orders.py
+++ b/src/pretix/control/forms/orders.py
@@ -105,6 +105,10 @@ class ConfirmPaymentForm(ForceQuotaConfirmationForm):
     pass
 
 
+class ReactivateOrderForm(ForceQuotaConfirmationForm):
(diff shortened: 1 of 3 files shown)


raphaelm:
This sounds like a small recipe for later problems, for example there's the long-standing idea to add a "date of payment" to the ConfirmPaymentForm. Maybe rename it to an abstract name and create two empty subclasses, so the code structure makes clear where it is used?

pc-coholic:
- Renamed the `ConfirmPaymentForm` to `ForceQuotaConfirmationForm`
- Added `class ConfirmPaymentForm(ForceQuotaConfirmationForm)`
- Added `class ReactivateOrderForm(ForceQuotaConfirmationForm)`

PR #2003, opened 2021-03-19
Auto-scheduled emails

raphaelm:


src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/update_rule.html
```diff
@@ -0,0 +1,55 @@
+{% extends 'pretixcontrol/event/base.html' %}
+{% load i18n %}
+{% load static %}
+{% load bootstrap3 %}
+{% block title %}{% trans "Update Scheduled Email" %}{% endblock %}
```
raphaelm, line 5: The headlines are also a little inconsistent, "Update Scheduled Email" vs "Delete Rule". We should be clear to the end user about the name of this feature :)

src/pretix/plugins/sendmail/views.py
```diff
@@ -20,6 +24,8 @@
 from pretix.plugins.sendmail.tasks import send_mails
 
 from . import forms
+from .models import Rule
+from ...control.views import CreateView, UpdateView
```
raphaelm, line 28: Is there any reason not to import these from django.views.generic? Probably just a confused PyCharm

raphaelm (commented):
Great! :) This is really going somewhere. I noted a few detail things inline.

To be done, this will also need tests. I don't think we absolutely need tests for the GUI part, it's simple enough and should fail in simple ways if it does, but we'll absolutely need tests for all of the background parts, since they are kinda invisible to users and if we introduce a logic error, it will only show after it already caused quite some harm.

Specifically, we'll need extensive tests for:

- [ ] Making sure ScheduledMail objects are created and updated correctly if subevents are created, deleted, or changed
- [ ] ScheduledMail.compute_time() in all possible configurations of date_is_absolute, offset_to_event_end and offset_is_after, likely with a [parametrized test](https://docs.pytest.org/en/stable/parametrize.html). We should also include tests for odd cases to make sure "3 days before at 8:00" behaves correctly for an event in a timezone far off from UTC like Asia/Tokyo, and that it does the expected thing e.g. in Europe/Berlin on the day before and after daylight saving time starts.
- [ ] ScheduledMail.send() to make sure include_pending, limit_products, and send_to behave correctly.

raphaelm:
The CI errors on the branch are probably also best resolved by

1. deleting all migrations in the PR
2. rebasing onto master
3. rerunning makemigrations (you can then apply with --fake locally to keep your local DB as-is)

src/pretix/plugins/sendmail/forms.py
```diff
@@ -172,3 +175,116 @@ def __init__(self, *args, **kwargs):
...
+        if dia == 'abs':
+            if not sd:
+                raise ValidationError(_('Please specify the send date'))
+            d['date_is_absolute'] = True
+            d['send_offset_days'] = d['send_offset_time'] = None
+        else:
+            if not (sod and sot):
+                raise ValidationError(_('Please specify the offset days and time'))
+            d['offset_is_after'] = True if dia[4] == 'a' else False
+            d['offset_to_event_end'] = True if dia[6] == 'e' else False
```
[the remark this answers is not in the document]
julia-luna: How about this?

```suggestion
            d['offset_is_after'] = '_a_' in dia
            d['offset_to_event_end'] = dia.endswith('_e')
```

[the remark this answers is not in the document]
raphaelm: Even better :D

src/pretix/plugins/sendmail/signals.py
```diff
@@ -1,9 +1,26 @@
...
 from django.urls import resolve, reverse
 from django.utils.translation import gettext_lazy as _
 
+from pretix.base.models import SubEvent
 from pretix.base.signals import logentry_display
 from pretix.control.signals import nav_event
+from pretix.plugins.sendmail.models import ScheduledMail
+
+
+@receiver(post_save, sender=SubEvent)
```
raphaelm, line 12: note to self: will need to check if this works properly with bulk-insertion of subevents, but probably will

src/pretix/plugins/sendmail/signals.py
```diff
@@ -1,9 +1,26 @@
...
+def sendmail(sender, **kwargs):
+
+    subevent = kwargs.get('instance')
+    event = subevent.event
+    rules = event.rule_set
+    qs = ScheduledMail.objects.filter(subevent=subevent)
+
+    if not qs.exists():
+        for rule in rules.all():
+            if not qs.filter(rule=rule).exists():
```
raphaelm, line 22: if we have ``n`` rules and create a new subevent, this will cause `2n+2` database queries. And I'm not even sure if the second ``if`` statement could ever be true if the first one is.

An alternative could be:

```
existing_rules = ScheduledMail.objects.filter(subevent=subevent).values_list('rule_id', flat=True)
to_create = []
for rule in event.rule_set.all():
    if rule.pk not in existing_rules:
        to_create.append(ScheduledMail(rule=rule, event=event, subevent=subevent))
ScheduledMail.objects.bulk_create(to_create)
```

This would get us down to a constant 3 queries per subevent, regardless of the number of rules.

src/pretix/plugins/sendmail/views.py
```diff
@@ -237,3 +243,142 @@ def get_context_data(self, **kwargs):
                     pass
 
         return ctx
+
+
+class CreateRule(EventPermissionRequiredMixin, CreateView):
+    template_name = 'pretixplugins/sendmail/rule_create.html'
+    permission = 'can_change_orders'
```
raphaelm, line 250: I think ``can_change_event_settings`` would be most appropriate

src/pretix/plugins/sendmail/signals.py
julia-luna: it does. just checked

src/pretix/plugins/sendmail/models.py
```diff
@@ -0,0 +1,206 @@
...
+                    }
+            return s
+
+    @property
+    def total_mails(self):
+        return len(ScheduledMail.objects.filter(rule=self))
+
+    @property
+    def sent_mails(self):
+        return len(ScheduledMail.objects.filter(rule=self, sent=True))
```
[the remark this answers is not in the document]
julia-luna: :heavy_check_mark:

raphaelm:
Hi :) Heads up when continuing work on this, I rebased against master and fixed the merge conflicts.

Also,

* I re-created the migrations to make them cleaner
* I restructured the tests a little
* I added some more tests for time edge cases and fixed minor issues with DST handling
* I resolved many of the issues noted above

As I see it, missing for a merge is:

- [ ] Move the ``sendmail_run_rules`` into a ``periodic_task`` signal listener so it gets called automatically
- [ ] Add tests for the re-computation after a subevent changes
- [ ] Add tests to make sure send() sends the correct set of emails based on
  - [ ] order vs position
  - [ ] correct subevent
  - [ ] correct products
  - [ ] correct order status

raphaelm:
ok, and somehow figure out why the migration fails on CI:

>E               django.db.utils.ProgrammingError: column "last_modified" of relation "pretixbase_subevent" already exists

src/tests/plugins/sendmail/test_rules.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+@pytest.mark.django_db
+@pytest.mark.parametrize('send_to,amount_mails,recipients', [
+    (Rule.CUSTOMERS, 1, ['dummy@dummy.test']),
+    (Rule.ATTENDEES, 1, ['meow@dummy.test']),
+    (Rule.BOTH, 2, ['dummy@dummy.test', 'meow@dummy.test']),
+])
+@scopes_disabled()
+def test_sendmail_rule_send_order_vs_pos(send_to, amount_mails, recipients, order, event, item):
```
raphaelm, line 176: maybe extend the test scope with

- Rule.ATTENDEES with an order where attendee email isn't set but only the order email
- Rule.BOTH with an order where both addresses are the same

src/tests/plugins/sendmail/test_rules.py
```diff
@@ -0,0 +1,250 @@
...
+    periodic_task.send(sender=None)
+
+    assert len(djmail.outbox) == 1
+
+    assert djmail.outbox[0].to[0] == p1.attendee_email
+
+
+@pytest.mark.django_db
+@scopes_disabled()
+def test_sendmail_rule_send_correct_order_status(event, order):
```
raphaelm, line 242: this one tests that include_pending=True works, can you also test that

* orders with status=CANCELED, status=EXPIRED or require_approval=true never get mail
* include_pending=False works?

src/tests/plugins/sendmail/test_rules.py
```diff
@@ -0,0 +1,250 @@
...
+
+    rule = event.rule_set.create(date_is_absolute=False, offset_is_after=False, send_offset_days=2, send_offset_time=datetime.time(4, 30))
+
+    se1.date_from += datetime.timedelta(days=2)
+    se1.save()
+
+    expected = dt_now.replace(hour=4, minute=30, second=0, microsecond=0)
+
+    m = ScheduledMail.objects.filter(rule=rule).first()
+    m.recompute()
```
raphaelm, line 165: instead of recomputing manually, maybe run the periodic task to check that it does so automatically?

PR #2010, opened 2021-03-24
Move quota cache from database to redis

raphaelm:
Previously, we stored a cache of quota availabilities in the database. This had a number of benefits:

- We could use these numbers in queries
- We did not need any additional calls to a cache store, since we'd just SELECT them when we select the quotas

However, it turns out this causes a problem in a peak load scenario:

Take for example a calendar page with hundreds of timeslots (and therefore hundreds of quotas):

- The first user opens the page, there are no cached values. Quotas are calculated and results saved to the database in a huge UPDATE query.
- Lots of subsequent users opens the page in the next two minutes, there are cached values in the database, no calculation required.
- Exactly two minutes after the first request, the cached values are stale. Now, if we have hundreds of people loading the page at the same time, this will cause hundreds of quota calculations (not great, but okay), but also hundreds of conflicting UPDATE queries. Each of these is trying to acquire a lock on each of the rows in the quota table. In a high-load situation of a pretix Enterprise customer yesterday, this caused very long lock times on the database which brought the whole system down.

This PR replaces the cache in the cache in the database with a cache in redis. The main benefit is that we save a lot of database locks. Since we're not using any kind of transactions on redis, competing writes will just be harmless. However to save load on redis, we're additionally introducing a "naive" locking algorithm ensuring max 1 write per 10 seconds for the same set of quotas.

This PR also gets rid of the background job to update the cache. I don't think we need it, but let's see how that plays out in production :)

This has a few subtle side effects, e.g. the quota column in event and subevent lists in the backend will now show "availability" instead of "paid orders".


pc-coholic (approved):

raphaelm:
on second thought, maybe let's try to keep the UI as-is

PR #2022, opened 2021-04-10
Fix checkin-list pdf export subevent time formatting

felixrindt:
I was getting this formatting error when exporting a checkin list as pdf for a single subevent.

image

[Blame](https://github.com/pretix/pretix/blame/master/src/pretix/plugins/checkinlists/exporters.py#L279)


raphaelm:
Lol :D

PR #2024, opened 2021-04-12
Customer accounts & Memberships

raphaelm:
This PR introduces a major new feature of memberships, plus to other major features that are a prerequisite to do that. Specifically, it adds:

* Organizer-level settings for email sender addresses and SMTP servers including a new UI on event level for settings that are set on organizer level.

2021-04-12T12:57:54+02:00

* An optional concept of **customers**, which allows ticket buyers to sign up with their email address and a password. These accounts are per-organizer, so people do not sign up for "a pretix account", but rather "a <Company> account". For now, this mainly allows them to view all of their orders in one place,  but in the future (outside the scope of this PR) it could allow to store invoice addresses or attendee profiles to make registrations input easier.

* A concept of **memberships**. Any product can grant a membership, which will be saved to the customer account and – if logged in – can later be bought to buy a member-only product in the same or a different event. This can be used to build a all-year pass that grants access to all events for members, or products like a "month pass" or "10-time ticket" which works together with limited time slot availability. At a later point, this feature could be extended towards subscription models usually found in the theatre world.

2021-04-12T13:00:30+02:00

TODO:

- [x] tests for customer handling
  - [x] backend
  - [x] auth
    - [x] test organizer locality
  - [x] checkout
  - [x] order change
- [x] tests for membership handling
  - [x] backend
  - [x] base
    - [x] grant membership
    - [x] require membership
  - [x] presale
    - [x] grant membership
    - [x] require membership
  - [x] order change
- [x] REST API
  - [x] Customers (writable)
  - [x] Membership types
  - [x] Memberships
  - [x] New item properties
  - [x] Order creation?
- [x] test event/item copy
- [x] test event-level custom domains
- [x] think about implications between memberships and shredder

After merge:

- [ ] Adjust coronashredder
- [ ] Adjust pretix-gdpr

Later:

- [ ] Stored invoice addresses
- [ ] Stored attendee profiles
- [ ] Integrate with POS
- [ ] allow to select first day of membership


wiffbi:
This is really great! I had a look at the features and have a few remarks:

**Account creation**

A user account asks for a name. Should we allow settings to not ask, optionally ask and force input (as with name settings in events)? Currently it is optional, but always asked. Also from a UI perspective when creating a user account, it is not clear, whether the name is optional or mandatory. I know mandatory fields are usually marked with an asterisk*, but in this interface, email is also marked optional (I assume due to technical reasons as email-input is only mandatory when having the radio-box „create account“ selected – we might need some JavaScript-logic in there). Also, when I submit account creation without entering an email address, I get an error `You need to fill out all fields.` which indicates I need to give a name – which at least at the moment is not true. Also, from a A11y-perspective, the erroneous fields are not marked properly that they contain errors.

Very minor, but in the account creation confirmation email, there is an empty space after „Hello ,“.


After I created the account, according to the checkout form, I am logged in: „You are currently logged in with the following credentials.“, but in the header top right, there is a link „Anmelden“. When I click it, I get to a login page – so I am not really logged in, but in some kind of in between state? Might be a bit irritating for users.
On the log-in page localization seems odd. I changed the organizers’ language to German, but everything is still English – might be because of new localization strings, but the link „Anmelden“ in the header reads „Log in“. So that seems odd.


**Setting password** 

Minor, as one can assume that they are mandatory, but form inputs are not marked correctly as mandatory. Furthermore there should be a help_text indicating the minimum requirements for a valid password.

For password managers, we should find a way to specify the username/email in the password-reset form. Currently it tries to overwrite my password for pretix-control (which is an edge case as usually users do not have a pretix-control account), but if no account is given, only the password will be saved, not the username/email. I need to further investigate, but maybe a disabled input for the username might help? Also, this reinsures the user that he/she/them is/are resetting the password for the right user.


**Update account settings**

When updating my account profile, a name is suddenly mandatory.

** Create account **

When I am creating an account outside checkout, then a name is mandatory as well.


** Logged in/pretix session**

Cookies for sessions pretix_session are saved on path /. This still enables me to be logged into two different accounts on two different organizers, but when I log out on organizer, I get logged out on the other one as well. I guess the session is killed on logout. I know this is an edge case …


In pretix-control I have not yet tested much. It looks good so far, created a Membership, linked product, bought through membership.

raphaelm:
note to self:

>wenn ich ein Produkt kaufe, dass eine Membership braucht, ich aber kein Membership habe, dann sagt er mir, ich brauche ein Membership um fortzufahren, aber der „Fortfahren”-Button wird trotzdem angezeigt. Wenn ich den klicke, bekomme ich das hier:
KeyError at /membersonly/supergeheim/checkout/membership/
'membership'

raphaelm:
> A user account asks for a name. Should we allow settings to not ask, optionally ask and force input (as with name settings in events)? Currently it is optional, but always asked.

I don't think we should add a setting unless we have a real-world use case to require a setting. I think it's useful to always have a name, and it being optional during checkout is a bug that I've now fixed.

> Very minor, but in the account creation confirmation email, there is an empty space after „Hello ,“.

Should no longer happen due to the name being required.

> After I created the account, according to the checkout form, I am logged in: „You are currently logged in with the following credentials.“, but in the header top right, there is a link „Anmelden“. When I click it, I get to a login page – so I am not really logged in, but in some kind of in between state? Might be a bit irritating for users.

Yes, you're in an in-between state, you are only authorized for the purpose of that checkout. This is intentional because you did not confirm your email address yet, so we can't log you in (otherwise you'd gain unauthorized access to previous purchases with that email address).

> On the log-in page localization seems odd. I changed the organizers’ language to German, but everything is still English – might be because of new localization strings, but the link „Anmelden“ in the header reads „Log in“. So that seems odd.

I struggle to reproduce that. Can we look at that together?

> Minor, as one can assume that they are mandatory, but form inputs are not marked correctly as mandatory. Furthermore there should be a help_text indicating the minimum requirements for a valid password.

Asterisks are now there. I struggle with the help_text as this is kinda dynamically generated by django and includes things like "not a common password" or "not your email address". I think the only really helpful thing to note is "min 8 characters" and a minlength attr might be even more helpful than a help text.

> Cookies for sessions pretix_session are saved on path /. This still enables me to be logged into two different accounts on two different organizers, but when I log out on organizer, I get logged out on the other one as well. I guess the session is killed on logout. I know this is an edge case …

Yeah, I've thought about this for a while. The global logout makes it a little less likely to screw up security issues in the future, but it's really annoying. I've changed it now.

PR #2025, opened 2021-04-13
Clean up alignment on shop page

raphaelm:
Ok, so this alignment issue on the frontpage (eg text of "resend email" not aligned with its headline) has been bugging me for YEARS, but I thought there was no clean way out: Removing the padding from the "resend email" section would mis-align the button with the cart button, and removing the padding from everywhere would screw with the responsiveness. After a new customer just got pretty annoyed about this on a call, I decided to try again.

I needed some CSS hacks with ::before and ::after to make all borders look good, but I think the end result is pretty okay.

## Before

local pretix eu_democon_3vjrh_ (1)

## After

local pretix eu_democon_3vjrh_

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
+++ b/src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
@@ -17,10 +17,10 @@
         {% endif %}
         {% for item in tup.1 %}
             {% if item.has_variations %}
-                <article>
-                <details class="item-with-variations" {% if event.settings.show_variations_expanded %}open{% endif %}
+                <article class="item-with-variations">
+                <details {% if event.settings.show_variations_expanded %}open{% endif %}
                         id="item-{{ item.id }}">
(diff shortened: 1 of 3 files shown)


wiffbi:
Looks way better! Clever idea to use the :before/:after-pseudo elements for the border. Looking at the HTML-code it also looks more consistent regarding the use of bootstrap .row class instead of row-voucher, etc.

An alternative approach with only a change to the CSS would have been with :nth-last-child (:last-child does not work due to the div.clearfix-element) and resetting the paddings:

`*[class^="col-"]:first-child {
    padding-left: 0;
}
*[class^="col-"]:nth-last-child(2) {
    padding-right: 0;
}
.row *[class^="col-"]:first-child,
*[class^="col-"]:first-child:nth-last-child(2) {
    padding-left: 15px;
}
.row *[class^="col-"]:nth-last-child(2) {
    padding-right: 15px;
}`

But your approach seems cleaner and makes the HTML more consistent, so I’d say we go with yours!

PR #2030, opened 2021-04-18
Check if celery channel has client before getting metrics

### LINKED ISSUE #2029: Celery error in metrics endpoint
**Describe the bug**
When using the metrics endpoint while celery is configured I get a 500 and the following error in the log:

```
ERROR 2021-04-18 15:19:39,868 django.request log Internal Server Error: /metrics
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/usr/local/lib/python3.8/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.8/contextlib.py", line 75, in inner
    return func(*args, **kwds)
  File "/pretix/src/pretix/base/views/metrics.py", line 74, in serve_metrics
    m = metrics.metric_values()
  File "/pretix/src/pretix/base/metrics.py", line 289, in metric_values
    client = app.broker_connection().channel().client
AttributeError: 'Channel' object has no attribute 'client'
```
Relevant line: https://github.com/pretix/pretix/blob/1844a1fe6d8cfb6e0337fd404276286ca0e32f39/src/pretix/base/metrics.py#L289

**Expected behavior**
The metrics endpoint should return the metrics with status code 200, not an Internal Server Error.

neumantm:
The current implementation only works if redis is used as broker.
Otherwise it would throw an Exception, which causes a 500 to be returend.
Now we just don't have celery metrics in those cases.

Fixes #2029

### DIFF
--- a/src/pretix/base/metrics.py
+++ b/src/pretix/base/metrics.py
@@ -286,17 +286,19 @@ def metric_values():
             metrics['pretix_model_instances']['{model="%s"}' % m._meta] = estimate_count_fast(m)
 
     if settings.HAS_CELERY:
-        client = app.broker_connection().channel().client
-        for q in settings.CELERY_TASK_QUEUES:
-            llen = client.llen(q.name)
-            lfirst = client.lindex(q.name, -1)
-            metrics['pretix_celery_tasks_queued_count']['{queue="%s"}' % q.name] = llen
-            if lfirst:
-                ldata = json.loads(lfirst)
-                dt = time.time() - ldata.get('created', 0)
-                metrics['pretix_celery_tasks_queued_age_seconds']['{queue="%s"}' % q.name] = dt
-            else:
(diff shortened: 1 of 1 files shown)


raphaelm:
`except: pass` is usually a bad idea. If this suddenly stops working for an unrelated reason (think IOError), monitoring setups could be dropped silently. *If* we want to fix this (see comment in issue), we should either check for the broker explicitly or only catch eg AttributeError

neumantm:
You are right. I will push a new proposal shortly.

raphaelm (approved):

PR #2039, opened 2021-04-22
add support for check-in into overlapping events

wiffbi:
When events overlap, check-in only worked for the currently running event. If events run back-to-back, it can happen, that admission should start earlier and overlaps the currently running event. This checks if an overlapping event has started even if the current event is still running.

TODO: The test regarding „Switch at half-time“ currently fails. I am actually not so sure whether that behavior is actually useful. In the current test, at 15:29 it expects to have the check-in list selected for the first event, which actually ends at 15:00 – IMHO that does not make much sense. Or is that maybe needed for the optional check-out process?


raphaelm (approved):
LGTM, but needs a rebase

PR #2040, opened 2021-04-23
Fix #1567 -- Per-subevent availability of items

### LINKED ISSUE #1567: Relative availablities for products
When setting up product availabilities right now, we only accept absolute dates.

However, in the context of eventseries that might want to sell early bird tickets, this is not very useful, if there is early bird tickets for every single date/performance.

We could counter this by extending the absolute dates with our relative date/time-picker and tie the relative dates to the subevent's date.

Z#2355699

raphaelm:
We've wanted to make ``Item.available_from`` and ``Item.available_until`` useful for event series for a long time.

So far, the idea was to turn them into something like our relative date fields (or similar to #2003), so you could express them relative to the event date. The main problem is that storing a duration like "24 hours before" is not good enough, it will for example give you 9am as the end of sale for a 10am event if daylight saving time is switched in the night, which is totally unexpected for users who are told "event sale ends exactly a day before the event".

As @felixrindt  already noted in #1567, this is far from easy since you need to express that logic in SQL. *Theoretically* that's no problem, "9am on the day before" could be expressed as ``TruncDay('start_date') - F('number_of_days') * datetime.timedelta(days=1) + F('time_of_day')`` with ``time_of_day`` set to 9am and ``number_of_days`` set to one. **Practically**, @zanderle did [some research](https://github.com/pretix/django-dateadd/pull/1/files) and it looks like there's no way to do this that works across all three supported databases.

This PR therefore proposes to avoid this problem alltogether and approach this from a completely different point of view: Attaching datetimes to the relation between subevents and items that we already have!

This has one major downside: The relative date needs to be set during bulk-creation of the subevent and there's no good way of changing it later. But it's better than nothing, and we can always spend more energy on the bulk editing to make it more clever.

- [x] data model
- [x] subevent create/update
- [x] subevent bulk create
- [x] subevent bulk update
- [x] check on event index page
- [x] check in cartmanager
- [x] check in order creation
- [x] check for calendar
- [x] REST API
- [x] tests


pc-coholic:
LGTM and seems to be working fine, too.

But I do have some other issue with that PR: I feel like the handling of the availability is not consistent with it anymore.

Currently, for a regular event, I am only able to set the availability for the whole item, not for variations.

For an event-series, with this PR, I can set a global availability for an item and all variations that will take precedence for all subevents, as well as a per-variation-availability for every single subevent.

I am not sure, what the best solution here is, but I would expect that the availabilities I can set are of the same type across all options:
- Add variation-availabilities on an item-level
- Remove variation-availabilities on a subevent-level
- Add item-including-all-variations-availabilities on a subevent-level

I think one possible solution might be:
- Regular Event / Regular Item: Availabilities as they are right now
- Regular Event / Item with Variations: Availabilities only on Variations
- Event Series / Regular Item: No availabilities on the item, only at a subevent-level
- Event Series / Item with Variations: Availabilities on Variations, only at a subevent-level.

raphaelm:
> Currently, for a regular event, I am only able to set the availability for the whole item, not for variations.

Yes. On the first glance, this is a mere technical consequence of the fact that the other subevent-item-override relations work the same way and it would be very confusing in both the code and the UI if we'd it differently.

However, the reasoning goes deeper. For the past years, I've hold the strong opinion that we should allow as little options on variation level as possible to keep configuration manageable.

I think I've lost that battle. With the decision to make seating categories bound to products, a whole lot of important use cases require more flexibility on variation level. Therefore, I've stopped holding back to introduce new options on variation level, for example in this PR or `require_membership` in #2024.

In the forseeable future, I think we'll also need to implement ``available_from``, ``available_until``, ``require_voucher``, ``hide_without_voucher``, ``checkin_attention``, and maybe even ``min_per_order`` on variation level, since it's the only way to do early bird pricing, voucher pricing or group pricing in a seated event. But I think that change warrants its own PR, and a pretty fundamental change in the settings UI for variations (I have an idea already, happy to discuss, but maybe not here).

> For an event-series, with this PR, I can set a global availability for an item and all variations that will take precedence for all subevents, as well as a per-variation-availability for every single subevent.

I'm not sure "take precedence" is the correct description, *both* restrictions will be active at the same time and both need to be satisfied for the product to be available.

> Add variation-availabilities on an item-level
> Remove variation-availabilities on a subevent-level

As outlined above, as much as I dislike adding so many options, I'd rather go with the first one.

> Add item-including-all-variations-availabilities on a subevent-level

This would indeed bring more consistency, even if we add variation-availability on item level, but it would also add *even more fields* to the already very complex subevent editor. Honestly not sure if it's worth it.

> I think one possible solution might be:
> Regular Event / Regular Item: Availabilities as they are right now

Sure.

> Regular Event / Item with Variations: Availabilities only on Variations

That would be a pretty breaking change, and also an inconvenient one. If you have an early bird ticket in the variations "adult", "youth", "child", and "member", you'd need to set the early-bird cutoff four times. Just to avoid a "duplicate field"? I'd rather redesign the form and add lots of sanity checks.

> Event Series / Regular Item: No availabilities on the item, only at a subevent-level.
> Event Series / Item with Variations: Availabilities on Variations, only at a subevent-level.

Same as above, I'm not sure we'd want to take that away from people if we need to maintain the code for it anyways (for the non-subevent case). I can totally come up with use cases for both. If you have a week-long exhibition with 10 timeslots a day, you might still want to have the same early-bird deadline for all subevents.

I see that keeping it on both levels might lead to even more "why doesn't my product show up?" on the user end, and you're right that there will continue to be an inconsistency between the different levels, but it feels like the most feasible compromise to me. What do you (and others) think?

PR #2044, opened 2021-04-26
Remove redundant mime-type from nginx config

IrmantasM:
Small change to remove warning noise from logs:

```sh
nginx: [warn] duplicate MIME type "text/html" in /etc/nginx/nginx.conf:43
```

http://nginx.org/en/docs/http/ngx_http_gzip_module.html#gzip_types 

```sh
Enables gzipping of responses for the specified MIME types in addition to “text/html”. The special value “*” matches any MIME type (0.8.29). Responses with the “text/html” type are always compressed.
```

### DIFF
--- a/deployment/docker/nginx.conf
+++ b/deployment/docker/nginx.conf
@@ -33,7 +33,7 @@ http {
 
 	gzip on;
 	gzip_disable "msie6";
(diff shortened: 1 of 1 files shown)


IrmantasM:
Not sure why GitHub doesn't like my identity, git client is configured to use same email as in my settings :/

raphaelm:
Weird. I'm happy to ignore it for this one-liner though, especially since it's not in the license-sensitive part :D

PR #2048, opened 2021-04-27
Fix type error in license check view

felixrindt:
Should check for license being None first.

### DIFF
--- a/src/pretix/control/views/global_settings.py
+++ b/src/pretix/control/views/global_settings.py
@@ -236,6 +236,14 @@ class LicenseCheckView(StaffMemberRequiredMixin, FormView):
         for entry_point in pkg_resources.iter_entry_points(group='pretix.plugin', name=None):
             license, url = self._get_license_for_pkg(entry_point.dist.key)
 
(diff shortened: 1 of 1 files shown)


src/pretix/control/views/global_settings.py
```diff
@@ -235,6 +235,14 @@ def _check_results(self, input):
...
         for entry_point in pkg_resources.iter_entry_points(group='pretix.plugin', name=None):
             license, url = self._get_license_for_pkg(entry_point.dist.key)
+            
+            if not license or not any(l in license for l in ('Apache', 'MIT', 'BSD', 'pretix Enterprise', 'GPL')):
+                res.append((
+                    'muted', 'warning',
+                    _('We found the plugin "{plugin}" with license "{license}" which this tool does not know about and '
+                      'therefore cannot give any recommendations.').format(plugin=entry_point.dist.key, license=license)
+                ))
+                continue
```
felixrindt, line 245: Could avoid the `continue` and build an `elif` stack, which might also avoid multiple warnings for the same license, although that shouldn't happen.

raphaelm:
@cla-bot check

raphaelm (approved):

PR #2050, opened 2021-04-28
Check-in rules: Make logic results understandable

raphaelm:
While our custom check-in logic is very flexible, its main problem is that is is pretty intransparent during execution. If the logic causes an entry to be forbidden, the result of the logic evaluation is just a simple ``False``, which is very unhelpful to explain to attendees why they don't get into the event.

The main problem with fixing this is that there is no correct answer for this, it is always up for interpretation. A good example is the following set of rules:

- Attendees with a regular ticket can enter the venue between 09:00 and 17:00 on three days 
- Attendees with a VIP ticket can enter the venue between 08:00 and 18:00 on three days

If an attendee with a regular ticket now shows up at 17:30 on the first day, there are three possible error messages:

a) You do not have a VIP ticket
b) You can only get in before 17:00
c) You can only get in after 09:00 tomorrow

All three of them are just as valid, and "fixing" either one of them would get the attendee in. Showing all three is too much, especially since the list can get very long with complex logic.

We therefore make an opinionated choice based on a number of assumptions. An example for these assumptions is "it is very unlikely that the attendee is unable to change their ticket type". Additionally, we favor a "close failure". Therefore, in the above example, we'd show "You can only get in before 17:00". In the middle of the night it would switch to "You can only get in after 09:00".

Todo:

- [x] expose to api
- [ ] show in apps

### DIFF
--- a/src/pretix/base/services/checkin.py
+++ b/src/pretix/base/services/checkin.py
@@ -46,6 +46,7 @@ from django.db.models import (
 )
 from django.db.models.functions import Coalesce, TruncDate
 from django.dispatch import receiver
+from django.utils.formats import date_format
 from django.utils.functional import cached_property
 from django.utils.timezone import make_aware, now, override
 from django.utils.translation import gettext as _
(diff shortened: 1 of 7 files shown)


src/pretix/base/services/checkin.py
```diff
@@ -55,27 +56,198 @@
...
+        return ev.date_from
+    elif t == 'date_to':
+        return ev.date_to or ev.date_from
+    elif t == 'date_admission':
+        return ev.date_admission or ev.date_from
+
+
+def _logic_explain(rules, ev, rule_data):
+    """
+    Explains whe the logic denied the check-in. Only works for a denied check-in.
```
wiffbi, line 79:
```suggestion
    Explains when the logic denied the check-in. Only works for a denied check-in.
```

src/pretix/base/services/checkin.py
```diff
@@ -55,27 +56,198 @@
...
+
+    While our custom check-in logic is very flexible, its main problem is that is is pretty
```
wiffbi, line 81:
```suggestion
    While our custom check-in logic is very flexible, its main problem is that it is pretty
```

wiffbi (commented):
The assumptions for the opinionated choice sound reasonable to me. Maybe outputting 2 or even 3 hints might be possible? In the current scenario and generally, maybe give one error-message with the highest weight („no VIP-ticket“ – the kind of super show-stopper) and the other with the lowest weight (check-in only until 17:00 – your ticket would have given you access, but only until 17:00).

raphaelm:
> The assumptions for the opinionated choice sound reasonable to me. Maybe outputting 2 or even 3 hints might be possible? In the current scenario and generally, maybe give one error-message with the highest weight („no VIP-ticket“ – the kind of super show-stopper) and the other with the lowest weight (check-in only until 17:00 – your ticket would have given you access, but only until 17:00).

Interesting idea to show both "extrema", but I think it doesn't work well:

* The goal of this is to allow decisions at the entry to be explained to avoid long conversations, there's not much time reading long texts.
* Rules often take the structure ``(product = X AND …) OR (product = Y AND …) OR (product=Z AND …)``. In this cae, there would possibly be *a lot* error message with the highest weight, which is just too much.
* There will often be some kind of "employee pass" that always gets in, but it's none of the attendees business that such a thing even exists.

wiffbi:
Thank you for your explanation, makes total sense!

PR #2054, opened 2021-05-02
License check: improve license detection for DistInfoDistribution packages.

MaicoTimmerman:
The license finder assumes that installed packages are in `.egg` or `.egg-info` format and does not consider `.dist-info` packages, which uses the `METADATA` field for package info. See [pkg_resources.DistInfoDistribution](https://github.com/pypa/pip/blob/main/src/pip/_vendor/pkg_resources/__init__.py#L3000) for reference.

### DIFF
--- a/src/pretix/control/views/global_settings.py
+++ b/src/pretix/control/views/global_settings.py
@@ -169,7 +169,7 @@ class LicenseCheckView(StaffMemberRequiredMixin, FormView):
         license, url = None, None
         try:
             pkg = pkg_resources.get_distribution(pkg)
-        except pkg_resources.DistributionNotFound:
+        except:
             return None, None
         try:
             for line in pkg.get_metadata_lines(pkg.PKG_INFO):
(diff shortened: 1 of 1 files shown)


pc-coholic:
@cla-bot check

src/pretix/control/views/global_settings.py
```diff
@@ -168,12 +168,11 @@ def get_context_data(self, **kwargs):
     def _get_license_for_pkg(self, pkg):
         license, url = None, None
         try:
-            pkgs = pkg_resources.require(pkg)
-            pkg = pkgs[0]
-        except:
+            pkg = pkg_resources.get_distribution(pkg)
+        except pkg_resources.DistributionNotFound:
```
raphaelm, line 172: The full except was intentional, we had e.g. dev installs crash in some cases because packages weren't correctly installed at all, which was annoying. Maybe this PR fixes all these, but maybe it doesn't, I wouldn't want to take the risk. Otherwise great PR!

MaicoTimmerman: Alright, I've reverted to the broader exception!

PR #2066, opened 2021-05-07
Waiting list: Cronjob should skip events with no waiting entries

raphaelm:


wiffbi (commented):
Looks good! Only isort with imports need fixing.

PR #2068, opened 2021-05-07
Visual: improve number inputs padding and make spinners bigger

wiffbi:


wiffbi:
Fails probably due to it being based on „old“ master. Fixed by rebasing

PR #2072, opened 2021-05-09
Allow to send all invoices to a specific email address

raphaelm:


pc-coholic (approved):
Found a missing whitespace - but apart from that I think this looks good.

There is only one thing that I ran into, that could use some thought (or a stamp of approval/ignore):
- `sent_to_organizer` is nullable and will be null for example when an invoice is regenerated. On the next `runperiodic`-run, it will be properly set to `0` where required. However: Since there is no middleground and either an invoice has been sent or not, I wouldn't have expected to find any `null` values at all.

raphaelm:
> sent_to_organizer is nullable and will be null for example when an invoice is regenerated. On the next runperiodic-run, it will be properly set to 0 where required. However: Since there is no middleground and either an invoice has been sent or not, I wouldn't have expected to find any null values at all.

So first of all there's obviously no good way to handle regeneration, which is part of my motivation for #2071.

The intention for having both a null and a false state was mainly backwards compatibility to avoid sending out thousands of old invoices if this feature is used for the first time, as well as to avoid checking whether an invoice needs to be sent multiple times if the feature is turned off. I'll at least add in a comment about that, though.

PR #2075, opened 2021-05-12
Add salutation "Mx", normalize salutations (save in English, localize on display)

wiffbi:
Salutations were saved localized, which already had its problems, when the customer’s language was different than the language of the user in pretix-control (when editing a customer’s profile, the salutation was not pre-selected). Furthermore, introducing the salutation „Mx“ needed special handling when generating the new placeholder {name_for_salutation}, which wasn’t as straight forward when salutations are saved localized.

This PR includes a migration that changes all salutations in the database to its normalized, english counterpart.

This PR introduces a new placeholder {name_for_salutation} that enables to correctly address people in emails either as:

- Dear Ms Doe
- Dear Mr Doe
- if salutation is „Mx" or if no salutation was selected: Dear Randy Dow


src/pretix/base/settings.py
```diff
@@ -2442,10 +2442,29 @@ def unserialize(cls, s):
 ])
 
 PERSON_NAME_SALUTATIONS = [
-    pgettext_lazy("person_name_salutation", "Ms"),
-    pgettext_lazy("person_name_salutation", "Mr"),
+    "Ms",
+    "Mr",
+    "Mx",
```
raphaelm, line 2447: `makemessages` is statically analyzing the code to find all gettext_usages, but it's not clever enough to evaluate the dynamic call you introduced in `questions.py`. Therefore, "Mr" will never show up in the .po file, and therefore our translators will never see or translate it.

We'll need to have the static pgettext-statment *somewhere* in our codebase. Usually, ``gettext_noop`` would solve that, but there unfortunately is no ``pgettext_noop``. To avoid changing the makemessages command in Django, the easiest solution would probably be to put this like a choice array:

```
PERSON_NAME_SALUTATIONS = [
    ("Ms", pgettext_lazy("person_name_salutation", "Ms")),
    ("Mr", pgettext_lazy("person_name_salutation", "Mr")),
    ("Mx", pgettext_lazy("person_name_salutation", "Mx")),
]
```

and then use it appropriately in the form

raphaelm (changes requested):
Looks good, I also tested the migration locally, works fine. I added one request for change on the gettext usage.

src/pretix/base/settings.py
wiffbi: Good idea, I changed it accordingly.

raphaelm (approved):

PR #2081, opened 2021-05-19
WIP: A11y improvements

wiffbi:
Working on improvements regarding BITV-test in May 2021.


src/pretix/presale/templates/pretixpresale/event/index.html
```diff
@@ -134,28 +134,28 @@ <h2 class="content-header">
...
-                            {{ ev.get_date_range_display }}
+                        <p><span class="sr-only">{% trans "When?" %}</span>
+                            <time datetime="{{ ev.date_from.isoformat }}">{{ ev.get_date_range_display }}</time>
                             {% if event.settings.show_times %}
                                 <br>
                                 <span data-time="{{ ev.date_from.isoformat }}" data-timezone="{{ request.event.timezone }}">
-                                    {% blocktrans trimmed with time=ev.date_from|date:"TIME_FORMAT" %}
-                                        Begin: {{ time }}
+                                    {% blocktrans trimmed with time=ev.date_from|date:"TIME_FORMAT" time_24=ev.date_from|time:"H:i" %}
+                                        Begin: <time datetime="{{ time_24 }}">{{ time }}</time>
```
raphaelm, line 151: We should avoid bothering translators with HTML. Not at any cost, but whenever it's feasible. It only causes mistakes when translators get it wrong, and changing the markup breaks all translations.

The following could work (untested):

```
{% with ev.date_from|time:"H:i" as time_24h %}
    {% blocktrans trimmed with time="<time datetime='"|add:time_24|add:"'>"|add:time|add:"</time>"|safe %}
        Begin: {{ time }}
    {% endblocktrans %}
{% endwith %}

Same for the other fields of course

```

src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
```diff
@@ -4,331 +4,353 @@
...
+                {% trans "The items in your cart are no longer reserved for you. You can still complete your order as long as they’re available." %}
+            {% endif %}
+        </p>
+    {% else %}
+        <p class="sr-only" id="cart-deadline">
+        {% if editable %}
+            {% trans "Overview of your ordered products. You can still change your order." %}
+        {% else %}
+            {% trans "Overview of your ordered products. You cannot change your order anymore." %}
         {% endif %}
```
raphaelm, line 351: The combination of `is_ordered` and `editable` can never happen. If orders are change-able, this will be in different ways through cancellation / rebook buttons. I'd suggest just

```suggestion
        {% trans "Overview of your ordered products." %}
```

src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
```diff
@@ -261,29 +275,31 @@ <h4>
...
                         </div>
                     {% elif item.cached_availability.0 == 100 %}
                         <div class="col-md-2 col-xs-6 availability-box available">
 							{% if item.order_max == 1 %}
                                 <label class="item-checkbox-label">
                                     <input type="checkbox" value="1" {% if itemnum == 1 %}checked{% endif %}
                                            {% if not ev.presale_is_running %}disabled{% endif %}
-                                           name="item_{{ item.id }}" id="item_{{ item.id }}">
+                                           name="item_{{ item.id }}" id="item_{{ item.id }}"
+                                           aria-label="{% blocktrans with item=item.name %}Add {{ item }} to basket{% endblocktrans %}"
```
raphaelm, line 291: basket → cart

src/pretix/presale/templates/pretixpresale/event/fragment_subevent_calendar.html
```diff
@@ -1,41 +1,62 @@
...
-        {% endif %}
-    {% endfor %}
     <div class="row">
         <div class="col-sm-4 hidden-xs text-left flip">
-            <a href="?{% url_replace request "year" subevent_list.before.year "month" subevent_list.before.month %}"
-                    class="btn btn-default">
-                <span class="fa fa-arrow-left" aria-hidden="true"></span>
-                {{ subevent_list.before|date:"F Y" }}
-            </a>
+            <form method="get" action="{{ request.path }}">
```
raphaelm, line 6: Converting this (any the many others like it) from a link into a form will probably stop Google and other search engines from scraping the calendar which might negatively impact future events to show up in search engines. Do we want that? Do we need to make this change?

src/pretix/presale/templates/pretixpresale/event/fragment_subevent_calendar_week.html
```diff
@@ -1,44 +1,64 @@
...
-        {% endif %}
-    {% endfor %}
     <div class="row">
         <div class="col-sm-4 hidden-xs text-left flip">
-            <a href="?{% url_replace request "year" subevent_list.before.isocalendar.0 "week" subevent_list.before.isocalendar.1 %}"
-                    class="btn btn-default">
-                <span class="fa fa-arrow-left" aria-hidden="true"></span>
-                {{ before|date:subevent_list.week_format }}
-            </a>
+            <form method="get" action="{{ request.path }}">
```
raphaelm, line 6: crawler issue, see above

src/pretix/presale/templates/pretixpresale/event/index.html
```diff
@@ -137,45 +159,53 @@ <h2 class="content-header">
...
                             {% if ev.date_admission %}
                                 <br>
                                 {% if ev.date_admission|date:"SHORT_DATE_FORMAT" == ev.date_from|date:"SHORT_DATE_FORMAT" %}
-                                    <span data-time="{{ ev.date_admission.isoformat }}" data-timezone="{{ request.event.timezone }}">
-                                        {% blocktrans trimmed with time=ev.date_admission|date:"TIME_FORMAT" %}
-                                            Admission: {{ time }}
+                                    <span data-time="{{ ev.date_admission.isoformat }}" data-timezone="{{ request.event.timezone }}">  
+                                    {% with time_human=ev.date_admission|date:"TIME_FORMAT" time_24=ev.date_admission|time:"H:i" %}
+                                        {% blocktrans trimmed with time='<time datetime="'|add:time_24|add:'">'|add:time_human|add:"</time>"|safe %}
+                                            Admission: <time datetime="{{ time }}">{{ time }}</time>
```
raphaelm, line 198: Whoops, I think this tag is twice

src/pretix/presale/templates/pretixpresale/event/waitinglist.html
```diff
@@ -3,36 +3,38 @@
 {% block title %}{% trans "Waiting list" %}{% endblock %}
 {% block content %}
     <h2>{% trans "Add me to the waiting list" %}</h2>
-    <form action="" method="post">
+    <form action="{{ request.get_full_path }}" method="post">
```
raphaelm, line 6: Is there any benefit to this change? Just in case I can learn something from it :)

raphaelm (changes requested):
So, a few notes / bugs / questions inline, and here are a few visual things I've noticed

## Focus outline

There's some weirdness going on with the +/- buttons in cart. The outline is in a different position when in `:active:focus` and in `:focus`. Not a big deal, but looks a little weird when clicking and holding the mouse slowly:

2021-07-10-201510_179x179_scrot
2021-07-10-201521_174x178_scrot

## Underlined links

As we already discussed, I totally get that links should be underlined when the link is embedded in text, but it often looks weird in other contexts. I think the one that aesthetically bugs me most is this:
2021-07-10-201647_1400x151_scrot

It looks so asymmetric! I also don't think it's a big deal if people do not notice that those are clickable – they're still accessible through the "Back" button at the bottom of the page.

Also, I think this one looks really horrible: especially because of the underlined space between the icon and the text, but also generally:

2021-07-10-201706_534x117_scrot

It also bugs me for the "Show all events of …" because it makes it much more visually prominent than it was (and should be?). Funnily, I don't care as much with the languages. Maybe because they are just one word and underlined text looks especially weird if there are spaces?

shop_frontpage_desktop_custom_sharp_largeheader_title_ar,de,en,fr_link

The waiting list also has the "underlined space after icon" issue:

2021-07-10-204002_136x51_scrot


I also don't like this one but I think we already discussed that there's no better option:

2021-07-10-201601_489x142_scrot

## Margins and paddings

I feel that this one looks a little off, although it's technically symmetric. I think the issue is that the vertical padding of this box is visibly larger than the horizontal one, and it's *also* larger than the margins inside, such as between "2 Produkte" and "The items in your cart…"

2021-07-10-201749_1525x469_scrot

Since you introduced many ``<p>`` tags, we should probably also remove the horizontal margin after a ``<p>`` tag if it's the last child of a panel:

2021-07-10-201809_1527x254_scrot

Here's the order confirmation, do you think it might look better with increased margin between the "Bitte speichern Sie…" text and the "Your order" headline?

2021-07-10-201853_1550x712_scrot

## Order confirmation page

We already talked about this before, and even though it will take me a long time to get used to it, you are probably right about moving the order status from the right to the left. However, I feel it doesn't look totally clean with both the order code and the order status between the headline and the first panel. What do you think about going only half-way, i.e. moving the order status but keeping the order code in the headline, like this?

2021-07-10-202120_1550x327_scrot


## Alert backgrounds

I generally like bringing the background colors back to alert boxes (after you explained the reasoning to me):

2021-07-10-203359_1540x202_scrot

However I feel it looks a little weird as soon as they contain buttons. For example, we use alert-info on the order confirmation page as some sort of a lighter panel that doesn't have a headline and uses less space:

2021-07-10-202006_1554x445_scrot

We could either drop the background for alert-info, which would be slightly inconsistent, or we could re-design this box into something else entirely. Any ideas?

I was also worried it would look weird in the backend where we have some with longer texts but I think it is okay:

2021-07-10-203546_2138x288_scrot

``alert-legal`` is still missing, though, unless that was intentional:

2021-07-10-202158_2144x157_scrot

src/pretix/presale/templates/pretixpresale/event/waitinglist.html
wiffbi: AFAIK regarding browser implementation: No. But the HTML-validator complained about it. And in the A11y-audit they complained about invalid HTML here and there (that’s why I also added loads of <p>-elements).

src/pretix/presale/templates/pretixpresale/event/fragment_subevent_calendar.html
wiffbi: No, not really. I think they are fine as links – even just to see where it is taking me. The only reason I changed it, was because it was mentioned in the audit that these UI-elements need to be <button>s. But I will try to debate whether this is really true.

src/pretix/static/lightbox/js/lightbox.js
```diff
@@ -78,7 +78,7 @@
...
+    $('<div id="lightboxOverlay" class="lightboxOverlay"></div><div id="lightbox" class="lightbox" role="dialog" aria-labelledby="lightboxLabel" aria-modal="true"><div class="lb-outerContainer"><div class="lb-container"><img class="lb-image" src="data:image/gif;base64,R0lGODlhAQABAIAAAP///wAAACH5BAEAAAAALAAAAAABAAEAAAICRAEAOw==" /><div class="lb-nav"><a class="lb-prev" href="" ></a><a class="lb-next" href="" ></a></div><div class="lb-loader"><a class="lb-cancel"></a></div></div></div><div class="lb-dataContainer"><div class="lb-data"><div class="lb-details" id="lightboxLabel"><span class="lb-caption"></span><span class="lb-number"></span></div><div class="lb-closeContainer"><button aria-label="'+gettext("close")+'" class="lb-close"></button></div></div></div></div>').appendTo($('body'));
```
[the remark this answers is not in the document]
wiffbi: You are right, not very good to edit a third-party script. I will add a comment.

src/pretix/static/pretixpresale/scss/main.scss
```diff
@@ -20,31 +20,68 @@
...
     text-align: center;
     font-size: 11px;
 }
+footer nav ul {
+    display: flex;
+    justify-content: center;
+    list-style: none;
+}
+footer nav li:not(:first-child):before {
+    content: "•";
```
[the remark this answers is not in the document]
wiffbi: No, I usually take the middot, but probably have messed it up here. Will change it to middle dot.

src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/pending.html
```diff
@@ -16,34 +16,43 @@
...
+                    <dt>{% trans "IBAN" %}:</dt><dd>{{ settings.bank_details_sepa_iban|ibanformat }}</dt>
+                    <dt>{% trans "BIC" %}:</dt><dd>{{ settings.bank_details_sepa_bic }}</dt>
+                    <dt>{% trans "Bank" %}:</dt><dd>{{ settings.bank_details_sepa_bank }}</dt>
+                {% if details %}
+                </dl>
+                {% endif %}
             {% endif %}
             {% if details %}
-                {{ details|linebreaksbr }}<br>
+                {{ details|rich_text }}
```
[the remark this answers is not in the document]
wiffbi: Good catch! I actually just changed it because of multiple line breaks being „forbidden“ (they were marked as an „Verstoß“ in A11y-testing) and rich_text converts multiple line breaks to <p> elements. Is the a filter that das linebreaksbr and turns multiple linebreaks into <p> and nothing else?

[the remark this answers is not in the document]
wiffbi: As far as I can remember it actually wasn’t my intention to add markdown-support, but to add support for paragraphs – which `|linebreaks` does just fine. I removed `rich_text` and changed it to `linebreaks`. Changed it in checkout_payment_form.html as well.

src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/checkout_payment_form.html
```diff
@@ -23,4 +25,4 @@
             {% trans "We will assign you a personal reference code to use after you completed the order." %}
         </strong>
     {% endif %}
-</address>
+</p>
```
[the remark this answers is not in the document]
wiffbi: Good point! I styled it accordingly.

src/pretix/static/pretixpresale/scss/main.scss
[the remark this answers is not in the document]
wiffbi: „middle dot“ is rendered narrower/lighter. I added font-wheight: bold to it.

src/pretix/presale/templates/pretixpresale/fragment_week_calendar.html
```diff
@@ -42,9 +42,9 @@ <h3><span class="fa fa-fw" aria-hidden="true"></span> {{ day.day_formatted }}</h
                                     {% if not show_names|default_if_none:True %}
                                         <strong>
                                     {% endif %}
-                                    {{ event.time|date:"TIME_FORMAT" }}
+                                    <time datetime="{{ event.time|date:"H:i" }}">{{ event.time|date:"TIME_FORMAT" }}</time>
```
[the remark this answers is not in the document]
wiffbi: As this is part of a list of time/time-ranges for a given day, this would be tedious for screen-reader users to always have the full date read out. First they navigate through the days (details/summary) and then decide to drill down on a day.

src/pretix/static/pretixpresale/scss/_cart.scss
```diff
@@ -68,23 +64,124 @@
...
+    .cart .checkout-button-row > div,
+    .cart .checkout-button-row p {
+        margin-bottom: 0;
+    }
+    .cart .checkout-button-row .btn {
+        white-space: nowrap;
+        width: auto;
+    }
+}
+*/
```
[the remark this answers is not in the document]
wiffbi: No, I removed it.

src/pretix/presale/templates/pretixpresale/event/checkout_addons.html
```diff
@@ -193,27 +207,27 @@ <h5>
                                                                                    id="cp_{{ form.cartpos.pk }}_variation_{{ item.id }}_{{ var.id }}"
                                                                                    name="cp_{{ form.cartpos.pk }}_variation_{{ item.id }}_{{ var.id }}"
                                                                                    data-exclusive-prefix="cp_{{ form.cartpos.pk }}_variation_{{ item.id }}_"
-                                                                                   title="{% blocktrans with item=item.name var=var.name %}Amount of {{ item }} – {{ var }} to order{% endblocktrans %}">
+                                                                                   aria-label="{% blocktrans with item=item.name var=var %}Add {{ item }}, {{ var }} to basket{% endblocktrans %}">
```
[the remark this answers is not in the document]
wiffbi: Good catch! Changed all occurrences of basket to cart.

src/pretix/presale/templates/pretixpresale/event/checkout_addons.html
```diff
@@ -300,23 +318,25 @@ <h4>
                                                                 <input type="checkbox" value="1"
                                                                        {% if item.initial %}checked="checked"{% endif %}
                                                                        name="cp_{{ form.cartpos.pk }}_item_{{ item.id }}"
-                                                                       id="cp_{{ form.cartpos.pk }}_item_{{ item.id }}">
+                                                                       id="cp_{{ form.cartpos.pk }}_item_{{ item.id }}"
+                                                                       aria-label="{% blocktrans with item=item.name %}Add {{ item }} to basket{% endblocktrans %}"
```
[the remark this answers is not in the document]
wiffbi: changed to cart.

src/pretix/static/pretixpresale/js/ui/cart.js
```diff
@@ -35,16 +35,20 @@ var cart = {
         var now = cart._get_now();
         var diff_minutes = Math.floor(cart._deadline.diff(now) / 1000 / 60);
         var diff_seconds = Math.floor(cart._deadline.diff(now) / 1000 % 60);
+
+        if (diff_minutes < 2 || diff_minutes == 5) $("#cart-deadline").get(0).setAttribute("aria-live", "polite");
+        else $("#cart-deadline").get(0).removeAttribute("aria-live");
+
         if (diff_minutes < 0) {
-            $("#cart-deadline").text(gettext("The items in your cart are no longer reserved for you."));
+            $("#cart-deadline").text(gettext("The items in your cart are no longer reserved for you. You can still complete your order as long as they’re available."));
```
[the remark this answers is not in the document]
wiffbi: This is an error, there should only be one #cart-deadline. I removed the inline, em.sr-only.

src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
wiffbi: I changed the surrounding HTML a bit, too, but added your suggestion.

src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
wiffbi: Done.

src/pretix/presale/templates/pretixpresale/event/index.html
wiffbi: Thanks, fixed!

src/pretix/presale/templates/pretixpresale/event/fragment_subevent_calendar.html
wiffbi: It might be the wording, A11y-tester wanted an „action“ – e.g. „to previous month“. Still waiting for feedback regarding my new wording as an aria-label on the link: „to previous month (April 2021)“.

wiffbi:
### Focus outline

This is fixed.

### Underlined links

This is tough. Due to the icons moste of the time being inserted via HTML there is always a space in between. I found one thing though regarding alternatives to underline: 

- Either bold (but we have other bold text, so not applicable to pretix)
- add an icon – it is not said, whether this has to be always the same icon (link-icon) or if it just has to be an icon.

Going with the icon route, this would mean, one could remove the underline for for „Tickets hinzufügen oder entfernen“ as this has an icon. Maybe it would even be better to just use an icon with an aria-label and move the link directly after the „Warenkorb“.

Removing underlines inside the step-navigation are probably okay, as this is a graphically dedicated navigation.

Regarding „add to calender“ could also be a button – as this usually triggers an action (calendar) and not a page load, semantically a button may even be better suited. What do you think? A small .btn-default maybe?

Speaking of buttons: I think the pure white of a .btn-default is something we might consider a light grey instead to better differentiate a button from an input element. See e.g. githubs light grey buttons.

### Margins and paddings

This is going to be a tough one. `<p>` elements might be nested several levels deep, even surrounded by `<dd>` or `<li>` which makes it hard to impossible to get the selector for this right. I will look into this.

I increased the margin after the thank-you message.

### Order confirmation page

I changed it how you suggested. Order-code is now in the headline.

### Alert backgrounds

I do not thinks it looks that bad. The thing is especially with the white button that it should be slightly tinted. The thing that makes it look off, is that the border (actually a box-shadow!?) is not tinted – instead of a fixed hex-color grey it should be an rgba black with very light alpha (of course, optically best would be to mix light-grey with the background to give it a tint – maybe with SASS? Otherwise rgba will at least be better than what we have now).

I will also look into changing that box into something different. 

alert-legal is fixed.

src/pretix/presale/templates/pretixpresale/event/fragment_subevent_calendar.html
wiffbi: Feedback confirms link, button is just a recommendation and is not a violation any more. I will revert to link.

wiffbi:
Regarding the ticket-download alert, it might be an option to exclude the buttons from the alert-box. To make this look more compact, I added bold and normal text with a line break and set the box to a smaller width. This might be technically not possible, so maybe we just move the buttons outside of the alert-box. I will try, when I have the new icons ready.

Bildschirmfoto 2021-09-27 um 13 19 48

wiffbi:
Regarding Ticket-Download, what if we omit the alert-box altogether and put it on the same level as the status-Info? Like so:
Bildschirmfoto 2021-09-28 um 17 03 58
Bildschirmfoto 2021-09-28 um 17 11 09
Bildschirmfoto 2021-09-28 um 17 10 55

src/pretix/static/pretixbase/scss/_variables.scss
```diff
@@ -40,14 +40,18 @@ $navbar-inverse-link-hover-color: $gray-lighter;
...
+$state-success-bg:               lighten($brand-success, 48%) !default;
+$state-success-border:           lighten($brand-success, 30%);
+$state-success-text:             darken($brand-success, 15%);
+$state-info-bg:                  lighten($brand-info, 33%) !default;
+$state-info-border:              lighten($brand-info, 20%);
+$state-info-text:                darken($brand-info, 20%);
+$state-warning-bg:               lighten($brand-warning, 41%) !default;
+$state-warning-border:           lighten($brand-warning, 20%);
+$state-warning-text:             darken($brand-warning, 25%);
+$state-danger-bg:               lighten($brand-danger, 43%) !default;
```
raphaelm, line 52: Very weirdly, this works for me in control, but not in presale:

src/pretix/static/pretixpresale/scss/_event.scss
```diff
@@ -1,12 +1,22 @@
...
+}
+.with-plus-circle-before:before {
+    display: block;
+    font-family: "FontAwesome";
+    content: "";
+    text-decoration: none;
+    position: absolute;
+    left: -1.2em;
+    top: 0.1em;
+}
```
raphaelm, line 12: Since we dropped underlined links from this PR, we can also revert this hack probably?

In the specific case of the waiting list, I think a button could be fine as well. (Also, wouldn't be including a <span> element in the link and underlining only that possible much simpler to go forward? But we should discuss that in a new issue

src/pretix/presale/templates/pretixpresale/fragment_week_calendar.html
```diff
@@ -4,11 +4,11 @@
         <details class="weekday {% if day.events %}has-events{% else %}no-events{% endif %} {% if day.today %}today{% endif %}"
                 data-date="{{ day.date|date:"SHORT_DATE_FORMAT" }}" open>
             <summary>
-                <h3><span class="fa fa-fw" aria-hidden="true"></span> {{ day.day_formatted }}</h3>
+                <h3><span class="fa fa-fw" aria-hidden="true"></span> <time datetime="{{ day.date|date:"Y-m-d" }}">{{ day.day_formatted }}</time></h3>
```
raphaelm, line 7: Somehow, these headlines are now much larger and even break into multiple lines, which causes a mess

src/pretix/static/pretixpresale/scss/_calendar.scss
```diff
@@ -2,58 +2,80 @@
...
+    &:focus {
+      outline-color: inherit;
     }
 
-    &.soon {
-      opacity: 0.8;
+    &.continued, &.over {
+      background: lighten(#767676, 48%);
+      border-color: lighten(#767676, 30%);
+      border-left-color: #767676;
```
raphaelm, line 39: In a calender full of dozens of unavailable or sold out events, it's now much harder to spot the available ones, at least for me as someone with a very very small degree of red-green color blindness. The old layout was really easy for me to spot.

Maybe we should lighten the border color for the "over" and maybe even "soldout" states even more so that the green ones stick out?

src/pretix/static/pretixpresale/scss/_cart.scss
```diff
@@ -94,23 +95,88 @@
...
 }
 
-.apply-voucher {
-    input {
-        height: 32px;
+.cart [role=rowgroup]:last-child {
+    border-top: 1px solid $table-border-color;
+}
+.cart .total strong {
+    font-size: $font-size-large;
```
raphaelm, line 102: I generally like the larger total, but the font sizes and paddings become a bit messy if you turn "display net prices":

src/pretix/presale/templates/pretixpresale/event/checkout_payment.html
```diff
@@ -13,43 +13,49 @@
         <div class="panel-group" id="payment_accordion">
             {% for p in providers %}
                 <div class="panel panel-default" data-total="{{ p.total|floatformat:2 }}">
-                    <label class="accordion-radio">
+                    <div class="accordion-radio">
```
raphaelm, line 16: To me, it looks like the text is not vertically centered with the radio button (screenshot taken in firefox):



Also, these panel heads are now much higher than other ones (which is fine for me, just something to be aware of)

raphaelm (changes requested):

raphaelm:
Uh, and one more thing: The accordions during checkout (e.g. CustomerStep) now require you to click directly on their label. Previously you could click anywhere on the header, which was much easier to use.

src/pretix/presale/templates/pretixpresale/event/base.html
```diff
@@ -66,19 +70,19 @@
         <div class="{% if not event_logo or not event_logo_image_large %}pull-left flip{% endif %}">
             {% if event_logo and event_logo_image_large %}
                 <a href="{% eventurl event "presale:event.index" cart_namespace=cart_namespace|default_if_none:"" %}"
-                   title="{{ event.name }}">
-                    <img src="{{ event_logo|thumb:'1170x5000' }}" alt="{{ event.name }}" class="event-logo" />
+                   aria-label="{% trans 'Homepage' %}">
```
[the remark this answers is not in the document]
wiffbi: title and aria-label depending on screen reader are both read, which is not ideal. I could probably just use title instead on aria-label, but would stick to `title="Homepage"` as this tells you where the link is taking you.

src/pretix/presale/templates/pretixpresale/event/checkout_base.html
```diff
@@ -12,15 +12,15 @@
     {% trans "Checkout" %}
 {% endblock %}
 {% block content %}
-<aside aria-label="{% trans "Your cart" %}">
+<aside aria-label="{% trans "Your cart." %}">
```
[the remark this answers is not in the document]
wiffbi: Yeah, I will revert this. I think I added forther info to that aside, which I later removed and the "." stayed.

src/pretix/static/bootstrap/scss/bootstrap/_alerts.scss
```diff
@@ -64,6 +64,10 @@
   @include alert-variant($alert-info-bg, $alert-info-border, $alert-info-text);
 }
 
+.alert-legal {
+  @include alert-variant($alert-info-bg, $alert-info-border, $alert-info-text);
+}
```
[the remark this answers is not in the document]
wiffbi: Sorry, I just added this where all the other alerts are defined and did not look what file it was. Of course, I will add somewhere appropriate.

src/pretix/presale/templates/pretixpresale/fragment_week_calendar.html
wiffbi: I will look into it.

src/pretix/static/pretixpresale/scss/_cart.scss
wiffbi: Yes, probably the text „Net total“ should not be that big. Or the „€ 110“ should be big as well. Will try both.

src/pretix/presale/templates/pretixpresale/event/checkout_payment.html
wiffbi: Good catch! On staging this is currently not as bad, but not perfect either – panel heads are 41px instead of 38px high and the centering of the radio buttons is a bit off as well. I will fix this.

wiffbi: Regarding your comment clicking the whole header instead of just the label/text: `<div>` is not allowed inside a `<label>`, so I had to move it. If it is okay, I will add a JavaScript helper for those label-headers. I am not a big fan of using JavaScript for this, but for being A11y compliant we need to be HTML compliant.

wiffbi: Layout is fixed. JavaScript to trigger change-event on radio when clicking .accordion-radio is added.

src/pretix/static/pretixpresale/scss/_cart.scss
wiffbi: I think making „Net total“ as big/small as the other text and only „Total“ bigger is best.

Bildschirmfoto 2021-09-29 um 22 03 46

src/pretix/static/pretixpresale/scss/_calendar.scss
wiffbi: Better? Already pushed.

Bildschirmfoto 2021-09-29 um 22 13 49

src/pretix/presale/templates/pretixpresale/fragment_week_calendar.html
wiffbi: fixed.

src/pretix/static/pretixpresale/scss/_event.scss
wiffbi: Removed that commit

src/pretix/presale/templates/pretixpresale/event/checkout_base.html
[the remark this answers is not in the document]
wiffbi: fixed.

src/pretix/static/bootstrap/scss/bootstrap/_alerts.scss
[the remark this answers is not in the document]
wiffbi: fixed.

src/pretix/static/bootstrap/scss/bootstrap/_panels.scss
```diff
@@ -22,7 +22,6 @@
 .panel-heading {
   padding: $panel-heading-padding;
   border-bottom: 1px solid transparent;
-  @include border-top-radius(($panel-border-radius - 1));
```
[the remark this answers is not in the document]
wiffbi: fixed.

src/pretix/presale/templates/pretixpresale/event/base.html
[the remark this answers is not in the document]
wiffbi: As further research turned out, that screen readers handle this differently whereas some do not read the title-attribute. So if we want this for mouse users, we should include both.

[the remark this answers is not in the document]
wiffbi: I included both with the text „Homepage“ – that way at least in VoiceOver „Homepage“ gets only read once.

src/pretix/static/pretixpresale/scss/_cart.scss
raphaelm: agreed!

src/pretix/static/pretixpresale/scss/_calendar.scss
raphaelm: yes, much better!

src/pretix/presale/templates/pretixpresale/event/checkout_payment.html
raphaelm: amazing :)

raphaelm:
I made a few last changes pre-merge, most notably I reverted the very light colors on the alert boxes – I haven't seen them in any of the previous reviews and I didn't like them, if you think we should still change them let's revisit this in a separate discussion.

PR #2084, opened 2021-05-23
Allow to save invoice addresses and attendee profiles to customer account

raphaelm:
- [x] data model
- [x] Allow to save
- [x] Allow to use
  - [x] advanced matching of questions
  - [x] (multiple) choice questions
  - [x] test with question dependencies
  - [x] unset boolean questions
  - [x] state in address
  - [x] date/time fields
- [x] Allow to show/delete in customer portal
- [ ] Allow to show/delete in backend
- [ ] check a11y
  - [ ] give feedback when auto-fill has been applied (either through an aria-live region if using select.change or on the button somehow?)
  - [ ] improve customer profile page(s) when managing profiles/addresses (e.g. `<dl>`)
- [x] UI: make clear whether you are updating or creating a new record
- [x] post merge: integrate in pretix-gdpr, write test for answer save (append, update of exisiting answers)
- [ ] BUG: Name-field in invoice-address gets overwritten with Customer’s name (if format is full_name) or emptied even if field was filled before, when returning from a checkout-step after the question-step.

### DIFF
--- a/src/pretix/presale/forms/renderers.py
+++ b/src/pretix/presale/forms/renderers.py
@@ -29,7 +29,7 @@ from django.utils.safestring import mark_safe
 from django.utils.translation import pgettext
 
 
-def render_label(content, label_for=None, label_class=None, label_title='', label_id='', optional=False, is_valid=None, attrs = {}):
+def render_label(content, label_for=None, label_class=None, label_title='', label_id='', optional=False, is_valid=None, attrs={}):
     """
     Render a label with content
     """
@@ -149,7 +149,7 @@ class CheckoutFieldRenderer(FieldRenderer):
             label_for=label_for,
             label_class=self.get_label_class(),
             label_id=label_id,
-            attrs = attrs,
+            attrs=attrs,
             optional=not required and not isinstance(self.widget, CheckboxInput),
             is_valid=is_valid
         ) + html
--- a/src/pretix/presale/forms/renderers.py
+++ b/src/pretix/presale/forms/renderers.py
@@ -29,10 +29,11 @@ from django.utils.safestring import mark_safe
 from django.utils.translation import pgettext
 
 
-def render_label(content, label_for=None, label_class=None, label_title='', label_id='', optional=False, is_valid=None, attrs={}):
+def render_label(content, label_for=None, label_class=None, label_title='', label_id='', optional=False, is_valid=None, attrs=None):
     """
     Render a label with content
     """
+    attrs = attrs or {}
     if label_for:
         attrs['for'] = label_for
     if label_class:
--- a/src/pretix/static/pretixpresale/js/ui/questions.js
+++ b/src/pretix/static/pretixpresale/js/ui/questions.js
@@ -117,7 +117,6 @@ function questions_init_photos(el) {
 function questions_init_profiles(el) {
     /*
     TODO:
-    - add dropdown for saved_id to select which profile/address to save to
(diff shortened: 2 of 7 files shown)


wiffbi:
To actually fill the form, it currently uses a button to do so. This is to really force the intent. Alternative would be to fill the form when the select-box changes – for this to work, one needs to add an empty option as otherwise there is no way to fire an onchange-event on select-elements.

Alternative if only one profile exists: only show the button and the description of values to be filled in.

src/pretix/presale/forms/renderers.py
```diff
@@ -29,11 +29,10 @@
 from django.utils.translation import pgettext
 
 
-def render_label(content, label_for=None, label_class=None, label_title='', label_id='', optional=False, is_valid=None):
+def render_label(content, label_for=None, label_class=None, label_title='', label_id='', optional=False, is_valid=None, attrs={}):
     """
     Render a label with content
     """
```
raphaelm, line 35:
```suggestion
def render_label(content, label_for=None, label_class=None, label_title='', label_id='', optional=False, is_valid=None, attrs=None):
    """
    Render a label with content
    """
    attrs = attrs or {}
```

src/pretix/static/pretixpresale/js/ui/questions.js
```diff
@@ -113,3 +113,356 @@ function questions_init_photos(el) {
         })
     });
 }
+
+function questions_init_profiles(el) {
+    /*
+    TODO:
+    – in the original profile, strikethrough which answer will be overwritten, followed by the new answer
+    – add new answers with a + in front
+    */
```
raphaelm, line 122: note from call: add some comments to explain all this

src/pretix/static/pretixpresale/js/ui/questions.js
```diff
@@ -380,26 +380,32 @@ function questions_init_profiles(el) {
...
+        if (matchedProfiles.length === 1) {
             $(".profile-select-control", scope).hide().parent().addClass("form-control-text");
-            $help.html(describeProfileHTML(selectedProfile)).addClass("single-profile-desc").after($button);
+            $help.html(describeProfileHTML(matchedProfiles[0])).addClass("single-profile-desc").after($button);
         }
         else {
+            var i = 0;
             for (p of matchedProfiles) {
-                $select.append("<option>" + labelForProfile(p, matchedProfiles, scope) + "</option>");
+                $select.append($("<option>").text(labelForProfile(p, matchedProfiles, scope)).attr("value", i));
```
wiffbi, line 394: Good catch with the `.text()`. The value-Attribute also makes it more robust instead of using the selectedIndex.

src/pretix/static/pretixpresale/js/ui/questions.js
```diff
@@ -380,26 +380,32 @@ function questions_init_profiles(el) {
...
+                i++;
             }
             $select.change(function() {
                 _updateDescription(this, matchedProfiles[this.selectedIndex], $help);
```
wiffbi, line 398: change to 
`selectedProfile = matchedProfiles[this.value];`
`_updateDescription(this, selectedProfile, $help);`
Then the other changes regarding selectedProfile vs. matchedProfiles are not needed

wiffbi (commented):
Ah, sorry. In my refactor I forgot to add a

`selectedProfile = matchedProfiles[this.selectedIndex];`

to the $select.change callback.

PR #2088, opened 2021-05-24
Event API: Add search parameter

MaicoTimmerman:
The events API is paginated with 50 events. When searching for a specific event within a large Pretix installations UX becomes very bad. Using this filter users are able to quickly find their event by name.

### DIFF
--- a/doc/api/resources/events.rst
+++ b/doc/api/resources/events.rst
@@ -88,6 +88,8 @@ Endpoints
 
     The ``with_availability_for`` parameter has been added.
 
+    The ``search`` query parameter has been added to filter events by their slug, name, or location in any language.
+
 .. http:get:: /api/v1/organizers/(organizer)/events/
 
    Returns a list of all events within a given organizer the authenticated user/token has access to.
@@ -170,6 +172,7 @@ Endpoints
                                  attribute with values of 100 for "tickets available", values less than 100 for "tickets sold out or reserved",
                                  and ``null`` for "status unknown". These values might be served from a cache. This parameter can make the response
                                  slow.
+   :query search: Only return events matching a given search query.
    :param organizer: The ``slug`` field of a valid organizer
    :statuscode 200: no error
    :statuscode 401: Authentication failure
--- a/src/pretix/api/views/event.py
+++ b/src/pretix/api/views/event.py
@@ -55,15 +55,18 @@ from pretix.base.models.event import SubEvent
 from pretix.base.services.quotas import QuotaAvailability
 from pretix.base.settings import SETTINGS_AFFECTING_CSS
 from pretix.helpers.dicts import merge_dicts
+from pretix.helpers.i18n import i18ncomp
 from pretix.presale.style import regenerate_css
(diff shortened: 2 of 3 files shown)


raphaelm (commented):
I think this is good, I'm only a little worried about the parameter name since it is called ``name`` but doesn't match exact names. Maybe ``search`` would be better? That would be consistent with e.g. orders (although we search in multiple fields over there).

MaicoTimmerman:
> I think this is good, I'm only a little worried about the parameter name since it is called `name` but doesn't match exact names. Maybe `search` would be better? That would be consistent with e.g. orders (although we search in multiple fields over there).

I didn't know orders had a search field. I think `search` is indeed a good naming, I'll update that.

The Orders filter is quite extensive:
- event slug
- code
- email
- invoices:
    - number full and short
    - address name
    - address company
- positions:
    - attendee name and email,
    - secret
    - voucher code
- comment
- primary key

And there is a signal that can be used to insert more disjunctional filters. 

The signal is something I'd like to have for events, however they would need have an conjunctive normal form structure (AND of ORs). With a CNF returned from the signal we could also exclude results based on some condition. Perhaps something that returns:
```py
{"all_of": Q(slug__icontains=u), "any_of": Q(name__icontains=u)}
```
which results in a query:
```sql
SELECT * from "event" where slug ilike '%{query}%' AND (name ilike '%{query}%' or <rest of the default query>)
```

How do you feel about this functionality?

In this PR I could expand the `search` parameter to include some more fields. I thought to include:
- name
- event slug
- location
- comment
- subevent
    - names

Any fields missed or that you'd like to see?

raphaelm:
> How do you feel about this functionality?

Honestly, I fear it is more complex than necessary. The order search is nice, but its downside is very poor performance on large databases (since OR can bascially not make use of indexes in PostgreSQL) and we'll need to swap it out for an entirely different approach at some point. However, there is a very clear need for plugin-supplied search functionality in orders: Things like shipping addresses.

Do we have a convincing use case why we need this for events, at all?

> In this PR I could expand the search parameter to include some more fields. I thought to include:

I'd personally only include name, slug and location. Comment can get confusing, meta data values can be better filtered through other means and subevents will make it slow.

MaicoTimmerman:
> > How do you feel about this functionality?
> 
> Honestly, I fear it is more complex than necessary. The order search is nice, but its downside is very poor performance on large databases (since OR can bascially not make use of indexes in PostgreSQL) and we'll need to swap it out for an entirely different approach at some point. However, there is a very clear need for plugin-supplied search functionality in orders: Things like shipping addresses.
> 
> Do we have a convincing use case why we need this for events, at all?

Let's discuss this in a separate issue to prevent scope creep of this PR.

> > In this PR I could expand the search parameter to include some more fields. I thought to include:
> 
> I'd personally only include name, slug and location. Comment can get confusing, meta data values can be better filtered through other means and subevents will make it slow.

I've updated the PR to include name, slug and location and updated the docs to mention 4.1, as 4.0 has just been released.

raphaelm (approved):

PR #2089, opened 2021-05-24
Call plugin's installed() method upon event creation

### LINKED ISSUE #2087: plugin api: add a signal for new event created
I'd like to add a new signal, that fires when a new event is created to set some sane default settings for our plugin. 

My suggestion would be to let `pretix.base.models.event.Event.set_defaults` fire a new event named `pretix.base.signals.event_created` to be in line with the `checkin_created` signal.

Before I start working on this, are you open to new plugin signals?

MaicoTimmerman:
Use the `event.set_active_plugins` to set the plugins to make sure that the installed function gets called for all plugin enable routes.

Note that in this commit all calls are made with `allow_restricted`, as there may be existing restricted plugins enabled for the copied event.  This is fine these locations do not accept user input to select plugins, they come from existing events or from the system configured django settings.

Fixes #2087

### DIFF
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -634,7 +634,7 @@ class Event(EventMixin, LoggedModel):
             Quota,
         )
 
-        self.plugins = other.plugins
+        self.set_active_plugins(other.plugins.split(","), allow_restricted=True)
         self.is_public = other.is_public
         if other.date_admission:
             self.date_admission = self.date_from + (other.date_admission - other.date_from)
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -634,11 +634,11 @@ class Event(EventMixin, LoggedModel):
             Quota,
         )
 
-        # Note: avoid self.set_active_plugins() causes trouble for BadgeApp plugin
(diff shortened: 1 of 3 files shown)


src/pretix/base/models/event.py
```diff
@@ -634,6 +634,11 @@ def copy_data_from(self, other):
             Quota,
         )
 
+        # Note: avoid self.set_active_plugins() causes trouble for BadgeApp plugin
+        #  Plugins can create data in installed() hook based on existing data of the event.
+        #  Calling set_active_plugins() results in defaults being created while actually data
+        #  should come from the copied event. Instead events should use they might be copied
+        #  using the event_copy_data-signal for new events.
```
raphaelm, line 641: Can you… read over this comment again? It looks like three incomplete sentences.

raphaelm (changes requested):
Almost perfect :)

src/pretix/base/models/event.py
raphaelm, line 641:
```suggestion
        #  Note: avoid self.set_active_plugins(), it causes trouble e.g. for the badges plugin.
        #  Plugins can create data in installed() hook based on existing data of the event.
        #  Calling set_active_plugins() results in defaults being created while actually data
        #  should come from the copied event. Instead plugins should use event_copy_data to move
        #  over their data.
```

PR #2099, opened 2021-05-31
Sendmail: Force preview and show number of orders

raphaelm:
Yesterday, a customer accidentally sent thousands of emails instead of a few.

This PR proposes to force mass emails to go through the "preview" step and showing the number of matching orders on the "send" button. It also makes the button red :rotating_light:

### DIFF
--- a/src/pretix/control/forms/__init__.py
+++ b/src/pretix/control/forms/__init__.py
@@ -153,7 +153,7 @@ class CachedFileInput(forms.ClearableFileInput):
 
         @property
         def is_img(self):
-            return any(self.file.filename.lower().endswith(e) for e in ('.jpg', '.jpeg', '.png', '.gif'))
+            return False  # thumbnailing doesn't work since the file isn't available publicly
 
         def __str__(self):
             return self.file.filename
--- a/src/pretix/plugins/sendmail/forms.py
+++ b/src/pretix/plugins/sendmail/forms.py
@@ -177,7 +177,7 @@ class MailForm(FormPlaceholderMixin, forms.Form):
         self.fields['sendto'] = forms.MultipleChoiceField(
             label=_("Send to customers with order status"),
             widget=forms.CheckboxSelectMultiple(
-                attrs={'class': 'scrolling-multiple-choice'}
+                attrs={'class': 'scrolling-multiple-choice no-search'}
             ),
             choices=choices
         )
--- a/src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/send_form.html
+++ b/src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/send_form.html
@@ -1,6 +1,7 @@
 {% extends "pretixcontrol/event/base.html" %}
 {% load i18n %}
 {% load bootstrap3 %}
+{% load humanize %}
 {% block title %}{% trans "Send out emails" %}{% endblock %}
 {% block content %}
     <h1>{% trans "Send out emails" %}</h1>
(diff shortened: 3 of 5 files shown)


wiffbi:
Hm, it seems a bit complicated to me. As the red button initially is disabled, I thought I did something wrong in the form above. It took me a while to realize, I had to click the e-mail preview button  – might be because I use pretix-control in German and the new button is not localized yet or I am just dumb … When I click the preview button, I can then still change the form (e.g. selection of clients) and then directly submit. Not really a problem, but the added security through the extra preview step is circumvented.

As the preview is mandatory (which is a good thing), why not remove the red button completely and make that form a two step form? First step would be to select the recipients, then click button „Next: create email“, where you already will get feedback regarding your selection of clients (e.g. how many emails will be sent – with a button to „Change recipients“, which brings you to step 1). Then you have the inputs for the email and have two buttons: „Show preview“ or „Send 142 emails“ (primary).

Alternatively we keep it the way it is in one big form, but do not disable the big red submit button, but instead notify the user on click, that he first has to use the preview. Also, when I change something on the preview step and do not click preview, we need to block submit and notify the user, he has to use the preview first. Do we already have some notification system in place? Otherwise a simple `return confirm("You need to preview your email first. Do you want to preview now?")` would be enough for a start? But generally that kind of notification-workflow feels a bit clunky to me instead of a straight-forward two step form as described above.

raphaelm:
Hmm, I think doing an actual two-step process makes sense. My only concern would be that long-time users might not see the change and think their email is sent after the first click when it in fact isn't?

wiffbi:
By two step process you mean my first suggestion regarding a two steps form with 1. select recipients, 2. create email? Because then the problem with long-term users would not be a problem AFAICT, as in the first step, they have not supplied any infos regarding subject, content or attachment of the email – so basically have not written an email. So I do not think they will think they have sent an email. And step two is no different than the current buttons, an optional email-preview or send – with the added bonus that we now can tell in the button, how many emails will be sent.

raphaelm:
No, sorry, I misunderstood you then. What you describe would be a three-step process (if someone wants a preview):

1. recipients
2. content
3. preview and send

This is a little much, I was thinking about

1. recipients and content
2. preview and send

So basically forced preview again, but without the edit option on the preview screen

raphaelm:
I pushed a proposal for a two-step process now

wiffbi:
Looks good! Maybe also place the order/recipient count at the fieldset->legend „Recipients“. And on the button instead of „Send (X orders)“ make it say „Send X emails“? Also „orders“ is not plural/singular ready yet.

One thing: when my selection returns 0 recipients, I get the error message but still get shown the preview page. I think this should return the recipient/edit page.

Shall I add my suggestions to the code?

raphaelm:
> Looks good! Maybe also place the order/recipient count at the fieldset->legend „Recipients“. 

Doesn't hurt, although I'm not sure people would even notice it there. It is also a little bit against reading order, it would show output before input.

> And on the button instead of „Send (X orders)“ make it say „Send X emails“? 

No, that would be incorrect. The number of *emails* depends on factors not yet computed when the button is shown. For example, if you send emails to all attendees, an order with five tickets with five different attendee email addresses would cause five emails, while an order with five tickets with all the same attendee email address would cause only one email.

> Also „orders“ is not plural/singular ready yet.

Technically correct. Using a mass-mail feature to send a single mail is such an obscure situation that I'm not sure it's worth the effort, but for other languages with more plural forms it might be.

> One thing: when my selection returns 0 recipients, I get the error message but still get shown the preview page. I think this should return the recipient/edit page.

Good point.

> Shall I add my suggestions to the code?

Feel free, if you find time before me :)

PR #2100, opened 2021-05-31
Add day calendar to organizer page

raphaelm:
- [x] test many different configurations
- [x] test on mobile
- [x] make it pretty
- [x] make it accessible or make it easy for screen reader users to discover the other calendar types
- [x] scroll to current time
- [x] add vertical bar that reflects current-time
- [x] display time above vertical bar for current-time 
    - [x] swap side from right to left if near end of calendar timerange
- [x] add support for rtl-lang

### DIFF
--- a/src/pretix/presale/views/organizer.py
+++ b/src/pretix/presale/views/organizer.py
@@ -939,6 +939,18 @@ class DayCalendarView(OrganizerViewMixin, EventListMixin, TemplateView):
             if not placed_in_row:
                 rows_by_collection[collection].append([e])
 
+        # flatten rows to one stream of events with attribute row
+        # for better keyboard-tab-order in html
+        for collection in rows_by_collection:
+            for i, row in enumerate(rows_by_collection[collection]):
+                concurrency = i+1
+                for e in row:
+                    e["concurrency"] = concurrency
+            rows_by_collection[collection] = {
+                "concurrency": len(rows_by_collection[collection]),
+                "events": sorted([e for row in rows_by_collection[collection] for e in row], key=lambda d: d['time']),
+            }
+
         def sort_key(c):
             collection, row = c
             if collection is None:
--- a/src/pretix/presale/views/organizer.py
+++ b/src/pretix/presale/views/organizer.py
@@ -782,10 +782,7 @@ class DayCalendarView(OrganizerViewMixin, EventListMixin, TemplateView):
 
     def get(self, request, *args, **kwargs):
         self._set_date()
-        self.nonce = base64.b64encode(token_bytes(32)).decode()
-        r = super().get(request, *args, **kwargs)
-        r['Content-Security-Policy'] = f"style-src 'nonce-{self.nonce}'"
-        return r
+        return super().get(request, *args, **kwargs)
 
     def get_context_data(self, **kwargs):
         ctx = super().get_context_data()
@@ -831,7 +828,6 @@ class DayCalendarView(OrganizerViewMixin, EventListMixin, TemplateView):
 
         ctx['events'] = events
 
-        ctx['css_nonce'] = self.nonce
         ctx['collections'] = self._grid_for_template(events)
         ctx['multiple_timezones'] = self._multiple_timezones
         return ctx
--- a/src/pretix/presale/views/organizer.py
+++ b/src/pretix/presale/views/organizer.py
@@ -836,7 +836,9 @@ class DayCalendarView(OrganizerViewMixin, EventListMixin, TemplateView):
 
         ctx['events'] = events
 
-        ctx['collections'] = self._grid_for_template(events)
+        events_by_series = self._grid_for_template(events)
+        ctx['collections'] = events_by_series
+        ctx['no_headlines'] = not any([series for series, events in events_by_series])
         ctx['multiple_timezones'] = self._multiple_timezones
         return ctx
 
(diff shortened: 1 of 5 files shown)


wiffbi:
Looking into a bug with a single event spanning two days: not all info (time as well as the „Book now“) is not shown on the second day. Furthermore in the auto-selected „today/best date“-view this spanning event is not even recognized and the next date with an event is shown.

wiffbi:
Regarding time-format: as I wanted to the colon to blink to indicate live time, I needed to split the time-format into HTML-segments. Due to this, time-format is currently hard-coded as HH:mm in JS. Is this a bad idea?

I know we have the file _rtl.scss, but I added the styles for rtl-support in _calendar.scss so they stay close together. Should I move them to _rtl.scss?

raphaelm:
> Regarding time-format: as I wanted to the colon to blink to indicate live time, I needed to split the time-format into HTML-segments. Due to this, time-format is currently hard-coded as HH:mm in JS. Is this a bad idea?

Unfortunately, yes. US people really insist on their AM/PM stuff.

> I know we have the file _rtl.scss, but I added the styles for rtl-support in _calendar.scss so they stay close together. Should I move them to _rtl.scss?

I think that's fine.

wiffbi:
Okay, then I will use the provided time-format and maybe let the first non alphanumeric char blink. Will fix tomorrow.

wiffbi:
When an event is very long and shown next to a lot of short events the description gets out of view. Maybe we can have the description sticky somehow without it not taking up space – maybe a combination of visibility:hidden and a second description overlayed with position: sticky. This should only apply to overly long events in comparison to the tick-interval.

src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -501,6 +506,80 @@ $(function () {
...
+            if (currentTimeDisplay.get(0).getBoundingClientRect().width + dx >= thisCalendar.scrollWidth) {
+                currentTimeBar.addClass("swap-side");
+            }
+            else {
+                currentTimeBar.removeClass("swap-side");
+            }
+            thisCalendar.style.setProperty('--current-time-offset', dx + "px");
+        }
+        currentTimeInterval = window.setInterval(setCurrentTimeBar, 15*1000);
+        $(window).resize(setCurrentTimeBar);
```
raphaelm, line 579: I think ``$(window).on("resize", setCurrentTimeBar);`` would make this one think less that breaks if we ever upgrade jquery

src/pretix/static/pretixpresale/scss/_calendar.scss
```diff
@@ -113,6 +113,251 @@
...
+*/
+[data-concurrency="1"] {--concurrency: 1}
+[data-concurrency="2"] {--concurrency: 2}
+[data-concurrency="3"] {--concurrency: 3}
+[data-concurrency="4"] {--concurrency: 4}
+[data-concurrency="5"] {--concurrency: 5}
+[data-concurrency="6"] {--concurrency: 6}
+[data-concurrency="7"] {--concurrency: 7}
+[data-concurrency="8"] {--concurrency: 8}
+[data-concurrency="9"] {--concurrency: 9}
```
raphaelm, line 222: what if there are more?

src/pretix/presale/views/organizer.py
```diff
@@ -719,6 +735,331 @@ def _events_by_day(self, before, after):
...
+        raster_size = min(self._get_raster_size(events), tick_duration)
+        events, start, end = self._rasterize_events(events, tick_duration=tick_duration, raster_size=raster_size)
+        calendar_duration = self._get_time_duration(start, end)
+        ctx["calendar_duration"] = self._format_duration(calendar_duration)
+        ctx['time_ticks'] = self._get_time_ticks(start, end, tick_duration)
+        ctx['start'] = datetime.combine(self.date, start)
+        ctx['raster_size'] = raster_size
+        # ctx['end'] = end
+        # size of each grid-column in calendar is based on shortest event duration
+        ctx['shortest_duration_rastered'] = next((d for d in [5, 10, 15, 30, 60, 90, 120, 180] if d >= shortest_duration), 180)
```
raphaelm, line 833: Could we just re-use the tick_duration variable here or do I miss something?

src/pretix/presale/views/organizer.py
```diff
@@ -719,6 +735,331 @@ def _events_by_day(self, before, after):
...
+
+    # currently based on minutes, might be factored into a helper class with a timedelta as raster?
+    def _floor_time(self, t, raster_size=5):
+        minutes = t.hour * 60 + t.minute
+        if minutes % raster_size:
+            minutes = (minutes // raster_size) * raster_size
+            return t.replace(hour=minutes // 60, minute=minutes % 60)
+        return t
+
+    # currently based on minutes, might be factored into a helper class with a timedelta as raster?
```
raphaelm, line 884: I think it's generally more common in Python and less "risky" to put comments about the function **inside** the function, not before them, even if they're not actual "docstrings".

src/pretix/presale/views/organizer.py
wiffbi: I used to have them different. tick_duration was just up to 60 minutes (just increased it recently for fullday-events to only include 3-hour steps) and shortest_duration_rastered basically just sets a class to size the grid-columns accordingly – which you just discovered a bug as I changed from fixed 5-minute raster to a flexible raster to fix the Firefox 1px overflow bug. So I will fix this.

src/pretix/presale/views/organizer.py
wiffbi: Thanks for the hint, I will put them inside.

PR #2101, opened 2021-05-31
Docs: Update Python Version in Nginx Config

fabm3n:
Debian 10 is shipping with Python 3.7 so the Debian 10 based and tested guide should use Python 3.7 in the nginx config.


fabm3n:
Signing a CLA just for updating a number in the docs is a bit too much :D
It would be nice if you can just adjust the Python Version :)

raphaelm:
We've specifically excluded docs from our relicensing. The cla-bot just isn't clever enough yet, sorry :( https://github.com/pretix/pretix/blob/master/LICENSE

fabm3n:
Okay thanks :)

PR #2111, opened 2021-06-07
Sendmail: Allow to disable rules

julia-luna:


src/pretix/plugins/sendmail/models.py
```diff
@@ -201,6 +201,8 @@ class Rule(models.Model):
 
     send_to = models.CharField(max_length=10, choices=SEND_TO_CHOICES, default=CUSTOMERS, verbose_name='Send email to')
 
+    enabled = models.BooleanField(default=True, help_text=_('Only enabled rules are actually sent'))
```
raphaelm, line 204: Please add a `verbose_name` to ensure this is translatable

raphaelm (changes requested):

PR #2114, opened 2021-06-08
Exporters: Fix crash in error url determination

wiffbi:
This should fix error 49419. When exporting orders aysnc, the key „exporter“ is not available in `POST` (as the check-url is called via `GET`). Add exporter-identifier to check-url and make `@cachedproperty exporter` look for it based on `request.method`.

Please note: I could not figure out how to test this fix. Any hint how to force async behaviour for export would be appreciated.


raphaelm:
I think this won't work. I'll gather docs on testing the async stuff for you tomorrow.

PR #2131, opened 2021-06-18
Widget: Use absolute URL for poweredby URL

MaicoTimmerman:
When the `<pretix-widget>` was hosted on a separate domain, the source code URL was pointing to the wrong page. This updates the URL to always point to the Pretix installation.

### DIFF
--- a/src/pretix/base/context.py
+++ b/src/pretix/base/context.py
@@ -30,7 +30,7 @@ from pretix.base.settings import GlobalSettingsObject
 from pretix.base.templatetags.safelink import safelink as sl
 
 
-def get_powered_by(safelink=True):
+def get_powered_by(request, safelink=True):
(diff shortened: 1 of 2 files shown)


MaicoTimmerman:
@raphaelm Normally you respond very quick and you merged my other two recent PRs. Did you perhaps miss this PR or should I clarify anything on this PR?

raphaelm:
No, you did everything right, sorry. I didn't merge it right away, since we *usually* do not use ``request.build_absolute_uri`` but use ``settings.SITE_URL`` etc instead, and I wanted to think/research if there's any downside of using your solution. I don't think there is a downside, but I didn't have the time/energy to actually think about it yet ;)

raphaelm (approved):

PR #2135, opened 2021-06-29
Use upgrade strategy "eager"

buenaventure:
Pip's default strategy is to keep the version of all packages which do not explicitly require an upgrade. This caused issues for me for the second time now in the migration step, because some dependencies were not compatible with the new pretix version, but not explicitly listed as a such. Also I think the "eager" strategy better resembles what happens in the docker container build, as it always installs the newest versions.


raphaelm:
This is a very good point, thank you, I did not know this flag existed.

We don't need a CLA for docs.

PR #2137, opened 2021-06-29
PayPal: Add postfix-field and pass information to custom-field

pc-coholic:
Z#2378438

This PR addresses the following issues:
- Adding a postfix-field in addition to the already existing prefix-field
- Passing the `{prefix}{event}-{code}{postfix}`-construct to PayPal's `transaction.custom`-field for use in settlement reports.
- Rework all transfer reason fields and similar to also include the postfix

Semi-Breaking changes:
- So far, when a prefix was defined, the resulting string would have been 'PREFIX Order for...'. With this change, we are eliminating the whitespaces/moving the responsibility for whitespaces to the the prefix-setup. I'd be willing to ship it like this, or - if requested - to add a migration to add a whitespace to existing set up prefixes.

I'd also be open to remove the prefixes (and by extension postfixes) from all user-facing fields and only have them in the `transaction.custom`-field, as this is the designated field by PayPal for settlement-stuff. If we do this, we should however at least inform the few customers using this on pretix hosted.

### DIFF
--- a/src/pretix/plugins/paypal/payment.py
+++ b/src/pretix/plugins/paypal/payment.py
@@ -145,7 +145,7 @@ class Paypal(BasePaymentProvider):
             ('postfix',
              forms.CharField(
                  label=_('Reference postfix'),
-                 help_text=_('Any value entered here will be added in front of the regular booking reference '
+                 help_text=_('Any value entered here will be added behind the regular booking reference '
                              'containing the order number.'),
                  required=False,
              )),
@@ -296,9 +296,9 @@ class Paypal(BasePaymentProvider):
                             "items": [
                                 {
                                     "name": '{prefix}{orderstring}{postfix}'.format(
-                                        prefix=self.settings.prefix if self.settings.prefix else '',
+                                        prefix='{} '.format(self.settings.prefix) if self.settings.prefix else '',
                                         orderstring=__('Order for %s') % str(request.event),
-                                        postfix=self.settings.postfix if self.settings.postfix else ''
+                                        postfix=' {}'.format(self.settings.postfix) if self.settings.postfix else ''
                                     ),
                                 }
                             ]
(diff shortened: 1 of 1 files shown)


src/pretix/plugins/paypal/payment.py
```diff
@@ -141,7 +141,14 @@ def settings_form_fields(self):
                  help_text=_('Any value entered here will be added in front of the regular booking reference '
                              'containing the order number.'),
                  required=False,
-             ))
+             )),
+            ('postfix',
+             forms.CharField(
+                 label=_('Reference postfix'),
+                 help_text=_('Any value entered here will be added in front of the regular booking reference '
```
raphaelm, line 148: mööp

src/pretix/plugins/paypal/payment.py
```diff
@@ -398,9 +410,13 @@ def _execute_payment(self, payment, request, payment_obj):
...
-                                "name": ('{} '.format(self.settings.prefix) if self.settings.prefix else '') +
-                                __('Order {slug}-{code}').format(
-                                    slug=self.event.slug.upper(), code=payment_obj.order.code
+                                "name": '{prefix}{orderstring}{postfix}'.format(
+                                    prefix=self.settings.prefix if self.settings.prefix else '',
+                                    orderstring=__('Order {slug}-{code}').format(
+                                        slug=self.event.slug.upper(),
+                                        code=payment_obj.order.code
+                                    ),
+                                    postfix=self.settings.postfix if self.settings.postfix else ''
```
raphaelm, line 419: Hmm, that will look weird:

``PREFIXOrder ABC-1234POSTFIX``

The old code added a space after the prefix, I think we should keep that and add it before the postfix and in the other places as well.

raphaelm (changes requested):

src/pretix/plugins/paypal/payment.py
raphaelm: Sorry, apparently I did not read the full PR body, you explain it. I'm not sure what the best way forward is.

A migration would solve it but only caus problems in the future, IIRC our input fields usually strip whitespace so it would be impossible to configure a new prefix with a trailing space.

Is there really a reason that makes us *need* to drop the whitespace?

raphaelm: Weird magical idea: Auto-add whitespace after the prefix / before the postfix if it ends/starts alphanumeric.

i.e. if your prefix is ABC, do "ABC EVENT-ORDER", but if your prefix is ABC-, do "ABC-EVENT-ORDER".

Would probably be the desired thing in all practical situations, but it would be a bit surprising behaviour

pc-coholic: Feeling lazy - omitting the magic for now...

PR #2138, opened 2021-06-29
Email: make responsive and show header image in MS Outlook

wiffbi:
Email-Template simple_logo.html had a mso-specific comment, which basically removed the header image in MS Outlook clients. I also optimized the html header to the settings recommended on [campaignmonitor](https://www.campaignmonitor.com/blog/email-marketing/correct-doctype-to-use-in-html-email/) and [emailonacid](https://www.emailonacid.com/blog/article/email-development/which-code-should-i-include-in-every-email/).

To make the HTML-email more robust and responsive (ie not that small on smartphones when using a full-width header image), I changed the header-image to always have 600px width – for this to work with smaller images, they need to be padded to 600px width. Therefore I added a new scaling operator: `_` for the size-argument of the thumb-filter in django-templates.

min-size-operator for size-argument of thumb-filter works for width and height independently, so the following is possible:

`image|thumb:"100_x100"` resizes the image to a maximum height of 100px (if it is lower, it does not upscale) and makes it at least 100px wide (if the resized image would be less than 100px wide it adds a white background to both sides to make it at least 100px wide).

`image|thumb:"100x100_"` resizes the image to a maximum width of 100px (if it is lower, it does not upscale) and makes it at least 100px high (if the resized image would be less than 100px high it adds a white background to top and bottom to make it at least 100px high).

`image|thumb:"100_x100_"` resizes the image proportionally to either a width or height of 100px – it takes the smaller side and resizes that to 100px, so the longer side will at least be 100px. So the resulting image will at least be 100px wide and at least 100px high. If the original image is bigger than 100x100 then no padding will occur. If the original image is smaller than 100x100, no resize will happen but padding to 100x100 will occur.

### DIFF
--- a/src/pretix/base/services/mail.py
+++ b/src/pretix/base/services/mail.py
@@ -640,14 +640,14 @@ def convert_image_to_cid(image_src, cid_id, verify_ssl=True):
             image_src = normalize_image_url(image_src)
 
             path = urlparse(image_src).path
-            guess_subtype = os.path.splitext(path)[1][1:]
+            image_type = os.path.splitext(path)[1][1:]
 
             response = requests.get(image_src, verify=verify_ssl)
             mime_image = MIMEImage(
-                response.content, _subtype=guess_subtype)
+                response.content, _subtype=image_type)
 
         mime_image.add_header('Content-ID', '<%s>' % cid_id)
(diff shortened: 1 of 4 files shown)


wiffbi:
One still needs to make the image responsive in Outlook, if it exceeds 600px in width.

raphaelm (approved):

PR #2142, opened 2021-07-05
Fix celery progress: assign state/info once to avoid race-conditions

wiffbi:


src/pretix/base/views/tasks.py
```diff
@@ -75,40 +76,40 @@ def _return_ajax_result(self, res, timeout=.5):
...
         data = self._ajax_response_data()
         data.update({
             'async_id': res.id,
             'ready': ready,
             'started': False,
         })
         if ready:
-            if res.successful() and not isinstance(res.info, Exception):
-                smes = self.get_success_message(res.info)
+            if res.successful() and not isinstance(info, Exception):
```
raphaelm, line 87: Shouldn't we stop using .successful() then as well? After all, it's just a

```
    def successful(self):
        """Return :const:`True` if the task executed successfully."""
        return self.state == states.SUCCESS
```

wiffbi: Yes, makes sense. Also that is more in line with the code/if-checks following after that line.

PR #2144, opened 2021-07-05
Widget: remove limit of 50 events in list-view by adding a load-more-button

wiffbi:


src/pretix/presale/views/widget.py
```diff
@@ -537,10 +540,21 @@ def _get_event_list(self, request, **kwargs):
...
                     filter_qs_by_attr(self.request.event.subevents_annotated(self.request.sales_channel.identifier), self.request)
                 )
+                ordering = self.request.event.settings.get('frontpage_subevent_ordering', default='date_ascending', as_type=str)
+                data['has_more_events'] = False
+                if ordering in ("date_ascending", "date_descending"):
+                    # fetch one more result than needed to check if more events exist
+                    evs = evs[offset:offset + limit + 1]
+                    if len(evs) > limit:
+                        data['has_more_events'] = True
+                        evs = evs[:limit]
```
raphaelm, line 556: This will cause more database queries than necessary, because both ``[offset:offset + limit + 1]`` as well as ``[:limit]`` will be executed in SQL.

```suggestion
                    evs = list(evs[offset:offset + limit + 1])
                    if len(evs) > limit:
                        data['has_more_events'] = True
                        evs = evs[:limit]
```
should force early evaluation and fix that

src/pretix/static/pretixpresale/js/widget/widget.js
```diff
@@ -1408,11 +1417,12 @@ var shared_root_methods = {
                 root.name = data.name;
                 root.frontpage_text = data.frontpage_text;
             } else if (data.events !== undefined) {
-                root.events = data.events;
+                root.events = (root.events || []).concat(data.events);
```
raphaelm, line 1420: Can you test if this behaves correctly in the following edge case?

* All list styles are set to "list", no calendars
* You embed the widget for an **organizer**, not an event
* You click on an event series in that list

This should open a list of dates in that event series. I have a hunch that it instead will append the list to the existing one :(

raphaelm (changes requested):

wiffbi:
TODO: during loading of more results at least disable button, maybe even show loading spinner?

src/pretix/presale/views/widget.py
wiffbi: Ah, I thought the `len(evs)` would execute the evs-queryset once and after that it’s a list. But your suggestion is more explicit – and so better!

raphaelm:
yeah, a spinner (we already ship one for loading the entire widget) would be great!

src/pretix/static/pretixpresale/js/widget/widget.js
wiffbi: Thanks, I did not think about the organizer-view! I have a solution ready but it expects the organizer view to not be paginated/not have load-more-functionality – which it currently does not have, but might? Currently working solution is:

`root.events = root.has_more_events && "has_more_events" in data && root.events ? root.events.concat(data.events) : data.events;`

But that looks a bit magic to me. So I would prefer setting a append-option when hitting the load-more button. This should be more robust e.g. when later adding load-more to the organizer list-view as well.

raphaelm (commented):
Almost perfect!

I found one edge case:

* If I have a widget with `style="list"` that shows an organizer overview,
* then select an event series,
* then hit "load" more to get the second page
* then hit "Back" at the top
* then select an event series again,
* I will receive a list starting at page 2 instead of page 1

I think we need to reset `$root.offset` and likely also `root.$append_events` whenever "Back" is pressed.

wiffbi:
Good catch! I added a fix.

raphaelm (approved):

PR #2154, opened 2021-07-20
Always register ApplePay domains in livemode

pc-coholic:
This is a re-re-re-visitation of the curious case of ApplePay not working for some users of pretix with Stripe - even though their domains had been registered through the corresponding API call.

Stripe says regarding this issue:
> The reason you’re getting this error is because, on acct_redacted, you’ve registered your Apple Pay domains with your testmode secret key. In order to correct for this, you’ll need to register your Apple Pay domain using the livemode Platform secret key, along with the Stripe-Account header of the Connected account, i.e. stripe_account: acct_redacted. You can then initialize Stripe.js with the testmode Platform publishable key along with the Stripe-Account header on their webpage, to test out Apple Pay functionality. 

This PR should implement the domain registration as proposed by Stripe.

However, in the past, I've found that even though domains seemed to be registered correctly in live-mode, it still might not work properly because the domain was (contrary to the API-response) actually registered in Testmode.

So I am wondering if we should provide a migration which re-registers the domains, ignoring the database-cache that we've built locally...

### DIFF
--- a/src/pretix/plugins/stripe/tasks.py
+++ b/src/pretix/plugins/stripe/tasks.py
@@ -54,13 +54,28 @@ def stripe_verify_domain(event, domain):
     prov = StripeCC(event)
     account = get_stripe_account_key(prov)
 
+    # Yes, we could just use the **prov.api_kwargs
+    # But since we absolutely need to always issue this call with live keys,
+    # we're building our api_kwargs here by hand.
+    # Only if no live connect secret key is set, we'll fall back to the testmode keys.
+    # But this should never happen except in scenarios where pretix runs in devmode.
+    if prov.settings.connect_client_id and prov.settings.connect_user_id:
+        api_kwargs = {
+            'api_key': prov.settings.connect_secret_key or prov.settings.connect_test_secret_key,
+            'stripe_account': prov.settings.connect_user_id
+        }
+    else:
+        api_kwargs = {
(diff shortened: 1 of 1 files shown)


src/pretix/plugins/stripe/tasks.py
```diff
@@ -54,13 +54,28 @@ def stripe_verify_domain(event, domain):
...
+    # Only if no live connect secret key is set, we'll fall back to the testmode keys.
+    # But this should never happen except in scenarios where pretix runs in devmode.
+    if prov.settings.connect_client_id and prov.settings.connect_user_id:
+        api_kwargs = {
+            'api_key': prov.settings.connect_secret_key or prov.settings.connect_test_secret_key,
+            'stripe_account': prov.settings.connect_user_id
+        }
+    else:
+        api_kwargs = {
+            'api_key': prov.settings.secret_key,
```
raphaelm, line 69: What if we're on a self-hosted system (no Stripe Connect) and the user entered a test secret key? Shouldn't we skip creating the Apple Pay domain entirely in this case?

pc-coholic: As is, it would still create the domain in testmode and not save to our model - which is fine, since it does enable testing of apple pay... At least that's how I understand it...

raphaelm (approved):

PR #2191, opened 2021-09-05
Display subevent location and end time in cart

raphaelm:
This is an attempt to fix two customer complaints:
- End time not being shown in cart (but we'll still only show it for events <24h, otherwise things get confusing)
- Location not being shown in cart / order confirmation

**Before**:

Screenshot 2021-09-05 at 22-08-10 Museum

**End**:

Screenshot 2021-09-05 at 22-42-49 Start only

I'm still a little unsatisfied with how full this gets when there's attendee names and addresses and questions and stuff, but there's probably little room to fix that.

### DIFF
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -156,6 +156,29 @@ class EventMixin:
             return _date(self.date_from.astimezone(tz), "DATE_FORMAT")
         return daterange(self.date_from.astimezone(tz), self.date_to.astimezone(tz))
 
+    def get_time_range_display(self, tz=None, force_show_end=False) -> str:
+        """
+        Returns a formatted string containing the start time and sometimes the end time
+        of the event with respect to the current locale and to the ``show_date_to``
+        setting. Dates are not shown. This is usually used in combination with get_date_range_display
+        """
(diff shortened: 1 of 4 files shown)


wiffbi:
I changed the toggle to a `<button>` for better A11y. This is still complicated as when the location/text is truncated via CSS (text-overflow: ellipsis) it is still read aloud completely (at least with VoiceOver on macOS) – so for screen reader users that button does not make much sense. But for general keyboard-users, the button still needs to be accessible by keyboard. The thing is, this also is not a usual „show more“ button so it only kind of works with aria-controls and aria-expanded. I am still thinking about it, how to best implement this in an accessible way. But for now it works. I anyway need to merge this into the A11y branch and update it there as cart_fragment.html needs some changes regarding table navigation, etc.

raphaelm:
Should we just add aria-hidden to the button then if it doesn'T need to be read?

wiffbi:
I tried, but at least VoiceOver did not ignore the button – probably because it is a button/input. I could remove it from the tab-order but that breaks for all other keyboard-users. So Screen-Reader-Users probably just have to live with it. I still need to manage focus once the button has been pressed and removed. I will look at it today!

PR #2199, opened 2021-09-09
Add config options for max file upload sizes

### LINKED ISSUE #2198: Allow configuring the max upload size for files
### Discussed in https://github.com/pretix/pretix/discussions/2188

<div type='discussions-op-text'>

<sup>Originally posted by **neumantm** September  1, 2021</sup>
I would  like to set it from `pretix.cfg`.
The current hard coded value could be used as the default.</div>

<div type='discussions-op-text'>

<sup>Originally posted by **rphaelm** September  2, 2021</sup>
The 10 MB limit for file upload is in different places throughout the codebase, see e.g. this search:

https://github.com/pretix/pretix/search?q=max_size

neumantm:
Closes #2198

### DIFF
--- a/doc/admin/config.rst
+++ b/doc/admin/config.rst
@@ -444,9 +444,9 @@ You can configure the maximum file size for uploading various files::
     ; Max upload size for images in MiB, defaults to 10 MiB
     max_size_image = 12
     ; Max upload size for favicons in MiB, defaults to 1 MiB
-    max_size_image = 2
+    max_size_favicon = 2
     ; Max upload size for email attachments in MiB, defaults to 10 MiB
-    max_size_image = 15
+    max_size_email_attachment = 15
     ; Max upload size for other files in MiB, defaults to 10 MiB
     ; This includes all file upload type order questions
-    max_size_image = 100
+    max_size_other = 100
--- a/doc/admin/config.rst
+++ b/doc/admin/config.rst
@@ -434,3 +434,19 @@ pretix can make use of some external tools if they are installed. Currently, the
 
 .. _Python documentation: https://docs.python.org/3/library/configparser.html?highlight=configparser#supported-ini-file-structure
 .. _Celery documentation: http://docs.celeryproject.org/en/latest/userguide/configuration.html
+
+Maximum upload file sizes
+-------------------------
+
+You can configure the maximum file size for uploading various files::
+
+    [pretix_file_upload]
+    ; Max upload size for images in MiB, defaults to 10 MiB
+    max_size_image = 12
+    ; Max upload size for favicons in MiB, defaults to 1 MiB
+    max_size_image = 2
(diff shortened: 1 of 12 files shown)


doc/admin/config.rst
```diff
@@ -434,3 +434,19 @@ pretix can make use of some external tools if they are installed. Currently, the
...
+Maximum upload file sizes
+-------------------------
+
+You can configure the maximum file size for uploading various files::
+
+    [pretix_file_upload]
+    ; Max upload size for images in MiB, defaults to 10 MiB
+    max_size_image = 12
+    ; Max upload size for favicons in MiB, defaults to 1 MiB
+    max_size_image = 2
```
wiffbi, line 447:
```suggestion
    max_size_favicon = 2
```

doc/admin/config.rst
```diff
@@ -434,3 +434,19 @@ pretix can make use of some external tools if they are installed. Currently, the
...
+    ; Max upload size for email attachments in MiB, defaults to 10 MiB
+    max_size_image = 15
```
wiffbi, line 449:
```suggestion
    max_size_email_attachment = 15
```

doc/admin/config.rst
```diff
@@ -434,3 +434,19 @@ pretix can make use of some external tools if they are installed. Currently, the
...
+    ; Max upload size for other files in MiB, defaults to 10 MiB
+    ; This includes all file upload type order questions
+    max_size_image = 100
```
wiffbi, line 452:
```suggestion
    max_size_other = 100
```

wiffbi (changes requested):
I think this is a very useful addition to pretix! Thank you!

src/pretix/settings.py
```diff
@@ -845,4 +845,10 @@ def traces_sampler(sampling_context):
 DATA_UPLOAD_MAX_NUMBER_FIELDS = 25000
 DATA_UPLOAD_MAX_MEMORY_SIZE = 10 * 1024 * 1024  # 10 MB
 
+# File sizes are in MB
```
raphaelm, line 848: Thecomment says MB, the new docs say MiB. I always get confused with the two, which is correct here and should we use it consistently? ;)

wiffbi: I guess it is MiB due to the 1024*1024 instead of 1000*1000

doc/admin/config.rst
neumantm: oops

src/pretix/settings.py
neumantm:
```suggestion
# File sizes are in MiB
```

Other occurrences of MiB in this file are also wrong.
But I think as MB and MiB are not that far apart, it is not such a big problem.

raphaelm:
Thank you! :sparkles:

PR #2231, opened 2021-09-30
Track if invoices have been sent via email

raphaelm:


src/pretix/control/templates/pretixcontrol/order/index.html
```diff
@@ -234,7 +234,24 @@ <h3 class="panel-title">
...
+                                        </span>
+                                    {% elif i.sent_to_customer %}
+                                        <span class="fa-stack fa-stack-small" data-toggle="tooltip" title="{% trans "Invoice was emailed to customer" %}">
+                                            <span class="fa fa-background fa-envelope text-muted fa-stack-1x"></span>
+                                            <span class="fa fa-check text-success fa-stack-1x fa-stack-shifted"></span>
+                                        </span>
+                                    {% else %}
+                                        <span class="fa-stack fa-stack-small" data-toggle="tooltip" title="{% trans "Invoice was not yet emailed to customer" %}">
+                                            <span class="fa fa-background fa-envelope text-muted fa-stack-1x"></span>
+                                            <span class="fa fa-ban text-danger fa-stack-1x fa-stack-shifted"></span>
```
wiffbi, line 252: I am missing some kind of pending send state. Right after I send the invoce, I get the success message „Message has been queued … “ but further down under invoices, I get the „not yet emailed“ icon again. The text itself works for both states (waiting in queue to be send as well as user has not added invoice mail to queue). Not sure, if `.text-danger`is the right icon to display here. The danger icon to me suggests it failed to send, which does not reflect neither states this icon is shown.

I see that currently it is not possible to add a pending state. So this icon needs to work for both states. Maybe just the .text-muted/light-gray open envelope-icon for both states? So basically:

```suggestion
                                        <span class="fa fa-envelope-open-o text-muted data-toggle="tooltip" title="{% trans "Invoice was not yet emailed to customer" %}">
```

wiffbi (approved):
Just a minor comment regarding icon, but just IMHO. Other than that, looks good!

Issue #1674, opened 2020-05-11
Clarify E-mail vs Email spelling


ligi:
Forwarding a suggestion I got in QA for an event. I think this makes sense - especially the consistency of E-Mail vs E-Mail address.

![Selection_498](https://user-images.githubusercontent.com/111600/81523965-dd235080-934f-11ea-9a91-b3f91605ae10.png)

raphaelm:
Yes, but we should research and fix this on a system-wide basis, if we go for it

morrme:
Has there been a consensus on the accepted style of the term? I am happy to help with the edits.

trysten:
I found [this post](https://grammarist.com/style/e-mail-email/) rather convincing: it doesn't matter. Either is fine, we just need to make sure we are consistent. I would say e-mail is more formal than email, so I guess we could go with the former.

PR #2236, opened 2021-10-03
fix: change email confirmation label

yoqwerty:


yoqwerty:
Hi @raphaelm , I raised this pull request to improve the copy for the email confirmation as part of the issue mentioned above. Please take a look when u have a moment. Thanks!

Also let me know if I need to send a signed copy of this prefix agreement, as `verification/cla-signed` is failing.

raphaelm:
Hi there, thank you so much for contributing to pretix!

Honestly, I don't actually think "confirm E-Mail" is better than what we had before. Any other thoughts?

 I think I didn't even notice that part of #1674 because I only ever read the text in my notifications and did not see the image :D The *main* issue of #1674 is to decide on "E-mail" vs "E-Mail" vs "Email" and use one consistently across the platform, which is still a valid issue.

yoqwerty:
> Hi there, thank you so much for contributing to pretix!
> 
> Honestly, I don't actually think "confirm E-Mail" is better than what we had before. Any other thoughts?
> 
> I think I didn't even notice that part of #1674 because I only ever read the text in my notifications and did not see the image :D The _main_ issue of #1674 is to decide on "E-mail" vs "E-Mail" vs "Email" and use one consistently across the platform, which is still a valid issue.

As per this label - `E-mail address (repeated)`, I think both are just fine, however, `Confirm E-mail` is more widely used I believe. But I will definitely leave this up to you :)  And as per the consistency of the label across the platform, I can update it to `E-mail` if that works for you. Did a quick look up in the grammarly, looks like `E-mail` and `Email` are both used interchangeably - https://www.grammarly.com/blog/spelling-e-mail-email/

Issue #2210, opened 2021-09-16
[A11y] make reason why add-to-basket button is disabled available to keyboard users


wiffbi:
**Describe the bug**
Currently the add-to-basket button is disabled if no product has a quantity set. The explanation why that button is disabled is only shown on hover of that button. Due to being disabled, keyboard users never reach that button and thus never get any feedback as to why they cannot use that button.

**Expected behavior**
Make explanation available to keyboard users.

wiffbi:
The more I research the subject, it is clear that disabling a button is not ideal regarding accessibility. There is no way for the user to focus the button or otherwise submit the form to get any hint as of why he can’t submit the form. In our case it is rather obvious, but making it explicitly clear through an error-message will be better. Furthermore the current design of the disabled button does not meet the contrast criteria. To meet contrast criteria, the button would either look to similar to an active button or it has to look completely different, which is irritating as well.

TLDR; I would try to have the button work normally. Having the form block onsubmit creates the problem, where to show the message, that one needs to add at least one item to the basket. So blocking the click-event on the button explcitly tells us, that the focus is on the button and we can savely show the message (and maybe have read aloud via aria-live?). This still let’s the form be submitted by hitting enter when inside an form-input, but then the normal error-message mechanism kicks in. I think this an okayish compromise.

Issue #1128, opened 2019-01-09
Labels for inputs on pretixpresale/event/index.html


RaphaelKimmig:
We've had feedback from multiple users that they were confused by the unlabelled input fields for item counts on the ticket page (see e.g. https://pretix.eu/demo/democon/). The issue may be more pronounced with multiple tickets with longer descriptions.

The fact that a user has to enter a number into that field to order a ticket isn't evident and the purpose of the field isn't visible anywhere on the page. Using the current design one could for example misuse the place holder as a label or maybe add an input group add-on to the count input as well.

![Examples with the current design](https://user-images.githubusercontent.com/226708/50910019-411cc880-142d-11e9-8628-1493cb078a43.png)

HDValentin:
I had last year some contact with some confused people about the missing Label. They all did not understand to rise the amount of tickets to get the "put to basket"-button enabled.

I think it could be an accessibility issue for blind users too.

rixx:
I'm not sure a label would help all that much. I've run into this issue on the support side repeatedly, too. When I thought about the problem, I always thought that direct action buttons per input would be nice.
![deepinscreenshot_select-area_20190223213059](https://user-images.githubusercontent.com/2657547/53291370-5e75de00-37b2-11e9-8dab-c6eb2aad99cd.png)
![deepinscreenshot_select-area_20190223213021](https://user-images.githubusercontent.com/2657547/53291371-5e75de00-37b2-11e9-8952-f2e820158c8b.png)

raphaelm:
Hmm, that looks pretty, but I see two downsides with individual actions:

* It gets crowded quickly. For English, "to Cart" is still fine, but a German "in den Warenkorb" already requires more space than we really have for both the input field and the button.

* It changes the semantics quite heavily, as it would no longer be possible to add multiple things to the cart at once. Even if we would keep the "global" button as an option, it would then be very inconvenient that everything is filled with a "1" by default.

I think replacing the placeholder "0" with "Number"/"Anzahl" already helps a little, but maybe we do need a more radical change like this one. Happy for more ideas :)

HDValentin:
It is necessary to have a label for every input field!
Because of accessibility reasons. Every tutorial starts with labels:

https://www.barrierefreies-webdesign.de/knowhow/formulare/label.html
https://webaim.org/techniques/forms/
https://www.w3.org/WAI/tutorials/forms/
https://medium.com/net-magazine/create-simple-accessible-web-forms-535cd253dee2
[...]

PR #2237, opened 2021-10-04
[WIP] Stick add-to-cart button to the bottom of form

wiffbi:
This PR is simply here to test what a sticky add-to-cart button would look like as suggested in #1128. Sticking the button to the bottom of the screen and contain it inside the product-listing form sounds like a possible solution for long forms with lots of products, but IMHO it does not work well – IMHO it actually confuses even more as the add-to-cart button is visible all the time, even if e.g. the first product is sold out. The button isn’t even clickable as it is currently disabled (see #2210). Furthermore it could overlay the next product, so users might think they are done and there are not more products to explore.

Bildschirmfoto 2021-10-04 um 10 56 53

Bildschirmfoto 2021-10-04 um 10 57 10

So generally the sticky-approach has the problem of communicating to the user, that there are/might be additional products/tickets to explore. I think if we really need to make the product-listing/cart-page easier to understand, I think we should explore something different.

### DIFF
--- a/src/pretix/static/pretixpresale/scss/_event.scss
+++ b/src/pretix/static/pretixpresale/scss/_event.scss
@@ -172,6 +172,12 @@ article.item-with-variations:last-child, .product-row:last-child {
 }
 section.front-page {
     margin-top: 30px;
+    position: sticky;
+    bottom: 0;
+    top: 100px;
+    padding: 10px 0;
+    background: rgba(255,255,255,.8);
+    z-index: 10;
 }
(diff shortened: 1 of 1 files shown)


raphaelm:
Yes, I do not think this looks good, this will be more confusing. I was thinking of this in the context of a larger redesign where I think would be possible to make it work if the button is on some kind of "card" that maybe contains other information as well and sticks to the bottom. For now, I'd rather not continue this.

PR #2240, opened 2021-10-06
Allow to hide products that require membership

raphaelm:


src/pretix/base/models/items.py
```diff
@@ -523,6 +523,12 @@ class Item(LoggedModel):
         verbose_name=_('Allowed membership types'),
         blank=True,
     )
+    require_membership_hidden = models.BooleanField(
+        verbose_name=_('Hide without a valid membership'),
+        help_text=_('Do not show this unless the customer is logged in and has a valid membership. This also means this '
+                    'will never visible in e.g. the widget.'),
```
wiffbi, line 529: Basically just the „be“ missing in `'will never visible'`, but I changed it a bit more. Regarding the „e.g.“: this suggests there are other places, where the product will not be shown? If not, I removed the „e.g.“

```suggestion
        help_text=_('Do not show this unless the customer is logged in and has a valid membership. Be aware that '
                    'this means it will never be visible in the widget.'),
```

src/pretix/base/models/items.py
```diff
@@ -802,6 +808,12 @@ class ItemVariation(models.Model):
         verbose_name=_('Membership types'),
         blank=True,
     )
+    require_membership_hidden = models.BooleanField(
+        verbose_name=_('Hide without a valid membership'),
+        help_text=_('Do not show this unless the customer is logged in and has a valid membership. This also means this '
+                    'will never visible in e.g. the widget.'),
```
wiffbi, line 814: See above.

```suggestion
        help_text=_('Do not show this unless the customer is logged in and has a valid membership. Be aware that '
                    'this means it will never be visible in the widget.'),
```

wiffbi (changes requested):
Looks good. Tested even with seating. Just one minor typo (just a „be“ missing), which resulted in me changing the wording completely. ;)

PR #2242, opened 2021-10-07
[WIP] Control: Add drag and drop to sort categories and products

wiffbi:
Sorting categories is straight forward.

Sorting products can only happen inside the constraints of categories. I added multiple `<tbody>` to enable sorting by drag and drop within each category. This automagically adds a strong line between each  `<tbody>`, which is a nice little detail to tell the user the drag boundaries. Maybe we should make it more clear though.

Related to this: the first and last up/down-buttons for reordering should be disabled as they do not do anything. Regarding accessibility I would not necessarily remove these button as drag and drop could be complicated or even impossible for certain users. Visually we probably should group these three buttons together.

- [x] I added a light background `$table-bg-hover` to the currently dragged element `(.sortable-ghost)` so you can better see where you are dragging the element to.
- [x] I added the option `revertOnSpill`, which enables the ESC to abort a drag operation. This also cancels when dragging outside the drop-area, which is a _somewhat bigger_ change to the current behaviour that allows to drag completely outside the drop-area and have the element get placed at the beginning/end of the list.
- [x] turn sorting up/down links into POST forms (as they change stuff)
- [x] disable up on first and down on last element in sort-group
- [x] visual defect on the first line (thick to thin)

### DIFF
--- a/src/tests/control/test_items.py
+++ b/src/tests/control/test_items.py
@@ -94,12 +94,12 @@ class CategoriesTest(ItemFormTest):
         self.assertIn("Entry tickets", doc.select("table > tbody > tr")[0].text)
         self.assertIn("T-Shirts", doc.select("table > tbody > tr")[1].text)
 
-        self.client.get('/control/event/%s/%s/categories/%s/down' % (self.orga1.slug, self.event1.slug, c1.id))
+        self.client.post('/control/event/%s/%s/categories/%s/down' % (self.orga1.slug, self.event1.slug, c1.id))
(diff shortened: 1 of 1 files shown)


wiffbi:
I am not 100% happy with the behaviour `revertOnSpill` introduces regarding dragging outside the drop-area, but am willing to accept this for the win of having ESC-support to cancel a drag-operation. What do you think, @raphaelm?

The other three open issues in the opening comment, I would rather turn into its own PRs. They do improve sort/ordering, but have nothing to do with drag-n-drop. Okay?

PR #2278, opened 2021-11-03
Allow to print event location on invoices

raphaelm:


src/pretix/base/settings.py
```diff
@@ -310,6 +310,17 @@ def unserialize(cls, s):
...
     },
+    'invoice_event_location': {
+        'default': 'False',
+        'type': bool,
+        'form_class': forms.BooleanField,
+        'serializer_class': serializers.BooleanField,
+        'form_kwargs': dict(
+            label=_("Show event location on invoices"),
+            help_text=_("The event location will be shown below the list of products if it is the same for all "
+                        "lines. It will shown on every line if there are different locations.")
```
wiffbi, line 321:
```suggestion
                        "lines. It will be shown on every line if there are different locations.")
```

wiffbi (approved):
Just one missing „be“ in the help-text and a very constructed case when joining the lines of multi-line locations.
Looks good!

PR #2283, opened 2021-11-03
Allow customers to change add-ons on existing orders

raphaelm:
- [x] Render form
- [x] Deal with bundles and add-ons on the same ticket
- [ ] Deal with required questions
- [x] Make sure we enforce all item constraints
  - [x] Deal with payment_term_last
  - [x] Deal with memberships
- [x] Respect change_allow_user_price
- [x] Deal with payment fees? Service fees?
- [x] Review step
- [x] Tests
  - [x] free price


pc-coholic:
- If an order has been manually edited in the backend to include more AddOns as expected, selfservice-rebooking will remove the extra items that had been manually added.
- The selfservice-rebooking is distinguishing between bundled items and AddOns - but it is not exposed anywhere in the backend. Im not sure if we need a "Add as AddOn" and "Add as Bundled Item" in the Admin-view - but perhaps we should add some kind of visual indicator what each item is. That way, an Admin can easier see what they are about to change.

src/pretix/base/services/orders.py
```diff
@@ -1515,6 +1521,190 @@ def split(self, position: OrderPosition):
...
+            # Fetch all quotas. If there are no quotas, this item is not allowed to be sold.
+            quotas = list(item.quotas.filter(subevent=op.subevent)
+                          if variation is None else variation.quotas.filter(subevent=op.subevent))
+            if not quotas:
+                raise OrderError(error_messages['unavailable'])
+
+            if (a['item'], a['variation']) in input_addons[op.id]:
+                raise OrderError(error_messages['addon_duplicate_item'])
+
+            if item.require_voucher or (op.item.hide_without_voucher or (op.variation and op.variation.hide_without_voucher)):
```
wiffbi, line 1603: Is there a reason for the extra `()`? I thought about a special intent, but did not find one. Shouldn’t this do the same?

```suggestion
            if item.require_voucher or op.item.hide_without_voucher or (op.variation and op.variation.hide_without_voucher):
```

src/pretix/presale/templates/pretixpresale/event/order_change_confirm.html
```diff
@@ -0,0 +1,219 @@
...
+        <div class="row-fluid">
+            <div class="panel panel-primary">
+                <div class="panel-heading">
+                    <h3 class="panel-title">
+                        {% trans "Change summary" %}
+                    </h3>
+                </div>
+                <table class="panel-body table table-hover">
+                    {% for op in operations %}
+                        {% if op|classname == "ItemOperation" %}
```
wiffbi, line 29: I see you do string-based class-name comparison to differentiate between change-operations. Usually I’d go with `isinstance`, but due to the string-based nature of params in template-filters, this is probably the best way. Of course it does not test for inheritance, but that is probably not needed.

Or would something like `{% if op is ItemOperation %}` work? I did not get it to work locally, but might be missing something.

src/pretix/static/pretixcontrol/scss/_forms.scss
```diff
@@ -569,6 +569,7 @@ table td > .checkbox input[type="checkbox"] {
     padding: 5px 0;
   }
 }
+
```
wiffbi, line 572: Is that newline needed?
```suggestion
```

src/pretix/static/pretixpresale/scss/_checkout.scss
```diff
@@ -117,7 +124,7 @@
 .profile-pre-select .addon-list {
   margin-bottom: 0;
 }
-.profile-select-container {
+.profile-select-container,  {
```
wiffbi, line 127:
```suggestion
.profile-select-container {
```

wiffbi (approved):
Looks good! I haven’t tested the most obscure setups though - just „normal“ dedicated products, that are bundled, not mixed stuff.

src/pretix/presale/templates/pretixpresale/event/order_change_confirm.html
raphaelm: No, ``is`` checks for object identity, not class hierarchy. Since these operation classes are namedtuples, inheritance is not really relevant.

wiffbi: Thanks you for clarifying!

PR #2284, opened 2021-11-04
Add check to force users to change password

ser8phin:
- [x] move checkbox down
- [x] change warning text
- [x] set variable to false after pw changed
- [x] make variable accessible to set later on
- [x] reroute to original target after pw change
- [x] check whether old and new pw entered are the same
- [x] tests
- [x] fix issues


src/pretix/base/models/auth.py
```diff
@@ -130,6 +132,8 @@ class User(AbstractBaseUser, PermissionsMixin, LoggingMixin):
                                    verbose_name=_('Is site admin'))
     date_joined = models.DateTimeField(auto_now_add=True,
                                        verbose_name=_('Date joined'))
+    needs_password_change = models.BooleanField(default=False,
+                                                verbose_name=_('Needs to change password on login'))
```
raphaelm, line 136: Technically, we don't ask "on login", so maybe

```suggestion
                                                verbose_name=_('Force user to select a new password'))
```

would be more correct? This is really nitpicky, though, and I don't really care about it

src/pretix/control/forms/users.py
```diff
@@ -42,7 +42,7 @@ class UserEditForm(forms.ModelForm):
     error_messages = {
         'duplicate_identifier': _("There already is an account associated with this e-mail address. "
                                   "Please choose a different one."),
-        'pw_mismatch': _("Please enter the same password twice"),
+        'pw_mismatch': _("Please enter the same password twice")
```
raphaelm, line 45: I think we can drop that change ;) The trailing comma is also generally considered a good idea in python, since it makes diffs easier to read. We currently do not enforce it, though.

src/pretix/control/middleware.py
```diff
@@ -69,6 +69,11 @@ class PermissionMiddleware:
         "user.settings.notifications.off",
     )
 
+    EXCEPTIONS_PW = (
```
raphaelm, line 72: Maybe ``EXCEPTIONS_FORCED_PW_CHANGE`` to be more precise about what this does?

src/pretix/control/views/user.py
```diff
@@ -243,7 +244,7 @@ def form_valid(self, form):
         return sup
 
     def get_success_url(self):
-        return reverse('control:user.settings')
+        return self.request.GET.get('next', reverse('control:user.settings'))
```
raphaelm, line 247: To prevent an [unvalidated redirect](https://cheatsheetseries.owasp.org/cheatsheets/Unvalidated_Redirects_and_Forwards_Cheat_Sheet.html), you should validate the URL first:

```suggestion
        if "next" in request.GET and url_has_allowed_host_and_scheme(request.GET.get("next"), allowed_hosts=None):
            return request.GET.get("next")
        return reverse('control:user.settings')
```

You'll also need

```
from django.utils.http import url_has_allowed_host_and_scheme
```

src/tests/control/test_auth.py
```diff
@@ -787,7 +787,8 @@ def test_update_session_activity(self):
         assert self.client.session['pretix_auth_last_used'] > t1
 
     def test_pinned_user_agent(self):
-        self.client.defaults['HTTP_USER_AGENT'] = 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.140 Safari/537.36'
+        self.client.defaults[
+            'HTTP_USER_AGENT'] = 'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/64.0.3282.140 Safari/537.36'
```
raphaelm, line 791: I'm pretty sure this was your editor, not you, but it looks less-readbale for me. If we want to break this line, maybe like this?

```suggestion
        self.client.defaults['HTTP_USER_AGENT'] = (
            'Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) '
            'Chrome/64.0.3282.140 Safari/537.36'
        )
```

src/tests/control/test_auth.py
```diff
@@ -927,3 +928,45 @@ def test_enabled_2fa_setup_enabled(self):
...
+        self.assertIn('/control/login/2fa?next=/control/events/', response['Location'])
+
+        d = TOTPDevice.objects.create(user=self.user, name='test')
+        totp = TOTP(d.bin_key, d.step, d.t0, d.digits, d.drift)
+        totp.time = time.time()
+
+        self.client.post('/control/login/2fa?next=/control/events/'.format(d.pk), {
+            'token': str(totp.token())
+        })
+        response = self.client.post('/control/events/')
```
raphaelm, line 969: Technically, a browser would send a ``GET`` after a redirect, not a ``POST``, so maybe we should adjust the test accordingly just for a little more realism. Not that it makes a difference in this case :)

src/tests/control/test_auth.py
ser8phin: I used 3 lines, but it does look better, and yes, my editor was at fault. It "corrected" it.

src/tests/control/test_auth.py
ser8phin: Appreciate more realism in tests, good idea!

PR #2288, opened 2021-11-07
Change default settings for background color, invoice attachmentes and name scheme

raphaelm:
This PR introduces the non-controversial changes discussed in #2146

### DIFF
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -565,6 +565,8 @@ class Event(EventMixin, LoggedModel):
         self.settings.ticketoutput_pdf__enabled = True
         self.settings.ticketoutput_passbook__enabled = True
         self.settings.event_list_type = 'calendar'
+        self.settings.invoice_email_attachment = True
+        self.settings.name_scheme = 'given_family'
 
     @property
     def social_image(self):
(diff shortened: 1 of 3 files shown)


wiffbi:
Looks good! The only thing that caught my eye, was, that existing events will be updated too – but I think that was intended anyway, so … One minor thing: I think with the light grey background, the layout could benefit from a bit more padding as it looks a bit cramped in its box. But this should probably be a different PR.

Bildschirmfoto 2021-11-08 um 11 40 07

wiffbi (approved):

wiffbi:
> For the background color, the situation is more complicated as doing it only for newly created events would require building custom CSS for every new event forever which would be a significant performance overhead. Therefore this PR makes a compromise by changing the default for "all events which did not touch their design settings on purpose by eg changing their primary color", which I think is fair.

Yes, I think so too!



> > One minor thing: I think with the light grey background, the layout could benefit from a bit more padding as it looks a bit cramped in its box. But this should probably be a different PR.

> Honestly, if it's easily possible we should fix this first before we roll out a change to thousands of events that we don't fully like. I'll try to create a new PR.

If you think it is okay, I could add the necessary changes to the CSS to this PR.

raphaelm:
See #2301

PR #2301, opened 2021-11-09
Increase padding if background color is set

raphaelm:
Open questions:

- Is 25 a good value? Our paddings are usually 15px or 30px, but 30px looked to much.

- Previously, the container width for content was constant for every desktop screen size. Now it will be inconsistent between ticket shops which *could* make it harder to work on future layout stuff. Possible courses of action:
	1. Accept that fact
	2. Increase the padding regardless of background color, making shops on white background narrower as well
	3. Increase container width

I think (iii) is too hard since that would require messing up bootstrap's breakpoints which could have all sorts of unintended consequences. (ii) is interesting since white background is no longer the default or preferred option anyways, however it's a waste of space in some cases and it might look weird depending on header styling. so just (i)?

cc @wiffbi

### DIFF
--- a/src/pretix/static/pretixpresale/scss/_theme.scss
+++ b/src/pretix/static/pretixpresale/scss/_theme.scss
@@ -40,7 +40,6 @@
 }
 
 .page-header.logo-large {
-  margin: 0 -15px;
 
   img {
     width: 100%;
@@ -62,6 +61,11 @@ h2.content-header {
 }
 
 @media (max-width: $screen-xs-max) {
+  .page-header.logo-large {
+    margin-left: -15px;
+    margin-right: -15px;
+    margin-top: 0;
+  }
   h2.content-header small {
     display: block;
   }
@@ -76,8 +80,10 @@ h2.content-header {
     }
   }
   .page-header-links-outside {
-    padding-left: 0;
-    padding-right: 0;
+    @if ($body-bg != #FFFFFF) {
+      padding-left: 0;
+      padding-right: 0;
+    }
   }
   .page-header-links > div.header-part {
     @if ($body-bg != #FFFFFF) {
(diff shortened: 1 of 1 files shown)


wiffbi:
Option iii. is indeed only possible up to the breakpoints, of which the largest is 1200px – to increase padding from 15px to 25 (I thought the same, something along 24px – usually 1.5rem@16px base fontsize), we need 20px more in width, which results in 1190px width of `.container`. The next smaller breakpoint is 992, which we would need `.container` to have a width of 990px. These small margins outside of `.container`would look strange, so option iii is not really that good.

Option ii is actually my favorite because this does not introduce another extra special case to the layout – and even with the extra padding it is wide enough. And currently the header-image-style is anyways not ideal regarding sizing especially the full-width images on white background – the image isn’t visually aligned anywhere – see screenshot. With a background-color it works as the image is full-width with the visual container. On white background there is no visual container.

Bildschirmfoto 2021-11-10 um 12 45 06

Option i. I do not see as a big drawback regarding future layout. The only thing I can think of right now are image-sizes and pre-rendered previews. But I think, the layout should be robust enough to handle these cases. In a far future there might be container queries and it could happen, that layout shifts happen, but then it is in good faith as we decided especially for that component to look different if there isn’t enough space.

Generally I even think we could remove the fixed-width steps and go with a max-width of 1200px instead of width and increased padding on `.container`, add padding to `body` and let the content freely resize up to 1200px width (could even be `75rem` instead). Just for thought.

wiffbi:
As discussed we go with option i. Mainly to maintain compatibilty with a users custom header image, so the full-width header-image is always shown as 1140px wide (note, there currently is a bug that scales it to 1170px, which should not happen).

wiffbi (approved):
As noted via DM pretix-control might need a change in the helptext regarding using the full height of the header image. Other than that, looks good!

PR #2308, opened 2021-11-10
Subevent calendar: Improve heuristic on when to show names

raphaelm:


src/pretix/presale/views/event.py
```diff
@@ -569,9 +569,14 @@ def _subevent_list_context(self):
...
-                len(i) for i in ebd.values() if isinstance(i, list)
-            ) < 2
+            # Hide names of subevents in event series where it is always the same.  No need to show the name of the museum thousands of times
+            # in the calendar. We previously only looked at the current time range for this condition which caused weird side-effects, so we need
+            # an extra query to look at the entire series. For performance reasons, we have a limit on how many different names we look at.
+            context['show_names'] = sum(len(i) for i in ebd.values() if isinstance(i, list)) < 2 or self.request.event.cache.get_or_set(
+                'has_different_subevent_names',
+                lambda: len(set(str(n) for n in self.request.event.subevents.values_list('name', flat=True).annotate(c=Count('*'))[:250])) != 1,
+                timeout=120,
+            )
```
wiffbi, line 579: I tried to understand the code, but probably failed. For me, changing both Boolean parts into its on variables is easier to understand – also to understand, what is actually asked for in the expression. So as far as I understand the code, it counts the number of events available in ebd:

`num_events_total = sum(len(i) for i in ebd.values() if isinstance(i, list))`

But `context['show_names']` is true, so names are only shown when one event is available or – if more than one event is available – their names differ. So when multiple events are available and have the same name, no name is shown?

Regarding the second part: 
`lambda: len(set(str(n) for n in self.request.event.subevents.values_list('name', flat=True).annotate(c=Count('*'))[:250])) != 1,`

What do you annotate the `Count(*)` for? I do not see it used. I only see the values_list-names flattened and converted to `set()`. Also, why do you do this in Python and not on the DB with a `DISTINCT`? Untested, probably totally wrong:
`self.request.event.subevents.order_by('name').distinct('name')`
or maybe event add `count()`?
`self.request.event.subevents.order_by('name').distinct('name').count()`

raphaelm: Good questions :)

The basic idea is "do not show names if all subevents in the event have the same name". This is the use-case of e.g. a museum where the title of the subevent is also the name of the museum and repeating it thousands of times is useless. So we hide the names when all names are the same … 

… *unless* there's *exactly* one subevent in the visible timeframe. In that case there's no repetition of the name, and this solves the use case of someone creating their first subevent for testing and then wondering why the name does not show up. I'm not sure if this part of the logic is good, but it has been in there  before this PR and I didn't have a reason to change it.

The new stuff is the query.

Previously, the code checked whether "all subevents in the visible timeframe have the same name", which was confusing because behaviour was different between months for the same event series. So the goal for this PR was to look at all subevents regardless of them being in the visible timeframe.

My first instinct was ``SELECT DISTINCT name`` as well, but what if we have an event series with millions of subevents with actually unique names? In that case, the query would return a million rows! ``SELECT DISTINCT name … LIMIT 250`` does not work, because it only *looks* at the first 250 rows to check if the names are unique, not at all rows.

What I'm doing with the ``Count(*)`` is forcing Django into a ``SELECT … GROUP BY name LIMIT 250`` which causes the database to stop after 250 **unique** names.

The reason I'm not stopping at 2 names already and use an additional ``set()`` call is to catch names that are identical to the user, but look different on DB level, e.g. ``{"de": "Museum"}`` vs ``{"en": "Museum"}``.

wiffbi: Ah, that makes total sense and is very clever! Thank you for clarifying, now I get it. :)

wiffbi (approved):
Thanks for explaining the code! I tested it locally and first thought it did not work, but found my event series had one event half a year ago that had a slightly different name. And then the 120s timeout kicked in.

PR #2330, opened 2021-11-17
Add central cookie consent mechanism

raphaelm:
- [x] Python API
- [x] Dialog
- [x] refine JS API
- [x] document JS and Python API
- [x] Widget integration
- [x] Adjust our plugins

### DIFF
--- a/doc/development/api/cookieconsent.rst
+++ b/doc/development/api/cookieconsent.rst
@@ -1,0 +1,122 @@
+.. highlight:: python
+   :linenothreshold: 5
+
+.. _`cookieconsent`:
+
+Handling cookie consent
+=======================
+
+pretix includes an optional feature to handle cookie consent explicitly to comply with EU regulations.
+If your plugin sets non-essential cookies or includes a third-party service that does so, you should
+integrate with this feature.
+
+Server-side integration
+-----------------------
+
+First, you need to declare that you are using non-essential cookies by responding to the following
+signal:
+
+.. automodule:: pretix.presale.signals
+   :members: register_cookie_providers
+
+You are expected to return a list of ``CookieProvider`` objects instantiated from the following class:
+
+.. class:: pretix.presale.cookies.CookieProvider
+
+   .. py:attribute:: CookieProvider.identifier
+
+      A short and unique identifier used to distinguish this cookie provider form others (required).
+
+   .. py:attribute:: CookieProvider.provider_name
+
+      A human-readable name of the entity of feature responsible for setting the cookie (required).
+
+   .. py:attribute:: CookieProvider.usage_classes
+
+      A list of enum values from the ``pretix.presale.cookies.UsageClass`` enumeration class, such as
+      ``UsageClass.ANALYTICS``, ``UsageClass.MARKETING``, or ``UsageClass.SOCIAL`` (required).
+
+   .. py:attribute:: CookieProvider.privacy_url
+
+      A link to a privacy policy (optional).
+
+Here is an example of such a receiver:
+
+.. code-block:: python
+
+   @receiver(register_cookie_providers)
+   def recv_cookie_providers(sender, request, **kwargs):
+       return [
+           CookieProvider(
+               identifier='google_analytics',
+               provider_name='Google Analytics',
+               usage_classes=[UsageClass.ANALYTICS],
+           )
+       ]
+
+JavaScript-side integration
+---------------------------
+
+The server-side integration only causes the cookie provider to show up in the cookie dialog. You still
+need to care about actually enforcing the consent state.
+
+You can access the consent state through the ``window.__pretix_cookie_consent`` variable. Whenever the
+value changes, a ``pretix:cookie-consent-updated`` event is fired on the ``document`` object.
+
+The variable will generally have one of the following states:
+
+.. rst-class:: rest-resource-table
+
+===================================================== =====================================================
+State                                                 Interpretation
+===================================================== =====================================================
+``__pretix_cookie_consent === undefined``             Your JavaScript has loaded before the cookie consent
+                                                      script. Wait for the event to be fired, then try again,
+                                                      do not yet set a cookie.
+``__pretix_cookie_consent === null``                  The cookie consent mechanism has not been enabled. This
+                                                      usually means that you can set cookies however you like.
+``__pretix_cookie_consent[identifier] === undefined`` The cookie consent mechanism is loaded, but has no data
+                                                      on your cookie yet, wait for the event to be fired, do not
+                                                      yet set a cookie.
+``__pretix_cookie_consent[identifier] === true``      The user has consented to your cookie.
+``__pretix_cookie_consent[identifier] === false``     The user has actively rejected your cookie.
+===================================================== =====================================================
+
+If you are integrating e.g. a tracking provider with native cookie consent support such
+as Facebook's Pixel, you can integrate it like this:
+
+.. code-block:: javascript
+
+     if (window.__pretix_cookie_consent !== null && !(window.__pretix_cookie_consent || {}).facebook) {
+         fbq('consent', 'revoke');
+     }
+     fbq('init', ...);
+     document.addEventListener('pretix:cookie-consent-updated', function (e) {
+         fbq('consent', window.__pretix_cookie_consent.facebook ? 'grant' : 'revoke')
+     })
+
+If you have a JavaScript function that you only want to load if consent for a specific ``identifier``
+is given, you can wrap it like this:
+
+.. code-block:: javascript
+
+     if (typeof window.__pretix_cookie_consent === "undefined") {
+         // Cookie consent tool has not loaded yet, wait for it to load, then try again
+         document.addEventListener('pretix:cookie-consent-updated', function () {
+             execute_function()
+         })
+     } else if (window.__pretix_cookie_consent === null) {
+         // Cookie consent tool is disabled
+         execute_function()
+     } else {
+         // Cookie tool is available
+         if (window.__pretix_cookie_consent[identifier] === true) {
+             execute_function()
+         } else {
+             document.addEventListener('pretix:cookie-consent-updated', function () {
+                 if (window.__pretix_cookie_consent[identifier] === true) {
+                     execute_function()
+                 }
+             })
+         }
+     }
--- a/doc/development/api/cookieconsent.rst
+++ b/doc/development/api/cookieconsent.rst
@@ -60,8 +60,8 @@ JavaScript-side integration
 The server-side integration only causes the cookie provider to show up in the cookie dialog. You still
 need to care about actually enforcing the consent state.
 
-You can access the consent state through the ``window.__pretix_cookie_consent`` variable. Whenever the
-value changes, a ``pretix:cookie-consent-updated`` event is fired on the ``document`` object.
+You can access the consent state through the ``window.pretix.cookie_consent`` variable. Whenever the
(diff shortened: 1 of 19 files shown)


wiffbi:
Regarding our discussion around pushing onto a global array and how Google Analytics does it with its window.dataLayer, a quick `console.log` reveals they overload the `push-`method – exactly how we guessed.

```
console.log([].push);
function push() {
    [native code]
}
```

vs.

```
console.log(dataLayer.push);
function (){var e;if(0< … return rp()&&n}
```

From an API-point of view I still like the idea of a custom event better. Somehow it feels more _natural_ to `window.addEventListener("pretix:changecookieconsent")` (or `$(window).on("pretix:changecookieconsent")`) instead of pushing a callback-function onto an array that I have to be careful not to overwrite and do not know what arguments the callback expects. With `$.on` I can expect an event-object to be passed to my function.

Regarding loading-order and event-triggering: anyone listening to `pretix:changecookieconsent` would initially need to check if `window.pretix_cookieconsent` is available and check the consent status if it is. This could even be done with an IFEE inside the `$.on("pretix:changecookieconsent", (function update(){…})()());` although this code is not the easiest to understand and I would advise against it.

src/pretix/presale/templates/pretixpresale/fragment_modals.html
```diff
@@ -15,3 +18,95 @@ <h3 id="loadingmodal-label"></h3>
...
+                                        <span class="text-muted">
+                                            {% trans "Functional cookies (e.g. shopping cart, login, payment, language preference) and technical cookies (e.g. security purposes)" %}
+                                        </span>
+                                    </label>
+                                </div>
+                                {% for cp in cookie_providers %}
+                                    <div class="checkbox">
+                                        <label>
+                                            <input type="checkbox" name="{{ cp.identifier }}">
+                                            {{ cp.provider_name }}<br>
```
wiffbi, line 58: Hm, this feels a bit strange. I can allow cookies per provider, but then only allow the cookie for all the usage-classes the provider needs/wants/offers? How do I activate a provider that offers „Functionality needed“ and optionally offers „Analytics“? As a plugin-author I would need to register multiple providers. Also, from a user-perspective I would expect this the other way around: group by usage_class and list the providers per usage_class?

So instead I would loop over all requested usage_classes and do (sorry, psudeo-code-filter `|csv_or_json`):

`<input type="checkbox" name="{{ usage_class }}" value="{{ list_of_providers|csv_or_json }}">`

src/pretix/static/pretixpresale/js/ui/cookieconsent.js
```diff
@@ -0,0 +1,103 @@
...
+        e.initEvent(type, true, true)
+        document.dispatchEvent(e)
+    }
+
+    if ($("#cookie-consent-storage-key").length) {
+        window.__pretix_cookie_consent = {}
+    } else {
+        window.__pretix_cookie_consent = null
+        return
+    }
```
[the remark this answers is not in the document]
wiffbi: The more I think about it, we could even make `window.pretix.cookie_consent` a function. Also read my comment above, instead of keying by provider, key by usage_class (and maybe additionally provider?). That way, if given only one or two param `key`, returns the cookie-status regarding that `key`. Given three params `window.pretix.cookie_consent('marketing', 'facebook', true)`, it not only sets the value but also emits an update-event. That way the cookie-dialog is basically just another consumer of the `document.addEventListener(pretix:cookie_consent_change)`-event. Setting the whole cookie_consent in one go could also work, as the first param then is not a string, but an object.
To avoid firing multiple upate-events in a row, I would run the update-event after setting the value in a 50ms window.setTimeout.

Of course, if a script wants to set a cookie_consent, it needs to wait for the newly added `document.addEventListener("pretix:cookie_consent:ready")`-event to do so.

src/pretix/presale/templates/pretixpresale/fragment_modals.html
wiffbi: One question regarding privacy: if we have e.g. analytics-consent, do we really need to ask for consent again, if we add a different analytics provider? I think this is fair to the user so the user can decide to allow analytics based on where the analytcs-data goes, but to be honest, probably only 0.1% of users will do that.

src/pretix/static/pretixpresale/js/ui/cookieconsent.js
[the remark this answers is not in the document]
wiffbi: Or course, we could also just keep it a simple object and do not care about 3rd-parties setting values inside `__pretix_cookie_consent` and not triggering an update.

src/pretix/static/pretixpresale/js/ui/cookieconsent.js
```diff
@@ -0,0 +1,103 @@
...
+    } else {
+        storage_val = JSON.parse(storage_val)
+        $("#cookie-consent-details input[type=checkbox][name]").each(function () {
+            if (typeof storage_val[$(this).attr("name")] === "undefined") {
+                // A new cookie type has been added that we haven't asked for yet
+                show_dialog = true
+            } else if (storage_val[$(this).attr("name")]) {
+                $(this).prop("checked", true)
+            }
+        })
```
wiffbi, line 43: A very rare edge-case: in an earlier session, I set the consent through the widget, then continue to pretix, the consent is saved to localStorage. Later I get back to the website with the widget embedded, but in the meantime another provider to the cookie consent was added. The widget triggers a new cookie-consent-dialog (due to the new provider) and – as I navigate to pretix – transfers its consent over to pretix through the `#cookie-consent-from-widget`. But this time, as a cookie-consent already exists for pretix from a previous session, but without the newly added provider, I get the cookie dialog that I answered in the widget again.

src/pretix/static/pretixpresale/js/ui/cookieconsent.js
```diff
@@ -0,0 +1,103 @@
...
+    var storage_val = window.localStorage[storage_key]
+    var show_dialog = false
+    if (!storage_val) {
+        show_dialog = true
+        storage_val = {}
+
+        if ($("#cookie-consent-from-widget").length) {
+            var consented = JSON.parse($("#cookie-consent-from-widget").text())
+            $("#cookie-consent-details input[type=checkbox][name]").each(function () {
+                storage_val[$(this).attr("name")] = consented.indexOf($(this).attr("name")) > -1
```
wiffbi, line 28: No need to jquery here – also in a few other places.

```suggestion
                storage_val[this.name] = consented.indexOf(this.name) > -1
```

wiffbi (commented):
I have not yet fully tested it, but commented especially on the JS-API-part and whether cookies-consent needs to be persisted (or requested?) per provider instead of per-usage-class?

src/pretix/static/pretixpresale/js/ui/cookieconsent.js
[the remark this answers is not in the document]
raphaelm: > Also read my comment above, instead of keying by provider, key by usage_class (and maybe additionally provider?).

Our legal counsel says people need to be able to turn off individual providers, so I do not think keying of provider gives us anything, it just makes things harder.

src/pretix/presale/templates/pretixpresale/event/base.html
```diff
@@ -169,6 +169,12 @@ <h1>
     {% if request.event.settings.contact_mail %}
         <li><a href="mailto:{{ request.event.settings.contact_mail }}">{% trans "Contact event organizer" %}</a></li>
     {% endif %}
+    {% if request.event.settings.privacy_url %}
+        <li><a href="{% safelink request.event.settings.privacy_url %}" target="_blank" rel="noopener">{% trans "Privacy policy" %}</a></li>
+    {% endif %}
+    {% if request.event.settings.cookie_consent and cookie_providers %}
+        <li><a href="#" id="cookie-consent-reopen">{% trans "Cookie settings" %}</a></li>
```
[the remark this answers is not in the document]
raphaelm: Is that… even possible to have a button there without screwing up the styling of the footer entirely?

src/pretix/static/pretixpresale/js/ui/cookieconsent.js
raphaelm: Well… yes. But in 90% of cases the widget runs on a different domain and gets an isolated localStorage by modern browsers, so it's even more rare.

src/pretix/presale/templates/pretixpresale/fragment_modals.html
raphaelm: > One question regarding privacy: if we have e.g. analytics-consent, do we really need to ask for consent again, if we add a different analytics provider?

As far as I understand the legal situation, yes.

src/pretix/presale/templates/pretixpresale/event/base.html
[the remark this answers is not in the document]
wiffbi: It just needs to be a button semantically, so 
```suggestion
        <li class="js-only"><button id="cookie-consent-reopen" class="btn btn-link">{% trans "Cookie settings" %}</button></li>
```

doc/development/api/cookieconsent.rst
```diff
@@ -0,0 +1,122 @@
...
+       ]
+
+JavaScript-side integration
+---------------------------
+
+The server-side integration only causes the cookie provider to show up in the cookie dialog. You still
+need to care about actually enforcing the consent state.
+
+You can access the consent state through the ``window.__pretix_cookie_consent`` variable. Whenever the
+value changes, a ``pretix:cookie-consent-updated`` event is fired on the ``document`` object.
```
wiffbi, line 64: Should we name the event slightly different to match common JS-DOM-events? E.g. `pretix:cookie-consent-change`? Or to make it event more matching with _namespacing_ with our global `window.pretix` element replace the `:` with `.` and  call it `pretix.cookie-consent.change`?

wiffbi:
I added click-to-load and overlay-support. click-to-load works by giving a link a data-attribute `data-click-to-load` to which element to load the content from its href-attribute – currently this only works with iframe, but could easily be extended to `fetch` and DOM-nodes. In its current form, it could be implemented with `<a name="name-of-iframe">` but it needs JS to fadeout the overlay to work properly, so the fallback is not to load into the iframe, but on its own page.

Regarding overlay: this is independent from click-to-load and can be used like this: Any element (usually a button or link) with class-attribute `overlay-remove` triggers its `closest("contains-overlay").find("overlay")` to fadeout.

raphaelm (commented):
minor nitpick

doc/development/api/cookieconsent.rst
```diff
@@ -0,0 +1,119 @@
...
+value changes, a ``pretix:cookie-consent:change`` event is fired on the ``document`` object.
+
+The variable will generally have one of the following states:
+
+.. rst-class:: rest-resource-table
+
+===================================================== =====================================================
+State                                                 Interpretation
+===================================================== =====================================================
+``pretix.cookie_consent === undefined``               Your JavaScript has loaded before the cookie consent
```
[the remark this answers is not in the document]
wiffbi: Currently that is true, but it might happen sometime that we define `window.pretix` somewhere else and cookie_consent is loaded later. Should we include this in the docs so that both cases (either `window.pretix` is undefined or even `window.pretix.cookie_consent`) are documented? The JS-code examples already cater for this by using `(window.pretix || {}).cookie_consent`.

```suggestion
``pretix === undefined``                              Your JavaScript has loaded before the cookie consent
``pretix.cookie_consent === undefined``               Your JavaScript has loaded before the cookie consent
```

PR #2340, opened 2021-11-19
Fix: calculate day calendar grid in JS as chrome does not support calc-division in CSS-grid

wiffbi:


src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -512,7 +512,32 @@ $(function () {
...
+        $(".day-calendar").each(function() {
+            var rasterSize = this.getAttribute("data-raster-size");
+            var duration = this.getAttribute("data-duration").split(":");
+            var cols = duration[0]*60/rasterSize + duration[1]/rasterSize;
+
+            $(".day-timeline", this).css("grid-template-columns", "repeat(" + cols + ", minmax(var(--col-min-size, 3em), 1fr))");
+
+            $(".day-timeline > li", this).each(function() {
+                var s = window.getComputedStyle(this);
+                console.log(s.getPropertyValue('grid-column-start'));
```
raphaelm, line 530:
```suggestion
```

PR #2358, opened 2021-12-01
Ensure uniqueness of question identifiers

raphaelm:
This should have already been the case, but validation in the form was broken and DB-level validation was missing.

Apparently, this might break a process for an important enterprise customer :( But I think we should still do it?


src/pretix/base/migrations/0205_auto_20211201_1155.py
```diff
@@ -0,0 +1,28 @@
+# Generated by Django 3.2.4 on 2021-12-01 11:55
+
+from django.db import migrations
+from django.db.models import Count
+
+
+def change_unique_identifiers(apps, schema_editor):
+    # We cannot really know if a position was bundled or an add-on, but we can at least guess
```
pc-coholic, line 8:
```suggestion
```

pc-coholic (approved):

pc-coholic:
> Apparently, this might break a process for an important enterprise customer :( But I think we should still do it?

Yes, I do think this change should be made.

We should probably inform the affected customers ahead of time and let them know of the change so they can deal with the change in due time.

The migration could also output the names of the affected organizers/events/questions to the console so that self-hosted customers can see this change, too.

raphaelm:
I did a lookup on pretix Hosted and found only ~5 events in violation that mostly looked like accidents, but we should double-check them. So it's probably just one Enterprise customer plus maybe 1 or 2 Community users we don't know about. But yes, we should try to announce it.

raphaelm:
Out of courtesy for the affected enterprise customer, we'll wait with this for the February release.

raphaelm:
Oops, we should finally rebase and solve this

PR #2368, opened 2021-12-06
Email: Allow to override backend for custom SMTP connections

MaicoTimmerman:
Whenever we specify a backend in the django settings, there is no
control over the backend used with custom SMTP settings. This change
reuses the e-mail backend configured.

Our SMTP EmailBackend is a custom class, subclassed of the `'django.core.mail.backends.smtp.EmailBackend'`, however with a local fix implemented for a long standing django bug. With the custom backend we are able to use correct use SMTP-relaying with EHLO authentication.

Context:
- https://github.com/django/django/pull/13728
- https://code.djangoproject.com/ticket/6989#no1.

### DIFF
--- a/src/pretix/base/email.py
+++ b/src/pretix/base/email.py
@@ -25,6 +25,7 @@ from datetime import timedelta
 from decimal import Decimal
 from itertools import groupby
 from smtplib import SMTPResponseException
+from typing import TypeVar
 
 import css_inline
 from django.conf import settings
@@ -49,23 +50,23 @@ from pretix.base.templatetags.rich_text import markdown_compile_email
 
 logger = logging.getLogger('pretix.base.email')
 
+T = TypeVar("T", bound=EmailBackend)
 
-class CustomSMTPBackend(EmailBackend):
 
-    def test(self, from_addr):
-        try:
-            self.open()
-            self.connection.ehlo_or_helo_if_needed()
-            (code, resp) = self.connection.mail(from_addr, [])
-            if code != 250:
-                logger.warn('Error testing mail settings, code %d, resp: %s' % (code, resp))
-                raise SMTPResponseException(code, resp)
-            (code, resp) = self.connection.rcpt('testdummy@pretix.eu')
-            if (code != 250) and (code != 251):
-                logger.warn('Error testing mail settings, code %d, resp: %s' % (code, resp))
-                raise SMTPResponseException(code, resp)
-        finally:
-            self.close()
+def test_custom_smtp_backend(backend: T, from_addr: str) -> None:
+    try:
+        backend.open()
+        backend.connection.ehlo_or_helo_if_needed()
+        (code, resp) = backend.connection.mail(from_addr, [])
+        if code != 250:
+            logger.warning('Error testing mail settings, code %d, resp: %s' % (code, resp))
+            raise SMTPResponseException(code, resp)
+        (code, resp) = backend.connection.rcpt('testdummy@pretix.eu')
+        if (code != 250) and (code != 251):
+            logger.warning('Error testing mail settings, code %d, resp: %s' % (code, resp))
+            raise SMTPResponseException(code, resp)
+    finally:
+        backend.close()
 
 
 class BaseHTMLMailRenderer:
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -670,16 +670,17 @@ class Event(EventMixin, LoggedModel):
         Returns an email server connection, either by using the system-wide connection
         or by returning a custom one based on the event's settings.
         """
-        from pretix.base.email import CustomSMTPBackend
 
         if self.settings.smtp_use_custom or force_custom:
-            return CustomSMTPBackend(host=self.settings.smtp_host,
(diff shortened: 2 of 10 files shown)


raphaelm:
note to self: this might actually break something we do for pretix hosted, will need to check. we can still go forward with this, but carefully ;)

MaicoTimmerman:
I just update the failed `isort` CI check, however did think of a potential weird UX. 

If the `settings.EMAIL_BACKEND` is not a SMTP-based backend, the organizer/event can still configure the host, port, etc. However, during the connection test it will fail with:

```
An error occurred while contacting the SMTP server: 'EmailBackend' object has no attribute 'connection'
```
Perhaps we could add a failsafe with fallback to the original django smtp email backend (which `CustomSMTPBackend` was):

```diff
diff --git a/src/pretix/base/models/event.py b/src/pretix/base/models/event.py
index 004968598..5670fe94c 100644
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -46,6 +46,7 @@ from django.conf import settings
 from django.core.exceptions import ValidationError
 from django.core.files.storage import default_storage
 from django.core.mail import get_connection
+from django.core.mail.backends.smtp import EmailBackend
 from django.core.validators import (
     MaxValueValidator, MinValueValidator, RegexValidator,
 )
@@ -58,6 +59,7 @@ from django.utils.crypto import get_random_string
 from django.utils.formats import date_format
 from django.utils.functional import cached_property
 from django.utils.html import format_html
+from django.utils.module_loading import import_string
 from django.utils.timezone import make_aware, now
 from django.utils.translation import gettext, gettext_lazy as _
 from django_scopes import ScopedManager, scopes_disabled
@@ -672,7 +674,11 @@ class Event(EventMixin, LoggedModel):
         """
 
         if self.settings.smtp_use_custom or force_custom:
-            return get_connection(host=self.settings.smtp_host,
+            backend = import_string(settings.EMAIL_BACKEND)
+            if not isinstance(backend, EmailBackend):
+                backend = "django.core.mail.backends.smtp.EmailBackend"
+            return get_connection(backend=backend,
+                                  host=self.settings.smtp_host,
                                   port=self.settings.smtp_port,
                                   username=self.settings.smtp_username,
                                   password=self.settings.smtp_password,

```

raphaelm:
Yes, this is exactly what I'm a little bit worried about. I also don't really like the failsafe, since it feels like "magic". 

To be fair, "using a non-standard email backend" currently requires "running your own settings.py" which is not documented or officially supported by pretix, so people are on their own either way, the question is just how much we want to screw it up for them ;) Maybe we should rather introduce a setting `EMAIL_CUSTOM_SMTP_BACKEND` that is used for credentials and needs to be configured separately?

MaicoTimmerman:
I agree on the "magic", that's why I've not included it in the PR as is.

I think from a consistency point of view it is better to use the same SMTP-backend for both the default mails and custom SMTP settings. It simply removes a potentional point of error from pretix' logic. For supporting the "running your own settings.py" case, I think the before mentioned UX problem will then simply be the risk of running an unsupported setup.

raphaelm:
Hmm, what you call "UX problem" is a significant, newly introduced breakage with no workaround for someone who rolls there own backend, isn't it?

Think e.g. of someone who currently sets `EMAIL_BACKEND = 'django_ses.SESBackend'` which uses the Amazon SES API to send emails. In that case, current pretix will send all emails through SES if no SMTP settings have been set, but will use regular SMTP if an organizer actively configures SMTP settings.

With your change, *all* email will now be sent to SES, which will likely break since sender addresses are not authorized.

In the SES case, your ``isinstance`` proposal might help, yes, but e.g. in our pretix Hosted case our default `EMAIL_BACKEND` **is** a subclass of `EmailBackend`, since it it SMTP-based, but does some special that routing to select different SMTP servers depending on the target MX (to avoid some spam problems). Our backend will thus just ignore custom SMTP settings in some cases and we'd not really have a good way to work around it.

I'm not sure what the problem is with an additional setting, you could then set

```
EMAIL_BACKEND = EMAIL_CUSTOM_SMTP_BACKEND = 'my.custom.fixed.backend'
```

and someone like us or someone using SES could set


```
EMAIL_BACKEND = 'our.custom.routing.backend'
EMAIL_CUSTOM_SMTP_BACKEND = 'django.core.mail.backends.smtp.EmailBackend'
```

MaicoTimmerman:
Alright, thanks for the clarifications. I interpreted your statement about "not documented or officially supported by pretix" little more strictly.

I've updated the patch to include a new `EMAIL_CUSTOM_SMTP_BACKEND` setting and added a test+test setting that validates both branches in `get_mail_backend`.

As you mentioned, custom settings.py is not officially supported. For documentation of this setting, should we add a section somewhere on custom `settings.py` and the availability of `EMAIL_CUSTOM_SMTP_BACKEND`? Currently, the only mention of a custom settings.py in the docs is in docs/admin/scaling.rst.

raphaelm:
I think it's appropriate to keep it undocumented. I just did not want to break anything if I can help it.

raphaelm (approved):

PR #2372, opened 2021-12-07
Add filter for revoked devices

pc-coholic:
Works as intended - but I seem to be hitting a wall when it comes to setting the initial value on the FilterForm, so that the "Active Devices"-filter is set by default.

### DIFF
--- a/src/pretix/control/forms/filter.py
+++ b/src/pretix/control/forms/filter.py
@@ -2018,6 +2018,15 @@ class DeviceFilterForm(FilterForm):
         ],
         required=False,
     )
+    state = forms.ChoiceField(
+        label=_('Device Status'),
+        choices=[
+            ('', _('All Devices')),
+            ('active', _('Active Devices')),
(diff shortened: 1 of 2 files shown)


src/pretix/control/forms/filter.py
```diff
@@ -2018,6 +2018,15 @@ class DeviceFilterForm(FilterForm):
         ],
         required=False,
     )
+    state = forms.ChoiceField(
+        label=_('Device Status'),
+        choices=[
+            ('', _('All Devices')),
+            ('active', _('Active Devices')),
+            ('revoked', _('Revoked Devices'))
```
raphaelm, line 2026:
```suggestion
        label=_('Device status'),
        choices=[
            ('', _('All devices')),
            ('active', _('Active devices')),
            ('revoked', _('Revoked devices'))
```

I'm not sure which is more correct, but this would be more consistent with the rest of the system

raphaelm (approved):

PR #2373, opened 2021-12-10
Organizer: improve calender navigation on organizer page (Z#177488)

wiffbi:
This PR adds several  improvements to the calendar navigation on organizers pages:

- unify GET-parameters to `date` instead of a mix of `week`, `month`, `year` and `date`
- remove calendar-type icons as they do not add distinction. This brings the +iCal button back on the same line most of the time
- move „show upcoming/past events“ links to their corresponding notification texts instead of in the navigation
- on month view, change month/year-selects to a single select
- on week view, change week/year-selects to a single select

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/fragment_calendar_nav.html
+++ b/src/pretix/presale/templates/pretixpresale/fragment_calendar_nav.html
@@ -5,7 +5,6 @@
 <div class="btn-group" role="group">
     <a href="?{% url_replace request "style" "list" "date" "" %}" type="button" 
             class="btn btn-default{% if style == "list" %} active{% endif %}">
-        <span class="fa fa-list hidden-xs" aria-hidden="true"></span>
         {% trans "List" %}
     </a>
     <a href="?{% url_replace request "style" "week" "old" "" "page" "" "date" date|date:"Y-\WW" %}" type="button"
--- a/src/pretix/presale/templates/pretixpresale/fragment_calendar_nav.html
+++ b/src/pretix/presale/templates/pretixpresale/fragment_calendar_nav.html
@@ -7,16 +7,16 @@
             class="btn btn-default{% if style == "list" %} active{% endif %}">
         {% trans "List" %}
     </a>
-    <a href="?{% url_replace request "style" "week" "old" "" "page" "" "date" date|date:"o-\WW" %}" type="button"
+    <a href="?{% if "date" in request.GET %}{% url_replace request "style" "week" "old" "" "page" "" "date" date|date:"o-\WW" %}{% else %}{% url_replace request "style" "week" "old" "" "page" "" %}{% endif %}" type="button"
             class="btn btn-default{% if style == "week" %} active{% endif %}">
         {% trans "Week" %}
     </a>
-    <a href="?{% url_replace request "style" "calendar" "old" "" "page" "" "date"  date|date:"Y-m" %}"
+    <a href="?{% if "date" in request.GET %}{% url_replace request "style" "calendar" "old" "" "page" "" "date" date|date:"Y-m" %}{% else %}{% url_replace request "style" "calendar" "old" "" "page" "" %}{% endif %}"
             type="button"
             class="btn btn-default{% if style == "calendar" %} active{% endif %}">
         {% trans "Month" %}
     </a>
-    <a href="?{% url_replace request "style" "day" "old" "" "page" ""  "date"  date|date:"Y-m-d" %}" type="button"
+    <a href="?{% if "date" in request.GET %}{% url_replace request "style" "day" "old" "" "page" ""  "date" date|date:"Y-m-d" %}{% else %}{% url_replace request "style" "day" "old" "" "page" ""  "date" %}{% endif %}" type="button"
        class="btn btn-default{% if style == "day" %} active{% endif %}">
         {% trans "Day" %}
     </a>
--- a/src/pretix/presale/templates/pretixpresale/fragment_calendar_nav.html
+++ b/src/pretix/presale/templates/pretixpresale/fragment_calendar_nav.html
@@ -7,7 +7,7 @@
             class="btn btn-default{% if style == "list" %} active{% endif %}">
         {% trans "List" %}
     </a>
-    <a href="?{% url_replace request "style" "week" "old" "" "page" "" "date" date|date:"Y-\WW" %}" type="button"
+    <a href="?{% url_replace request "style" "week" "old" "" "page" "" "date" date|date:"o-\WW" %}" type="button"
             class="btn btn-default{% if style == "week" %} active{% endif %}">
         {% trans "Week" %}
     </a>
--- a/src/pretix/presale/templates/pretixpresale/fragment_calendar_nav.html
+++ b/src/pretix/presale/templates/pretixpresale/fragment_calendar_nav.html
@@ -1,0 +1,29 @@
+{% load i18n %}
+{% load eventurl %}
+{% load urlreplace %}
+
+<div class="btn-group" role="group">
+    <a href="?{% url_replace request "style" "list" "date" "" %}" type="button" 
+            class="btn btn-default{% if style == "list" %} active{% endif %}">
+        <span class="fa fa-list hidden-xs" aria-hidden="true"></span>
+        {% trans "List" %}
+    </a>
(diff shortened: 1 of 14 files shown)


src/pretix/presale/templates/pretixpresale/fragment_calendar_nav.html
```diff
@@ -0,0 +1,28 @@
+{% load i18n %}
+{% load eventurl %}
+{% load urlreplace %}
+
+<div class="btn-group" role="group">
+    <a href="?{% url_replace request "style" "list" "date" "" %}" type="button" 
+            class="btn btn-default{% if style == "list" %} active{% endif %}">
+        {% trans "List" %}
+    </a>
+    <a href="?{% url_replace request "style" "week" "old" "" "page" "" "date" date|date:"Y-\WW" %}" type="button"
```
raphaelm, line 10: I think you need to use `o` instead of `Y` here to  handle the year boundary correctly. For example, ``1977-01-02`` is part of the week ``1976-W53``

src/pretix/presale/templates/pretixpresale/organizers/calendar_week.html
```diff
@@ -14,69 +14,42 @@ <h3>{{ date|date:"F Y" }}</h3>
...
+                {% endfor %}
                 </select>
                 <button type="submit" class="js-hidden btn btn-default">
                     {% trans "Go" %}
                 </button>
             </div>
             <div class="col-md-4 hidden-sm hidden-xs text-right flip">
                 {% if has_before %}
-                    <a href="?{% url_replace request "year" before.isocalendar.0 "week" before.isocalendar.1 %}"
+                    <a href="?{% url_replace request "date" before|date:"Y-\WW" %}"
```
raphaelm, line 45: see above

src/pretix/presale/templates/pretixpresale/organizers/calendar_week.html
```diff
@@ -14,69 +14,42 @@ <h3>{{ date|date:"F Y" }}</h3>
...
                        class="btn btn-default">
                         <span class="fa fa-arrow-left" aria-hidden="true"></span>
                         {{ before|date:week_format }}
                     </a>
                 {% endif %}
                 {% if has_after %}
-                    <a href="?{% url_replace request "year" after.isocalendar.0 "week" after.isocalendar.1 %}"
+                    <a href="?{% url_replace request "date" after|date:"Y-\WW" %}"
```
raphaelm, line 52: see above

src/pretix/presale/templates/pretixpresale/organizers/calendar_week.html
```diff
@@ -86,16 +59,16 @@ <h3>{{ date|date:"F Y" }}</h3>
         </div>
     </form>
     {% include "pretixpresale/fragment_week_calendar.html" with show_avail=request.organizer.settings.event_list_availability %}
-    <div class="col-sm-4 visible-xs text-center">
+    <div class="col-sm-12 visible-sm visible-xs text-center">
         {% if has_before %}
-            <a href="?{% url_replace request "year" before.isocalendar.0 "week" before.isocalendar.1 %}"
+            <a href="?{% url_replace request "date" before|date:"Y-\WW" %}"
```
raphaelm, line 64: see above

src/pretix/presale/templates/pretixpresale/organizers/calendar_week.html
```diff
@@ -86,16 +59,16 @@ <h3>{{ date|date:"F Y" }}</h3>
...
                class="btn btn-default">
                 <span class="fa fa-arrow-left" aria-hidden="true"></span>
                 {{ before|date:week_format }}
             </a>
         {% endif %}
         {% if has_after %}
-            <a href="?{% url_replace request "year" after.isocalendar.0 "week" after.isocalendar.1 %}"
+            <a href="?{% url_replace request "date" after|date:"Y-\WW" %}"
```
raphaelm, line 71: see above

src/pretix/presale/templates/pretixpresale/organizers/index.html
```diff
@@ -147,11 +139,11 @@ <h3>{% trans "Upcoming events" %}</h3>
         {% empty %}
             {% if "old" in request.GET %}
                 <div class="col-md-12">
-                    <em>{% trans "No archived events found." %}</em>
+                    <p><em>{% trans "No archived events found." %} <a href="?{% url_replace request "old" "" %}">{% trans "Show upcoming" %}</a></em></p>
                 </div>
             {% else %}
                 <div class="col-md-12">
-                    <em>{% trans "No public upcoming events found." %}</em>
+                    <p><em>{% trans "No public upcoming events found." %} <a href="?{% url_replace request "old" "1" %}">{% trans "Show past events" %}</a></em></p>
```
raphaelm, line 123: This change would mean we no longer allow finding past events if there is any upcoming event? I don't think that's a reasonable change without discussion :)

src/pretix/presale/views/organizer.py
```diff
@@ -231,6 +232,14 @@ def _set_month_year(self):
             except ValueError:
                 self.year = now().year
                 self.month = now().month
+        elif 'date' in self.request.GET:
+            self.tz = self.request.organizer.timezone
```
raphaelm, line 236: line 225 only sets ``tz``, this sets ``self.tz``, line 229 sets neither. So ``self.tz`` will exist or not exist based on what kind of GET parameters we send? This feels dangerous when we rely on it elsewhere in the code.

src/pretix/presale/views/organizer.py
```diff
@@ -289,14 +298,22 @@ def _set_week_year(self):
...
+            self.week = self.subevent.date_from.astimezone(tz).isocalendar()[1]
         if 'year' in self.request.GET and 'week' in self.request.GET:
             try:
                 self.year = int(self.request.GET.get('year'))
                 self.week = int(self.request.GET.get('week'))
             except ValueError:
                 self.year = now().isocalendar()[0]
                 self.week = now().isocalendar()[1]
+        elif 'date' in self.request.GET:
+            self.tz = self.request.organizer.timezone
```
raphaelm, line 310: see above

src/pretix/presale/templates/pretixpresale/fragment_calendar_nav.html
wiffbi: Good catch! Thank you!

src/pretix/presale/templates/pretixpresale/organizers/index.html
wiffbi: You are right, I have not realized this! I moved them outside the empy-part and unified the layout with the other views regarding text-placement. Or do you want this links to be at the top?

Bildschirmfoto 2021-12-13 um 10 22 28
Bildschirmfoto 2021-12-13 um 10 22 52

src/pretix/presale/views/organizer.py
wiffbi: Not sure where this came from, will set to `tz`. Thanks!

wiffbi: Ah, I copied it from `DayCalendarView` where it is used lateron.

raphaelm:
Looks mostly good, one little thing bugs me. If I go to ``?style=calendar`` it will show me the current month. If I then click on "day", it will take me to ``?style=day&date=2021-12-01`` instead of ``?style=day`` which means I will neither see "today" nor "next event", but "first day of month"

wiffbi:
Ah yes, thanks for reminding me. I stumbled upon this and wanted to fix it. The problem is the `url_replace` filter, which does not only replace, but also adds if not available. Ideally we have a filter, that only replaces if available. Because previously if I navigated to a certain month and switched to week or day view, I was thrown to the current date. Also not ideal. So what about I add a new filter `url_replace_only` or to make more explicit disticntion to the current filter `url_replace_dont_add`?

Also, what is the reason there is this `replace_url` filter in the first place? Could there be some session-ID in the GET-URL? Or would it even make sense to dump the GET-params and make a _readable_ URL? The question here is then what URL-keyword do we reserve for this? Hm, maybe stick with GET-params.

wiffbi:
As discussed I did not add a new filter but added an `if` to only replace date-GET-param if it is set in GET.

raphaelm (approved):

PR #2380, opened 2021-12-14
Reduce confusion about customer login with event level domains

raphaelm:
## Problem

Customer account login is an organizer-level thing. Specifically, we offer the login and profile management views as part of the organizer URLs, e.g. ``https://pretix.eu/org/account/login``. However, one can also log in or sign up during checkout. This is all fun while both organizer URLs as well as event URLs are on the same domain, e.g. ``https://pretix.eu/org/`` and ``https://pretix.eu/org/event/`` or ``https://tickets.org.com/`` and ``https://tickets.org.com/event/``.

However, once you use event-specific domains, this becomes incredibly confusing. If you visit ``https://tickets.eventdomain.com`` and click on the "login" button, you will be taken to ``https://tickets.org.com/account/login`` and after you logged in, you will see your profile on ``https://tickets.org.com/account/profile``. If you manually return back to the event, it will still not show you as login, since there's no way to do cookies across entirely different domains.

Similarly, if you start at the event domain and sign in during a purchase in the ``CustomerStep``, you will be logged in on the event-specific domain, but if you click on your name in the navigation to access your profile… you will need to login again, since you're on a different domain.

Also it's pretty unclear what "logout" means with these different levels of logins.

Currently, this is just annoying, but it will become a major blocker if we (or plugins) start adding event-level views that require a valid login.

## Solution

This PR adds a mechanism that allows cross-domain authentication. If you click "login" on the event domain, you will be redirected to the organizer-level login (this is unchanged), but after you logged in, you will now be redirected **back** to the place you came from. This redirect will also pass back a one-time authorization token. (Since all domains are handled by the same server we can do SSO much simpler than with OAuth/OIDC/…)

This token now creates a "child session": Your django session of the event domain now includes a reference to your django session on the organizer domain. On every page load, we check both sessions for validity. This has a major advantage over just logging you in on the second domain as well: If you hit logout on organizer level, you'll instantly be logged out from *all* event domains you ever logged in to with this mechanism, which is an important property.

Additionally, if you created a cart while logged in, we now store a flag in the cart session that makes sure the cart becomes inaccessible if you log out. (We cannot do this unconditionally as this would break the "sign up for a new account while purchasing" user journey.)

There's still one "unclean" situation: You can still log in on the event domain directly by logging in during checkout. Everything would be much cleaner if we'd handle this via the organizer domain as well, but that would require a redirect to a different domain in the middle of your checkout process which is really bad for the ticket purchasing user journey. So for now, we accept that there is still some situation where you could e.g. be logged in to two different accounts (or to the same account with two different sessions) on different domains. We try to reduce the risk of abandoned sessions as much as possible by introducing a redirect in the logout view to make sure you log out both on the domain you clicked logout on as well as the organizer's main domain.

### DIFF
--- a/src/pretix/presale/views/customer.py
+++ b/src/pretix/presale/views/customer.py
@@ -19,14 +19,17 @@
 # You should have received a copy of the GNU Affero General Public License along with this program.  If not, see
 # <https://www.gnu.org/licenses/>.
 #
-from urllib.parse import quote
+from urllib.parse import quote, urlsplit, urlparse, urlencode, urlunparse, parse_qs
 
+from django.conf import settings
 from django.contrib import messages
+from django.core.exceptions import ImproperlyConfigured
 from django.core.signing import BadSignature, dumps, loads
 from django.db import transaction
 from django.db.models import Count, IntegerField, OuterRef, Q, Subquery
 from django.http import Http404, HttpResponseRedirect
 from django.shortcuts import get_object_or_404, redirect
+from django.utils.crypto import get_random_string
 from django.utils.decorators import method_decorator
 from django.utils.functional import cached_property
 from django.utils.http import url_has_allowed_host_and_scheme
@@ -35,9 +38,11 @@ from django.views.decorators.cache import never_cache
 from django.views.decorators.csrf import csrf_protect
 from django.views.decorators.debug import sensitive_post_parameters
 from django.views.generic import DeleteView, FormView, ListView, View
+from django_redis import get_redis_connection
 
 from pretix.base.models import Customer, InvoiceAddress, Order, OrderPosition
 from pretix.base.services.mail import mail
+from pretix.multidomain.models import KnownDomain
 from pretix.multidomain.urlreverse import build_absolute_uri, eventreverse
 from pretix.presale.forms.customer import (
     AuthenticationForm, ChangeInfoForm, ChangePasswordForm, RegistrationForm,
@@ -57,9 +62,14 @@ class RedirectBackMixin:
             self.redirect_field_name,
             self.request.GET.get(self.redirect_field_name, '')
         )
+        hosts = list(KnownDomain.objects.filter(event__organizer=self.request.organizer).values_list('domainname', flat=True))
+        siteurlsplit = urlsplit(settings.SITE_URL)
+        if siteurlsplit.port and siteurlsplit.port not in (80, 443):
+            hosts = ['%s:%d' % (h, siteurlsplit.port) for h in hosts]
+
         url_is_safe = url_has_allowed_host_and_scheme(
             url=redirect_to,
-            allowed_hosts=None,
+            allowed_hosts=hosts,
             require_https=self.request.is_secure(),
         )
         return redirect_to if url_is_safe else ''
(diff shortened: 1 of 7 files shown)


wiffbi:
That setup regarding login only works for one event-domain at a time, right? Due to only the redirect having a one-time token. What happens if I login on the organizers page, then browse the events, which are all hosted on different domains? Can/should we add a one-time-token step inbetween before redirecting to the event’s website? Hm. Or at least we could add some JavaScript to check the login-status via a cross-domain request to the organizers page and do a redirect-step to initialize the one-time-token? This only works with JS enabled, but better than nothing?

Another problem: I see that redirecting to the ogranizers login-domain during checkout might not be ideal for user-flow. A problem this introduces though is password-manager auto-fill, which AFAIK are based on the domain. If it’s on a subdomain it might work, but if it is on a totally different domain the user would need to look through her/his username/password-store. I do not see a solution for this though other than taking the login-step completely outside the checkout-flow to login on the organizer-page. :(

Another idea: as we use a common button during checkout-login to continue, we cannot use the button’s formaction-attribute to redirect the login-action to the organizer page. During checkout we do the login on event level. What if after login on event-level we do a redirect with a one-time-token to the organizer page, which sets the cookie and redirects back to the event? Basically the other way around you described?

raphaelm:
First of all: Please keep in mind that this is a huge edge case that I don't think is very important. It should not be *broken* like it currently is, but it also does not need to be *perfect* at all cost. Event domains are rarely used, and if they are used, they are typically used by organizers who view their events very isolated from each other. Thus, customer accounts often are not a relevant feature for them.

> That setup regarding login only works for one event-domain at a time, right? 

Yes.

> Due to only the redirect having a one-time token. What happens if I login on the organizers page, then browse the events, which are all hosted on different domains? 

Yes, on each of the event pages I will be considered logged out. But I only need to click "Login" once and it will work instantly, I will not need to re-add my password. I think that's not only pretty acceptible, it also matches how users generally know SSO to work, i.e.  "login with facebook" is also a one-click experience and not a zero-click experience.

> Can/should we add a one-time-token step inbetween before redirecting to the event’s website? 

No, too dangerous. Think e.g. we would change all links in the event calendar to include your one time token on the event page. Not only would we need to create thousands of somewhat costly one-time-tokens that are never used, but there's also the risk of you hitting "right click > copy link" and sharing that secret elsewhere. (There are other approaches with redirects but they'd also break "copy link" in other ways.)

> Hm. Or at least we could add some JavaScript to check the login-status via a cross-domain request to the organizers page and do a redirect-step to initialize the one-time-token? This only works with JS enabled, but better than nothing?

Too dangerous as well, I think. This would require quote open CORS settings including ``Access-Control-Allow-Credentials``. It probably can be done in a safe way with origin checks, but honestly, I rather would not take this risk.

Additionally, any way to do cross-domain authentication checks in the background is probably impossible now or soon in all current browsers since any functionality that would allow this could also be used for tracking.

> Another problem: I see that redirecting to the ogranizers login-domain during checkout might not be ideal for user-flow. A problem this introduces though is password-manager auto-fill, which AFAIK are based on the domain.

Good point, but I think we can accept that as a downside of per-event domains. (Note that it only applies if you log in during checkout, the "manual" login form will always be rendered on the organizer domain.)

> If it’s on a subdomain it might work, but if it is on a totally different domain the user would need to look through her/his username/password-store. I do not see a solution for this though other than taking the login-step completely outside the checkout-flow to login on the organizer-page. :(

Yeah, and I don't think that's worth it (for now).

> Another idea: as we use a common button during checkout-login to continue, we cannot use the button’s formaction-attribute to redirect the login-action to the organizer page. During checkout we do the login on event level. What if after login on event-level we do a redirect with a one-time-token to the organizer page, which sets the cookie and redirects back to the event?

Interesting idea! This could work. It would simplify the session handling. At the same time it would make the checkout process more complex and introduce additional error scenarios there. However, it might cause problems with the widget. For example, it would be a pretty horrible breaking change for everyone who uses our widget and sets their Content-Security-Policy to ``frame-src https://event-domain.com`` because the checkout would just break once it starts redirecting to the organizer domain. Not sure if that risk is worth it, what do you think?

wiffbi:
Thank you for your thorough explanations – I totally agree with them. Regarding widget-redirect and frame-src CSP – this is a valid point! This will be harder to test for organizers as this step happens late in the checkout process, so it is indeed risky. As it does not add much value to 99% of use-cases (one event with checkout per session) I probably would not risk it.

src/pretix/presale/views/customer.py
```diff
@@ -96,11 +108,23 @@ def get_form_kwargs(self):
...
-        if getattr(self.request, 'event_domain', False):
-            default_url = '/'
-        else:
-            default_url = eventreverse(self.request.organizer, 'presale:organizer.customer.profile', kwargs={})
-        return url or default_url
+
+        if not url:
+            return eventreverse(self.request.organizer, 'presale:organizer.customer.profile', kwargs={})
+
+        if self.request.GET.get("request_cross_domain_customer_auth") == "true":
```
wiffbi, line 115: As discussed we should ignore `request_cross_domain_customer_auth` when it is not on a custom organizer/event domain.

wiffbi (changes requested):
Other than my comment regarding `request_cross_domain_customer_auth`it looks good to me! :)

PR #2381, opened 2021-12-14
Add approval requirement option to product variations

ser8phin:
- [x] Extend variation model
- [x] Extend variation backend view/templates
- [x] Adjust checkout process for variation approval
- [x] Test whether approval requirements are applied
- [x] Test if approval is propagated to order
- [x] API support for the new require_approval field


src/pretix/base/models/items.py
```diff
@@ -799,6 +801,13 @@ class ItemVariation(models.Model):
         help_text=_('If set, this will be displayed next to the current price to show that the current price is a '
                     'discounted one. This is just a cosmetic setting and will not actually impact pricing.')
     )
+    require_approval = models.BooleanField(
+        verbose_name=_('Buying this product requires approval'),
+        default=False,
+        help_text=_('If this product is part of an order, the order will be put into an "approval" state and '
+                    'will need to be confirmed by you before it can be paid and completed. You can use this e.g. for '
+                    'discounted tickets that are only available to specific groups.'),
+    )
```
raphaelm, line 810: The new field should also be exposed in the REST API, which requires you to change the serializer (ItemVariationSerializer) as well as the test cases and docs for the applicable API endpoints

src/pretix/base/models/items.py
```diff
@@ -799,6 +801,13 @@ class ItemVariation(models.Model):
         help_text=_('If set, this will be displayed next to the current price to show that the current price is a '
                     'discounted one. This is just a cosmetic setting and will not actually impact pricing.')
     )
+    require_approval = models.BooleanField(
+        verbose_name=_('Buying this product requires approval'),
```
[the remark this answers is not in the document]
wiffbi: I’d say, we should make it consistent. Directly above it is a checkbox „This variation will only be shown if a voucher matching …“ Switching to „product“ then reads a bit weird. Another alternative might be to shorten it to something like „Require approval“ (similar to the „Require a valid membership“ further down the page). The help-text could even be shortened to something like: „Buying this will put the order into an "approval" state …“

I think the help-text above „This variation will only be shown if a voucher matching the product is redeemed.“ has an error and should read „This variation will only be shown if a voucher matching the products’ variation is redeemed.“ Or maybe shorter: „only show if a matching voucher is redeemed“?

Shortening all these checkboxes makes the language more consistent. Currently all three: „This variation will only be shown if a voucher matching the product is redeemed.“, „Buying this product requires approval“ and „Require a valid membership“ have a different type of _language_.

src/pretix/base/models/items.py
wiffbi: Regarding „product“ in help-text, see above.

src/pretix/base/models/items.py
```diff
@@ -764,6 +764,8 @@ class ItemVariation(models.Model):
     :type default_price: decimal.Decimal
     :param original_price: The item's "original" price. Will not be used for any calculations, will just be shown.
     :type original_price: decimal.Decimal
+    :param require_approval: If set to ``True``, orders containing this product can only be processed and paid after approved by an administrator
```
wiffbi, line 767: Maybe make it consistent regarding variation here as well?
```suggestion
    :param require_approval: If set to ``True``, orders containing this product's variation can only be processed and paid after approved by an administrator
```

src/pretix/base/models/orders.py
```diff
@@ -1441,6 +1441,13 @@ def address_format(self):
         lines = [r.strip() for r in lines if r]
         return '\n'.join(lines).strip()
 
+    def requires_approval(self):
+        if self.item.require_approval:
+            return True
+        if self.variation is not None and self.variation.require_approval:
```
wiffbi, line 1447: Shouldn’t this be enough?

```suggestion
        if self.variation and self.variation.require_approval:
```

wiffbi: P.S.: btw. good idea to move this to `AbstractPosition`!

wiffbi (commented):
Looks good so far. Regarding naming see my comment.

src/pretix/base/models/items.py
ser8phin: I think I found a nice solution, but please check again :)

raphaelm (approved):
LGTM :christmas_tree:

PR #2383, opened 2021-12-15
Web checkin: Redirect user to login if session expired

pc-coholic:
Z#2391402

### DIFF
--- a/src/pretix/plugins/webcheckin/static/pretixplugins/webcheckin/components/app.vue
+++ b/src/pretix/plugins/webcheckin/static/pretixplugins/webcheckin/components/app.vue
@@ -453,7 +453,7 @@ export default {
       fetch(this.$root.api.lists + this.checkinlist.id + '/positions/?ignore_status=true&expand=subevent&expand=item&expand=variation&check_rules=true&search=' + encodeURIComponent(this.query))
           .then(response => {
             if (!response.ok && [401, 403].includes(response.status)) {
-              window.location.href = '/control/login?next=' + window.location.pathname + window.location.hash;
+              window.location.href = '/control/login?next=' + encodeURIComponent(window.location.pathname + window.location.search + window.location.hash);
             }
             response.json()
           })
--- a/src/pretix/plugins/webcheckin/static/pretixplugins/webcheckin/components/app.vue
+++ b/src/pretix/plugins/webcheckin/static/pretixplugins/webcheckin/components/app.vue
@@ -361,7 +361,7 @@ export default {
               }
             }
             if (!response.ok && [401, 403].includes(response.status)) {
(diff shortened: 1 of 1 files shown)


src/pretix/plugins/webcheckin/static/pretixplugins/webcheckin/components/app.vue
```diff
@@ -360,6 +360,9 @@ export default {
                 reason: 'invalid',
               }
             }
+            if (!response.ok && [401, 403].includes(response.status)) {
+              window.location.href = '/control/login?next=' + window.location.pathname + window.location.hash;
```
raphaelm, line 364: you probably need to call escape() on pathname and hash for the hash to actually work. are we sure we never have a query string here?

also, can we imagine reasons for 401/403 where the redirect does not make sense?

pc-coholic: I tried it out like this and it worked as intended...

Not sure about if there could be 401 or 403 that do not warrant a re-login... Possibly if the user got its access-rights removed to that specific event... In that case, the relogin wouldn't change anything.

wiffbi: AFAIK not encoding e.g. "/" and most other chars works most of the time in current browsers, but it should be encoded nonetheless. See https://datatracker.ietf.org/doc/html/rfc3986#section-2.2

Also I don’t think the hash works as intended – it will not be included in the next-GET-param but will be interpreted as a hash (Sprungmarke?) on the /control/login page. So I think – as Raphael pointed out as well – it should be encoded. So basically:

```suggestion
              window.location.href = '/control/login?next=' + encodeURIComponent(window.location.pathname + window.location.search + window.location.hash);
```

raphaelm: > Also I don’t think the hash works as intended

It should not work, but if @pc-coholic  says he tested, I believe that – browser have a weird behaviour of carrying the hash across redirects in some cases.

We should still use `encodeURIComponent` and `window.location.search` to be future-proof, this would be a really annoying bug to find if we missed that

wiffbi: > It should not work, but if @pc-coholic says he tested, I believe that – browser have a weird behaviour of carrying the hash across redirects in some cases.

Yes, I believe so too – sorry @pc-coholic if that did not come across.

pc-coholic: Everything is fine - no worries :-)

raphaelm (approved):

PR #2384, opened 2021-12-15
Allow to attach files to order confirmation email

raphaelm:
This feature has been requested by users hundreds of times. So far, I always refused. I think it is generally a bad idea, as it will hurt email deliverability of the most important email we send and it will lead to users misusing the feature.

However, after many years, I'm now breaking under the recent pressure applied by multiple customers and implementing it. The two intended use cases are:

* Many customers want to send their terms of service and similar documents, since there might be a legal requirement to do so. Hard to argue with that
* One customer wants to send a 1-page FAQ sheet on the booked event to avoid lots of support overhead

To keep the misuse of this feature small, I applied two restrictions:

* I only added one setting. If we wanted symmetry with the email texts we would need 6 (for the events order_placed, order_free, order_placed_require_approval + the same 3 for the attendee emails).
* I only added support for attaching to the order confirmation email, not to the order paid email, so people can't use this as a very bad replacement for the digital content module.
* I limited file size to 1MB. If your terms of service PDF has more than 1MB, that's a problem you should fix. I'm not sure, this might be a little harsh, but I'm really worried we'll have significant delivery problems otherwise if people upload their 10MB PDF files. Any other opinions here?

### DIFF
--- a/src/pretix/base/services/mail.py
+++ b/src/pretix/base/services/mail.py
@@ -35,6 +35,7 @@
 import hashlib
 import inspect
 import logging
+import mimetypes
 import os
 import re
 import smtplib
@@ -51,6 +52,7 @@ from bs4 import BeautifulSoup
 from celery import chain
 from celery.exceptions import MaxRetriesExceededError
 from django.conf import settings
+from django.core.files.storage import default_storage
 from django.core.mail import (
     EmailMultiAlternatives, SafeMIMEMultipart, get_connection,
 )
@@ -94,7 +96,7 @@ def mail(email: Union[str, Sequence[str]], subject: str, template: Union[str, La
          context: Dict[str, Any] = None, event: Event = None, locale: str = None, order: Order = None,
          position: OrderPosition = None, *, headers: dict = None, sender: str = None, organizer: Organizer = None,
          customer: Customer = None, invoices: Sequence = None, attach_tickets=False, auto_email=True, user=None,
-         attach_ical=False, attach_cached_files: Sequence = None):
+         attach_ical=False, attach_cached_files: Sequence = None, attach_other_files: list=None):
     """
     Sends out an email to a user. The mail will be sent synchronously or asynchronously depending on the installation.
 
@@ -141,6 +143,8 @@ def mail(email: Union[str, Sequence[str]], subject: str, template: Union[str, La
     :param customer: The user this email is sent to
 
     :param attach_cached_files: A list of cached file to attach to this email.
+
+    :param attach_other_files: A list of file paths on our storage to attach.
 
     :raises MailOrderException: on obvious, immediate failures. Not raising an exception does not necessarily mean
         that the email has been sent, just that it has been queued by the email backend.
@@ -301,6 +305,7 @@ def mail(email: Union[str, Sequence[str]], subject: str, template: Union[str, La
             organizer=organizer.pk if organizer else None,
             customer=customer.pk if customer else None,
             attach_cached_files=[(cf.id if isinstance(cf, CachedFile) else cf) for cf in attach_cached_files] if attach_cached_files else [],
+            attach_other_files=attach_other_files,
         )
 
         if invoices:
@@ -338,7 +343,8 @@ class CustomEmail(EmailMultiAlternatives):
 def mail_send_task(self, *args, to: List[str], subject: str, body: str, html: str, sender: str,
                    event: int = None, position: int = None, headers: dict = None, bcc: List[str] = None,
                    invoices: List[int] = None, order: int = None, attach_tickets=False, user=None,
-                   organizer=None, customer=None, attach_ical=False, attach_cached_files: List[int] = None) -> bool:
+                   organizer=None, customer=None, attach_ical=False, attach_cached_files: List[int] = None,
+                   attach_other_files: List[str] = None) -> bool:
     email = CustomEmail(subject, body, sender, to=to, bcc=bcc, headers=headers)
     if html is not None:
         html_message = SafeMIMEMultipart(_subtype='related', encoding=settings.DEFAULT_CHARSET)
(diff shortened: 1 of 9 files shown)


src/pretix/base/services/orders.py
```diff
@@ -941,7 +941,10 @@ def _order_placed_email(event: Event, order: Order, pprov: BasePaymentProvider,
             log_entry,
             invoices=[invoice] if invoice and event.settings.invoice_email_attachment else [],
             attach_tickets=True,
-            attach_ical=event.settings.mail_attach_ical
+            attach_ical=event.settings.mail_attach_ical,
+            attach_other_files=filter(lambda a: a, [
+                event.settings.get('mail_attachment_new_order', as_type=str, default='')[7:]
```
wiffbi, line 946: I know I often stumble upon those _magic_ numbers. Why do we need an offset here and why is it 7? If it is due to `file://` shouldn’t we strip the protocol off more explicitly?

src/pretix/base/services/orders.py
```diff
@@ -958,7 +961,10 @@ def _order_placed_email_attendee(event: Event, order: Order, position: OrderPosi
             invoices=[],
             attach_tickets=True,
             position=position,
-            attach_ical=event.settings.mail_attach_ical
+            attach_ical=event.settings.mail_attach_ical,
+            attach_other_files=filter(lambda a: a, [
+                event.settings.get('mail_attachment_new_order', as_type=str, default='')[7:]
```
wiffbi, line 966: See above.

src/pretix/base/services/mail.py
```diff
@@ -455,6 +461,20 @@ def mail_send_task(self, *args, to: List[str], subject: str, body: str, html: st
                         logger.exception('Could not attach invoice to email')
                         pass
 
+        if attach_other_files:
+            for fname in attach_other_files:
+                ftype, _ = mimetypes.guess_type(fname)
+                data = default_storage.open(fname).read()
+                try:
+                    email.attach(
+                        re.sub(r'^(.*)(\.[^.]+)(\.[^.]+)$', r'\1\3', os.path.basename(fname)),
```
wiffbi, line 470: Why do we need to trim the last part separated by a dot before the file-extensions? This would turn `file.tar.gz` into `file.gz`

wiffbi (commented):
Looks good! I totally agree with your reasoning only allowing one file up to 1MB and only on order_placed.

One thing missing is documentation in doc/admin/config.rst regarding `max_size_auto_email_attachment` (or if you update according to my comment `max_size_email_auto_attachment`)

src/pretix/helpers/hierarkey.py
```diff
@@ -0,0 +1,17 @@
+def clean_filename(fname):
+    """
+    hierarkey.forms.SettingsForm appends a random value to every filename. However, it keeps the
+    extension around "twice". Tis leads to:
```
wiffbi, line 4: Tiny typo:

```suggestion
    extension around "twice". This leads to:
```

src/pretix/helpers/hierarkey.py
```diff
@@ -0,0 +1,17 @@
...
+    In pretix Hosted, our storage layer also adds a hash of the file to the filename, so we have
+
+    "Terms.pdf" → "Terms.pdf.OybgvyAH.22c0583727d5bc.pdf"
+
+    This function reverses this operation:
+
+    "Terms.pdf.OybgvyAH.22c0583727d5bc.pdf" → "Terms.pdf"
+    """
+    ext = fname.split('.')[-1]
+    return fname.rsplit(ext, 2)[0] + ext
```
wiffbi, line 17: What happens if the original file did not have a filename-extension? Then the new extension would be `22c0583727d5bc`. Not really a problem, but just to make sure. 

Also I think you mean to only split once (so you get 2 parts), not twice (which could lead to 3 parts). And also make sure to split at an „ext-point“ separated by a "." and lateron add the ext separated by a ".":

```suggestion
    ext = '.' + fname.split('.')[-1]
    return fname.rsplit(ext, 1)[0] + ext
```

wiffbi (changes requested):
Mini typo and a change regarding cleaning the filename.

src/pretix/helpers/hierarkey.py
raphaelm: > What happens if the original file did not have a filename-extension? 

Then we would never allow for that file to be uploaded in the first place.  But yes, we can apply your fix.

> Also I think you mean to only split once (so you get 2 parts), not twice (which could lead to 3 parts). And also make sure to split at an „ext-point“ separated by a "." and lateron add the ext separated by a ".":

Nope. Pretty sure my code works correctly. With my code,
``clean_filename("fooasdas.asdasd.pdf.0123456.abcdef00.pdf") == "fooasdas.asdasd.pdf"``.

With your code, ``clean_filename("fooasdas.asdasd.pdf.0123456.abcdef00.pdf") == "fooasdas.asdasd.pdf.0123456.abcdef00.pdf"``.

This is because of:

```
>>> "fooasdas.asdasd.pdf.0123456.abcdef00.pdf".rsplit(".pdf", 1)
['fooasdas.asdasd.pdf.0123456.abcdef00', '']
```

wiffbi: Ah sorry, you are right. My brain was splitting on the filename without extension, which is not the case here.

wiffbi (approved):

src/pretix/helpers/hierarkey.py
```diff
@@ -0,0 +1,40 @@
...
+    In pretix Hosted, our storage layer also adds a hash of the file to the filename, so we have
+
+    "Terms.pdf" → "Terms.pdf.OybgvyAH.22c0583727d5bc.pdf"
+
+    This function reverses this operation:
+
+    "Terms.pdf.OybgvyAH.22c0583727d5bc.pdf" → "Terms.pdf"
+    """
+    ext = '.' + fname.split('.')[-1]
+    return fname.rsplit(ext, 2)[0] + ext
```
wiffbi, line 40: One super edge-edge-case of one of random values matches the ext: Terms.pdf.pdfOybgvyAH.22c0583727d5bc.pdf
So maybe we should do?
```suggestion
    return fname.rsplit(ext+".", 1)[0] + ext
```

PR #2385, opened 2021-12-16
Add fixed scroll position when navigating calendar views (Z#177488)

wiffbi:
This was suggested from a customer when navigating the calendar-views. It adds a bit of JavaScript that listens to navigation-events in the top-calendar-navigation (prev/next and date-dropdown) and records the scrollposition. When the next page is loaded it will scroll/jump to the same position. If loading is fast enough (most of the time) it makes for a seamless navigation in the calendar.

### DIFF
--- a/src/pretix/static/pretixpresale/js/ui/main.js
+++ b/src/pretix/static/pretixpresale/js/ui/main.js
@@ -180,10 +180,10 @@ $(function () {
 
     $("body").removeClass("nojs");
 
-    var scrollpos = localStorage.getItem('scrollpos');
+    var scrollpos = sessionStorage.getItem('scrollpos');
     if (scrollpos) {
         window.scrollTo(0, scrollpos);
-        localStorage.removeItem('scrollpos');
+        sessionStorage.removeItem('scrollpos');
     }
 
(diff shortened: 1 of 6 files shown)


src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -180,6 +180,12 @@ $(function () {
 
     $("body").removeClass("nojs");
 
+    var scrollpos = localStorage.getItem('scrollpos');
```
raphaelm, line 183: Can you check if we might want to use sessionStorage instead of localStorage? This would e.g. prevent race conditions if I have multiple tabs open:
https://developer.mozilla.org/en-US/docs/Web/API/Window/sessionStorage

wiffbi: That is a good idea! Although a race-condition might be very rare here, as the scrollpos is saved to localStorage just before leaving the page and cleared onDOMReady for the new page. So one would need to be quick switching tabs – or at least quickly abort loading and then switch tab. But nonetheless, sessionStorage is the way to go here! Thanks!

raphaelm (approved):

PR #2395, opened 2021-12-22
Fix #555 -- Preselect a single required add-on

### LINKED ISSUE #555: Preselect Addons if they have to be bought
Using AddOns, we can already force a customer to buy an AddOn...

Even though this specific usecase might be outdated when CheckIn-lists arrive (#515), this could be used (for example) to give users a second ticket for swag-redemption (badges, T-Shirts, etc.).

`if (minAddons == maxAddons) && num(AddOns) == (minAddons || maxAddons)`, I would propose to either skip the AddOns-page completely and silently add the items to the cart or at least set the `checked`-property for the items...

If we are forcing the users to take the items, we might as well save them a few clicks ;-)

ser8phin:
If an item contains a single add-on that is also required to be selected, this change autoselects the corresponding checkbox, so users don't have to click it themselves.

TODO:
- [x] Need to make sure no other add-on options exist besides the supposed single one

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/fragment_addon_choice.html
+++ b/src/pretix/presale/templates/pretixpresale/event/fragment_addon_choice.html
@@ -285,7 +285,7 @@
                             {% if c.max_count == 1 or not c.multi_allowed %}
                                 <label class="item-checkbox-label">
                                     <input type="checkbox" value="1"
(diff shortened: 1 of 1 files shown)


wiffbi (approved):
Looks good!

PR #2399, opened 2021-12-30
Fix #2320 - Move file upload "required" attrs manipulation from init to rendering

### LINKED ISSUE #2320: Regression: File question required when editing ticket
This is the same issue as #1408, I'm not quite sure whether this should be a new bug because it's a regression or if the old bug should be reopened?

Long story short: We noticed in Pretix 4.3 that customers editing their order will be faced with file questions being `required` even though they have uploaded a file for that question already. The requirement comes from the `input` being marked as `required` and the browsers refusing to submit the form without something in there.

This was fixed in #1443, but it seems this doesn't work anymore. I haven't had time to look into this more, so I don't have a clue where it might fail.

thegcat:
Having the attrs manipulation in the `__init__` leaves this open to additional manipulation after the Widget was created. This would happen when the question was dependent on another question's answers.

Moving the manipulation to `get_context` as done in other places in Pretix solves this issue.

Fixes #2320

### DIFF
--- a/src/pretix/base/forms/widgets.py
+++ b/src/pretix/base/forms/widgets.py
@@ -86,14 +86,6 @@ class TimePickerWidget(forms.TimeInput):
 
 class UploadedFileWidget(forms.ClearableFileInput):
     def __init__(self, *args, **kwargs):
-        # Browsers can't recognize that the server already has a file uploaded
-        # Don't mark this input as being required if we already have an answer
-        # (this needs to be done via the attrs, otherwise we wouldn't get the "required" star on the field label)
-        attrs = kwargs.get('attrs', {})
-        if kwargs.get('required') and kwargs.get('initial'):
-            attrs.update({'required': None})
-        kwargs.update({'attrs': attrs})
-
         self.position = kwargs.pop('position')
         self.event = kwargs.pop('event')
         self.answer = kwargs.pop('answer')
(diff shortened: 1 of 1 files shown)


thegcat:
I'm still a little perplex as to why this is necessary at all. From my understanding of https://docs.djangoproject.com/en/2.2/ref/forms/widgets/#django.forms.Widget.use_required_attribute and the associated source code in `ClearableFileInput` the desired behaviour should already be happening on its own. I'm afraid I'm not proficient enough in Django nor Python to be able to grok why `use_required_attribute` doesn't get applied there though.

raphaelm:
@cla-bot recheck cla

raphaelm:
Thanks! Patch looks good, but as the mere existance of the bug clearly shows, this is a complex area. Would you be able to add a small regression test to make sure we notice if we re-introduce the bug?

thegcat:
> but as the mere existance of the bug clearly shows, this is a complex area.

As I said earlier, this is per se not a bug with the original patch, just that the original patch did not also take into account dependent fields. This was not a regression as I first thought.

> Would you be able to add a small regression test to make sure we notice if we re-introduce the bug?

I am neither familiar with Python nor Django, much less with the test system or what seems to be needed to be added to get tests with Pretix (I have not understood what the scopes chicanery does for example), and I'm not quite sure where interface tests or tests for the form widgets would go? Furthermore this probably would need 4 tests to check the presence of the `required` keyword when there already is a file or not and when the file question is dependent or not?

raphaelm:
Sorry for the delay. I wanted to look into this more deeply, and I think we can skip the additional tests as this only affects form rendering and not form validation and thus is less of a risk if we get it wrong in the other direction now. I will try to find time for a little manual testing (unless e.g. @wiffbi gets around to first) and then go ahead and merge.

wiffbi:
I will look into this now.

wiffbi (approved):
Works as intended on the order-change details page for customers. I have not tested it elsewhere but do not expect it to break. Just for completeness as this is a bit buried in the original bug-report: this fixes the required-bug that happens, when a file-question is dependent on another question.

raphaelm:
Great, thanks everyone!

PR #2400, opened 2021-12-31
Add static select2 to choice widgets

### LINKED ISSUE #2020: Optional search for dropdown/select questions
We have a select question with a fairly large number of options (>150), it would be nice to either optionally (in the question settings) or when the question has a certain number of options (maybe over 20 oder 30) be able to add a search option to the select question.

thegcat:
This is a rough proof of concept (the code is mostly copied from pretixcontrol, the custom select2 css from pretixcontrol has not been copied, I have no idea how necessary it is for presale) to add select2 features to single choice questions. This can be refined by only enabling select2 when there are more than a threshold number of choices or by having a setting to enable this on a per-question level (I don't see the harm having this on all the time though).

This uses the in-place version of select2 to gracefully fall back to a normal select with all options available should the user use a non-js client or a screen reader or similar. Even if the question has lots of choices (we regularly have more than 100) it should be efficient enough without requiring the dynamic/server-query based version of select2.

Fixes  #2020

### DIFF
--- a/src/pretix/base/forms/questions.py
+++ b/src/pretix/base/forms/questions.py
@@ -706,7 +706,7 @@ class BaseQuestionsForm(forms.Form):
                     queryset=q.options,
                     label=label, required=required,
                     help_text=help_text,
-                    widget=forms.Select,
+                    widget=forms.Select(attrs={'class': 'select2-static'}),
                     to_field_name='identifier',
                     empty_label='',
                     initial=initial.options.first() if initial else None,
--- a/src/pretix/presale/context.py
+++ b/src/pretix/presale/context.py
@@ -36,7 +36,7 @@ import logging
 from django.conf import settings
 from django.core.files.storage import default_storage
 from django.utils import translation
-from django.utils.translation import get_language_info
+from django.utils.translation import get_language_info, get_language
 from django_scopes import get_scope
 from i18nfield.strings import LazyI18nString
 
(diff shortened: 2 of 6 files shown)


raphaelm:
This is in the critical checkout path, as noted in the issue this can only move forward if we clarify accessibility impact. @wiffbi is on vacation next week but maybe can try to find out more after that.

raphaelm:
After we've had specific feedback from blind users now that select2 is unusable for them, I'll close this. We can revisit when (and if) we have a select2 alternative that fulfills all of our other requirements.

thegcat:
This is surprising to me, as select2 as used only injects behaviour when js is active (i.e. should work with screenreaders) and at least on [the select2 docs](https://select2.org/getting-started/basic-usage#single-select-boxes) it seems to go out of its way to add `aria` tags to the generated markup, but well. Thanks for investigating this.

wiffbi:
Screenreaders don’t disable JavaScript, they just use the browser API to access what is on the page. I read the docs on select2 and it seems to use all the right aria-attributes and roles, but in practice it still does not work. I don’t know why they go all the way to add these aria-attributes and roles, but it seems they do not test with a screenreader. I tested with VoiceOver with Safari on Mac and can confirm the reported findings from JAWS (Chrome, Firefox) on Windows, that select2 does not really work with a screenreader as the options are not read aloud. One can type into the search box and even use the cursor-keys, but nothing is read until confirmed with return/enter. So not really useful. What works with a screen-reader is the example on W3-ARIA examples, e.g. https://www.w3.org/WAI/ARIA/apg/example-index/combobox/combobox-autocomplete-list.html I will look into this whether that is something that can be fixed in select2 with a PR or if we should look into something else.

PR #2409, opened 2022-01-07
Allow tax rules to trigger approval requirement

raphaelm:
This change adds a new action that can be triggered by tax rules for specific customer groups (defined by country and type of customer). You can now set e.g. that all customers from non-EU countries will be placed in the order review process:

This is useful e.g. if you have a specific situation where you want to handle taxation of unknown countries manually.


ser8phin:
Looks good all in all :sparkles:

ser8phin (approved):
Looks good all in all :sparkles: 
(ps: finally, github is showing me the review tools I was missing just seconds before…)

PR #2413, opened 2022-01-11
Improve add-to-cart checkbox for items with max. 1 per order (Z#178704)

wiffbi:
When having products with a max. quantity of one per order, we render a checkbox instead of a number-input to mark an item to be added to the cart. This checkbox on its own can easily be overlooked, especially combined with a free-price input next to it.

This PR adds an auto-check to the checkbox, when the free-price input is changed as well adds an icon next to the checkbox.

### DIFF
--- a/src/pretix/static/pretixpresale/js/ui/main.js
+++ b/src/pretix/static/pretixpresale/js/ui/main.js
@@ -636,13 +636,11 @@ $(function () {
     lightbox.init();
 
     // free-range price input auto-check checkbox
-    $(".input-item-price").each(function() {
-        var input = document.getElementById("item_" + this.id.substr("price-item-".length));
-        if (input && input.type == "checkbox") {
-            $(this).on("change", function() {
-                input.checked = true;
-            });
-        }
+    $("[data-checked-onchange]").each(function() {
+        var checkbox = this;
+        $("#" + this.getAttribute("data-checked-onchange")).on("change", function() {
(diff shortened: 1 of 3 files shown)


wiffbi:
Bildschirmfoto 2022-01-11 um 11 37 22

src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -634,6 +634,16 @@ $(function () {
 
     // Lightbox
     lightbox.init();
+
+    // free-range price input auto-check checkbox
+    $(".input-item-price").each(function() {
+        var input = document.getElementById("item_" + this.id.substr("price-item-".length));
```
raphaelm, line 640: Should we support variations here as well to be consistent?

raphaelm (commented):
I think this is a good pragmatical approach

src/pretix/static/pretixpresale/js/ui/main.js
wiffbi: Argh, that is what I forgot about to test – free price on variations. Of course we should be consistent. I will update accordingly.

wiffbi: I changed it to a data-attribut that links to the ID of the element and set it onto the checkbox-element instead of the free-price input. Somehow seemed a more logical place.

PR #2414, opened 2022-01-12
Add phone to customer profile (Z#178346)

wiffbi:
When a customer registers an account with an organizer, they can optionally add a phone number. In the customer’s account-settings page the phone number is always visible and editable – even when no event asks for a phone number. When registering on the organizer’s page, the phone number is an optional input. When registering during the checkout process of an event, whether the phone number is diplayed or marked as required in the embedded registration form, is depending on the events settings `order_phone_asked` and `order_phone_required`.

Should we make the availability of phone numbers in customer accounts a setting on the organizer’s page? That way if no phone numbers are used in the events, they could be removed from the accounts as well. It could lead to confusion when no phone numbers are allowed in accounts but an event uses phone numbers – although this will then be no different to what it currently is (no support for phone numbers in customer accounts).

- [x] add phone number in organizer back-end view
- [x] anonymize should remove phone number as well


raphaelm:
> When a customer registers an account with an organizer, they can optionally add a phone number. In the customer’s account-settings page the phone number is always visible and editable – even when no event asks for a phone number. When registering on the organizer’s page, the phone number is an optional input. 

Let's default the other way around – let's just never ask for the phone-number in the standalone registration and only ask during checkout? Most registrations will be done during checkout anyways

> When registering during the checkout process of an event, whether the phone number is diplayed or marked as required in the embedded registration form, is depending on the events settings order_phone_asked and order_phone_required.

This makes sense.

> Should we make the availability of phone numbers in customer accounts a setting on the organizer’s page? That way if no phone numbers are used in the events, they could be removed from the accounts as well. It could lead to confusion when no phone numbers are allowed in accounts but an event uses phone numbers – although this will then be no different to what it currently is (no support for phone numbers in customer accounts).

I don't think we should add a setting before we see real-world evidence that is needed. Given the changes above I don't think it will be a problem.

src/pretix/base/models/customers.py
```diff
@@ -45,6 +46,7 @@ class Customer(LoggedModel):
     organizer = models.ForeignKey(Organizer, related_name='customers', on_delete=models.CASCADE)
     identifier = models.CharField(max_length=190, db_index=True, unique=True)
     email = models.EmailField(db_index=True, null=True, blank=False, verbose_name=_('E-mail'), max_length=190)
+    phone = PhoneNumberField(null=True, blank=True, verbose_name=_('Phone number'))
```
raphaelm, line 49: `anonymize()` should clear the phone number as well

raphaelm (changes requested):
This generally looks fine. Apart from the comment above, this is still missing the display and editing of the phone number in the backend from an organizer perspective.

raphaelm:
Note to self: After merge, we need to adjust pretix-gdpr

src/pretix/base/models/customers.py
wiffbi: Thanks for pointing that out. I added it to `models.Customer.anonymize`

wiffbi:
> This generally looks fine. Apart from the comment above, this is still missing the display and editing of the phone number in the backend from an organizer perspective.

I missed the organizer’s customer pages in pretix-control. I meant the customers account page on the organizers pretix-page. This is fixed/added.

As you suggested, I removed the phone number from the registration form when registering through the organizers main page instead of during checkout.

src/pretix/control/templates/pretixcontrol/organizers/customer.html
```diff
@@ -48,6 +48,8 @@ <h3 class="panel-title">
                             </dd>
                             <dt>{% trans "Name" %}</dt>
                             <dd>{{ customer.name }}</dd>
+                            <dt>{% trans "Phone" %}</dt>
+                            <dd>{{ customer.phone }}</dd>
```
raphaelm, line 52: This should be wrapped by an {% if %} like in the presale view, otherwise it will show a literal, untranslated "None"

raphaelm (changes requested):
I noticed that all the new phone number fields have no country selected by default. Can we default-select the country specified by ``organizer.settings.region`` or based on the language? We do it like this in ``BaseQuestionForm``:

```
                with language(get_babel_locale()):
                    default_country = guess_country(event)
                    default_prefix = None
                    for prefix, values in _COUNTRY_CODE_TO_REGION_CODE.items():
                        if str(default_country) in values:
                            default_prefix = prefix
                    try:
                        initial = PhoneNumber().from_string(initial.answer) if initial else "+{}.".format(default_prefix)
                    except NumberParseException:
                        initial = None
                    field = PhoneNumberField(
                        label=label, required=required,
                        help_text=help_text,
                        # We now exploit an implementation detail in PhoneNumberPrefixWidget to allow us to pass just
                        # a country code but no number as an initial value. It's a bit hacky, but should be stable for
                        # the future.
                        initial=initial,
                        widget=WrappedPhoneNumberPrefixWidget()
                    )
```

src/pretix/control/templates/pretixcontrol/organizers/customer.html
wiffbi: Ah, thanks – I tested it with an existing account and deleting the phone number just sets it blank, but initially creating an account sets it `None`. Fixed.

src/pretix/control/forms/organizer.py
```diff
@@ -535,11 +539,19 @@ class CustomerUpdateForm(forms.ModelForm):
...
     class Meta:
         model = Customer
-        fields = ['is_active', 'name_parts', 'email', 'is_verified', 'locale']
+        fields = ['is_active', 'name_parts', 'email', 'is_verified', 'phone', 'locale']
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        if not self.instance.phone and (self.instance.organizer.settings.region or self.instance.locale):
+            country_code = self.instance.organizer.settings.region or get_country_by_locale(self.instance.locale)
+            self.initial['phone'] = "+{}.".format(get_phone_prefix(country_code))
```
raphaelm, line 548: Might not be a new problem, but this will lead to ``"+None."`` in the case where we know nothing. Does that cause issues?

raphaelm (approved):

src/pretix/control/forms/organizer.py
wiffbi: Good catch, will probably throw an NumberParse-Error. So I will better handle this …

PR #2419, opened 2022-01-17
Add Organizer-Attribute to ical

pc-coholic:
Z#2392882

Since this apparently a rather new addition to the ical-standard, one could argue that we should leave the cleartext-mention of the organizer in the event-description intact.


raphaelm:
No, bad idea. Whenever someone imports the calendar entry into some calender tools and "accepts" the invite, the organizer will get an email!

PR #2422, opened 2022-01-19
TaxRules: Add internal_name and keep_gross_if_rate_changes

raphaelm:
This brings three improvements to tax rules:

- An internal name. This has been missing for a long time to differentiate tax rules with subtle differences when configuring products.

- An option to keep the gross price of a product constant even if the rule specifies a varying tax rate.

- Some additional unit tests for the low level implementation


src/pretix/control/templates/pretixcontrol/order/change.html
```diff
@@ -145,7 +145,12 @@ <h3 class="panel-title">
                                 <strong>{% trans "Tax rule" %}</strong>
                             </div>
                             <div class="col-sm-5">
-                                {{ position.tax_rule.name }} ({{ position.tax_rule.rate }} %)
+                                {% if poition.tax_rule.internal_name %}
```
ser8phin, line 148:
```suggestion
                                {% if position.tax_rule.internal_name %}
```

PR #2423, opened 2022-01-19
Remove disabled attribute on checkbox

ser8phin:
Removing the disabled attribute on the checkbox, since it also prevents the value from being submitted by POST

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/fragment_addon_choice.html
+++ b/src/pretix/presale/templates/pretixpresale/event/fragment_addon_choice.html
@@ -285,7 +285,9 @@
                             {% if c.max_count == 1 or not c.multi_allowed %}
                                 <label class="item-checkbox-label">
                                     <input type="checkbox" value="1"
-                                            {% if c.max_count == 1 and c.min_count == 1 and c.items|length == 1 or item.initial %}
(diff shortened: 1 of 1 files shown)


src/pretix/presale/templates/pretixpresale/event/fragment_addon_choice.html
```diff
@@ -285,10 +285,7 @@ <h4 id="cp-{{ form.pos.pk }}-item-{{ item.pk }}-legend">{{ item.name }}</h4>
                             {% if c.max_count == 1 or not c.multi_allowed %}
                                 <label class="item-checkbox-label">
                                     <input type="checkbox" value="1"
-                                            {% if c.max_count == 1 and c.min_count == 1 and c.items|length == 1 %}
-                                                checked="checked"
-                                                disabled="disabled"
-                                            {% elif item.initial %}
+                                            {% if c.max_count == 1 and c.min_count == 1 and c.items|length == 1 or item.initial %}
```
raphaelm, line 288: I'd prefer not mixing "and" and "or" in templates as it is impossible to see in which order they are executed.

Can we maybe add a property/it to the object that is `c` here to collapse the "and" part?

Another way might be keeping the two conditions and just removing the disabled line. Some repetition is okay if it improves clarity.

raphaelm (approved):

PR #2447, opened 2022-02-03
Make template tag getitem not fail if field does not exist

pc-coholic:
Tiny change for the `getitem` template tag to not come down crashing if the requested key is not in the dictionary.

For current usage this change should not have any adverse effects, for new developments this brings adds the possibility of chaining the `|getitem` with `|default`.

### DIFF
--- a/src/pretix/control/templatetags/getitem.py
+++ b/src/pretix/control/templatetags/getitem.py
@@ -29,4 +29,7 @@ def getitem_filter(value, itemname):
     if not value:
         return ''
 
+    if itemname not in value:
+        return ''
+
     return value[itemname]


raphaelm:
:joy: I did the same thing today, just did not commit it yet

raphaelm:
However, I did ``value.get('itemname')`` which returns ``None`` instead of ``""``. I assume ``""`` could have some problems with chained getitem calls? Not sure though

pc-coholic:
I guess we could absolutely shorten this to your solution - I'll gladly close this issue in its favour :-)

Just a word on equally expected results...

Right now, we also return `''` if the dict is None... I feel there might be an argument, that a failed result should always be `None` or ` `. Or not - it's not like anyone else beside us is using this ;-)

PR #2448, opened 2022-02-04
Fix guard in mail service

thegcat:
I'm guessing this is a typo as `logger` is not used in the condition body.


raphaelm:
Good catch, thanks!

PR #2451, opened 2022-02-08
Fix #2449 -- Format variation-price with Intl.NumberFormat

### LINKED ISSUE #2449: Correctly format variation price
Right now, prices in the variation-accordeon in pretixcontrol are only displayed based on the raw value in the input field.

Depending on the amount of effort given by the user when entering those prices, they might have anywhere from zero to n decimals:
Screenshot from 2022-02-04 16-51-38

I don't see this as a burning issue but rather as a small UI improvement to have a properly formatted decimal value there and perhaps event a currency symbol.

https://github.com/pretix/pretix/blob/c65883b32889a1fa6c022a177a01a79f6bde17da/src/pretix/static/pretixcontrol/js/ui/variations.js#L10-L14

wiffbi:
If Intl.NumberFormat is available, try to format price with locale and currency – all in try-block, in case formatting fails due to unknown currency or locale.

### DIFF
--- a/src/pretix/static/pretixcontrol/js/ui/main.js
+++ b/src/pretix/static/pretixcontrol/js/ui/main.js
@@ -16,22 +16,36 @@ function ngettext(singular, plural, count) {
 
 function formatPrice(price, currency, locale) {
     if (!window.Intl || !Intl.NumberFormat) return price;
-    if (isNaN(price) && price.replace) {
-        price = price.replace(/\s/g, "")
-        if (price.indexOf(",") > -1) {
-            price = price.replace(/\./g, "").replace(",", ".")
-        }
-    }
-
+    var priceToFormat = price
     if (currency === undefined) {
         currency = $("[data-currency]").data("currency")
     }
     if (locale === undefined) {
         locale = $("[data-locale]").data("locale") || $("[data-pretixlocale]").data("pretixlocale");
     }
+
     var opt = currency ? {style: "currency", currency: currency} : null;
+    var nf = new Intl.NumberFormat(locale, opt)
+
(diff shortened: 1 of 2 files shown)


raphaelm:
I'm still a little bit worried that this might cause weird issues since the locale data is defined by the browser and not specified further, and might theoretically be different with what we use server-side. This can be dangerous, an hypothetical example woul dbe if we use Australian dollars with en-au locale, there could be disagreement whether to write `A$` or `$` which might cause confusion. Admittedly, it's not that important for the location we're dealing with here, but it might become a problem if we roll out Intl.NumberFormat to other places like the widget. With Firefox, I was **not** able to produce an edge case that is different between Python and JS, so I feel fine with moving this forward, but I wanted to at least voice my concerns again. ECMAScript specification **recommends* that browsers use the data from cldr.unicode.org, which babel (our python library) uses as well, but it's not a strict requirement of the spec.

raphaelm:
Some googling suggest that Chrome, FF and  Apple all use CLDR data, at least for some things. So that suggests this is fine for us to do.

Do we maybe want to move the full try-block into its own function we can reuse in other places?

wiffbi:
Regarding $ and A$ is a good catch! Good to know that it seems to be handled „correctly.“
I was more concerned regarding our custom-locale `de-informal`. But it works just as well (I guess it falls back to just `de`). The try-block is there to just make sure.
I will add a function `formatPrice` to main.js

wiffbi:
Made `formatPrice` more robust regarding number-input. If no locale is given, it tries to read locale from data-attributes available in pretix-presale and pretix-control. Same with currency.

raphaelm (approved):

PR #2453, opened 2022-02-09
Fix #2452 -- Disallow invalid values in min_per_order/max_per_order

### LINKED ISSUE #2452: Max Ticket Amount per Order not working
Max amount is set to 0 in my setup, but the shop does not allow more than one. Changing the value to something > 0 works (e.g. 2,3, etc.), but 0 only allows 1 Ticket.

The max amount field also accepts values like -1, which might actually break things in a worse way, just a side note though.

pretix version 4.6.0

image
image

wiffbi:
This fixes a bug where min_per_order/max_per_order on an item can be set < 0. The fix is in the widget, alternatively we might migrate from `models.IntegerField` to `models.PositiveIntegerField`.

### DIFF
--- a/src/pretix/control/forms/item.py
+++ b/src/pretix/control/forms/item.py
@@ -628,8 +628,8 @@ class ItemUpdateForm(I18nModelForm):
             }),
             'generate_tickets': TicketNullBooleanSelect(),
             'show_quota_left': ShowQuotaNullBooleanSelect(),
-            'max_per_order': forms.widgets.NumberInput(attrs={'min':0}),
-            'min_per_order': forms.widgets.NumberInput(attrs={'min':0}),
+            'max_per_order': forms.widgets.NumberInput(attrs={'min': 0}),
+            'min_per_order': forms.widgets.NumberInput(attrs={'min': 0}),
         }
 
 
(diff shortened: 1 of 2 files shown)


src/pretix/control/forms/item.py
```diff
@@ -627,7 +627,9 @@ class Meta:
                 'class': 'scrolling-multiple-choice'
             }),
             'generate_tickets': TicketNullBooleanSelect(),
-            'show_quota_left': ShowQuotaNullBooleanSelect()
+            'show_quota_left': ShowQuotaNullBooleanSelect(),
+            'max_per_order': forms.widgets.NumberInput(attrs={'min': 0}),
+            'min_per_order': forms.widgets.NumberInput(attrs={'min': 0}),
```
raphaelm, line 632: I think only doing this in the widget is bad, as it will only be client side and not validated server side. The correct level in the form layer would be the `forms.IntegerField`, but even that would not affect the API etc. As discussed I don't want to use PositiveIntegerField since the migration will be dangerous, but I think we should add a [MinValueValidator](https://docs.djangoproject.com/en/4.0/ref/validators/#minvaluevalidator) to the IntegerField on the model – maybe combined with the widget (or form field) change.

raphaelm (changes requested):

src/pretix/control/forms/item.py
wiffbi: Good idea, I will add this as well!

wiffbi: I added the MinValueValidator to the model-field. I also tried setting the form-field in the UpdateForm to an IntegerField with validators, but that did not translate to a min-attribute on the input[type=number]. So in the form I left the forms.widget stuff in there for the min-attribute in HTML.

PR #2466, opened 2022-02-14
Refs #2465 -- Introduce unique identifiers for authentication backends

raphaelm:


raphaelm:
Necessary plugin changes are demonstrated in https://github.com/pretix-unofficial/pretix-auth-okta/pull/1

src/pretix/base/models/auth.py
```diff
@@ -83,6 +87,111 @@ def create_superuser(self, email: str, password: str = None):  # NOQA
...
+        This method will never return users managed by a different authentication backend.
+        If you try to create an account with an email address already blocked by a different
+        authentication backend, :py:class:`EmailAddressTakenError` will be raised. In this case,
+        you should display a message to the user.
+
+        :param backend: The `identifier` attribute of the authentication backend
+        :param identifier: The unique, immutable identifier of this user, max. 190 characters
+        :param email: The user's email address
+        :param set_always: A dictionary of fields to set on the user model if it's newly created
+        :param set_on_creation: A dictionary of fields to update on the user model on every login
```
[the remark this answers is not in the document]
raphaelm: Thanks!

src/pretix/base/models/auth.py
```diff
@@ -152,6 +265,7 @@ class User(AbstractBaseUser, PermissionsMixin, LoggingMixin):
     )
     notifications_token = models.CharField(max_length=255, default=generate_notifications_token)
     auth_backend = models.CharField(max_length=255, default='native')
+    auth_backend_identifier = models.CharField(max_length=190, db_index=True, null=True, blank=True)
```
thegcat, line 268: So I've been thinking about this 190 chars limit and a few things came across my mind.

First of all this might be too short for some cases (e.g. DNs from an LDAP source can get pretty long pretty quickly, even email addresses might get unreasonably long). I then thought that if an auth plugin author fears the backend identifier might be too long it can simply be hashed and it will automatically fit.

Which led me to: I can't think of a scenario where Pretix would need the backend identifier for whatever reason, and a DN for example might contain PII that Pretix probably doesn't want to keep in its DB anyway.

I'm unclear wether there's a Hash scheme that will accommodate strings with such wildly differing lengths (might be 1 if using and external ID, might be long if it's an LDAP DN or similar) efficiently though, so this might be a bad idea.

thegcat (commented):
Sorry for the delayed response, I'm down with a bad cold 🤧

src/pretix/base/models/auth.py
```diff
@@ -83,6 +87,111 @@ def create_superuser(self, email: str, password: str = None):  # NOQA
...
+                if getattr(u, k) != v:
+                    setattr(u, k, v)
+                    dirty = True
+            if dirty:
+                try:
+                    with transaction.atomic():
+                        u.save(update_fields=set_always.keys())
+                except IntegrityError:
+                    # This might only raise IntegrityError if the email address is used
+                    # by someone else
```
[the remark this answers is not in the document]
raphaelm: Uhm, well, but `auth_backend`, and `auth_backend_identifier` should never change, so passing them in `set_always` is already wrong. But we can add a check for that, I guess.

src/pretix/base/models/auth.py
raphaelm: So first of all, the reason for the 190 characters is a very sad one. We absolutely want an index for this field so there's no performance problem with large user databases. However, depending on system configuration and version, MariaDB and MySQL have an upper limit of 767 bytes on indexed columns. Since the worst case for UTF-8 is four bytes per character, this means the field can not be more than 191 characters. Up until recently, this was a very hard issue, but it seams that now all supported MySQL/MariaDB versions at least *allow* to increase that upper limit, although it might require some configuration (setting of row format and the `large_prefix_sizes` option). If we'd be in a PostgreSQL-only dream world, I'd have just chosen TextField with no limit whatsoever.

> I then thought that if an auth plugin author fears the backend identifier might be too long it can simply be hashed and it will automatically fit.

If you think LDAP DNs easily get longer than this, we might need to reconsider this decision!

Hashing is a good workaround, although it might be useful in some cases to trace accounts back to IdP records, e.g. as part of a permission audit.

> Which led me to: I can't think of a scenario where Pretix would need the backend identifier for whatever reason, and a DN for example might contain PII that Pretix probably doesn't want to keep in its DB anyway.

This is a fun one. I think yes, pretix might want this data, because if a pretix account starts doing weird stuff, a sysadmin might want to find out which person that is in their main database.

**However,** it also means that ``UserAnonymizeView`` should wipe the identifier if the account should be permanently anonymized (and deactivated) as part of a "deletion" (pretix doesn't allow full deletion of user accounts).

Hashing will only very marginally help with privacy, since it's probably not too hard to just build a rainbow table of hashes if you have IdP database access (or just a list of people working at the company and constructing the DNs yourself).

> I'm unclear wether there's a Hash scheme that will accommodate strings with such wildly differing lengths (might be 1 if using and external ID, might be long if it's an LDAP DN or similar) efficiently though, so this might be a bad idea.

Every cryptographic hash scheme should be fine with that?

thegcat: > So first of all, the reason for the 190 characters is a very sad one. We absolutely want an index for this field so there's no performance problem with large user databases. However, depending on system configuration and version, MariaDB and MySQL have an upper limit of 767 bytes on indexed columns. Since the worst case for UTF-8 is four bytes per character, this means the field can not be more than 191 characters. Up until recently, this was a very hard issue, but it seams that now all supported MySQL/MariaDB versions at least _allow_ to increase that upper limit, although it might require some configuration (setting of row format and the `large_prefix_sizes` option). If we'd be in a PostgreSQL-only dream world, I'd have just chosen TextField with no limit whatsoever.

I am, unfortunately, very aware of this issue. I think I'd have searched by email address and then rejected the login if the external attribute didn't match, but well.

> > I then thought that if an auth plugin author fears the backend identifier might be too long it can simply be hashed and it will automatically fit.
> 
> If you think LDAP DNs easily get longer than this, we might need to reconsider this decision!

There's enough other LDAP attributes to go around, or it could be cut off at the search base or whatever. It might be that in some fringe cases this is not enough, but again there's the possibility of hashing too, so if supporting MySQL/MariaDB with very conservative update cycles is a goal just keep it as is.

> Hashing is a good workaround, although it might be useful in some cases to trace accounts back to IdP records, e.g. as part of a permission audit.

You'd still have the email address to go back to as is the case today already?

> > Which led me to: I can't think of a scenario where Pretix would need the backend identifier for whatever reason, and a DN for example might contain PII that Pretix probably doesn't want to keep in its DB anyway.
> 
> This is a fun one. I think yes, pretix might want this data, because if a pretix account starts doing weird stuff, a sysadmin might want to find out which person that is in their main database.
> 
> **However,** it also means that `UserAnonymizeView` should wipe the identifier if the account should be permanently anonymized (and deactivated) as part of a "deletion" (pretix doesn't allow full deletion of user accounts).
> 
> Hashing will only very marginally help with privacy, since it's probably not too hard to just build a rainbow table of hashes if you have IdP database access (or just a list of people working at the company and constructing the DNs yourself).

If an attacker already has IdP access you don't gain anything hashing it on Pretix side. If you consider the scenario that Pretix data might leak, then making it harder to link the email addresses to the external identifier might be a goal.

Again, regarding the audit, you'd already be limited to only email addresses on the Pretix side today, and I'd say whomever needs audits probably has an audit trail for the email addresses in the IdP already, so hashing the external identifier wouldn't make it worse than today.

> > I'm unclear wether there's a Hash scheme that will accommodate strings with such wildly differing lengths (might be 1 if using and external ID, might be long if it's an LDAP DN or similar) efficiently though, so this might be a bad idea.
> 
> Every cryptographic hash scheme should be fine with that?

Yes and no. They don't (and can't as you're mapping an input space on a smaller output space) guarantee uniqueness in the output. Practically they'd probably be fine, but I am not a ~~lawyer~~ security engineer, so I'm pretty cautious when giving advice in that domain 😅

src/pretix/base/models/auth.py
```diff
@@ -115,6 +115,11 @@ def get_or_create_for_backend(self, backend, identifier, email, set_always, set_
             raise ValueError('You need to supply a custom, unique identifier for this user.')
         if email is None:
             raise ValueError('You need to supply an email address for this user.')
+        if 'auth_backend_identifier' in set_always or 'auth_backend_identifier' in set_on_creation or \
+                'auth_backend' in set_always or 'auth_backend' in set_always:
```
thegcat, line 119: The 2nd `set_always` should be `set_on_creation`

src/pretix/base/models/auth.py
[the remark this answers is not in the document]
thegcat: As I said, constructed case. They should never change and never be changed, someone unfamiliar with the code might still be tempted to do it.

thegcat:
Trank you!

Issue #2477, opened 2022-02-20
Crash when deleting voucher in use in cart position that is addon.


MaicoTimmerman:
**Describe the bug**
I'm not sure how someone managed to do this, as I was unable to find how to apply a voucher to an add-on product. However, a crash is caused when deleting a voucher that is currently applied to a `CartPosition` which is an add-on to another `CartPosition`, using the same voucher. 

```pytb
ProtectedError: ("Cannot delete some instances of model 'CartPosition' because they are referenced through protected foreign keys: 'CartPosition.addon_to'.", {<CartPosition: item 934, variation 0 for cart>})
  File "django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "pretix/control/permissions.py", line 69, in wrapper
    return function(request, *args, **kw)
  File "django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "pretix/control/views/vouchers.py", line 552, in post
    obj.cartposition_set.all().delete()
  File "django/db/models/query.py", line 745, in delete
    collector.collect(del_query)
  File "django/db/models/deletion.py", line 302, in collect
    raise ProtectedError(
```

**Expected behavior**
No crashes.

**Screenshots**

**Additional context**
Sentry: https://sentry.io/share/issue/512c0cc5dda34b329e30da190d09ce4b/

MaicoTimmerman:
Fixed in #2478

PR #2478, opened 2022-02-20
vouchers: allow deleting vouchers that are used on addon cartpositions

MaicoTimmerman:
Proposed fix for #2477.

I'm open to construct a test-case, if needed.

### DIFF
--- a/src/pretix/control/views/vouchers.py
+++ b/src/pretix/control/views/vouchers.py
@@ -56,7 +56,7 @@ from django.views.generic import (
     CreateView, DeleteView, ListView, TemplateView, UpdateView, View,
 )
 
-from pretix.base.models import CartPosition, LogEntry, OrderPosition, Voucher
+from pretix.base.models import CartPosition, LogEntry, Voucher
 from pretix.base.models.vouchers import generate_codes
 from pretix.base.services.locking import NoLockManager
 from pretix.base.services.vouchers import vouchers_send
@@ -550,7 +550,7 @@ class VoucherBulkAction(EventPermissionRequiredMixin, View):
             for obj in self.objects:
                 if obj.allow_delete():
                     obj.log_action('pretix.voucher.deleted', user=self.request.user)
-                    OrderPosition.objects.filter(addon_to__voucher=obj).delete()
+                    CartPosition.objects.filter(addon_to__voucher=obj).delete()
                     obj.cartposition_set.all().delete()
                     obj.delete()
                 else:


raphaelm:
The bug seems valid, I'm hesitant if this is the best fix, will need to think about it briefly.

MaicoTimmerman:
> The bug seems valid, I'm hesitant if this is the best fix, will need to think about it briefly.

For reference, according to [the docs ](https://docs.djangoproject.com/en/4.0/ref/models/fields/#django.db.models.RESTRICT) the behaviour should remain the same, except the protection is lifted by delete ting the related object in the same action. Removing the non add-on CartPosition by itself should remain impossible.

raphaelm:
Yes, I agree that from the docs it sounds like this is the solution. I'm hesitant with lifting this requirement on the model because I'm worried it might have unexpected consequences in other places, although I can currently not come up with any.

The reason I was surprised by this issue is mostly because we're already handling it here for single vouchers:

https://github.com/pretix/pretix/blob/069dd02ebc2847852daa8467fdffbc48dcd01411/src/pretix/control/views/vouchers.py#L246-L248

And here for bulk deletion:

https://github.com/pretix/pretix/blob/069dd02ebc2847852daa8467fdffbc48dcd01411/src/pretix/control/views/vouchers.py#L553-L554

I'd like to understand first why the issue shows up in the first place and I currently do not :D

MaicoTimmerman:
I was also surprised this showed up when looking at the lines you mentioned, however Django does not rely on database foreign key constraints. Instead, Django checks for related objects itself by issuing `SELECT` statements. The logic is contained [here](https://github.com/django/django/blob/main/django/db/models/deletion.py).

With `PROTECT` the error is raised regardless of the related object also being deleted. With `RESTRICT` the related objects are actually inspected, to see if they are also being deleted.

To re-iterate the actual issue here, this only happens if we are deleting 2 simultaneous `CartPostion`. `CartPosition` 1 is a product using the voucher and `CartPosition` 2 is using the voucher and has `CartPosition.addon_to_id == 1`. When only one of both `CartPosition` are using the voucher, no errors is thrown.

If you which to retain `PROTECT`, I think we can also change the lines you mentioned to first delete the CartPositions that are add-ons and afterwards delete the non add-ons products. Although, I've not verified this.

raphaelm:
> To re-iterate the actual issue here, this only happens if we are deleting 2 simultaneous CartPostion. CartPosition 1 is a product using the voucher and CartPosition 2 is using the voucher and has CartPosition.addon_to_id == 1. When only one of both CartPosition are using the voucher, no errors is thrown.

If that were correct, would not CartPosition 2 be already deleted first by `CartPosition.objects.filter(addon_to__voucher=self.object).delete() ` and then cartposition 1 is no longer referenced and can be deleted by `self.object.cartposition_set.all().delete() ` just fine?

Could it be that your case includes two different vouchers?

MaicoTimmerman:
Right, so I think we've both missed the actually mistake. In the bulk deletion case, it actually says `OrderPosition`, not `CartPosition`!

raphaelm:
Ooooh. This is bad, OrderPositions should obviously never be deleted (luckily, there is a .is_allowed() check that should prevent that).

PR #2489, opened 2022-02-23
Change restricted plugins from event-level action to org-level whitelist

raphaelm:


pc-coholic:
LGTM, but there is one breaking change:

Plugins that are restricted and were previously activated by a staff account for a specific event now cannot be en- or disabled by the same staff user unless they give the organizer-level permission to the organizer.

I think this is something we can ignore, as this should not touch too many accounts.

pc-coholic (approved):

raphaelm:
> Plugins that are restricted and were previously activated by a staff account for a specific event now cannot be en- or disabled by the same staff user unless they give the organizer-level permission to the organizer.

Yeah, I had a quick look what kinds of restricted plugins actually exist and I don't think this will be an issue.

PR #2490, opened 2022-02-23
Layout refresh for plugin settings page 

raphaelm:
Here's something else about the plugin management that has been bugging me for a while.

First, here's how this currently looks:

before

And here is what bugs me:

* The plugin page was designed with technical, self-hosting users in mind and exposes many technicalities. For example, a user who does not manage their own pretix instance but uses pretix Hosted or some other pretix managed by some other person will likely not care about the individual versions and authors of the plugins. I suggest getting rid of that information at least for pretix Hosted.
* The large green blocks with black-on-green text did not really look nice.
* The sorting is currently "pretix-bundled plugins first, then all other plugins, alphabetically within these two categories". However, that's not really a good sorting. Our plugins have significant differences in relevance which is not tied to how they are technically distributed. We have plugins that currently show up very close to the top like pretix-agenda that only provide a micro-feature relevant to 0,1% of events. We have other plugins that are way at the bottom like pretix-seating that we'd rather want every new pretix user to find out about quickly. We need another way of highlighting the most important plugins.

Here's how my proposal for a changed version looks like:

after

There are some things that are also on my mind and not solved by this:

- We really need to write better plugin descriptions.
- 
- We're starting to publish the first plugins who add additional cost to pretix Hosted users, such as pretixLEAD. We need to warn users about this at some point, which probably can just be added into the `description` by the respective plugins.


pc-coholic:
Just from looking at the screenshots, I do like what I'm seeing and I do like the idea of "promoting" plugins.

However, I would like to advocate to use the two-column layout for all plugins - not just promoted ones. Perhaps we can add some kind of box around the promoted section - or just sort them towards them to the top.

raphaelm:
I believe there must be a very, very obvious distinction between the promoted and the non-promoted ones. Currently, on the phone, I often instruct customers like "to activate seating, look at S, it's alphabetical" to speed up their search. This is obviously a lie currently since the two sections are sorted independently. If we "promote" more plugins, this will become worse. So (a) there must not be more than ~6 promoted plugins in one category before it becomes hard to read and (b) it must be absolutely clear without reading a word that the list restarts after the promoted plugins.

If we decide that we do want a two-column layout for everything, we'd need some big line or headline in between.

However, I'm not sure a two-column layout for everything is actually a good idea:

- It's much harder to scan a long two-column list for an alphabetical value than a single column one, at least for me
 
- If we do get better at writing descriptions and they get longer, the boxes will have different heights, leading to lots of empty spaces if the left box is larger than the right one. For 3-6 featured plugins this is okay-ish, for 20+ this looks very unclean

raphaelm:
Here's a new proposal, all with two columns, but only on very large (1650px+) screens:

after4

Also looks a little weird this way with the different logo sizes but I think it's fine?

I also added in a beta flag

after3


For context, on hosted, I would currently mark the following as featured:

- Features: Badges, Digital Content, Exhibitors, Pages, Seating, maybe sendmail
- Payment: Bank transfer, PayPal, Stripe
- Integrations: none? maybe digital covid certificates, venueless
- Customizations: none
- Output: PDF, Passbook
- API: none
- 
And the following as experimental:

- webcheckin, webinar, vacc-utils, keyper, portalum, semigator, plus some customer-specific ones

wiffbi:
As @raphaelm already pointed out:

> - It's much harder to scan a long two-column list for an alphabetical value than a single column one, at least for me
>
> - If we do get better at writing descriptions and they get longer, the boxes will have different heights, leading to lots of empty spaces if the left box is larger than the right one. For 3-6 featured plugins this is okay-ish, for 20+ this looks very unclean

I think I like the original approach of 3 to 6 featured plug-ins in two-column (or even more columns) mode with a different box-layout with a list of „normal“ plugins below better – although I would try to make theses featured plug-ins less „boxy“ (even just adding a purple border only to the left and let the other borders stay the same might help visually lessen the boxiness).

One advantage of the two-column approach is that the „activate“-button is not that far away from the rest of the description. I will look into this, whether we can have a bit of both world with single-column scanability and the buttons not being so far offset.
Other than that we could make use of two columns with one column for inactive and the other column for active plug-ins?

wiffbi:
Just an idea and quick sketch regarding multi-column layout for featured plug-ins and a reduced list-view for the rest. Is that a direction we might want to put some more effort into?

Bildschirmfoto 2022-03-14 um 11 28 20

raphaelm:
Yes, I like that a lot!

raphaelm:
@wiffbi If you already have that draft made, do you want to push it? :)

wiffbi:
Oh sorry, I just wanted to check how it looks on smaller screens, but will push ASAP

raphaelm:
Restricted plugins should probably also not have centered text any more:

Screenshot 2022-03-18 at 14-09-12 Landesgartenschau Bad Neuenahr-Ahrweiler 2022 pretix eu

And more importantly, the image size for non-featured plugins is… a little off ;)

Screenshot 2022-03-18 at 14-08-32 Landesgartenschau Bad Neuenahr-Ahrweiler 2022 pretix eu

wiffbi:
Ah, thanks, I did not have a non-featured plugin with an image. IMHO non-featured plugins should not have an image as in to make the featured ones pop out more. What do you think?

I will fix the centered-text bits.

raphaelm:
>Ah, thanks, I did not have a non-featured plugin with an image. IMHO non-featured plugins should not have an image as in to make the featured ones pop out more. What do you think?

I disagree. Many plugins, non-featured or featured, are integrations, and putting logos in makes it much easier for users to find the correct integration

wiffbi:
> I disagree. Many plugins, non-featured or featured, are integrations, and putting logos in makes it much easier for users to find the correct integration

You are right. I left image support for non-featured plugins in and just added some safe-guards regarding sizing off images.

PR #2493, opened 2022-02-24
PayPal Order v2 / ISU

pc-coholic:
This is the long awaited upgrade from the old PayPal Payment API to the new Orders v2 API, Smart Payment Buttons and ISU.

I tried keeping with the existing code as much as possible, but there are a few changes (subject to change)
- Webhooks do not process refund signals, as they only transmit the ID of the refund and only have HATEOAS towards the capture. The capture then will have a HATEOAS towards the Order (which we save in the ReferencedPayPalObject). So going from a refund webhook to the original payment will require at least 2 additional API-calls.
- There is no check right now for ISU, if the connecting PayPal account is ready to receive funds/the email address has been confirmed. We could add this to the isu_return, but I'm not sure we should keep people from linking their accounts.
- The transaction search is not yet implemented.
- QRC is not yet implemented (but we should probably still request the scopes for future-proofing)
- We do not transmit any shipping information - all orders are "No shipping". We could add this however with slight dependencies to pretix_shipping.
- Goes hand in hand with the one above: We do request `TRACKING_SHIPMENT_READWRITE` but don't use it yet.
- The quality and usability of the paypalcheckoutsdk (replaying paypalrestsdk) seems questionable at least.
- PayPal now absolutely requires Javascript as the redirect flow has been phased out. We could in theory still offer it - although only for Wallet payments and not for APMs.

Things that we definitely need to take another look before merging this:
- Compatibility with existing API-Key/non-SSO integrations
- Compatibility with existing SSO-Integrations
  - We already know that ISU will be a breaking change for SSO-users, since ISU is a special scope that is not widely available.
- Webhooks


src/pretix/plugins/paypal/payment.py
```diff
@@ -549,136 +777,92 @@ def execute_refund(self, refund: OrderRefund):
...
-                'error': str(e)
+                'error': str(ioe)
             })
-            raise PaymentException(_('Refunding the amount via PayPal failed: {}').format(str(e)))
-        if not pp_refund.success():
+            raise PaymentException(_('Refunding the amount via PayPal failed: {}').format(str(ioe)))
+        if response.result.status == 'COMPLETED':
+            req = OrdersGetRequest(refund.payment.info_data['id'])
+            response = self.client.execute(req)
+            refund.payment.info = json.dumps(response.result.dict())
```
raphaelm, line 820: This does not get saved. But also, it should not, refunding should not touch the historical orderpayment.

src/pretix/plugins/paypal/payment.py
```diff
@@ -549,136 +777,92 @@ def execute_refund(self, refund: OrderRefund):
...
+            refund.info = json.dumps(response.result.dict())
+            refund.done()
+        elif response.result.status == 'PENDING':
+            req = OrdersGetRequest(refund.payment.info_data['id'])
+            response = self.client.execute(req)
+            refund.payment.info = json.dumps(response.result.dict())
```
raphaelm, line 826: This does not get saved. But also, it should not, refunding should not touch the historical orderpayment.

src/pretix/plugins/paypal/payment.py
```diff
@@ -549,136 +777,92 @@ def execute_refund(self, refund: OrderRefund):
...
-                payee = {
-                    "email": request.event.settings.payment_paypal_connect_user_id,
-                    # If PayPal ever offers a good way to get the MerchantID via the Identifity API,
-                    # we should use it instead of the merchant's eMail-address
-                    # "merchant_id": request.event.settings.payment_paypal_connect_user_id,
-                }
+            try:
+                response = self.client.execute(req)
+            except IOError as ioe:
+                if isinstance(ioe, HttpError):
```
raphaelm, line 849: see above

src/pretix/plugins/paypal/static/pretixplugins/paypal/pretix-paypal.js
```diff
@@ -0,0 +1,280 @@
...
+    ready: function () {
+        if ($("input[name=payment][value=paypal_apm]").length > 0) {
+            pretixpaypal.renderAPMs();
+        }
+
+        $("input[name=payment][value^='paypal']").change(function () {
+            pretixpaypal.renderButton($(this).val());
+        });
+
+        $("input[name=payment]").not("[value^='paypal']").change(function () {
```
raphaelm, line 131: Is this safe? change is probably also called when one of the others is **un**checked which would disable paypal in the wrong moment?

src/pretix/plugins/paypal/payment.py
```diff
@@ -218,114 +276,168 @@ def settings_content_render(self, request):
...
+
+    def get_link(self, links, rel):
+        for link in links:
+            if link.rel == rel:
+                return link
+
+        return None
+
+
+class PaypalMethod(BasePaymentProvider):
```
raphaelm, line 336: Can `matching_id` and `api_payment_details` really stay the same if the info_data format changes a lot?

src/pretix/plugins/paypal/views.py
```diff
@@ -31,36 +31,71 @@
...
+                raise Http404('Unknown order')
+        except Order.DoesNotExist:
+            # Do a hash comparison as well to harden timing attacks
+            if 'abcdefghijklmnopq'.lower() == hashlib.sha1('abcdefghijklmnopq'.encode()).hexdigest():
+                raise Http404('Unknown order')
+            else:
+                raise Http404('Unknown order')
+        return super().dispatch(request, *args, **kwargs)
+
+    @property
```
raphaelm, line 84: we usually use a @cached_property for properties like this unless we really want a database query every time we use it

src/pretix/plugins/paypal/views.py
```diff
@@ -76,40 +111,134 @@ def redirect_view(request, *args, **kwargs):
...
+    template_name = ''
+
+    def post(self, request, *args, **kwargs):
+        if 'order' in self.kwargs:
+            order = self.request.event.orders.filter(code=self.kwargs['order']).select_related('event').first()
+            if order:
+                if order.secret.lower() == self.kwargs['secret'].lower():
+                    pass
+                else:
+                    order = None
```
raphaelm, line 127:
```suggestion
                if order.secret != self.kwargs['secret']:
                    order = None
```
I don't think these are case-insensitive in other places?

src/pretix/plugins/paypal/views.py
```diff
@@ -76,40 +111,134 @@ def redirect_view(request, *args, **kwargs):
...
+
     try:
-        tokeninfo = Tokeninfo.create(request.GET.get('code'))
-        userinfo = Tokeninfo.create_with_refresh_token(tokeninfo['refresh_token']).userinfo()
-    except paypalrestsdk.exceptions.ConnectionError:
-        logger.exception('Failed to obtain OAuth token')
-        messages.error(request, _('An error occurred during connecting with PayPal, please try again.'))
+        response = prov.client.execute(req)
+    except IOError as ioe:
+        if isinstance(ioe, HttpError):
```
raphaelm, line 208: see above

src/pretix/plugins/paypal/views.py
```diff
@@ -76,40 +111,134 @@ def redirect_view(request, *args, **kwargs):
...
+    getparams = ['merchantId', 'merchantIdInPayPal', 'permissionsGranted', 'accountStatus', 'consentStatus', 'productIntentID', 'isEmailConfirmed']
+    sessionparams = ['payment_paypal_isu_event', 'payment_paypal_isu_tracking_id']
+    if not any(k in request.GET for k in getparams) or not any(k in request.session for k in sessionparams):
         messages.error(request, _('An error occurred during connecting with PayPal, please try again.'))
         return redirect(reverse('control:index'))
 
-    event = get_object_or_404(Event, pk=request.session['payment_paypal_oauth_event'])
+    event = get_object_or_404(Event, pk=request.session['payment_paypal_isu_event'])
 
+    gs = GlobalSettingsObject()
```
raphaelm, line 196: I know we also didn't have this in the past, but maybe it makes sense here to check if you're still logged in as a user that is allowed to manipulate the settings of this event? it would be too bad to have an attack where I can reconfigure your event to my paypal account and get all your payments just because I somehow got the session manipulated (should be impossible, but defense in depth, bla)

src/pretix/plugins/paypal/views.py
```diff
@@ -216,58 +342,28 @@ def webhook(request, *args, **kwargs):
...
-            payment_info = p.info_data
-            for res in payment_info['transactions'][0]['related_resources']:
-                for k, v in res.items():
-                    if k == 'sale' and v['id'] == sale['id']:
-                        payment = p
-                        break
+            # Legacy PayPal info-data
+            if "purchase_units" not in p.info_data:
+                req = OrdersGetRequest(p.info_data['cart'])
+                response = prov.client.execute(req)
```
raphaelm, line 348: this call might be missing exception handling

raphaelm (commented):
had a first code read, did not actually test it, here are some preliminary comments :)

> This is the long awaited upgrade from the old PayPal Payment API to the new Orders v2 API, Smart Payment Buttons and ISU.

Yaaay :heart_eyes: 

> Webhooks do not process refund signals, as they only transmit the ID of the refund and only have HATEOAS towards the capture. The capture then will have a HATEOAS towards the Order (which we save in the ReferencedPayPalObject). So going from a refund webhook to the original payment will require at least 2 additional API-calls.

Hmm I think that effort is worth it. Not getting these webhooks would be kinda sad.

>    There is no check right now for ISU, if the connecting PayPal account is ready to receive funds/the email address has been confirmed. We could add this to the isu_return, but I'm not sure we should keep people from linking their accounts.

Agreed. People may just want to prepare their shop even though the account is not ready yet. I we want, we can add a warning, but I would not want to block it.

>    The transaction search is not yet implemented.
>    QRC is not yet implemented (but we should probably still request the scopes for future-proofing)

Yeah, these are wide out of scope for this PR

>    We do not transmit any shipping information - all orders are "No shipping". We could add this however with slight dependencies to pretix_shipping.

This hasn't been a problem for now, so I'd consider it out of scope for this PR.
We really do not want to introduce dependencies on plugins in core, so it would require quite some effort.

>    The quality and usability of the paypalcheckoutsdk (replaying paypalrestsdk) seems questionable at least.

Anything specific we should be aware of?

src/pretix/plugins/paypal/payment.py
```diff
@@ -134,6 +147,49 @@ def settings_form_fields(self):
                  )),
             ]
 
+        methods = [
+            ('method_wallet',
+             forms.BooleanField(
+                 label=_('PayPal'),
+                 required=False,
+                 help_text=_('ToDo: Explain that PayPal is always required (APM-only is not possible) since PayPal will be fallback.')
```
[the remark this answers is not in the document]
pc-coholic: Apparently Checkboxes cannot be `readonly`, so I'll go with `disabled` and set a hierarkey default instead.

src/pretix/plugins/paypal/payment.py
```diff
@@ -191,12 +255,6 @@ def settings_content_render(self, request):
                     }),
                     _('Disconnect from PayPal')
                 )
-        else:
-            settings_content = "<div class='alert alert-info'>%s<br /><code>%s</code></div>" % (
-                _('Please configure a PayPal Webhook to the following endpoint in order to automatically cancel orders '
```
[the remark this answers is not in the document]
pc-coholic: A victim of my misguided attempts of First Party ISU...

src/pretix/plugins/paypal/views.py
pc-coholic: Copy/Pasted from other pprov - seems like we are a little bit lax about this and have both `@property` and `@cached_property` in concurrent use across multiple payment providers.

pc-coholic:
> > Webhooks do not process refund signals, as they only transmit the ID of the refund and only have HATEOAS towards the capture. The capture then will have a HATEOAS towards the Order (which we save in the ReferencedPayPalObject). So going from a refund webhook to the original payment will require at least 2 additional API-calls.
> 
> Hmm I think that effort is worth it. Not getting these webhooks would be kinda sad.

I'll give them another try then.

However I did realize after drafting this PR, that this might be more challenging/require more entries in the `ReferencedPayPalObject`: While it is true that the refund-webhook have the HATEOAS-links pointing up to the charge, we  cannot just poll the charge as we do not have determined an event yet and - by extension - have not initialized the pprov and API with the correct credentialset.

One idea I had about this was to not only write the OrderID to the `ReferencedPayPalObject`, but also the ChargeID and any RefundIDs. If they all point to payment, I can then initialize the pprov with the event of that payment.

So even if we get a refund notification of a refund that we did not yet see (and have no trace of the refundID), we can extract the ChargeID from the HATEOAS-link, look with that for the payment and go from there.

> > The quality and usability of the paypalcheckoutsdk (replaying paypalrestsdk) seems questionable at least.
> 
> Anything specific we should be aware of?

Well, all the `except` calls are verbatim from their "best practise" (or to be more fair: examples)... And I did prefer the shorthand of the `paypalrestsdk` where I was able to just initialize an object with an ID and get the data - not having to manually querying stuff and handling exception-trees. Also, (and I'll have to check again), I think everything is a `HttpError` - no matter if the connection dropped out, the API is returning an undefined result or something else...

src/pretix/plugins/paypal/views.py
pc-coholic: This is verbatim from `/presale/views/order.py`: 
https://github.com/pretix/pretix/blob/19881dff443c5503afaaedc9b06b5504b6a4346e/src/pretix/presale/views/order.py#L102-L115

So I guess your choice how to proceed :-)

src/pretix/plugins/paypal/static/pretixplugins/paypal/pretix-paypal.js
pc-coholic: `.change()` is only firing for the element that is being selected. As an experiment, I added a `console.log()` to both events (line 127 and line 131) - it was only the newly selected item that fired.

The only issue I could see arise is if we have multiple payment providers that do these kinds of shenanigans and en-/disable the pretix-provided payment button at the same time...

src/pretix/plugins/paypal/payment.py
```diff
@@ -549,136 +825,92 @@ def execute_refund(self, refund: OrderRefund):
...
-                    # If PayPal ever offers a good way to get the MerchantID via the Identifity API,
-                    # we should use it instead of the merchant's eMail-address
-                    # "merchant_id": request.event.settings.payment_paypal_connect_user_id,
-                }
+            try:
+                req = OrdersGetRequest(paypal_order_id)
+                response = self.client.execute(req)
+            except IOError as e:
+                messages.warning(request, _('We had trouble communicating with PayPal'))
+                logger.error('OrdersGetRequest: {}'.format(str(e)))
```
raphaelm, line 896:
```suggestion
                logger.exception('PayPal OrdersGetRequest: {}'.format(str(e)))
```

src/pretix/plugins/paypal/views.py
```diff
@@ -76,40 +111,141 @@ def redirect_view(request, *args, **kwargs):
...
+                    pass
+                else:
+                    order = None
+        else:
+            order = None
+
+        # urlkwargs = {}
+        # if 'cart_namespace' in kwargs:
+        #     urlkwargs['cart_namespace'] = kwargs['cart_namespace']
+
```
raphaelm, line 134:
```suggestion
```

src/pretix/plugins/paypal/views.py
```diff
@@ -76,40 +111,141 @@ def redirect_view(request, *args, **kwargs):
...
+        r._csp_ignore = True
+        return r
+
+
+@method_decorator(xframe_options_exempt, 'dispatch')
+class PayView(PaypalOrderView, TemplateView):
+    template_name = ''
+
+    def get(self, request, *args, **kwargs):
+        if self.payment.state not in [OrderPayment.PAYMENT_STATE_CREATED, OrderPayment.PAYMENT_STATE_PENDING]:
```
raphaelm, line 167: Is PENDING really allowed here?

src/pretix/plugins/paypal/views.py
```diff
@@ -76,40 +111,141 @@ def redirect_view(request, *args, **kwargs):
...
-    except paypalrestsdk.exceptions.ConnectionError:
-        logger.exception('Failed to obtain OAuth token')
+        req = PartnersMerchantIntegrationsGetRequest(
+            gs.settings.get('payment_paypal_connect_partner_merchant_id'),
+            request.GET.get('merchantIdInPayPal')
+        )
+        response = prov.client.execute(req)
+    except IOError as e:
         messages.error(request, _('An error occurred during connecting with PayPal, please try again.'))
+        logger.error('PartnersMerchantIntegrationsGetRequest: {}'.format(str(e)))
```
raphaelm, line 209:
```suggestion
        logger.exception('PayPal PartnersMerchantIntegrationsGetRequest: {}'.format(str(e)))
```

src/pretix/plugins/paypal/views.py
```diff
@@ -76,40 +111,141 @@ def redirect_view(request, *args, **kwargs):
...
+            else:
+                messages.error(request, _('An error occurred during connecting with PayPal, please try again.'))
+        else:
+            if response.result.tracking_id != request.session['payment_paypal_isu_tracking_id']:
+                messages.error(request, _('An error occurred during connecting with PayPal, please try again.'))
+            else:
+                if request.GET.get("isEmailConfirmed") == "false":  # Yes - literal!
+                    messages.warning(
+                        request,
+                        _('The E-Mail address on your PayPal account has not yet been confirmed. You will need to do '
```
raphaelm, line 224:
```suggestion
                        _('The e-mail address on your PayPal account has not yet been confirmed. You will need to do '
```

src/pretix/plugins/paypal/views.py
```diff
@@ -176,18 +312,19 @@ def webhook(request, *args, **kwargs):
...
     # We do not check the signature, we just use it as a trigger to look the charge up.
-    if event_json['resource_type'] not in ('sale', 'refund'):
+    if event_json['resource_type'] not in ["checkout-order", "refund"]:
         return HttpResponse("Not interested in this resource type", status=200)
 
-    if event_json['resource_type'] == 'sale':
-        saleid = event_json['resource']['id']
+    # Retrieve the Charge ID of the refunded payment
+    if event_json['resource_type'] == 'refund':
+        payloadid = get_link(event_json['resource']['links'], 'up')['href'].split('/')[-1]
```
raphaelm, line 320: Actual handling of refunds seems to be missing

raphaelm (changes requested):

src/pretix/plugins/paypal/signals.py
```diff
@@ -67,19 +77,94 @@ def pretixcontrol_logentry_display(sender, logentry, **kwargs):
...
+            (url.url_name == "event.checkout" and url.kwargs['step'] == "payment") or
+            (url.namespace == "plugins:paypal" and url.url_name == "pay")
+    ):
+        if 'Content-Security-Policy' in response:
+            h = _parse_csp(response['Content-Security-Policy'])
+        else:
+            h = {}
+
+        csps = {
+            'script-src': ['https://www.paypal.com', "'nonce-{}'".format(_nonce(request))],
```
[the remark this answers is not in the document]
pc-coholic: As discussed via Slack: postponed for another day.

src/pretix/plugins/paypal/payment.py
```diff
@@ -134,6 +148,68 @@ def settings_form_fields(self):
...
+                     attrs={
+                         'data-checkbox-dependency': '#id_payment_paypal_method_apm',
+                     }
+                 )
+             )),
+            ('enable_method_paylater',
+             forms.BooleanField(
+                 label=_('Enable Buy Now Pay Later'),
+                 help_text=_(
+                     'Offer your customers the possibility to buy now (up to 1.000€) and pay in 3/6/12/24 installments '
```
raphaelm, line 200:
```suggestion
                     'Offer your customers the possibility to buy now (up to a certain limit) and pay in multiple installments '
```
Let's maybe not hardcode things that are likely country/currency-specific or could be changed by paypal any day?

raphaelm (approved):

PR #2494, opened 2022-02-24
Fix #1749 -- Stripe: Rewrite for Payment Methods and Payment Intents

### LINKED ISSUE #1749: Support for stripe SEPA
Hi team,

first of all, thanks for the great product and continous development! We have recently started using stripe SEPA on another system and think it is a great way to automate SEPA payments. The burden to get activated is quiet low as well, so I was wondering if it would be possible to add it to pretix (as Germans still prefer the SEPA-way over credit card?). I have found issue #407 which is marked as closed, but it seems like SEPA was never actually added. 

Thanks for your consideration!

pc-coholic:


src/pretix/plugins/stripe/payment.py
```diff
@@ -698,59 +778,33 @@ def shred_payment_info(self, obj: OrderPayment):
...
-            'settings': self.settings,
-            'is_moto': self.is_moto(request)
-        }
-        return template.render(ctx)
+class StripePaymentIntentMethod(StripeMethod):
+    identifier = ''
+    method = ''
 
     def payment_is_valid_session(self, request):
-        return request.session.get('payment_stripe_payment_method_id', '') != ''
```
raphaelm, line 723: Changing this and similar things is of course possible and creates a cleaner codebase, but we have a risk to think about with people who are performing a checkout while we ware deploying the change. They might have already entered their card number and then the session gets lost during purchase? We might want to deploy this at night, but still, people are buying with Stripe all the time. (Similar concerns with the PayPal PR.)

src/pretix/plugins/stripe/payment.py
```diff
@@ -698,59 +778,33 @@ def shred_payment_info(self, obj: OrderPayment):
...
-        payment_method_id = request.POST.get('stripe_payment_method_id', '')
-        request.session['payment_stripe_payment_method_id'] = payment_method_id
-        request.session['payment_stripe_brand'] = request.POST.get('stripe_card_brand', '')
-        request.session['payment_stripe_last4'] = request.POST.get('stripe_card_last4', '')
+        payment_method_id = request.POST.get('stripe_{}_payment_method_id'.format(self.method), '')
+        request.session['payment_stripe_{}_payment_method_id'.format(self.method)] = payment_method_id
+
         if payment_method_id == '':
-            messages.warning(request, _('You may need to enable JavaScript for Stripe payments.'))
+            messages.warning(request, _('You may need to enable JavaScript for Stripe payments.2'))
```
raphaelm, line 793:
```suggestion
            messages.warning(request, _('You may need to enable JavaScript for Stripe payments.'))
```

src/pretix/plugins/stripe/payment.py
```diff
@@ -950,6 +1048,109 @@ def payment_presale_render(self, payment: OrderPayment) -> str:
...
+            'form': self.payment_form(request),
+            'email': order.email if order else cs.get('email', '')
+        }
+        return template.render(ctx)
+
+    @property
+    def payment_form_fields(self):
+        return OrderedDict(
+            [
+                ('accountname',
```
raphaelm, line 1074: Can we find a way to prefill this and all of the other fields from the invoice address from the cart or order?

src/pretix/plugins/stripe/static/pretixplugins/stripe/pretix-stripe.js
```diff
@@ -107,25 +112,65 @@ var pretixstripe = {
...
+                                },
+                            },
+                            supportedCountries: ['SEPA'],
+                            classes: {
+                                focus: 'is-focused',
+                                invalid: 'has-error',
+                            }
+                        });
+                        pretixstripe.sepa.on('change', function(event) {
+                           if (['AD', 'PF', 'TF', 'GI', 'GB', 'GG', 'VA', 'IM', 'JE', 'MC', 'NC', 'BL', 'PM', 'SM', 'CH', 'WF'].includes(event.country)) {
```
raphaelm, line 143: Can we add a comment where this list of countries comes from, how it is defined and under what conditions it might need to be updated?

src/pretix/plugins/stripe/static/pretixplugins/stripe/pretix-stripe.js
```diff
@@ -196,28 +241,47 @@ $(function () {
         return;
 
     if ($("input[name=payment][value=stripe]").is(':checked') || $(".payment-redo-form").length) {
-          pretixstripe.load();
+        pretixstripe.load();
     } else {
         $("input[name=payment]").change(function () {
-            if ($(this).val() === 'stripe') {
+            if (['stripe', 'stripe_sepa_debit'].includes($(this).val())) {
```
raphaelm, line 247: We have not yet dropped IE11 support for the most fundamental features like "paying for a ticket". `includes` is unfortunately not supported by IE11, use indexOf or a plain ||

src/pretix/plugins/stripe/static/pretixplugins/stripe/pretix-stripe.js
```diff
@@ -196,28 +241,47 @@ $(function () {
...
             return false;
         }
     );
 
     if ($("#stripe-current-card").length) {
-        $("#stripe-elements").hide();
+        $("#stripe-card-elements").hide();
+    }
+
+    $("#stripe_other_account").click(
```
raphaelm, line 268: Should we include sepa_debit in the name of these IDs as well in case we have other stripe methods later that also use the word "account"?

raphaelm (commented):
some comments after reading the code, no manual test yet

pc-coholic:
As discussed, there is still one issue: `StripePaymentIntentMethod` assumes right now, that there is always a `payment_method` collected through the customer facing pages which is of course not true.

For now, we could just go ahead with the code "as-is", since except Credit Card and SEPA all payments are still handled using `sources`.

If we want to migrate all payment methods to Payment Intents (which is probably a good idea in the long run), we would need to
a) Either: Split into `StripePaymentIntentMethod` and `StripePaymentIntentWithPaymentMethodMethod`
b) Or: Add a `pm_required`-property to `StripePaymentMethod` and set it accordingly.

As a quick and dirty test, I trialed option B by
- Adding `pm_required = False` to `StripePaymentIntentMethod`
- Added  `pm_required = True` to `StripeCC`
- Added  `pm_required = True` to `StripeSEPADirectDebit`
- Modified `StripePaymentIntentMethod`:`execute_payment()` to only delete the payment-method-id session-key if `self.pm_required` is set
- Modified `StripePaymentIntentMethod`:`_handle_payment_intent()` to only set `payment_method` if `self.pm_required` is set
- Modified `StripePaymentIntentMethod`:`_handle_payment_intent()` to build the `idempotency_key` without the payment_method_id if `self.pm_required` is not set
- Changed the Base-method of `StripeGiropay` to `StripePaymentIntentMethod`
- Removed `StripeGiropay`'s `_create_source()`
- Added `_payment_intent_kwargs()`  to `StripeGiropay` to return a `payment_method_data`-dict with the account owners name

This seems to be working fine (minus a CSP-related issue when iframing the Stripe-generated payment approval page) - so we could potentially also add these fixes already in this PR...

src/pretix/plugins/stripe/static/pretixplugins/stripe/pretix-stripe.js
```diff
@@ -112,25 +117,67 @@ var pretixstripe = {
...
+                            if (['AD', 'PF', 'TF', 'GI', 'GB', 'GG', 'VA', 'IM', 'JE', 'MC', 'NC', 'BL', 'PM', 'SM', 'CH', 'WF'].indexOf(event.country) > 0) {
+                                $("#stripe_sepa_debit_country").prop('checked', true);
+                                $("#stripe_sepa_debit_country").change();
+                            } else {
+                                $("#stripe_sepa_debit_country").prop('checked', false);
+                                $("#stripe_sepa_debit_country").change();
+                            }
+                            if (event.bankName) {
+                                $("#stripe_sepa_debit_bank").val(event.bankName);
+                       }
```
raphaelm, line 159: Code formatting broke here

src/pretix/plugins/stripe/payment.py
```diff
@@ -993,6 +1103,131 @@ def payment_presale_render(self, payment: OrderPayment) -> str:
...
+                     label=_('Account Holder Country'),
+                     required=False,
+                     choices=CachedCountries(),
+                     widget=forms.Select(
+                         attrs={
+                             'data-display-dependency': '#stripe_sepa_debit_country',
+                             'data-required-if': '#stripe_sepa_debit_country'
+                         }
+                     ),
+                     initial=self.ia.country,
```
raphaelm, line 1198: If the invoice address does not exist, could we use ``guess_country()`` here as we do otherwise? Or could we guess the country from the IBAN?

src/pretix/plugins/stripe/templates/pretixplugins/stripe/checkout_payment_form_sepadirectdebit.html
```diff
@@ -0,0 +1,74 @@
...
+{% load bootstrap3 %}
+
+<div class="form-horizontal stripe-container">
+    {% if is_moto %}
+        <h1>
+            <span class="label label-info pull-right flip" data-toggle="tooltip_html" title="{% trans "This transaction will be marked as Mail Order/Telephone Order, exempting it from Strong Customer Authentication (SCA) whenever possible" %}">MOTO</span>
+        </h1>
+        <div class="clearfix"></div>
+    {% endif %}
+
```
raphaelm, line 11:
```suggestion
```

PR #2500, opened 2022-02-28
Group identical add-ons in cart

raphaelm:
This is an attempt to give a nicer view in the cart if someone buys the exact same addon 5x for the same top-level product, while at the same time keeping all other important properties of the grouping and sorting intact. @wiffbi can you check if you think this breaks anything?

### DIFF
--- a/src/pretix/base/email.py
+++ b/src/pretix/base/email.py
@@ -177,7 +177,7 @@ class TemplateBasedMailRenderer(BaseHTMLMailRenderer):
                     op.variation,
                     op.subevent,
                     op.attendee_name,
-                    (op.pk if op.addon_to_id else None),
+                    op.addon_to_id,
                     (op.pk if op.has_addons else None)
                 )
             )]
--- a/src/pretix/base/models/orders.py
+++ b/src/pretix/base/models/orders.py
@@ -1330,6 +1330,10 @@ class AbstractPosition(models.Model):
         else:
             return {}
 
+    @property
+    def item_and_variation(self):
+        return self.item, self.variation
+
     @meta_info_data.setter
     def meta_info_data(self, d):
         self.meta_info = json.dumps(d)
--- a/src/pretix/base/templates/pretixbase/email/order_details.html
+++ b/src/pretix/base/templates/pretixbase/email/order_details.html
@@ -90,13 +90,16 @@
                         {% for groupkey, positions in cart %}
                             <tr>
                                 <td>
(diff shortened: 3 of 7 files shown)


wiffbi:
I really like this. The cart looks better with this, yes! A few things I stumbled upon:

UI/UX: In cart-overview IMHO leaving out line.count 1 when there is another group with line.count > 1 looks a bit odd.

Bildschirmfoto 2022-03-01 um 09 27 38

Adding line.count==1 as well IMHO looks better. Although ideally we only add the line.count output for line.count==1 when any group has line.count > 1, so when only having one add-on of a kind, it looks the same as currently (ie. no line.count)

Bildschirmfoto 2022-03-01 um 09 27 17


When having added multiple add-ons and returning to the event’s homepage, I can remove add-ons until it reaches line.count==1. Then I cannot edit it anymore. Furthermore I cannot increase line.count – I get the error, that I can only add this product to the cart with another product (basically the addon-only-error-message). IMHO we should either allow all addons to be freely added/removed or have add-ons’ line.count not at all be editable in the cart-edit-view but only in its corresponding checkout-step. What do you think?

Bildschirmfoto 2022-03-01 um 09 27 02

raphaelm:
> Adding line.count==1 as well IMHO looks better.

I'm torn. I like that the column is empty in the "more likely case" currently, but we could drop it for consistency.

> Although ideally we only add the line.count output for line.count==1 when any group has line.count > 1, so when only having one add-on of a kind, it looks the same as currently (ie. no line.count)

That sounds a little complicated.

> When having added multiple add-ons and returning to the event’s homepage, I can remove add-ons until it reaches line.count==1. Then I cannot edit it anymore. Furthermore I cannot increase line.count – I get the error, that I can only add this product to the cart with another product (basically the addon-only-error-message). IMHO we should either allow all addons to be freely added/removed or have add-ons’ line.count not at all be editable in the cart-edit-view but only in its corresponding checkout-step. What do you think?

This is a bug, add-ons should not at all be editable

wiffbi:
>> Although ideally we only add the line.count output for line.count==1 when any group has line.count > 1, so when only having one add-on of a kind, it looks the same as currently (ie. no line.count)
>
> That sounds a little complicated.

Well, as we do not have a „group“ of items this is indeed complicated, especially inside a django-template. So let’s leave it optimized for the most usual case as you already did: one addon per add-on-kind.

I fixed the editable-bug in cart-view.

PR #2510, opened 2022-03-02
Discounts

raphaelm:
@ Reviewers if you are reading this PR, best start with the documentation changes :)

- [x] Data model
- [x] Backend UI
- [x] Offer REST API
- [x] CartPosition REST API
- [x] Make all CartPosition operations work again
  - [x] add_new_items
  - [x] extend_expired
  - [x] apply_voucher
  - [x] set_addons
- [x] Display a message to user after discount has been applied?
- [x] (How to) deal with self-service order changes?
- [x] check for problems with order changes
- [x] Verify tax rule changes are still fine
- [x] Verify bundled price handling is still fine
- [ ] Verify voucher budget handling is still fine and consistent between orderposition and cartposition
  - [x] Do we maybe want to change the behavior? How bad would be the impact?
- [x] Test enforcement on minimum free price in the same case as test_free_price_net_price_reverse_charge_keep_gross
- [x] Actual offers engine
- [x] Exclude unsupported combinations (min_value XOR min_count, distinct subevents XOR min_value, cheapest_one XOR min_value)
- [x] Developer docs on algorithm
- [x] Processing during order submit
- [x] Processing during API order creation
- [x] Implement logdisplay
- [x] Implement timeline
- [x] Tests, tests, tests
  - [x] cart management
  - [x] order creation
  - [x] backend views
- [x] Update product structure guide
- [x] Remove legacy fields
- [x] Remove print statements
- [ ] Prepare for migration
  - [ ] Document and announce behaviour changes
- [ ] combine migrations


src/pretix/control/navigation.py
```diff
@@ -186,6 +186,14 @@ def get_event_navigation(request: HttpRequest):
...
                     'active': 'event.items.questions' in url.url_name,
                 },
+                {
+                    'label': _('Discounts'),
+                    'url': reverse('control:event.items.discounts', kwargs={
+                        'event': request.event.slug,
+                        'organizer': request.event.organizer.slug,
+                    }),
+                    'active': 'event.items.discounts' in url.url_name,
+                },
```
pc-coholic, line 196: Before checking the source, it took me quite some time to find this menu item... I think we should consider moving this closer to the Vouchers menu item.

src/pretix/control/forms/discounts.py
```diff
@@ -0,0 +1,99 @@
...
+            'sales_channels',
+            'available_from',
+            'available_until',
+            'subevent_mode',
+            'condition_all_products',
+            'condition_limit_products',
+            'condition_min_count',
+            'condition_min_value',
+            'condition_apply_to_addons',
+            'benefit_discount_matching_percent',
```
pc-coholic, line 50: Eyecandy: Add a `attrs={'addon_after': _('%')}` to the widget

src/pretix/base/models/discount.py
```diff
@@ -0,0 +1,355 @@
...
+        default=Decimal('0.00'),
+        validators=[MinValueValidator(Decimal('0.00'))],
+    )
+    benefit_only_apply_to_cheapest_n_matches = models.PositiveIntegerField(
+        verbose_name=_('Apply discount only to this number of matching products'),
+        help_text=_('Keep the field empty to apply the discount to all matching products.'),
+        null=True,
+        blank=True,
+        validators=[MinValueValidator(1)],
+    )
```
pc-coholic, line 128: To me this reads more like a Jeopardy clue ;-)

I think it needs more explanation along the lines of "Setting this to two will apply the discount only to groups of 2 tickets. Buying three tickets would grant two tickets with discount, one at full price. Four tickets though would give 4 discounted tickets".

src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
```diff
@@ -212,6 +212,13 @@
                     {% else %}
                         {{ line.price|money:event.currency }}
                     {% endif %}
+                    {% if line.discount and line.line_price_gross != line.price %}
+                        <span class="text-success discounted" data-toggle="tooltip" title="{% trans "The price of this product was reduced because of an automatic discount." %}">
```
pc-coholic, line 216: Perhaps we could give the customer a little bit more details on to why this is discounted... We should be able to create a human readable version of the discount rule?

src/pretix/control/forms/discounts.py
```diff
@@ -0,0 +1,99 @@
...
+        fields = [
+            'active',
+            'internal_name',
+            'sales_channels',
+            'available_from',
+            'available_until',
+            'subevent_mode',
+            'condition_all_products',
+            'condition_limit_products',
+            'condition_min_count',
```
pc-coholic, line 47: Having this field empty will hide the `benefit_only_apply_to_cheapest_n_matches` field.

But having `condition_min_count` empty and `condition_min_value` set, will fail the client-side validation, as `min_count` cannot be empty.

pc-coholic (changes requested):
I do think that only granting the discounts in the cart is good - but I can also see that it might be confusing for customers too...

Assume you have a BOGO-discount and the customer adds only one item to their cart.

In this case, they could get a second ticket for free - but in the item list (landing page and AddOn-step) they are still displayed at full price. The user friendly thing would be to display the item price as free..

However I do see the issue, as if it was a "true" BOGO (only every 2 items), the first additional one would be free, the second would be full price. 

So yeah - probably no way to do this right without printing pricing charts for the customers ;-)

------

I'm missing in the backend some information for the event organizer as to how the priced came to be: if a voucher is redeemed, the OrderPosition is linked to that voucher. Perhaps the OP could also get another FK to the Discount.

------

Discounts are stacked on top of vouchers...

A BOGO (50% discount) on an 10€ item that I can already redeem a 50% discount for, will yield me 2 items for 5 Euros.

This might be desirable, but very often such discounts are not allowed. Perhaps we should add a "do not discount vouchers any further"-option to the discount or the vouchers.

------

I'm almost certain we'll see a FR for a maximum discount budget very, very quickly.

src/pretix/control/navigation.py
raphaelm: Uh… interesting. But … they are not vouchers. Happy for further input here, eg from @wiffbi

pc-coholic: "closer" not "put it in there" ;-)

Even though those are two very different concepts, I feel like they somehow work together... :shrug:

raphaelm: Hmm, I'm still not sure, especially since in another comment you ask for an option that prevents mixing of discounts and vouchers :see_no_evil: I feel that they are part of "pricing" and "pricing" is something that belongs as much to "products" as e.g. quotas do.

Also, what options do we have? If we don't want to put it under products and not under vouchers, it needs to be a top-level item? I don't feel that is warranted, more top-level items make it harder for new users to get an overview and this is *not* a feature you need to discover before you disovered quotas.

src/pretix/base/models/discount.py
raphaelm: I don't think your example is quite perfect yet, "Setting this to two will apply the discount only to groups of 2 tickets" is confusing, because with condition_min_count=3 and benefit_only_apply_to_cheapest_n_matches=2, it will *look* for groups of *three*. However, I pushed a new text inspired by it, do you think it's better?

src/pretix/control/forms/discounts.py
raphaelm: done

src/pretix/control/forms/discounts.py
raphaelm: Ah yes, since the field is not nullable it is expecting  0 in the form. Form is fixed.

src/pretix/base/models/discount.py
pc-coholic: :+1:

src/pretix/control/navigation.py
pc-coholic: Hm, if you say it like that, I do agree 100% - it is part of the pricing and that does indeed live within the products-area...

Sometimes it helps just talking about it ;-)

src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
raphaelm: Hmm, this sounds quite tough. Dynamical string building is very hard to do in a way that works across languages. Is this really something users will expect? Do other online shops do this? :thinking:

ser8phin: This was something I noticed as well. Is there a way to easily make it possible to enter custom strings when creating the discount rule, comparable to shop descriptions and such, so a shop owner can set their own explanation in their configured languages?
I do think other shops use discounts as incitement, so they would make a grand proclamation somewhere, that it is possible to get discounts. This is not the place to advertise them, so it would also be fine if left as is, I believe.

raphaelm: I did not get around to this before the merge now, but I think the custom-set explanation ("public description") is a great solution! We can always implement it later on. Would you maybe interested in doing that, @ser8phin? :)

PR #2511, opened 2022-03-03
Fix -- Custom favicon not showing correctly in Safari (#Z181607)

wiffbi:
This PR fixes some favicon URLs not using the custom favicon.

Furthermore it removes the unnecessary 194x194-size, which makes only sense, if a custom favicon for e.g. Android Chrome with e.g. a drop-shadow were designed, that we do not expect to be shown anywhere else. As this is not the case with neither default pretix-icons nor the generated custom icons, we can safely remove this icon size.

The same with mask-icon, as Safari >12 can use a normal favicon instead – even apple.com does not use the mask-icon anymore.

If we at some point drop support for older IEs, we could also drop the `<link rel="shortcut icon"` line. IMHO it currently does not hurt much to keep it in for now.

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/base.html
+++ b/src/pretix/presale/templates/pretixpresale/base.html
@@ -27,6 +27,8 @@
         <link rel="icon" type="image/png" sizes="32x32" href="{{ settings.favicon|thumb:'32x32^' }}">
         <link rel="icon" type="image/png" sizes="194x194" href="{{ settings.favicon|thumb:'194x194^' }}">
         <link rel="icon" type="image/png" sizes="16x16" href="{{ settings.favicon|thumb:'16x16^' }}">
+        <link rel="icon" type="image/png" sizes="192x192" href="{% settings.favicon|thumb:'192x192^' %}">
+        <link rel="apple-touch-icon" sizes="180x180" href="{% settings.favicon|thumb:'180x180^' %}">
     {% elif development_warning or debug_warning %}
         <link rel="shortcut icon" href="{% static "pretixbase/img/favicon-debug.ico" %}">
     {% else %}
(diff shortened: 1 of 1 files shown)


raphaelm (approved):

src/pretix/presale/templates/pretixpresale/base.html
```diff
@@ -24,22 +24,21 @@
     {% if settings.favicon %}
         <link rel="icon" href="{{ settings.favicon|thumb:'16x16^' }}">
         <link rel="shortcut icon" href="{{ settings.favicon|thumb:'16x16^' }}">
-        <link rel="icon" type="image/png" sizes="32x32" href="{{ settings.favicon|thumb:'32x32^' }}">
-        <link rel="icon" type="image/png" sizes="194x194" href="{{ settings.favicon|thumb:'194x194^' }}">
         <link rel="icon" type="image/png" sizes="16x16" href="{{ settings.favicon|thumb:'16x16^' }}">
+        <link rel="icon" type="image/png" sizes="32x32" href="{{ settings.favicon|thumb:'32x32^' }}">
+        <link rel="icon" type="image/png" sizes="192x192" href="{% settings.favicon|thumb:'192x192^' %}">
+        <link rel="apple-touch-icon" sizes="180x180" href="{% settings.favicon|thumb:'180x180^' %}">
```
raphaelm, line 30:
```suggestion
        <link rel="icon" type="image/png" sizes="192x192" href="{{ settings.favicon|thumb:'192x192^' }}">
        <link rel="apple-touch-icon" sizes="180x180" href="{{ settings.favicon|thumb:'180x180^' }}">
```

raphaelm (changes requested):

wiffbi:
Thanks for fixing the header output. I missed to push my latest local change.

PR #2512, opened 2022-03-03
Add the .mta file extension for Volksbank's MT940

henrybrink:
Volksbank's in Germany are exporting MT940-Documents with the file extension ".mta" instead of ".sta". While the format is valid, pretix would mark those files as invalid, as long the user does not rename it to one of the allowed.

This PR adds the .mta file to the list of recognized file extensions in the banktransfer plugin.


raphaelm (commented):
Thanks for your contribution!

Issue #2258, opened 2021-10-21
Error editing ticket amount after added to cart using Voucher


theschweitz:
**Describe the bug**
When using a Voucher code that is good for X number of redemptions, Pretix allows the user to add X number of tickets to their cart as displayed in Screenshot 1. But - if the user accidentally selects more/less tickets than desired & wants to use the - or + buttons after it is added to the cart, Pretix gives the error in Screenshot two, which is very confusing for the user, as the Voucher is approved to use up to X amount of tickets.

**Expected behavior**
Pretix should recognize the Voucher as redeemable up to X amount  & allow the user to add/remove tickets after the first screen. Currently if the user makes any mistakes in how many tickets they want to order, they will need to Empty their Cart, and start the process over, which is very inconvenient.

**Screenshots**
<img width="1153" alt="Screen Shot 2021-10-21 at 1 00 52 PM" src="https://user-images.githubusercontent.com/77123385/138277184-9dfae733-f016-46c1-988e-06d4d260d966.png">

<img width="1155" alt="Screen Shot 2021-10-21 at 1 01 12 PM" src="https://user-images.githubusercontent.com/77123385/138277225-492c79c7-8ef3-4cc7-b110-ab3081b431dc.png">

PR #2515, opened 2022-03-05
Add the _voucher_code field to the add to cart form

henrybrink:
As mentioned in #2258, currently a user won't be able to add an item via the plus button in the cart view, when this item requires a voucher.
This PR fixes this issue, by adding the "_voucher_code" field to the form the plus button uses to add an item to the cart.

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
+++ b/src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
@@ -253,6 +253,9 @@
                             <input type="hidden" name="price_{{ line.item.id }}"
                                     value="{% if event.settings.display_net_prices %}{{ line.bundle_sum_net }}{% else %}{{ line.bundle_sum }}{% endif %}" />
                         {% endif %}
+                        {% if line.voucher %}
+                            <input type="hidden" name="_voucher_code" value="{{ line.voucher.code }}">
+                        {% endif %}
                         <button class="btn btn-mini btn-link {% if line.seat %}btn-invisible{% endif %}" title="{% blocktrans with item=line.item %}Add one more {{item}} to your cart{% endblocktrans %}" {% if line.seat %}disabled{% endif %}>
                             <i class="fa fa-plus" aria-hidden="true"></i>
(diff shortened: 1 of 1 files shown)


raphaelm:
Hmm, this solves the case of "item requires the voucher", but it breaks the case of "single-use voucher that just gives a discount", so I'm not sure if it's worth it. @wiffbi Do you have an opinion here?

henrybrink:
Thats with the patch, when adding more items than the voucher is allowed for.
For example, when I have an item in my cart and the voucher can only be used once, this message would appear when I click on the "plus". When the voucher could be used multiple times, the plus works as long it won't reach the limit, after that this message would appear.

I maybe missed one case, but because I only modified the frontend code, even if there is a validation error with the voucher, it will be there without the code I added and someone could just add the "_voucher_code" field via the developer tools to the form and the system would use this voucher.

raphaelm:
> Thats with the patch, when adding more items than the voucher is allowed for.
For example, when I have an item in my cart and the voucher can only be used once, this message would appear when I click on the "plus". When the voucher could be used multiple times, the plus works as long it won't reach the limit, after that this message would appear

Yes. We're in agreement about the facts here. 

However, without your patch, it behaves differently. Currently, if you have a voucher that can only be used once, it does NOT give an error, but it adds the product without the error.

So all I'm saying is that your patch fixes an error message in one situation, but introduces a new error message in another situation and I'm not yet sure whether that is an improvement.

henrybrink:
Ah, I'm realizing what you meant. That's really not ideal, I would think this issue is more rare than the issue it would cause, you are right!

wiffbi:
Hm, it is indeed tricky. With this patch, you enable to increase the number of items with a voucher to be ordered from within the cart, but if you want to add more than the voucher allows, you need to go through the event’s start page/product list. Without this patch, if you increase the number of items with a voucher from within the cart, it does add them without the voucher. To add a product with the voucher, you need to go the original route again with the „Redeem a voucher“ and the adding the products. Both ways are reasonable although I prefer the „intentional“ way it is now, although in some cases it might seem a bit complicated. Ideally the frontend would know whether the voucher on the item would allow for one additional item and add your change (ir. the voucher.code in a hidden input) conditionally, so it adds the item with voucher if possible and without if not. But I guess this makes the code way more complex for a rather edge-case.

raphaelm:
Let's keep it as-is for now, then.

PR #2523, opened 2022-03-09
Unicode-aware sorting of phone prefixes

raphaelm:
This will put Österreich between Oman and Pakistan as expected, instead of to the end of the list. Should also highly improve country sorting for entirely-non-latin scripts like Arabic.

### DIFF
--- a/src/pretix/base/forms/questions.py
+++ b/src/pretix/base/forms/questions.py
@@ -41,7 +41,6 @@ from io import BytesIO
 import dateutil.parser
 import pycountry
 import pytz
-from babel import Locale
 from django import forms
 from django.conf import settings
 from django.contrib import messages
@@ -52,7 +51,6 @@ from django.core.validators import (
 )
 from django.db.models import QuerySet
 from django.forms import Select, widgets
-from django.utils import translation
 from django.utils.formats import date_format
 from django.utils.html import escape
 from django.utils.safestring import mark_safe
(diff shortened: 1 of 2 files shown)


src/pretix/base/forms/questions.py
```diff
@@ -264,17 +264,13 @@ class WrappedPhonePrefixSelect(Select):
 
     def __init__(self, initial=None):
         choices = [("", "---------")]
-        language = get_babel_locale()  # changed from default implementation that used the django locale
-        locale = Locale(translation.to_locale(language))
+
         for prefix, values in _COUNTRY_CODE_TO_REGION_CODE.items():
```
wiffbi, line 268: Thanks to your optimization and removing the country_code bit, we could skip this for-loop entirely if initial is None. Furthermore we could `break` after setting self.initial. I could not make a code suggestion as indenting the loop would go over a removed line. If you want, I can add the change as a whole commit myself. Here is what I would suggest:

```
    def __init__(self, initial=None):
        choices = [("", "---------")] + get_phone_prefixes_sorted_and_localized()
        if initial:
            for prefix, values in _COUNTRY_CODE_TO_REGION_CODE.items():
                if initial in values:
                    self.initial = "+%d" % prefix
                    break

        super().__init__(choices=choices, attrs={'aria-label': pgettext_lazy('phonenumber', 'International area code')})

```

wiffbi (commented):
Looks good! I only have a minor cleanup/optimization, which is now possible thanks to your update with `get_phone_prefixes_sorted_and_localized`

PR #2528, opened 2022-03-13
paypal/views: Check that resource_type exists on the webhook event payload.

augustin:
Otherwise we get exceptions on webhook calls without resource_type.

Example of this happening:
```
ERROR 2022-03-13 01:29:22,953 django.request log Internal Server Error: /_paypal/webhook/
  File ".../venv.pretix/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File ".../venv.pretix/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File ".../venv.pretix/python3.9/site-packages/django/views/decorators/csrf.py", line 54, in wrapped_view
    return view_func(*args, **kwargs)
  File ".../venv.pretix/python3.9/site-packages/django/views/decorators/http.py", line 40, in inner
    return func(request, *args, **kwargs)
  File "~/.pyenv/versions/3.9.10/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File ".../venv.pretix/python3.9/site-packages/pretix/plugins/paypal/views.py", line 179, in webhook
    if event_json['resource_type'] not in ('sale', 'refund'):
KeyError: 'resource_type'
```
Background tasks that hit this error seem to just "stack up" which eventually causes problems.

### DIFF
--- a/src/pretix/plugins/paypal/views.py
+++ b/src/pretix/plugins/paypal/views.py
@@ -176,7 +176,7 @@ def webhook(request, *args, **kwargs):
     event_json = json.loads(event_body)
 
     # We do not check the signature, we just use it as a trigger to look the charge up.
-    if event_json['resource_type'] not in ('sale', 'refund'):
+    if not ('resource_type' in event_json) or (event_json['resource_type'] not in ('sale', 'refund')):
         return HttpResponse("Not interested in this resource type", status=200)
 
     if event_json['resource_type'] == 'sale':


augustin:
FWIW I've ran this change in production on the version I had installed early last year (i.e. before the switch to AGPL.) You may well want to do something different entirely in order to prevent this error or write different logic than I have here (it's probably not up to your style standards.)

If you absolutely insist I can sign the CLA, but it likely makes more sense that you just re-create this change (which probably isn't copyrightable anyway) in whatever fashion you see fit and close this PR, I suppose...

raphaelm:
We don't strictly require CLA for oneliners, but we're also going to merge #2493 soon which is a full rewrite of that part of the code anyways.

However, I'm surprised about his bug since we haven't seen it before after processing millions of PayPal transactions. Do you know what kind of webhook message triggers it?

pc-coholic:
LGTM, but I would also be interested to see which Webhook has no resource type...

augustin:
```
[2022-03-16 21:35:55,568: INFO/ForkPoolWorker-7] Task pretix.api.webhooks.send_webhook[f5e9e9fe-ee83-4bf7-89d7-9e842d147374] succeeded in 0.07769250869750977s: None
{'notification_id': 4409, 'organizer': '<redacted>', 'event': '<redacted>', 'code': 'T0LHY', 'action': 'pretix.event.order.placed'}
```

Actually, it looks like someone set up our Pretix instance to send itself webhooks... No wonder they don't have `resource_type`, they're not coming from PayPal at all 😆 

I guess it would still make sense to have this, though, so we don't get 500 ISEs when invalid data is sent to the hook URL?

PR #2529, opened 2022-03-13
PDF editor improvements

raphaelm:
This PR brings 4 of the most-wanted missing features to the PDF layout engine:

- Support for multi-page tickets (by using multi-page backgrounds)
- Support for multilingual custom texts (multi-lingual backgrund are out of scope for now)
- Support for string interpolation in custom texts (``{attendee_name_given} {attendee_name_family}``)
- Support for creating QR codes with arbitrary content (i.e. every text field can now also be a QR code)

All of these are implemented in a way that is maybe not the cleanest API design ever, but has the advantage that it is fully downwards compatible with older layouts. Still, we should probably implement this in libpretixprint an drelease new pretixPRINT and pretixSCAN Desktop versions before we merge this PR.

The PR also adds convenience features that do not change the API:

- Support for Ctrl+A to select all objects
- Button for downloading the current background PDF
- Tooltip explaining what dynamic images are


pc-coholic (changes requested):
Some of these items are probably out of scope of this PR, but I'll write them down too, as I noticed them during testing:

- "Price including Add-Ons" will fail in preview:
```  File "/home/martin/PycharmProjects/pretix/pretix/src/pretix/base/pdf.py", line 694, in _get_text_content
    return self.variables[o['content']]['evaluate'](op, order, ev)
  File "/home/martin/PycharmProjects/pretix/pretix-shipping/pretix_shipping/signals.py", line 413, in <lambda>
    'evaluate': lambda op, order, event: money_filter(op.price + sum(
TypeError: unsupported operand type(s) for +: 'float' and 'decimal.Decimal'```
- The "QR code for  Check-In" and "QR code for Lead Scanning" are correct as such - but it did take me quite a few moments to figure out that I could insert one of those and only then change them to have another value encoded. I do see the advantage of having those two quick access buttons, but I am wondering if we could make it clearer, that there are more choices available.
- During testing, I ran into an error with a scanned 2-page PDF document I wanted to use as a background: `Could not find xref table at specified location`. Seems to be a corrupted background - but perhaps we can fail more graceful/fall back to a white background instead of error 500...
- Selecting all objects with CTRL-A on page 1, changing to page 2 and then back to page 1 will have moved all objects to the bottom right. Seems to be related to CTRL-A, since selecting just a single or multiple elements by mouse will not cause this issue.

raphaelm:
>     "Price including Add-Ons" will fail in preview:

fixed

> The "QR code for Check-In" and "QR code for Lead Scanning" are correct as such - but it did take me quite a few moments to figure out that I could insert one of those and only then change them to have another value encoded. I do see the advantage of having those two quick access buttons, but I am wondering if we could make it clearer, that there are more choices available.

fixed by adding a third button

> Selecting all objects with CTRL-A on page 1, changing to page 2 and then back to page 1 will have moved all objects to the bottom right. Seems to be related to CTRL-A, since selecting just a single or multiple elements by mouse will not cause this issue.

fixed by deselecting everything on page switch, that makes sense anyways

> Perhaps we could update the jquery.fileupload-component to fileuploadUIX, so that we can pass a file-extension filter. Since we're only accepting PDFs as backgrounds, no need to offer to upload all files available on the users device.

no need to swap out the library for that, committed

raphaelm:
> During testing, I ran into an error with a scanned 2-page PDF document I wanted to use as a background: Could not find xref table at specified location. Seems to be a corrupted background - but perhaps we can fail more graceful/fall back to a white background instead of error 500...

fixed with a check during upload and a more proper error message

PR #2544, opened 2022-03-18
Fix #2542 -- Validation of exporter forms

### LINKED ISSUE #2542: Error in single exporter form collapses all exporters
**Describe the bug**
When submitting the "checkinlistpdf" exporter, without a check-in list present, an error is created on all of the available forms. Due to the presence of the error on all forms, all exporters are shown, making it very confusing to see the actual problem with the exporter in use.

I've located https://github.com/pretix/pretix/commit/488273d5f2abd91910e89f01eae15ccf96d325a6 to introduce the current behaviour, however I'm not sure why the forms which were not submitted also render with `.has-error` on it's fields. 

**Expected behavior**
To have all exporters be closed, except for the one submitted.

**Screenshots**
scrot-2022-03-17T17:54:06+01:00

**Additional context**
-

wiffbi:


raphaelm (changes requested):
I think the same changes are necessary on organizer level, e.g. pretix/control/views/organizer.py

wiffbi:
> I think the same changes are necessary on organizer level, e.g. pretix/control/views/organizer.py

Yes and no. If you look at `ExportMixin` in pretix/control/views/organizer.py it does filter the exporters with 
`id = self.request.GET.get("identifier") or self.request.POST.get("exporter")`
so that only the active exporter is shown. I think we should do the same on event-level exporter. I will change my code accordingly

raphaelm (approved):

PR #2548, opened 2022-03-21
Fix voucher redemption page markup

raphaelm:
When we refactored the product list, we apparently forgot about the voucher redemption page, leading to minor design and functionality issues. Also, the new accessiblity improvements did not make it here.

@wiffbi please double-check if my patch makes sense :)


wiffbi (approved):
Looks good!

PR #2552, opened 2022-03-23
API: Allow to change orders

raphaelm:
- [x] REST-y operations on positions
  - [x] PATCH
  - [x] POST
  - [x] DELETE
  - [x] regenerate secret
- [x] Transactional API
  - [x] comprehensive tests
  - [x] docs


src/pretix/api/serializers/orderchange.py
```diff
@@ -0,0 +1,423 @@
...
+            return
+        self.fields['order'].queryset = self.context['event'].orders.all()
+        self.fields['item'].queryset = self.context['event'].items.all()
+        self.fields['subevent'].queryset = self.context['event'].subevents.all()
+        self.fields['seat'].queryset = self.context['event'].seats.all()
+        self.fields['variation'].queryset = ItemVariation.objects.filter(item__event=self.context['event'])
+        if 'order' in self.context:
+            del self.fields['order']
+
+    def validate(self, data):
```
wiffbi, line 70: Is this missing a `super().validate(data)`?

wiffbi (commented):
I have to admit this is a rather complex PR. I took some time to try to understand the flow of the API-requests, serializers, etc. and mainly concentrated on the tests to understand, how it works. I actually just stumbled upon one little thing regarding validation, but I am not sure whether that is actually a problem. I think everything should work as intended, but then again I did not/could not follow everything through deeply enough. Sorry!

src/pretix/api/serializers/orderchange.py
raphaelm: Hmm, I think that the parent function doesn't do anything, but it is more correct to still call it, will change it.

crackwitz:
The docs aren't clear on this: is the API capable of creating a position and also create addons on that position, in the same request?

PR #2555, opened 2022-03-24
Questions step: Only prefill customer name for first ticket

raphaelm:
see internal Ticket# 2396506

### DIFF
--- a/src/pretix/base/views/mixins.py
+++ b/src/pretix/base/views/mixins.py
@@ -58,7 +58,7 @@ class BaseQuestionsViewMixin:
     def _positions_for_questions(self):
         raise NotImplementedError()
 
-    def get_question_override_sets(self, position):
+    def get_question_override_sets(self, position, index):
         return []
 
     def question_form_kwargs(self, cr):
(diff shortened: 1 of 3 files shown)


wiffbi:
There is one edge-case where this does not work: when I have a non-admission product as the first product in my basket, it does not auto-fill the first admission-product with the name. Probably very very rare and so rare enough to be ignored?

src/pretix/base/views/mixins.py
```diff
@@ -96,7 +96,7 @@ def forms(self):
                 ))
             )
 
-            override_sets = self.get_question_override_sets(cr)
+            override_sets = self.get_question_override_sets(cr, idx)
```
wiffbi, line 99: Based on idx, there is one edge-case where this does not work: when I have a non-admission product as the first product in my basket, it does not auto-fill the first admission-product with the name. Probably very very rare and so rare enough to be ignored?

wiffbi (commented):
Looks good, just a comment regarding an edge-case having a non-admission-product as the first product in the cart.

src/pretix/base/views/mixins.py
raphaelm: Yeah, I think that's fine.

PR #2560, opened 2022-03-29
Allow to print expiry date of gift card in PDFs

raphaelm:


src/pretix/base/pdf.py
```diff
@@ -418,6 +419,11 @@
         "editor_sample": _("2017-05-31 19:00"),
         "evaluate": lambda op, order, ev: get_first_scan(op)
     }),
+    ("giftcard_expiry_date", {
+        "label": _("Gift card: Expiration date"),
+        "editor_sample": _("2017-05-31"),
```
wiffbi, line 424: I am not sure whtether a translateable string is right here? `SHORT_DATE_FORMAT` is 'm/d/Y' as default. Shouldn’t we just use a dummy `date_format(datetime.date(2017, 5, 17), "SHORT_DATE_FORMAT")`?

wiffbi (commented):
Looks good, only a question regarding the editor_sample for a date being translatable instead of using a dummy date formatted with the current locale.

wiffbi (approved):

src/pretix/base/pdf.py
wiffbi: Nevermind, that does not work in this context.

raphaelm: This is just copied from the other templates. Using lazy date_formatting would be more correct.

However, you're wrong, SHORT_DATE_FORMAT is 'Y-m-d' in pretix by default.

PR #2561, opened 2022-03-29
Allow users to see the number of checkins

raphaelm:


src/pretix/presale/views/order.py
```diff
@@ -129,6 +129,12 @@ def position(self):
             addon_to__isnull=True,
             order__code=self.kwargs['order'],
             positionid=self.kwargs['position']
+        ).annotate(
+            checkin_count=Subquery(
+                Checkin.objects.filter(
+                    successful=True, type=Checkin.TYPE_ENTRY, position_id=OuterRef('pk')
+                ).order_by().values('position').annotate(c=Count('*')).values('c')
+            )
```
wiffbi, line 137: Would it make sense to annotate the checkin_count only if `self.request.event.settings.show_checkin_number_user` is set? Not sure though if this is worth it from a performance perspective. Also, if we want to always show the checkin_count somewhere else (e.g. in pretix-control) it makes no sense to not always annotate.

src/pretix/presale/views/order.py
```diff
@@ -230,8 +236,15 @@ def download_buttons(self):
...
-            queryset=self.order.positions.prefetch_related('issued_gift_cards').select_related('tax_rule'),
+            answers=True,
+            downloads=ctx['can_download'],
+            queryset=self.order.positions.prefetch_related('issued_gift_cards').select_related('tax_rule').annotate(
+                checkin_count=Subquery(
+                    Checkin.objects.filter(
+                        successful=True, type=Checkin.TYPE_ENTRY, position_id=OuterRef('pk')
+                    ).order_by().values('position').annotate(c=Count('*')).values('c')
+                )
+            ),
```
wiffbi, line 247: See my comment above regarding annotating checkin_count only if `show_checkin_number_user` is set.

src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
```diff
@@ -94,6 +94,21 @@
                         </dd>
                     </div>
                 {% endif %}
+                {% if event.settings.show_checkin_number_user and line.checkin_count %}
```
wiffbi, line 97: If only the ticket-usage is shown, the surounding `<dl>` is not output. I will push a fix for this.

src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
```diff
@@ -94,6 +94,21 @@
...
+                    <div class="cart-icon-details">
+                        <dt class="sr-only">{% trans "Usage:" context "ticket_checkins" %}</dt>
+                        <dd class="text-success">
+                            <span class="fa fa-check-circle fa-fw text-success" aria-hidden="true"></span>
+                            <strong>
```
wiffbi, line 102: Not sure whether this needs to be `<strong>` as this then gets the same weight as the corresponding item-label. Thanks to coloring it with text-success, maybe a normal font-weight might be enough?

wiffbi (commented):
Looks good. Just some questions, nothing breaking.
I will push a change to fix the missing `<dl>`.

src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
raphaelm: I tried that first, but it given that text-success has even less contrast than regular text I felt it was not prominent enough.

My reasoning is this: If you see the order detail page for the first time, this does not appear at all, since the ticket wasn't used, so it does not matter. If you see the order detail page later, this is the thing that **changed** and possible the most important info you are looking for in case of a ticket than can be used X times. So it makes sense to give it a lot of weight.

wiffbi: Makes sense!

PR #2562, opened 2022-03-29
Improvements to PDF text placeholders

raphaelm:
We recently introduced placeholders for PDF texts, i.e. you can now use {attendee_name_full_name} as part of a custom text on your ticket.

This PR actually makes that usable:

- It adds a dynamic documentation page listing all placeholders available for an event
- It moves question placeholders from including the question's primary key to including the question's identifier. This has long been a pain point that we need to carefully rewrite the layout JSON when we copy the layout to a new event, since question primary keys are not stable between events. This patch basically adds the new fields and "hides" the old ones, while still retaining full functionality of the old ones. The downside is that for the forseeable future, the API will now expose *both* options to the devices, which seems like an inefficient but safe option. We cannot easily do a migration of all server-known layouts and drop the old fields, because the PDF engine is used by plugins as well (e.g. our pretix-certificates, but maybe also other community users). We should announce a deprecation of the old fields in the release notes.

### DIFF
--- a/src/pretix/base/pdf.py
+++ b/src/pretix/base/pdf.py
@@ -492,10 +492,17 @@ def variables_from_questions(sender, *args, **kwargs):
     for q in sender.questions.all():
         if q.type == Question.TYPE_FILE:
             continue
+        d['question_{}'.format(q.identifier)] = {
+            'label': _('Question: {question}').format(question=q.question),
+            'editor_sample': _('<Answer: {question}>').format(question=q.question),
+            'evaluate': partial(get_answer, question_id=q.pk),
+            'migrate_from': 'question_{}'.format(q.pk)
+        }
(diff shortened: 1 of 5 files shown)


wiffbi (approved):
Looks good and works as intended.

Would it make sense to update the layout to the new placeholders when it is opened for editing? That way, when someone copies the layout it already uses the new placeholders and at some point we could hopefully phase out the old ones?

raphaelm:
> Would it make sense to update the layout to the new placeholders when it is opened for editing? That way, when someone copies the layout it already uses the new placeholders and at some point we could hopefully phase out the old ones?

I don't think there is much gain in that. Either we – and all plugins – add a proper data migration, or we don't bother. There's no use in having 70% of the layouts converted at some point, we'll need 100% before we can turn off the old ones.

PR #2573, opened 2022-04-04
deps: Pillow minimum version 9.1.0

MaicoTimmerman:
The changes introduced in db39b89ae use a constant which isn't available until Pillow 9.1.0.

Reference: https://pillow.readthedocs.io/en/stable/releasenotes/9.1.0.html#constants

### DIFF
--- a/src/setup.py
+++ b/src/setup.py
@@ -205,7 +205,7 @@ setup(
         'packaging',
         'paypalrestsdk==1.13.*',
         'phonenumberslite==8.12.*',
-        'Pillow==9.*',
+        'Pillow==9.1.*',
         'protobuf==3.19.*',
         'psycopg2-binary',
         'pycountry',


src/setup.py
```diff
@@ -205,7 +205,7 @@ def run(self):
         'packaging',
         'paypalrestsdk==1.13.*',
         'phonenumberslite==8.12.*',
-        'Pillow==9.*',
+        'Pillow>=9.1.*',
```
raphaelm, line 208: I think we should stick with == instead of >= exactly to prevent this sort of problem ;)

raphaelm (changes requested):

src/setup.py
MaicoTimmerman: Ah, that was a remnant of me debugging this issue. Fixed!

raphaelm:
Sorry, I needed to fix it in d0474afdf almost simultaneously since there was another issue as well. Thanks anyways!

PR #2578, opened 2022-04-06
Add Czech translation

luto:
As discussed with @raphaelm, add the Czech translation.

### DIFF
--- a/src/pretix/settings.py
+++ b/src/pretix/settings.py
@@ -525,12 +525,13 @@ ALL_LANGUAGES = [
     ('es', _('Spanish')),
     ('tr', _('Turkish')),
     ('gl', _('Galician')),
+    ('cs', _('Czech')),
 ]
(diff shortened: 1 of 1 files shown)


src/pretix/settings.py
```diff
@@ -525,12 +525,13 @@
     ('es', _('Spanish')),
     ('tr', _('Turkish')),
     ('gl', _('Galician')),
+    ('cz', _('Czech')),
```
raphaelm, line 528: While the country code of the Czech Republic is CZ, I believe the ISO-assigned language code of the Czech language is cs ;) Not at all confusing, I know.

raphaelm (changes requested):

pc-coholic:
@cla-bot check

src/pretix/settings.py
luto: fixed.

PR #2580, opened 2022-04-08
Allow to add a comment when cancelling an order

raphaelm:
This is in analogy to denying an order and should be useful

See also internal ticket #2393113


src/pretix/base/settings.py
```diff
@@ -1914,6 +1914,8 @@ def unserialize(cls, s):
 
 your order {code} for {event} has been canceled.
 
+{comment}
```
pc-coholic, line 1917: I think we might need a manual adjustment of the .po-files to include this in all languages, so weblate does not flag all these strings as new/untranslated

src/pretix/control/forms/orders.py
```diff
@@ -167,6 +167,12 @@ class CancelForm(ForceQuotaConfirmationForm):
         initial=True,
         required=False
     )
+    comment = forms.CharField(
+        label=_('Comment (will be sent to the user)'),
+        help_text=_('Will be included in the notification email when the respective placeholder is present in the '
+                    'configured email text.'),
+        required=False,
+    )
```
pc-coholic, line 175: Should we make this a TextArea perhaps? I do see the advantage of a CharField (Keep you message simple and on point), but I could also see the advantage of a bigger text input area.

Second: Perhaps we should save the last message for easier recall if you are doing mass-cancellations?

pc-coholic (approved):

src/pretix/control/forms/orders.py
raphaelm: > Should we make this a TextArea perhaps? I do see the advantage of a CharField (Keep you message simple and on point), but I could also see the advantage of a bigger text input area.

First I wanted to agree and changed the widget, then I remembered that linebreaks and markdown don't always behave like users expect and unless we want to go through all that trouble of implementing a preview mode etc. it will be less error-prone with we stick to a single-line text (it's still not limited in length). I think we can leave it like this for now and always change it later if it turns out to be a problem in the real world.

> Second: Perhaps we should save the last message for easier recall if you are doing mass-cancellations?

Hmm, could be useful but is something we don't do anywhere else and if users do not expect it, they might not notice it and accidentally send the wrong comment to a user, so I think better not?

PR #2583, opened 2022-04-09
Allow to bulk-edit devices

raphaelm:


wiffbi (commented):
Looks good! I have only one device connected though, so I could not test pagination.

I initially stumbled upon not having a „cancel“-button on the bulk-edit form-page – but as this is consistent with bulk-editing sub-events I think we could leave it like that. Although IMHO we should consider at some point adding a „cancel“-button to both bulk-edit forms, especially as changing nothing and hitting save does not leave the page and gives no message at all – e.g. „you have not changed anything“.

PR #2590, opened 2022-04-12
Voucher redemption: Raise 404 error if subevent_pk is not an integer

wiffbi:
Due to some malconfiguration it could happen that the subevent-PK submitted via GET-params is not an integer, which currently results in a server error. This fix checks the subevent-GET-param and raises an 404 as if the subevent_pk would not be found.

### DIFF
--- a/src/pretix/presale/views/cart.py
+++ b/src/pretix/presale/views/cart.py
@@ -632,7 +632,7 @@ class RedeemView(NoSearchIndexViewMixin, EventViewMixin, CartMixin, TemplateView
             if request.GET.get('subevent'):
                 try:
                     subevent_pk = int(request.GET.get('subevent'))
-                    self.subevent = SubEvent.objects.get(event=request.event, pk=subevent_pk, active=True)
+                    self.subevent = request.event.subevents.get(pk=subevent_pk, active=True)
                 except (ValueError, SubEvent.DoesNotExist):
                     raise Http404(pgettext('subevent', 'We were unable to find the specified date.'))  
 


src/pretix/presale/views/cart.py
```diff
@@ -630,8 +630,12 @@ def dispatch(self, request, *args, **kwargs):
         self.subevent = None
         if request.event.has_subevents:
             if request.GET.get('subevent'):
-                self.subevent = get_object_or_404(SubEvent, event=request.event, pk=request.GET.get('subevent'),
-                                                  active=True)
+                try:
+                    subevent_pk = int(request.GET.get('subevent'))
+                except ValueError:
+                    raise Http404(_('No SubEvent matches the given query.'))
```
raphaelm, line 636: We should never ever use the word "SubEvent" when talking to users, it should also never show up in translations. 

I know you copied this from the auto-generated message by get_object_or_404 as well, but maybe we should change both then?

Correct would be e.g.:

```
raise Http404(pgettext('subevent', 'We were unable to find the specified date.'))
```

(We always use pgettext for subevent because "date" is ambigous in english between a "subevent" and a "calendar date", but we use different words eg in German (Termin vs Datum))

raphaelm (changes requested):

src/pretix/presale/views/cart.py
wiffbi: Thanks very much, your message makes way more sense – and yes, I copied it from `get_object_or_404` :) I will change it for both of them.

src/pretix/presale/views/cart.py
```diff
@@ -630,8 +630,11 @@ def dispatch(self, request, *args, **kwargs):
         self.subevent = None
         if request.event.has_subevents:
             if request.GET.get('subevent'):
-                self.subevent = get_object_or_404(SubEvent, event=request.event, pk=request.GET.get('subevent'),
-                                                  active=True)
+                try:
+                    subevent_pk = int(request.GET.get('subevent'))
+                    self.subevent = SubEvent.objects.get(event=request.event, pk=subevent_pk, active=True)
```
raphaelm, line 635:
```suggestion
                    self.subevent = request.event.subevents.get(pk=subevent_pk, active=True)
```

PR #2595, opened 2022-04-13
Calendar view: improve differentiation between events’ states (Z#173939)

wiffbi:
In calendar view it was hard to distinguish the availability state of events on a first glance especially with some type of color blindness. This tries to fix this by:

- make unavailable events background/border even lighter
- remove ticket-icon (or any icon at all) from unavailable events to make the distinction for „book now“ links with a ticket-icon
- remove bold font on unavailable events
- strike-through first line on unavailable events

Before:
before

After:
after

After (no color):
after-no-color

After (no red):
after-no-red

After (no green):
after-no-green


raphaelm:
I think this is a great improvement!

I was a little worried that the strike-through of the time could be interpreted as "event has been cancelled" event though it's only no longer on sale, but I think with the text below people will understand and we can keep it like this.

PR #2597, opened 2022-04-19
eMail: Ratelimit exponential backup for Amazon SES

pc-coholic:
Z#2396582


pc-coholic:
Wontfix for now, as 454 is also an `Authentication failed` error message in which case continued retries might land us on a denylist for credential stuffing...

PR #2600, opened 2022-04-22
PayPal: Patch API client to cache access tokens

raphaelm:


pc-coholic (approved):

raphaelm:
Not merging any more because we plan to merge #2493 tomorrow and there's no point in taking the risk

pc-coholic:
Reopening this, since we are still running those code in parallel...

PR #2605, opened 2022-04-26
Add note field and external identifier to customers

wiffbi:
This adds an optional „Notes“ textfield as well as an optional „external identifier“ to a customer record.


src/pretix/base/models/customers.py
```diff
@@ -59,6 +59,8 @@ class Customer(LoggedModel):
                               default=settings.LANGUAGE_CODE,
                               verbose_name=_('Language'))
     last_modified = models.DateTimeField(auto_now=True)
+    external_identifier = models.CharField(default="", max_length=255, verbose_name=_('External identifier'), blank=True)
+    notes = models.TextField(default="", verbose_name=_('Notes'), blank=True)
```
wiffbi, line 63: I somehow could not get the migration to work without the `default=""` – I thought this was the default (no pun intended) when `blank=True` and not provided otherwise.

src/pretix/control/forms/organizer.py
```diff
@@ -651,7 +651,7 @@ class CustomerCreateForm(CustomerUpdateForm):
 
     class Meta:
         model = Customer
-        fields = ['identifier', 'is_active', 'name_parts', 'email', 'is_verified', 'locale']
+        fields = ['identifier', 'is_active', 'name_parts', 'email', 'is_verified', 'locale', 'external_identifier', 'notes']
```
wiffbi, line 654: The Phone-Number text-input in the create-view is appended as the last input. Have not found a way yet, to change this – but have not looked that hard for it. Maybe just leave it as is?

src/pretix/control/templates/pretixcontrol/organizers/customers.html
```diff
@@ -62,6 +62,9 @@ <h3 class="panel-title">{% trans "Filter" %}</h3>
                     <th>{% trans "Name" %}
                         <a href="?{% url_replace request 'ordering' '-name' %}"><i class="fa fa-caret-down"></i></a>
                         <a href="?{% url_replace request 'ordering' 'name' %}"><i class="fa fa-caret-up"></i></a></th>
+                    <th>{% trans "External ID" %}
+                        <a href="?{% url_replace request 'ordering' '-external_identifier' %}"><i class="fa fa-caret-down"></i></a>
+                        <a href="?{% url_replace request 'ordering' 'external_identifier' %}"><i class="fa fa-caret-up"></i></a></th>
```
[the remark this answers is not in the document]
raphaelm: You probably need to extend the filter form for this

src/pretix/control/forms/organizer.py
raphaelm: Hmm, I think honestly that the external identifier should be *above* the name and the notes should be the last one. 

maybe

- identifier
- external_identifier
- is_active
- name_parts
- email
- is_verified
- phone
- locale
- notes

If it's too hard to resort the fields dictionary, the next-best option would be to have the fields explicitly one by one in the template. That's how we do it for every complex form anyways

src/pretix/base/models/customers.py
```diff
@@ -90,6 +92,8 @@ def anonymize(self):
         self.name_cached = ''
         self.email = None
         self.phone = None
+        self.external_identifier = ""
+        self.notes = ""
```
raphaelm, line 96: Should be null then as well

src/pretix/base/models/customers.py
raphaelm: Generally, you are right, it's no good to have two states for new data. However:

- Django will make it pretty impossible to have an empty string here, since if you enter an empty string into the form it will cause NULL to be saved

- NULL is semantically more correct for "there is no external identifier", so (a) it's nicer in the API (b) it would support e.g. future uniqueness constraints, if we'd say external identifiers must be unique (they probably should not be) then multiple "" values would be a problem, but multiple NULLs would not
 
- On some database (versions) adding a non-NULL-able column to a pre-existing table causes the entire table to be rewritten. If the table is large, this is a problem. For example, if we already have millions of customers in the pretix.eu production database, adding a non-nullable column would create a possibly **long** full downtime of pretix since the entire database goes down in locks! As a rule, we therefore avoid adding non-nullable columns to tables with large numbers of rows like the plague. (In this case, it would be probably fine since **recent** PostgreSQL versions can handle this quickly **if there is a default value**).

src/pretix/control/forms/organizer.py
wiffbi: Yes, makes more sense. I changed the order of fields like you suggested, except I moved `is_active` as the first item – it just looked better with the checkbox at the top and felt more „in line“ with other forms we have, where one can activate something.

src/pretix/base/models/customers.py
wiffbi: Yes, done! :)

src/pretix/control/templates/pretixcontrol/organizers/customers.html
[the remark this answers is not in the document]
wiffbi: Yes, that was it. Thanks for pointing me in the right direction! Fixed. :)

wiffbi:
> I think we should also expose the new fields in the API :)

I added the fields to the Meta-class of CustomerSerializer as well as to the docs in customer.rst – any other place I missed regarding the API?

src/pretix/base/models/customers.py
wiffbi: Thanks for explaining the reasoning behind nullable columns. Makes total sense!

raphaelm (commented):
LGTM, but let's merge after #2510. You will likely need to rebase and recreate the migration then.

PR #2610, opened 2022-04-27
Change npm install call in setup.py

henryk:
Fixes two problems: Windows python needs shell=True to find npm, and npm needs to be executed from within the directory with the package.json.
The second problem seems to should have existed everywhere, but for some reason the command appears to have worked on Linux?
Note: Using cwd= neatly circumvents potential escaping problems introduced by shell=True.


raphaelm:
@cla-bot check

raphaelm:
Thanks! Let's hope this breaks nothing else but I could not find anything.

PR #2612, opened 2022-04-28
Order list export: Include canceled positions

raphaelm:
This includes canceled lines in the "order positions" and "order fees" tabs of the "order list" export.

Before, the lines were just not included.

With this patch, the lines only marked by the "canceled" value in the status column that is set even when the order as a whole has a different status. This is logical and consistent, but I'm still not sure if it might be a bad surprise for some users of the list? Can't imagine a situation where it would be, though.

### DIFF
--- a/src/pretix/base/exporters/orderlist.py
+++ b/src/pretix/base/exporters/orderlist.py
@@ -424,13 +424,13 @@ class OrderListExporter(MultiSheetListExporter):
         ).values(
             'm'
         ).order_by()
-        qs = OrderFee.objects.filter(
+        qs = OrderFee.all.filter(
             order__event__in=self.events,
         ).annotate(
             payment_providers=Subquery(p_providers, output_field=CharField()),
         ).select_related('order', 'order__invoice_address', 'tax_rule')
         if form_data['paid_only']:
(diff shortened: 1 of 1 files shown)


src/pretix/base/exporters/orderlist.py
```diff
@@ -424,13 +424,13 @@ def iterate_fees(self, form_data: dict):
         ).values(
             'm'
         ).order_by()
-        qs = OrderFee.objects.filter(
+        qs = OrderFee.all.filter(
```
wiffbi, line 427: Thanks for this! I wondered why this change, but now learned `objects` is a Manager with a filter removing canceled fees. Good to know!

wiffbi (changes requested):
Looks good, except in src/pretix/base/exporters/orderlist.py:521  I think `OrderPosition.objects` needs to be `OrderPosition.all` as well, as `OrderPosition.objects` is an ActivePositionManager, that filters `canceled=False` already, so the filter with canceled in line 533 does not have any effect and always only non-canceled OrderPositions are returned.

wiffbi (approved):

PR #2617, opened 2022-04-29
Explicitly store whether checkins were offline

raphaelm:
This adds an additional field on the checkin model that explicitly saves whether the scan was made offline or online for future scans. Useful for debugging and statistics.

I wanted to avoid introducing a new API parameter that all API clients need to support, so I made it based on the "forced" field. I hence deliberately did not name the new field "offline" since that's not 100% in line with the semantics of the "force" parameter. For every implementation I am aware of (i.e. libpretixsync and pretixscan-ios), force is set to true if and only if the scan was offline, but who knows what other API consumers might do.


pc-coholic (changes requested):
> CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0215_checkin_forced_sent, 0215_customer_organizer_identifier_unique in pretixbase).

PR #2626, opened 2022-05-02
Widget: Fix trigger_close_callback()

MaicoTimmerman:
It seems I did some breaking changes before I submitted the #2569, that I did not correctly test. This minor change should fix that. Sorry for the inconvenience.


wiffbi (approved):
Good catch!

PR #2627, opened 2022-05-03
Badges: Remove trimBox from background PDF (Z#2398854)

wiffbi:
PyPDF 1.27.6 has introduced different handling of trimBox when merging pages – with 1.26 trimBox was ignored when the background-PDF was placed, but since 1.27.6 PDFs with a trimBox got a mask the size of the trimBox (ie. clipped) when merged. As of the release notes, this is intended behaviour for PyPDF, so we need to change the behaviour in our code. This PR sets the trimBox of any background-PDF in badges to the mediaBox.


wiffbi:
In my tests on my machine with PyPDF (pdftk not enabled) tickets (also multiple tickets in one PDF) are not affected.

PR #2639, opened 2022-05-10
Add placeholder name_for_salutation to editor/PDF to improve handling of salutation „Mx“

wiffbi:
There is an useful email placeholder `name_for_salutation` which this PR brings to the editor/PDF as well.

### DIFF
--- a/src/pretix/base/pdf.py
+++ b/src/pretix/base/pdf.py
@@ -570,6 +570,18 @@ def get_variables(event):
             "evaluate": partial(_get_ia_name_part, key)
         }
 
+    concatenation_for_salutation = scheme.get("concatenation_for_salutation", scheme["concatenation"])
+    v['attendee_name_for_salutation'] = {
+        'label': _("Attendee name for salutation"),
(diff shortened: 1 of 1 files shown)


src/pretix/base/pdf.py
```diff
@@ -546,6 +546,14 @@ def get_variables(event):
     v = copy.copy(DEFAULT_VARIABLES)
 
     scheme = PERSON_NAME_SCHEMES[event.settings.name_scheme]
+
+    concatenation_for_salutation = scheme.get("concatenation_for_salutation", scheme["concatenation"])
+    v['attendee_name_for_salutation'] = {
+        'label': _("Attendee name for salutation"),
+        'editor_sample': _("Mr Doe"),
+        'evaluate': lambda op, order, ev: concatenation_for_salutation(op.attendee_name_parts)
```
wiffbi, line 554: Is `op.attendee_name_parts` always a dict or could it be None? If so, this line should read

```suggestion
        'evaluate': lambda op, order, ev: concatenation_for_salutation(op.attendee_name_parts or {})
```

PR #2657, opened 2022-05-18
Orders API: Allow downloading tickets for pending orders

raphaelm:


doc/api/resources/orders.rst
```diff
@@ -609,13 +609,17 @@ Fetching individual orders
...
+   The API now supports ticket downloads for pending orders if allowed by the event settings.
+
 .. http:get:: /api/v1/organizers/(organizer)/events/(event)/orders/(code)/download/(output)/
 
    Download tickets for an order, identified by its order code. Depending on the chosen output, the response might
    be a ZIP file, PDF file or something else. The order details response contains a list of output options for this
    particular order.
 
-   Tickets can be only downloaded if the order is paid and if ticket downloads are active. Note that in some cases the
+   Tickets can be only downloaded if the order is paid (depending on event settings) and if ticket downloads are active. Note that in some cases the
```
wiffbi, line 622: I think if you do not know that tickets for pending orders are possible, the part „depending on event settings“ is a bit confusing as there is in „what is the alternative to paid orders?“. I am not 100% sure though as this setting does not seem to be available through the GUI – then I guess swapping around „can be only downloaded“ to „can only be downloaded“ is enough.

```suggestion
   Tickets can only be downloaded if ticket downloads are active and – depending on event settings – the order is either paid or pending. Note that in some cases the
```

doc/api/resources/orders.rst
```diff
@@ -1646,7 +1654,7 @@ Order position ticket download
    The referenced URL can provide a download or a regular, human-viewable website - so it is advised to open this URL
    in a webbrowser and leave it up to the user to handle the result.
 
-   Tickets can be only downloaded if the order is paid and if ticket downloads are active. Also, depending on event
+   Tickets can be only downloaded if the order is paid (depending on event settings) and if ticket downloads are active. Also, depending on event
```
wiffbi, line 1657:
```suggestion
   Tickets can only be downloaded if ticket downloads are active and – depending on event settings – the order is either paid or pending. Also, depending on event
```

wiffbi (approved):
Looks good. Just some minor comment regarding documentation. This setting `ticket_download_pending` is a hidden setting only we can set?

raphaelm:
@wiffbi 

>  This setting ticket_download_pending is a hidden setting only we can set?

No, everyone can set it at Settings > Tickets. We just did not respect it in the API before.

PR #2659, opened 2022-05-19
Fix error handling on add-ons selection step

raphaelm:


src/pretix/static/pretixbase/js/asynctask.js
```diff
@@ -152,6 +153,7 @@ function async_task_error(jqXHR, textStatus, errorThrown) {
 
             if (respdom.filter('#page-wrapper') && $('#page-wrapper').length) {
                 $("#page-wrapper").html(respdom.find("#page-wrapper").html());
+                setup_basics($("body"));
```
wiffbi, line 156: I think this needs to be constrained to `#page-wrapper` in this case, otherwise existing elements on the body might be reset (e.g. collapsed-state).

```suggestion
                setup_basics($("#page-wrapper"));
```

src/pretix/static/pretixbase/js/details.js
```diff
@@ -1,6 +1,7 @@
 /*global $ */
 
 setup_collapsible_details = function (el) {
+    console.log(el);
```
wiffbi, line 4:
```suggestion
```

wiffbi (approved):
Looks good, I did not have a local event setup to test this though! Just two minor changes, one regarding constraining the setup_basics to the page_wrapper if applicable. The other is a console.log.

Also, would a more general PR name be better? Of course it fixes the add-on step, but it actually generally fixes the setup when using asynctask. So something like „Fix JS page setup when XHR returns errors“ maybe? Hm, maybe to general.

PR #2678, opened 2022-06-09
Allow to bulk-select many tickets to check in or out

raphaelm:


src/pretix/control/templates/pretixcontrol/checkin/index.html
```diff
@@ -180,13 +200,16 @@ <h3 class="panel-title">{% trans "Filter" %}</h3>
...
             {% if "can_change_orders" in request.eventpermset %}
                 <button type="submit" class="btn btn-primary btn-save">
+                    <span class="fa fa-sign-in" aria-hidden="true"></span>
                     {% trans "Check-In selected attendees" %}
                 </button>
                 <button type="submit" class="btn btn-default btn-save" name="checkout" value="true">
+                    <span class="fa fa-sign-out" aria-hidden="true"></span>
                     {% trans "Check-Out selected attendees" %}
                 </button>
                 <button type="submit" class="btn btn-default btn-save" name="revert" value="true">
```
wiffbi, line 210: Not sure this is needed, but to make it consistent with e.g. bulk-deleting of events, this should be a red button.

```suggestion
                <button type="submit" class="btn btn-danger btn-save" name="revert" value="true">
```

wiffbi (approved):
Works as intended. Only minor comments I could add a IE fallback for event.submitter, if you want.

src/pretix/static/pretixbase/js/asynctask.js
```diff
@@ -218,10 +218,16 @@ $(function () {
...
+            // Not supported on IE, may lead to wrong results, but we don't support IE in the backend anymore
+            console.warn('Unknown submitter, IE unsupported');
+            var $submitter = $(e.originalEvent.submitter);
+            if ($submitter.attr("name") && $submitter.val()) {
+                formData.append($submitter.attr("name"), $submitter.val());
+            }
+        }
         $.ajax(
             {
                 'type': 'POST',
```
wiffbi, line 233:
```suggestion
                'type': this.method,
```

raphaelm: From JQuery documentation, I'm unclear if there *might* be an issue with lower-case "post" instead of "POST" which is probably technical possibly with this.method, so I'd skip touching it until we need to

src/pretix/static/pretixbase/js/asynctask.js
```diff
@@ -218,10 +218,16 @@ $(function () {
             'this page and try again.'
         ));
 
-
-        console.log($(this).get(0))
         var formData = new FormData($(this).get(0))
         formData.append('ajax', '1');
+        if (e.originalEvent.submitter) {
+            // Not supported on IE, may lead to wrong results, but we don't support IE in the backend anymore
+            console.warn('Unknown submitter, IE unsupported');
```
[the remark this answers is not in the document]
raphaelm: We do not need an IE fallback, since
- we've dropped caring about IE in the backend a year a go, we show a big fat warning to IE users since May 2021 and nobody complained
- we still support IE in the frontend, but we have no data-asynctask with ``<button name=…>`` in the current frontewnd
- IE11 is officially retired by Microsoft *next Wednesday*. We haven't formally decided or announced this yet, but I think we can also start stop caring about it in the frontend very very soon, and we're unlikely to introduce a new button in the frontend by then ;)

However, we *should* care on Safari. B ut as I understand you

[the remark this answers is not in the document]
wiffbi: Yes, we are fine with older Safari-versions. Even current Safari (15.5) shows the same behaviour regarding buttons name-value-pair being added to the FormData.

src/pretix/static/pretixbase/js/asynctask.js
wiffbi: Okay, might probably come from the original XMLHttpRequest, which also states methods in uppercase. Then we leave it as is and change later, if needed. Or we use `this.method.toUpperCase()`?

src/pretix/static/pretixbase/js/asynctask.js
raphaelm, line 225:
```suggestion
```

src/pretix/static/pretixbase/js/asynctask.js
```diff
@@ -218,10 +218,15 @@ $(function () {
...
         ));
 
-
-        console.log($(this).get(0))
         var formData = new FormData($(this).get(0))
         formData.append('ajax', '1');
+        if (e.originalEvent.submitter) {
+            // Not supported on IE, may lead to wrong results, but we don't support IE in the backend anymore
+            var submitter = e.originalEvent.submitter;
+            if (submitter.name) {
```
wiffbi, line 226: Sorry, I forgot we need to check if submitter itself is not null.

```suggestion
            if (submitter && submitter.name) {
```

wiffbi: How could I not see this!? Probably too late :)

src/pretix/control/views/checkin.py
```diff
@@ -206,14 +226,20 @@ def post(self, request, *args, **kwargs):
...
                     }, user=request.user)
                     checkin_created.send(op.order.event, checkin=ci)
+            return 'ok', request.POST.get('returnquery')
 
-            messages.success(request, _('The selected tickets have been marked as checked in.'))
+    def get_success_message(self, value):
+        if value[0] == 'reverted':
+            return _('The selected check-ins have been reverted.')
+        else:
+            return _('The selected tickets have been marked as checked in.')
```
wiffbi, line 235: This will also return the success message for checked-in for tickets marked as checked out. In line 229 we need a third option next to `'ok'`, or instead of ok, use something like `checked-in` and `checked-out`.

wiffbi: Sorry, I did not wait and just added the third option. I was too lazy to turn this into a dict of messages, but I think with three options that is kind of still okay. If there will be a fourth option, we should consider turning the if-elif-else block into a dict-lookup.

PR #2685, opened 2022-06-15
Add customized links to page footer

raphaelm:


src/pretix/control/views/event.py
```diff
@@ -195,6 +196,7 @@ def form_valid(self, form):
         self.save_meta()
         self.save_item_meta_property_formset(self.object)
         self.save_confirm_texts_formset(self.object)
+        self.save_footer_links_formset(self.object)
```
wiffbi, line 199: Why not just a `self.footer_links_formset.save()`? I guess it is for consistency with the other `save_*` above?

src/pretix/base/models/organizer.py
```diff
@@ -464,3 +465,25 @@ def get_events_with_permission(self, permission, request=None):
             return self.get_events_with_any_permission()
         else:
             return self.team.organizer.events.none()
+
+
+class OrganizerFooterLink(models.Model):
+    """
+    A footer link assigned to an organizer.
+    """
+    organizer = models.ForeignKey('Organizer', on_delete=models.CASCADE, related_name='footer_links')
```
wiffbi, line 474: How about adding a sort-field? Currently to change the order, I need to delete and add.

wiffbi (approved):
Looks good! Only minor text-label recommendations and an idea: maybe add a sort-order for footer links? And TIL about pretix’ `data-formset`. :)

On the event-settings, the propagated meta_noindex initially irritated me when looking at the footer-links as the propagated block somehow breaks the reading flow and I wasn’t sure, whether the footer links were affected as well. I moved the propagated meta_noindex block to the bottom and it is IMHO better.
Bildschirmfoto 2022-06-15 um 12 15 09
Bildschirmfoto 2022-06-15 um 12 14 57

A bit problematic IMHO is the layout=inline regarding missing labels and help_texts. The Link-Text being a I18nTextField only tells which language to put in and the URL-Field in inline-layout does not show help_text to describe, that only full URLs are supported. But this is definitely out of scope for this PR, just a general remark we might at some point need to think about.

src/pretix/control/views/event.py
raphaelm: Yes, in case we need future logic and to keep it consistent

src/pretix/base/models/organizer.py
raphaelm: We can always add that in later, my time budget did not allow for it now and I don't think it's critically important here. I don't want to encourage people adding loads of links here anyawys.

PR #2686, opened 2022-06-15
Thumbnails: Support animated GIFs

raphaelm:


src/pretix/helpers/thumb.py
```diff
@@ -171,24 +171,35 @@ def create_thumbnail(sourcename, size):
...
     if source.name.lower().endswith('.jpg') or source.name.lower().endswith('.jpeg'):
         # Yields better file sizes for photos
         target_ext = 'jpeg'
         quality = 95
+    elif source.name.lower().endswith('.gif'):
+        target_ext = 'gif'
+        quality = None
+        image_out.info = image.info
+        save_kwargs = {
+            'append_images': frames[1:] if len(frames) > 1 else [],
```
wiffbi, line 187: Isn’t a list with just one element returning an empty list on [1:] ? As we always have at least one frame, this should be enough:

```suggestion
            'append_images': frames[1:],
```

src/pretix/helpers/thumb.py
```diff
@@ -171,24 +171,35 @@ def create_thumbnail(sourcename, size):
...
+    frames = [resize_image(frame, size) for frame in ImageSequence.Iterator(image)]
+    image_out = frames[0]
+    save_kwargs = {}
 
     if source.name.lower().endswith('.jpg') or source.name.lower().endswith('.jpeg'):
         # Yields better file sizes for photos
         target_ext = 'jpeg'
         quality = 95
+    elif source.name.lower().endswith('.gif'):
+        target_ext = 'gif'
```
wiffbi, line 183: What about animated png? Browser-support is quite good and PIL supports it: https://caniuse.com/?search=apng

```suggestion
    elif source.name.lower().endswith('.gif') or source.name.lower().endswith('.png'):
        target_ext = source.name.lower()[-3:]
```

wiffbi (approved):
Looks good. Just suggested Animated-PNG support – browser support is rather good!

PR #2695, opened 2022-06-20
API: fix recalc of order total and fees after patch-operations (Z#23101774)

wiffbi:
When changing positions through the API, cached values were used for recalc of order total, etc. This PR fixes it by refreshing order from DB after performing patch operations in API.

### DIFF
--- a/src/pretix/base/services/orders.py
+++ b/src/pretix/base/services/orders.py
@@ -2317,6 +2317,7 @@ class OrderChangeManager:
                 self._check_and_lock_memberships()
                 try:
                     self._perform_operations()
+                    self.order.refresh_from_db()
                 except TaxRule.SaleNotAllowed:
                     raise OrderError(self.error_messages['tax_rule_country_blocked'])
             self._recalculate_total_and_payment_fee()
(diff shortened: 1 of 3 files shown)


raphaelm:
from my gut, this feels suboptimal, but if it helps with a bug, merge it, it's harmless, but take a note please that we analyse the root cause after my vacation, there should be more to this.

wiffbi:
Sory again, change_item should not update the price as this should be done explicitely with the price-attribute, which triggers a change_price operation.

This is tricky, a price-change does not seem to be reflected properly when `_recalculate_total_and_payment_fee` is executed. Although, if one calls a `self.order.refresh_from_db()` this overwrites e.g. the current order.status, which then fails the test `test_pending_free_order_stays_pending` as this relies on order.status not being changed. I regard this as an error, so added a `self.order.save()` after setting the status:

```
self.order.status = Order.STATUS_PENDING
self.order.save()
```

Similarly the `test_recalculate_reverse_charge` does not refresh the order from db when changing the InvoiceAddress.vat_id_validated to False and saving it.

I pushed a change to these tests. I do not think this is super urgent, so will wait after your vacation.

ser8phin (commented):
Looks good to me. Regarding previous points, if you want to wait before merging to look for the root cause, I support that decision though.

PR #2697, opened 2022-06-21
PDF render: enable word-wrap on hyphens (Z#23101879)

wiffbi:
In ReportLab the automatic word wrap only works on space characters. When rendering long strings (e.g. names) this creates ugly line-breaks. This PR adds an almost invisible space after hyphen characters to enable word wrapping in words with hyphens. Sadly the „zero width space“ is not supported/not recognized as a word boundary, but the „hair space“ works.

Before:
Bildschirmfoto 2022-06-21 um 09 12 04

After:
Bildschirmfoto 2022-06-21 um 09 11 22

### DIFF
--- a/src/pretix/base/pdf.py
+++ b/src/pretix/base/pdf.py
@@ -831,9 +831,10 @@ class Renderer:
             textColor=Color(o['color'][0] / 255, o['color'][1] / 255, o['color'][2] / 255),
             alignment=align_map[o['align']]
         )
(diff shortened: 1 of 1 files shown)


wiffbi:
As @pc-coholic suggested and @robbi5 confirmed: libpretixprint already supports word wrap on hyphens. So this PR makes rendering in pretix consistent with libpretixprint.

PR #2698, opened 2022-06-21
Checkout: label add-ons as free if price is 0 and add-ons are not included in the main product (Z#23101974)

wiffbi:
~~This PR adds a label „FREE“ in price columns to products that are free – previously addons have not had anything shown if they were free. I see how this works well with all addons being free, but IMHO is a bit confusing if free addons are combined with addons that have a price other than 0.~~

Furthermore this PR adds cart-icons to addons-checkboxes like they are used in the normal product selection. Sorry, if this mixes two only remotely related things into one PR.

UPDATE: this PR adds the label „free“ to addon-products, if their price is 0.00 and if the main product’s setting „price_included“ is set to false. If the main product’s setting „price_included“ is true, everything will stay as is and the label „free“ is only available to screen reader users.

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/fragment_addon_choice.html
+++ b/src/pretix/presale/templates/pretixpresale/event/fragment_addon_choice.html
@@ -127,29 +127,29 @@
                                             </del>
                                             <ins><span class="sr-only">{% trans "New price:" %}</span>
                                         {% endif %}
-                                    {% if item.free_price %}
-                                        <div class="input-group input-group-price">
-                                            <span class="input-group-addon">{{ event.currency }}</span>
-                                            <input type="number" class="form-control input-item-price"
-                                                    id="price-variation-{{form.pos.pk}}-{{ item.pk }}-{{ var.pk }}"
-                                                    placeholder="0"
-                                                    min="{% if event.settings.display_net_prices %}{{ var.display_price.net|money_numberfield:event.currency }}{% else %}{{ var.display_price.gross|money_numberfield:event.currency }}{% endif %}"
-                                                    name="cp_{{ form.pos.pk }}_variation_{{ item.id }}_{{ var.id }}_price"
-                                                    title="{% blocktrans trimmed with item=var.value %}Modify price for {{ item }}{% endblocktrans %}"
-                                                    step="any"
-                                                    value="{% if event.settings.display_net_prices %}{{ var.initial_price.net|money_numberfield:event.currency }}{% else %}{{ var.initial_price.gross|money_numberfield:event.currency }}{% endif %}"
-                                            >
-                                        </div>
-                                    {% elif not var.display_price.gross %}
-                                        <span class="text-uppercase">{% trans "free" context "price" %}</span>
-                                    {% elif event.settings.display_net_prices %}
-                                        {{ var.display_price.net|money:event.currency }}
-                                    {% else %}
-                                        {{ var.display_price.gross|money:event.currency }}
-                                    {% endif %}
-                                    {% if item.original_price or var.original_price %}
-                                        </ins>
-                                    {% endif %}
+                                        {% if item.free_price %}
+                                            <div class="input-group input-group-price">
+                                                <span class="input-group-addon">{{ event.currency }}</span>
+                                                <input type="number" class="form-control input-item-price"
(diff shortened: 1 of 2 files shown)


raphaelm:
I fear some customers will *not* like this at all, in the case you mention where every add-on is free this might be problematic in cases where add-ons are not actually "free" but "included" or "paid in a different way". For such a customer, this would be a breaking, u expected change that leads to bad communication so -1 from me :/

raphaelm:
Idea: We have an "add-ons are included in the price" flag in the ItemAddOn model. We could make the behaviour dependent on that and show "free" if the checkbox is unset and hide it for the entire category if it is set. Still a little worried because people might not have set it in the past if all products have price=0, but it would (a) offer a way out if someone complains and (b) would be logically consistent with the intent of that flag

wiffbi:
That makes total sense and aligns with what I felt when I wasn’t sure about the change. I think the price_included-flag is a good idea. That is actually how it is handled currently by addons with variants (I obviously did not see this). We could even go one step further and check, if all addons are free of charge and remove the „free“ label as well (at least visually, sr-only is still there) – this should be possible to check all prices when appending each `formsetentry['categories']`-dict. I will look into this.

raphaelm:
> We could even go one step further and check, if all addons are free of charge and remove the „free“ label as well

The downside in such an approach is that it is very "magic" and might confuse people e.g. when the "free" label suddenly disappears or even flickers, for example when all addons are free except one and then the paid one is sold out (and "hide sold out products" is set).

wiffbi:
You are right, to much magic and in combination with the sold out product confusing as well. So I will use the price_included flag only. Thanks!

src/pretix/presale/templates/pretixpresale/event/fragment_addon_choice.html
```diff
@@ -127,28 +127,29 @@ <h5 id="cp-{{ form.pos.pk }}-item-{{ item.pk }}-{{ var.pk }}-legend">{{ var }}</
...
+                                                        min="{% if event.settings.display_net_prices %}{{ var.display_price.net|money_numberfield:event.currency }}{% else %}{{ var.display_price.gross|money_numberfield:event.currency }}{% endif %}"
+                                                        name="cp_{{ form.pos.pk }}_variation_{{ item.id }}_{{ var.id }}_price"
+                                                        title="{% blocktrans trimmed with item=var.value %}Modify price for {{ item }}{% endblocktrans %}"
+                                                        step="any"
+                                                        value="{% if event.settings.display_net_prices %}{{ var.initial_price.net|money_numberfield:event.currency }}{% else %}{{ var.initial_price.gross|money_numberfield:event.currency }}{% endif %}"
+                                                >
+                                            </div>
+                                        {% elif not var.display_price.gross %}
+                                            <span class="text-uppercase">{% trans "free" context "price" %}</span>
```
wiffbi, line 144: All the changes in this block around that line are just indentation fixes, this line is the only actual addition

src/pretix/presale/templates/pretixpresale/event/fragment_addon_choice.html
```diff
@@ -248,23 +250,24 @@ <h4 id="cp-{{ form.pos.pk }}-item-{{ item.pk }}-legend">{{ item.name }}</h4>
...
+                                                id="price-item-{{ form.pos.pk }}-{{ item.pk }}"
+                                                min="{% if event.settings.display_net_prices %}{{ item.display_price.net|money_numberfield:event.currency }}{% else %}{{ item.display_price.gross|money_numberfield:event.currency }}{% endif %}"
+                                                name="cp_{{ form.pos.pk }}_item_{{ item.id }}_price"
+                                                title="{% blocktrans trimmed with item=item.name %}Modify price for {{ item }}{% endblocktrans %}"
+                                                value="{% if event.settings.display_net_prices %}{{ item.initial_price.net|money_numberfield:event.currency }}{% else %}{{ item.initial_price.gross|money_numberfield:event.currency }}{% endif %}"
+                                                step="any">
+                                    </div>
+                                {% elif not item.display_price.gross %}
+                                    <span class="text-uppercase">{% trans "free" context "price" %}</span>
```
wiffbi, line 265: All the changes in this block around that line are just indentation fixes, this line is the only actual addition

PR #2701, opened 2022-06-27
Banktransfer: add help-text to name of account holder

wiffbi:
Add some help_text regarding special characters in the account holder’s name. Punctuation though is not 100% correct as it seems e.g. ! can cause problems. Maybe use „some punctuation“ instead?

### DIFF
--- a/src/pretix/plugins/banktransfer/payment.py
+++ b/src/pretix/plugins/banktransfer/payment.py
@@ -80,6 +80,7 @@ class BankTransfer(BasePaymentProvider):
             )),
             ('bank_details_sepa_name', forms.CharField(
                 label=_('Name of account holder'),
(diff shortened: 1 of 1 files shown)


raphaelm (approved):

src/pretix/plugins/banktransfer/payment.py
```diff
@@ -80,6 +80,7 @@ def form_fields():
             )),
             ('bank_details_sepa_name', forms.CharField(
                 label=_('Name of account holder'),
+                help_text=_('Please note: special characters other than letters, numbers and some punctuation can cause problems with some banks.'),
```
raphaelm, line 83:
```suggestion
                help_text=_('Please note: special characters other than letters, numbers, and some punctuation can cause problems with some banks.'),
```

PR #2715, opened 2022-07-06
Add custom_field in orders API

mamico:
Add the `custom_field` in the `order.invoice_address` resource for REST API.


raphaelm:
@mamico Thanks for this PR again! Is it possible for you to sign our SLA so we're safe legal-wise here?

PR #2718, opened 2022-07-11
OpenID Connect support for customer accounts

raphaelm:
- [x] pretix as a relying party (RP)
  - [x] data model
  - [x] provider settings
  - [x] auto-discovery
  - [x] basic login
  - [x] allow to disable native auth
  - [x] integrate with checkout flow
    - [ ] test with widget
  - [x] support cross-domain auth
    - [ ] manual testing
  - [x] block password and email change
  - [x] think about email validation requirements
  - [x] REST API?
  - [x] Tests
- [x] pretix as a openid connect provider (OP)
  - [x] Client registration
  - [x] Authorization Code Flow
  - [x] Implicit Flow
  - [x] Hybrid Flow
  - [x] UserInfo endpoint
  - [x] Discovery
	  - [x] well-known endpoint
	  - [x] keys endpoint
  - [x] Tests and conformance checks
	  - [x] Comprehensive test coverage
	  - [x] Check against https://openid.net/specs/openid-connect-core-1_0.html#ServerMTI
	  - [x] Check against https://openid.net/wordpress-content/uploads/2018/06/OpenID-Connect-Conformance-Profiles.pdf
	  - [x] Test with https://www.certification.openid.net/login.html
- [ ] Docs


src/pretix/base/migrations/0219_auto_20220706_0913.py
```diff
@@ -0,0 +1,38 @@
...
+            ],
+            options={
+                'abstract': False,
+            },
+            bases=(models.Model, pretix.base.models.base.LoggingMixin),
+        ),
+        migrations.AddField(
+            model_name='customer',
+            name='provider',
+            field=models.ForeignKey(null=True, on_delete=django.db.models.deletion.CASCADE, related_name='customers', to='pretixbase.customerssoprovider'),
```
wiffbi, line 36: Shouldn’t this be `SET_NULL` instead of `CASCADE`? Or do we want the customer account to be deleted if we remove the SSOProvider as without without the SSOProvider that specific customer could not login (although currently email+password would still work)?

src/pretix/base/models/customers.py
```diff
@@ -39,12 +40,42 @@
...
+        return not self.customers.exists()
+
+
 class Customer(LoggedModel):
     """
     Represents a registered customer of an organizer.
     """
     id = models.BigAutoField(primary_key=True)
     organizer = models.ForeignKey(Organizer, related_name='customers', on_delete=models.CASCADE)
+    provider = models.ForeignKey(CustomerSSOProvider, related_name='customers', on_delete=models.CASCADE, null=True, blank=True)
```
wiffbi, line 78: See comment above regarding `CASCADE` vs `SET_NULL`

src/pretix/base/migrations/0219_auto_20220706_0913.py
raphaelm: I think this should actually be PROTECT

wiffbi: You are right! I thought that, if a SSOProvider is compromised, the organizer could simply remove it – but it is better to block delete, when that SSOProvider is in use and have the organizer set the SSOProvider’s `active` attribute to `False`.

src/pretix/presale/checkoutflow.py
```diff
@@ -301,15 +318,17 @@ def post(self, request):
                 self.cart_session['customer'] = request.customer.pk
                 self.cart_session['customer_cart_tied_to_login'] = True
                 return redirect(self.get_next_url(request))
-            elif self.login_form.is_valid():
+            elif "login-sso-data" in self.request.POST and self._handle_sso_login():
```
wiffbi, line 321: Should the `_handle_sso_login` be done in the if statement or would it be better to explicitly just check for the `"login-sso-data" in self.request.POST` and then check `self._handle_sso_login()` separately and output some error? Otherwise it will just skip to the next elif/else, which is probably not what we want? So something like:

```suggestion
            elif "login-sso-data" in self.request.POST:
                if not self._handle_sso_login():
                    messages.error(self.request, _("Login failed."))
```

src/pretix/presale/checkoutflow.py
```diff
@@ -301,15 +318,17 @@ def post(self, request):
...
+                return redirect(self.get_next_url(request))
+            elif self.login_form.is_valid() and self.event.settings.customer_accounts_native:
```
wiffbi, line 323: Tiny optimization, but check `customer_accounts_native` first to skip `is_valid` if it is not needed?

```suggestion
            elif self.event.settings.customer_accounts_native and self.login_form.is_valid():
```

wiffbi (commented):
Looks good so far, only minor comments. I will need to check with a SSOProvider next week.

src/pretix/presale/views/customer.py
```diff
@@ -541,3 +568,286 @@ def get(self, request, *args, **kwargs):
...
+                # This might either be a race condition or the email address is taken
+                # by a different customer account
+                try:
+                    customer = self.request.organizer.customers.get(
+                        provider=self.provider,
+                        identifier=identifier,
+                    )
+                except Customer.DoesNotExist:
+                    return self._fail(
+                        _('We were unable to use your login since the email address is already used for a different '
```
wiffbi, line 763: Should we output the email in this error message? I know this is a rare case, but I authorized pretix as a SSO-client in my Google account in a previous, failed attempt to login and now, when I press the „Login with Google“ Button I immediately get redirected from Google back to the login page with this error message. I need to go to the google homepage to see with what account/email I am logged in.

src/pretix/presale/views/customer.py
```diff
@@ -541,3 +568,286 @@ def get(self, request, *args, **kwargs):
...
+                        popup_origin,
+                    )
+        else:
+            if customer.is_active and customer.email != profile['email']:
+                customer.email = profile['email']
+                try:
+                    customer.save(update_fields=['email'])
+                except IntegrityError:
+                    return self._fail(
+                        _('We were unable to use your login since the email address is already used for a different '
```
wiffbi, line 774: See comment above

src/pretix/presale/views/customer.py
```diff
@@ -541,3 +568,286 @@ def get(self, request, *args, **kwargs):
...
+                return self._fail(
+                    _('Login was not successful. Error message: "{error}".').format(
+                        error='state parameter missing',
+                    ),
+                    popup_origin,
+                )
+
+            nonce, redirect_to = request.GET['state'].split('#')
+
+            if nonce != request.session[f'pretix_customerauth_{self.provider.pk}_nonce']:
```
wiffbi, line 683: I stumbled into this error, because of redirect error in my setup. So it should not happen IRL, but if something is wrong with the serssion whatsoever, this should not throw a 500 IMHO. So I suggest doing:

```suggestion
            if nonce != request.session.get(f'pretix_customerauth_{self.provider.pk}_nonce'):
```

wiffbi (approved):
Looks good, only moinor comments/questions! One general question: when a SSO-provider is deactivated, would it make sense to logout all currently logged in users/customers assigned to that SSO-provider? Or make this an option, when deactivating a SSO-provider?

raphaelm:
> Looks good, only moinor comments/questions! 

All resolved :) 

> One general question: when a SSO-provider is deactivated, would it make sense to logout all currently logged in users/customers assigned to that SSO-provider? Or make this an option, when deactivating a SSO-provider?

Yes, sounds like a good idea, fixed as well.

PR #2719, opened 2022-07-11
API: Add RPC-style check-in endpoints to support multi-event scan

raphaelm:
- [x] api operations 
  - [x] redeem
  - [x] search
- [x] security profiles
- [x] docs
- [x] tests
  - [x] regular tests
  - [x] test for performance impact
  - [x] test for permission consistency


raphaelm:
Results of some simple query performance analysis with PostgreSQL on two events with ~50k + ~10k orders:

- There is no difference between the old and new code for the redemption endpoint as performance is dominated by the index on orderposition.secret
- There is no difference between the old and new code for search in a single event
- Multi-event search has a performance penalty since PostgreSQL needs to replace an IndexScan with multiple BitmapIndexScan+BitmapOr

src/pretix/api/views/checkin.py
```diff
@@ -414,280 +755,143 @@ def redeem(self, *args, **kwargs):
...
+            nonce=s.validated_data.get('nonce'),
+            untrusted_input=True,
+            user=self.request.user,
+            auth=self.request.auth,
+            expand=self.request.query_params.getlist('expand'),
+            pdf_data=self.request.query_params.get('pdf_data', 'false') == 'true',
+            questions_supported=s.validated_data['questions_supported'],
+            canceled_supported=True,
+            request=self.request,  # this is not clean, but we need it in the serializers for URL generation
+            legacy_url_support=True,
```
robbi5, line 815: shouldn't this be `False`, because its used by the new checkin rpc url?

raphaelm: Oops, thanks!

PR #2728, opened 2022-07-24
Docs: Clarify customer name field editing

stapelberg:
Upon my first reading, I thought the name of a customer couldn’t be changed, but that’s not correct. PATCH requests that update the name_parts field work fine, and that influences the name field.

I think what the docs are trying to convey is that the name field is never settable because it’s derived from the name_parts field, so update the prose to make that more clear.


stapelberg:
I signed it!

raphaelm:
Thanks a lot! Sorry for the annoying CLA-bot, I wish I had time to write a better one :D

PR #2729, opened 2022-07-24
API: Allow to send activation email when creating customers

stapelberg:
Before this commit, customers whose accounts were created via the API would not
get any email, and would not be able to log in until they followed the reset
password flow — not a great experience overall.

With this commit, customers created via the API get the same activation email
that they would get when using the account creation form.

### DIFF
--- a/doc/api/resources/customers.rst
+++ b/doc/api/resources/customers.rst
@@ -131,7 +131,9 @@ Endpoints
 
 .. http:post:: /api/v1/organizers/(organizer)/customers/
 
-   Creates a new customer
+   Creates a new customer. In addition to the fields defined on the resource, you can pass the field ``send_email``
+   to control whether the system should send an account activation email with a password reset link (defaults to
+   ``false``).
 
    **Example request**:
 
@@ -143,7 +145,8 @@ Endpoints
       Content-Type: application/json
 
       {
-        "email": "test@example.org"
+        "email": "test@example.org",
+        "send_email": true
       }
 
    **Example response**:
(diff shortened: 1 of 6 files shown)


raphaelm:
Good point, we should add this. I think this will need to be opt-in though, similar to other endpoints (eg order creation, order cancellation) that have a send_email input value. The reason is that one intended use case of the customer API is bulk-creation of users for e.g. importing a legacy customer database ir syncing with a membership system – sending emails immediately might not be wanted behaviour in these cases (maybe the system isn't even live yet). With upcoming SSO support in #2718 there are also scenarios coming up where the email is not necessary.

(If you're not interested on doing that I think someone on our end can get around and build it)

stapelberg:
> Good point, we should add this. I think this will need to be opt-in though, similar to other endpoints (eg order creation, order cancellation) that have a send_email input value. The reason is that one intended use case of the customer API is bulk-creation of users for e.g. importing a legacy customer database or syncing with a membership system – sending emails immediately might not be wanted behaviour in these cases (maybe the system isn't even live yet). With upcoming SSO support in #2718 there are also scenarios coming up where the email is not necessary.
> 
> (If you're not interested in doing that I think someone on our end can get around and build it)

Makes sense! I think it might be easier (fewer round-trips) if one of you did the required changes. Feel free to take what’s in this PR and finish it, or do whatever is easiest for you :)

stapelberg:
Thanks for finishing this! It’s nice to see you properly refactored it and added a test :)

PR #2730, opened 2022-07-25
Change sort order in cart to keep subevents together

raphaelm:


wiffbi (commented):
Looks good, just some minor comments.

In line 176 there is `) + (0, ) * 10` – I think this should be `) + (0, ) * 11` or even *12 depending whether we add pos.subevent_id

PR #2735, opened 2022-07-25
Allow to hide cancellation fee from end-user in approval flow

raphaelm:


src/pretix/presale/templates/pretixpresale/event/order.html
```diff
@@ -373,7 +373,26 @@ <h3 class="panel-title">
...
+                                    {% blocktrans trimmed %}
+                                        If your request is approved, the organizer will determine if you will receive a
+                                        full refund or if a cancellation fee is deducted according to their cancellation
+                                        policy.
+                                    {% endblocktrans %}
+                                    {% if request.event.settings.cancel_allow_user_paid_refund_as_giftcard == "force" %}
+                                        {% trans "The refund will be issued in form of a gift card that you can use for further purchases." %}
+                                    {% elif request.event.settings.cancel_allow_user_paid_refund_as_giftcard == "option" %}
+                                        {% trans "The refund can be issued to your original payment method or as a gift card." %}
+                                    {% elif request.event.settings.cancel_allow_user_paid_refund_as_giftcard != "manually" %}
```
wiffbi, line 390: I can’t think of another option in these giftcard-settings, but would it be more future-proof to compare it to the actual value "off" for the of refund to original payment method?

```suggestion
                                    {% elif request.event.settings.cancel_allow_user_paid_refund_as_giftcard == "off" %}
```

src/pretix/presale/templates/pretixpresale/event/order_cancel.html
```diff
@@ -37,6 +50,7 @@ <h2>
             </p>
         {% endif %}
 
+        {% if not request.event.settings.cancel_allow_user_paid_require_approval or not request.event.settings.cancel_allow_user_paid_require_approval_fee_unknown %}
```
wiffbi, line 53: This doesn’t allow the combination of requiring approval and allowing the customer to choose a lower refund amount. AFAICT this was possible before? I think the `cancel_allow_user_paid_require_approval` is in there, because the `cancel_allow_user_paid_require_approval_fee_unknown` should only be in effect, when `cancel_allow_user_paid_require_approval_fee_unknown` is actually set. So shouldn’t this be:

```suggestion
        {% if request.event.settings.cancel_allow_user_paid_require_approval and request.event.settings.cancel_allow_user_paid_require_approval_fee_unknown %}
        {% else %}
```

wiffbi (commented):
Looks good, I have a comment regarding combining requiring approval and giving the customer the option to choose the refund amount.

src/pretix/presale/templates/pretixpresale/event/order.html
raphaelm: Not really part of this PR but why not ;)

src/pretix/presale/templates/pretixpresale/event/order_cancel.html
raphaelm: I think your code and my code are equivalent, as De Morgan says:
```
(not A) or (not b) == not(a and b)
```

wiffbi: You are absolutely right! Sorry for my confusion!

PR #2738, opened 2022-07-27
checkout: fix initial membership selection with only one choice

stapelberg:
Before this commit:
before

With this commit:
after


wiffbi (approved):
Thanks a lot!

PR #2743, opened 2022-07-29
custom URLs: decode idna for pretty URLs

stapelberg:
I noticed that setting an organizer’s custom domain to an internationalized domain name results in emails that use the `xn-` representation instead of the unicode representation. 

For example, setting the custom domain to `dömäin.example` results in links to `https:/xn--dmin-moa0i.example/account/pwrecover` in the password recovery email.

This PR fixes that, but I’m not sure if it’s complete (are there other places in pretix that need to be updated?) or correct (is this best practice for how to convert idna?). We should probably also have tests for this behavior. Feel free to take this PR and finish it, or let me know what needs to be done and I’ll try to find some time for it.


raphaelm:
Oh! Interesting that this has never come up before so far, seems like internationalized domains are still pretty uncommon.

> This PR fixes that, but I’m not sure if it’s complete (are there other places in pretix that need to be updated?)

Assuming that we only care about the event- or organizer-level domain, not the system domain: Yes, this PR is "complete" in the sense that it will apply your change **everywhere** we construct an URL from these domains, i.e. it will be applied to every single ``<a>`` tag on the page, every email link, every calendar invitation, …

A little bit better than ``eventreverse`` is probably the ``build_absolute_uri`` function in the same module, which is always used when the URL is communicated externally, the ``eventreverse`` function directly is usually called for internal use of the URL (say, an ``<a>`` tag within the system), as ``eventreverse`` will generate non-absolute URLs for organizers without a custom domain.

I wonder, though, if that still is **too** complete and might break stuff. For example, many payment plugins will pass a URL constructed this way to external systems as a callback URL. One example: https://github.com/pretix/pretix-saferpay/blob/11be0d4e1972f336d2aa91cf66f4af9424881575/pretix_saferpay/payment.py#L452

The average state of IT systems at payment providers is "old and fragile". Pure-ASCII representation of the domain probably worked fine, but I'm worried that if we deploy this change, payments might suddenly stop working for customers with an internationalized domain name.

A more complex alternative could be to, say, introduce a ``human_readable=True`` parameter to the ``build_absolute_uri`` function, to only change the representation for places where we care about user-visible URLs specifically, such as emails, calendar invites, or e.g. voucher redemption URLs shown in the backend. Finding and changing all of these places will obviously be more effort.

A cheap option might be implementing this on the level of our Markdown render, fixing (a) the HTML part of emails (b) everywhere else we show URLs as part of rich text. This won't help for other places like calendar invites where it would still be nice to have, though.

I'm a little unsure how best to proceed from here. Maybe @wiffbi also has some thoughts on the matter next week :)

> or correct (is this best practice for how to convert idna?).

I don't know, I havend worked with idna much, but people on the internet claim that the idna codec from the Python standard library only supports RFC 3490, which was later superseded by RFC 5891, recommending use of a third-party library instead:
https://pypi.org/project/idna/

Given that that this PR will add an additional urlparse()-encode()-decode() roundtrip to potentially hundreds of URL constructions on every single page view, we might also want to check if there's a significant performance impact and we need some thread-level caching.

raphaelm:
Oh, and one more thing, I'm wondering if there is any flavor of a [IDN homoglyph attack](https://en.wikipedia.org/wiki/IDN_homograph_attack) that we'll need to care about in the case of untrusted input (eg pretix SaaS)… I can't come up with an attack vector right now, but I'm also not sure.

pc-coholic:
This might be just hearsay - but I think remember from when IDNs were introduced, that they work fine for webbrowsers, but that email clients still need the xn-notation for To and From-addresses.

But: this might be way outdated...

raphaelm:
@pc-coholic You are correct, IDN are basically unsupported in email **addresses** in almost all email clients email infrastructure, but we're talking about URLs in the email **content** here which is hopefully better, since most clients just use a modern HTML engine for rendering. Except for Outlook, we'd need to double-check Outlook :/

stapelberg:
> A little bit better than `eventreverse` is probably the `build_absolute_uri` function in the same module, which is always used when the URL is communicated externally

Noted!

> A more complex alternative could be to, say, introduce a `human_readable=True` parameter to the `build_absolute_uri` function, to only change the representation for places where we care about user-visible URLs specifically, such as emails, calendar invites, or e.g. voucher redemption URLs shown in the backend. Finding and changing all of these places will obviously be more effort.

That sounds like the safest path forward. It would be sufficient for my needs, and unlikely to break existing payment plugins. Thanks also for mentioning that — I hadn’t checked before, but it looks like my payment provider handles idna domains without any problems :)

> I don't know, I havend worked with idna much, but people on the internet claim that the idna codec from the Python standard library only supports RFC 3490, which was later superseded by RFC 5891, recommending use of a third-party library instead:
> https://pypi.org/project/idna/

Good catch! https://www.unicode.org/reports/tr46/ is a good read regarding the differences between IDNA2003 and IDNA2008. I agree that we should use the third-party idna library so that we implement IDNA2008.

> Given that that this PR will add an additional urlparse()-encode()-decode() roundtrip to potentially hundreds of URL constructions on every single page view, we might also want to check if there's a significant performance impact and we need some thread-level caching.

We could probably just memoize the netloc→replacement mapping. I’m not sure how expensive the encoding/decoding is.

wiffbi:
AFAIK idna is basically just the visual representation the browser/viewer translates internally to pure ASCII domains – so as far as I understand it, internationalized domains do not (or even should not?) get passed to DNS. As these domains are purely visual I think it is a good idea as @raphaelm suggested to add a parameter `human_readable` to `build_absolute_uri` and only use it where appropriate/needed. Also, I think we should check the ASCII representation of `domain` first, whether we need to encode to idna at all and skip that step, if not. I guess generally enabling it through Markdown as @raphaelm suggested could be a good catch-all approach for all HTML we output through rich-text. I am not sure though, whether this catches all the places – not only in ics-files, etc. but also in email templates. I will need to double check. Otherwise, manually adding it in the places where we need it, IMHO could be doable as well – even if we do not find all places in the first try.

PR #2756, opened 2022-08-04
PDF-Editor: fix other/other-i18n input prefill bug (Z#23104692)

wiffbi:
When changing the selected object to either type `other` or `other-i18n`, the text-value of a previously selected object of the same type was still present in the input textarea. This PR resets the content of text inputs `other` and `other-i18n`.

Also when switching between type other and other-i18n, this PR adds the feature, that the content from other is copied to other-i18n. When switching from other_i18n to other, the value of the first i18n-textarea is copied to the other-textarea.

### DIFF
--- a/src/pretix/static/pretixcontrol/js/ui/editor.js
+++ b/src/pretix/static/pretixcontrol/js/ui/editor.js
@@ -520,7 +520,7 @@ var editor = {
         }
     },
 
-    _update_values_from_toolbox: function () {
+    _update_values_from_toolbox: function (e) {
         var o = editor.fabric.getActiveObject();
         if (!o) {
             o = editor.fabric.getActiveGroup();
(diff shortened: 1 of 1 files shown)


src/pretix/static/pretixcontrol/js/ui/editor.js
```diff
@@ -553,8 +553,16 @@ var editor = {
             $("#toolbox-content-other-help").toggle($("#toolbox-content").val() === "other" || $("#toolbox-content").val() === "other_i18n");
             o.content = $("#toolbox-content").val();
             if ($("#toolbox-content").val() === "other") {
+                if (this.id === "toolbox-content") {
```
raphaelm, line 556: - This is the first code introducing ``this`` in this function, which I'm not sure of whether it is a good idea. First of all, as a reader, I had to think quite a while what ``this`` is here since we're not *directly* in an event handler. This could be solved with a comment.
- This will also *overwrite* values of the other field if there currently was one. Should we only call val() if it was previously empty? Not sure what would be expected behaviour from users here, to be honest, we can also keep it like this.

wiffbi: I thought about using `event.target` instead of `this`, but as you said, we are not _directly_ in an event handler. So yes, a comment is a must here!

Regarding overwrite: yes, it will overwrite, but the content in there would be the text of the last content-other object selected – so not really helpful. Also, it should actually be empty anyways (see changes in Line 639 ff).

raphaelm: 👍

wiffbi: I actually changed it to `e.target` – `_update_values_from_toolbox` is bound to several events and only once used internally, which I passed the event to it. I also updated the comment.

PR #2761, opened 2022-08-09
Add last order modification time as an email placeholder

raphaelm:


src/pretix/base/email.py
```diff
@@ -469,6 +470,19 @@ def base_placeholders(sender, **kwargs):
                 }
             ),
         ),
+        SimpleFunctionalMailTextPlaceholder(
+            'last_order_modification_date_and_time', ['order', 'event'],
```
wiffbi, line 474: IMHO the name is a bit misleading as I initially read it as the last time the order was modified – I think either `latest_order_modification_date_and_time` or maybe even better something with _deadline_:
```suggestion
            'order_modification_deadline_date_and_time', ['order', 'event'],
```

wiffbi (approved):
Looks good and works as expected. I only stumbled upon the naming of the placeholder, see my comment.

PR #2767, opened 2022-08-16
Bank transfer: Minimal implementation of Swiss QR-bill

raphaelm:
This generates a new style of QR codes for bank transfers to Switzerland

Sample code has been tested with https://validation.iso-payments.ch/


stapelberg:
I have successfully verified that the QR codes generated with this PR are scanned correctly by the ZKB app and by the PostFinance app. The textual form also matches what my qrbill program generates, so this seems to work fine AFAICT! :)

wiffbi (approved):
Looks good. Just added some minor improvements regarding tabs and accessibility (see my commits) as well as added a space after the tabs-container so the retry-payment-button is not touching the tabs-container. I compressed the swiss-cross SVG.

PR #2773, opened 2022-08-26
Retain filenames when cloning an event

### LINKED ISSUE #2771: Cloning an event with a configured mail_attachment_new_order does change its name
**Describe the bug**
#2384 did introduce the option to add an email attachment to order confirmation emails. Greatly appreciated for attaching legally required documents.
The original filename is persisted by adding a random string while storing to filesystem and removing it again (https://github.com/pretix/pretix/blob/master/src/pretix/base/services/mail.py#L462) when attaching the file to the order confirmation.
But when creating a new event based on an existing one, the filename is lost due to the copied filename being generically generated in https://github.com/pretix/pretix/blob/9101b5b69dd87a34169cd8a9c689cb70651d2014/src/pretix/base/models/event.py#L923.

**Expected behavior**
It is expected that when an event is created based on an existing one, that the `mail_attachment_new_order` original filename is preserved.

**Screenshots**

**Additional context**
I'm happy to do a PR by e.g. calling `clean_filename` and adding a new random string if the key of the `other.settings._object` is `mail_attachment_new_order`, while copying files https://github.com/pretix/pretix/blob/9101b5b69dd87a34169cd8a9c689cb70651d2014/src/pretix/base/models/event.py#L923 in `copy_data_from` method.

vchrisb:
attempt to fix #2771 by retaining the original mail_attachment_new_order filename.

### DIFF
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -920,10 +920,7 @@ class Event(EventMixin, LoggedModel):
             if s.value.startswith('file://'):
                 fi = default_storage.open(s.value[len('file://'):], 'rb')
                 nonce = get_random_string(length=8)
-                if s.key == 'mail_attachment_new_order':
-                    fname_base = clean_filename(os.path.basename(s.value))
-                else:
-                    fname_base = s.key
+                fname_base = clean_filename(os.path.basename(s.value))
 
                 # TODO: make sure pub is always correct
                 fname = 'pub/%s/%s/%s.%s.%s' % (
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -921,7 +921,7 @@ class Event(EventMixin, LoggedModel):
                 fi = default_storage.open(s.value[len('file://'):], 'rb')
                 nonce = get_random_string(length=8)
                 if s.key == 'mail_attachment_new_order':
-                    fname_base = clean_filename(os.path.basename(s.value[len('file://'):]))
+                    fname_base = clean_filename(os.path.basename(s.value))
                 else:
                     fname_base = s.key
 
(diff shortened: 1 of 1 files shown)


pc-coholic:
@cla-bot check

src/pretix/base/models/event.py
```diff
@@ -917,11 +918,16 @@ def _walk_rules(rules):
             s.object = self
             s.pk = None
             if s.value.startswith('file://'):
-                fi = default_storage.open(s.value[7:], 'rb')
+                fi = default_storage.open(s.value[len('file://'):], 'rb')
```
wiffbi, line 921: I like that you removed that magical number. :)

src/pretix/base/models/event.py
```diff
@@ -917,11 +918,16 @@ def _walk_rules(rules):
...
                 nonce = get_random_string(length=8)
+                if s.key == 'mail_attachment_new_order':
+                    fname_base = clean_filename(os.path.basename(s.value[len('file://'):]))
```
wiffbi, line 924: AFAIK and in my tests I did not need to remove `file://` when using `os.path.basename`. So maybe shorten to this?

```suggestion
                    fname_base = clean_filename(os.path.basename(s.value))
```

wiffbi (commented):
Thanks a lot for this PR. Looks good, I just added a comment regarding `basename` and `file://`.

I am not going to approve this PR yet though, as I think it might be generally useful for all files to keep the original filename when copying events. I just want to make sure internally why `s.key` was used in the first place and not `os.path.basename(s.value)`. Due to vacation this will need to wait until next week.

wiffbi:
Thanks again for your PR. I checked and we would like to enable keeping the basename for all files in event settings. This would basically mean to just remove the if/else block regarding `if s.key == 'mail_attachment_new_order':`. Would you be willing to change your PR accordingly? I could do if you want, but I do not want to interfere with your contribution.

PR #2783, opened 2022-09-08
Fix discount for distinct dates when only certain products are allowed (Z#23107419)

wiffbi:
When a discount needs discounted products to be from different dates from an event series, the product-filter for this discount (`initial_candidates`) was not applied. This PR fixes this.

AFAICT tests in tests/base/test_pricing_discount.py currently do not test for the product filter at all. Is this somewhere else or something we should add?


raphaelm:
Yes please this absolutely should have tests! ;)

PR #2786, opened 2022-09-09
api: allow changing customer password

stapelberg:
To connect pretix to an external source-of-truth for customer accounts, setting the password at account creation time is a good first step (https://github.com/pretix/pretix/pull/2758).

In order to implement a password reset feature, we also need to change an existing customer’s password, which this PR implements.

### DIFF
--- a/doc/api/resources/customers.rst
+++ b/doc/api/resources/customers.rst
@@ -26,8 +26,9 @@ date_joined                           datetime                   Date and time o
 locale                                string                     Preferred language of the customer
 last_modified                         datetime                   Date and time of modification of the record
 notes                                 string                     Internal notes and comments (or ``null``)
-password                              string                     Can only be set during creation of a new customer, will
-                                                                 not be included in any responses.
+password                              string                     Can be set during creation of a new customer, or
+                                                                 changed when updating a customer. Will not be included
+                                                                 in any responses.
 ===================================== ========================== =======================================================
 
 .. versionadded:: 4.0
@@ -35,6 +36,10 @@ password                              string                     Can only be set
 .. versionchanged:: 4.3
 
    Passwords can now be set through the API during customer creation.
+
+.. versionchanged:: 4.13
+
+   Passwords can now be changed through the API.
 
 Endpoints
 ---------
--- a/src/pretix/api/serializers/organizer.py
+++ b/src/pretix/api/serializers/organizer.py
@@ -72,7 +72,19 @@ class CustomerSerializer(I18nAwareModelSerializer):
     class Meta:
         model = Customer
         fields = ('identifier', 'external_identifier', 'email', 'name', 'name_parts', 'is_active', 'is_verified', 'last_login', 'date_joined',
-                  'locale', 'last_modified', 'notes')
+                  'locale', 'last_modified', 'notes', 'password')
+        extra_kwargs = {
+            "password": {"write_only": True},
+        }
+
+    def update(self, instance, validated_data):
+        for attr, value in validated_data.items():
+            if attr == 'password':
+                instance.set_password(value)
+            else:
+                setattr(instance, attr, value)
+        instance.save()
+        return instance
 
 
 class CustomerCreateSerializer(CustomerSerializer):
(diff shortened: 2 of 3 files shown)


raphaelm:
I'm a little bit undecided on this one right now.

On one hand, it's a very easy patch and it's always good to make the API "more complete".

On the other hand, it feels like an invitation to use cases I would consider a bad practice. Setting the password through the API during customer creation has many valid use cases such as importing a legacy set of customer accounts when handling the distribution through a separate channel. Setting the password for existing accounts however seems to have just one use case, the one you are trying to build: Syncing with an external user database.

While I know you well enough that I believe you have a use case where the trade off is adequate, this is not a use case we want to promote in any way. First, transmitting plain-text passwords across multiple systems isn't really a best practice security-wise. Second, it includes quite some potential for confusion. If someone uses the password reset link of your central database, the password will also be updated in pretix. But if someone uses the password reset (or password change) feature *in pretix*, they will suddenly have two different passwords and there's no good way to sync them again (except with another password change on the other site).

I think there are only ways to handle all this that have a chance of being understood by the average user:

1. Explaining to the user that it's two different accounts, and just creating them at the same as a convenience. However, any password changes etc need to be done separately. This is what's currently possible.

2. Explaining to the user that it's just one account, but then also handling that consistently (there's only one password reset form, only one password change form, only one signup form). I don't think this should be implemented through syncing of plain-text passwords, but through a proper Single-Sign-On method. My PR #2718, which will likely get merged in just a few days, brings the basis for integrating such SSO methods to pretix and implements OpenID Connect as a first, well-standardized protocol.

@pc-coholic @wiffbi Do you have any opinions? I'm currently leaning towards "merging this, but adding some more warning flags, and no specific adjustments just for this use case (e.g. disabling password reset in pretix)".

pc-coholic:
> @pc-coholic @wiffbi Do you have any opinions? I'm currently leaning towards "merging this, but adding some more warning flags, and no specific adjustments just for this use case (e.g. disabling password reset in pretix)".

I agree with your assessment - but I can already see that the very next PRs regarding this feature will absolutely be the option to either disable password resets or to replace them with a custom redirect. So we would need to add that to our list of "hard nopes" and really, really stick to it and - more importantly - make sure that we cannot imagine ourselves a scenario where we might want the disabling of such functionality.

Alternatively we can also just say "nope, not opening this box of snakes" and ask OP to put that into a plugin. Would require a little bit more code, but that way we can at least discourage casual misuse of this feature.

stapelberg:
> While I know you well enough that I believe you have a use case where the trade off is adequate, this is not a use case we want to promote in any way. First, transmitting plain-text passwords across multiple systems isn't really a best practice security-wise.

Right — in our setup, we encrypt the password in the client’s browser for a public key that only the pretix server has, but I agree that it’s an icky situation overall.

Ideally, we could generate a password hash in the client’s browser that django accepts and send that to the server. But I think that will be a much more difficult feature request to get accepted across all parts of the stack :) 

> 2\. Explaining to the user that it's just one account, but then also handling that consistently (there's only one password reset form, only one password change form, only one signup form).

Yes, this is our current approach. We already block the signup form in nginx:
```
	location = /account/register {
		return 403;
	}
```
…and could easily do the same for the password reset form or any other potentially confusing handlers (including a custom redirect, so I wouldn’t worry too much about doing these at the pretix level).

> 2\. My PR [OpenID Connect support for customer accounts #2718](https://github.com/pretix/pretix/pull/2718), which will likely get merged in just a few days, brings the basis for integrating such SSO methods to pretix and implements OpenID Connect as a first, well-standardized protocol.

That’s fair! Maybe OpenID Connect really is the best long-term solution for this situation.

I’m not sure if/when I’ll have time to do an OpenID Connect integration, so perhaps we’ll just carry this patch locally for a while :)

wiffbi:
> @pc-coholic @wiffbi Do you have any opinions? I'm currently leaning towards "merging this, but adding some more warning flags, and no specific adjustments just for this use case (e.g. disabling password reset in pretix)".

I think so too. Especially with a hint to redirect `/account/register`, `/control/forgot`, etc. on another level outside pretix (e.g. nginx).

raphaelm:
I think we close it for now, it's easy enough for @stapelberg to carry on the patch as it has little potential for conflicts, it's easy enough for us to merge it if we want at any later point. And it would be hard to close the box again after we opened it...

PR #2788, opened 2022-09-13
Fix: display cleaned filename in upload forms

wiffbi:


raphaelm (commented):
let's quickly talk this through later, please

src/pretix/control/forms/__init__.py
```diff
@@ -127,7 +127,7 @@ def is_img(self):
         def __str__(self):
             if hasattr(self.file, 'display_name'):
                 return self.file.display_name
-            return os.path.basename(self.file.name).split('.', 1)[-1]
```
raphaelm, line 130: as discussed, use clean_filename

PR #2790, opened 2022-09-13
New retry logic for webhooks

raphaelm:
Long description is in the comment in the diff


src/pretix/control/views/organizer.py
```diff
@@ -1263,6 +1265,26 @@ def webhook(self):
...
+            manually_retry_all_calls.apply_async(args=(self.webhook.pk,))
+            messages.success(request, _('All requests will now be scheduled for an immediate attempt. Please '
+                                        'allow for a few minutes before they are processed.'))
+        elif request.POST.get("action") == "drop":
+            self.request.organizer.log_action('pretix.webhook.retries.dropped', user=self.request.user, data={
+                'webhook': self.webhook.pk,
+            })
+            self.webhook.retries.all().delete()
+            messages.success(request, _('All requests will now be scheduled for an immediate attempt. Please '
+                                        'allow for a few minutes before they are processed.'))
```
wiffbi, line 1282: The message needs to be about deleted retries. Something like:

```suggestion
            messages.success(request, _('All unprocessed webhooks have been stopped from retrying.'))
```

src/pretix/control/templates/pretixcontrol/organizers/webhook_logs.html
```diff
@@ -6,13 +6,42 @@ <h1>{% blocktrans with url=webhook.target_url %}Logs for webhook {{ url }}{% end
...
+                    <button type="submit" name="action" value="expedite" class="btn btn-success">
+                        {% trans "Retry now" %}
+                    </button>
+                    <button type="submit" name="action" value="drop" class="btn btn-danger">
+                        {% trans "Stop retrying" %}
+                    </button>
+                </p>
+                <p class="help-block">
+                    {% blocktranslate trimmed with minutes=5 %}
+                        Webhooks scheduled to be retried in less than {{ minutes }} minutes may be not listed here and can
```
wiffbi, line 30:
```suggestion
                        Webhooks scheduled to be retried in less than {{ minutes }} minutes may not be listed here and can
```

src/pretix/control/views/organizer.py
```diff
@@ -1252,6 +1253,7 @@ class WebHookLogsView(OrganizerDetailViewMixin, OrganizerPermissionRequiredMixin
     def get_context_data(self, **kwargs):
         ctx = super().get_context_data(**kwargs)
         ctx['webhook'] = self.webhook
+        ctx['retry_count'] = self.webhook.retries.count()
```
wiffbi, line 1256: Should we maybe add `retry_celery_cutoff` to ctx and use it to tell the minutes in webhook_logs.html? Probably not worth the effort (`retry_celery_cutoff` would need to be moved outside of `send_webhook`), but would help in the future if/when we change the cutoff-time for celery-based retries.

```suggestion
        ctx['retry_count'] = self.webhook.retries.count()
        ctx['retry_cutoff_minutes'] = retry_celery_cutoff // 60
```

raphaelm: I mostly put it into variable in the template so if we change it, translations do not need to change. I don't think it's worth moving it to a global variable for now.

PR #2794, opened 2022-09-14
Fix name placeholder not being used correctly in scheduled emails

raphaelm:


src/pretix/plugins/sendmail/models.py
```diff
@@ -155,12 +155,13 @@ def send(self):
                     positions = [p for p in positions if p.subevent_id == self.subevent_id]
 
                 for p in positions:
-                    email_ctx = get_email_context(event=e, order=o, position_or_address=ia, position=p)
                     try:
                         if p.attendee_email and (p.attendee_email != o.email or not o_sent):
+                            email_ctx = get_email_context(event=e, order=o, position_or_address=p, position=p)
```
wiffbi, line 160: Looking at the definition of `get_email_context`, if `position_or_address` is the same as `position`, we can omit it completely. So we might simplify – although I can see the benefit of clearly stating `position_or_address` here as to make the differentiation to line 164 more clear. I am undecided.

```suggestion
                            email_ctx = get_email_context(event=e, order=o, position=p)
```

wiffbi (commented):
Looks good. Just a minor comment regarding a parameter. One question regarding the algorithm though: we loop through all positions of an order and send an email for each position – except, when the `position.attendee_email` is the same as `o.email`, then we only send one email. So hypothetical when multiple positions have the same email-address, but that email-adress is not `o.email`, then multiple emails would be sent? If so, do we ant that? Or should we filter positions to only unique email-adresses with fallback being `o.email`?

src/pretix/plugins/sendmail/models.py
raphaelm: I made another change to actually *never* set ``position_or_address`` but always rely on the implementation inside ``get_email_context``. That feels much safer against future confusions!

raphaelm:
> So hypothetical when multiple positions have the same email-address, but that email-adress is not o.email, then multiple emails would be sent? If so, do we want that?

Yes, it **must** be that way! Hypothetical scenario:

- Order with address1@example.org
  - Position 1 with address1@example.org
  - Position 2 with address2@example.org
  - Position 3 with address2@example.org

address1@example.org will get **one** email with the **secret order link**. address2@example.org will get **two** emails with **two different secret ticket links**. If we were to deduplicate here, we'd either need to have send out the order link (in which case address2@example.org gets full access to cancellation of the order, download of position 1, … which is not allwed) or have a complex way of sending multiple links in one email… I'd rather not ;)

raphaelm:
> except, when the position.attendee_email is the same as o.email, then we only send one email

note also that this is only the case if ``send_to_orders`` is set. if only ``send_to_attendees`` is set, every ticket will get its own email

wiffbi:
Thanks for explaining the flow regarding attendee-tickets!

src/pretix/plugins/sendmail/models.py
wiffbi: Passing the `invoice_address` in this case is just an optimization to not have it „calculated“ on every invocation of `get_email_context`, right? Because otherwise we could leave the `invoice_address=ia` parameter out as well.

raphaelm: >Passing the invoice_address in this case is just an optimization to not have it „calculated“ on every invocation of get_email_context, right? 
Correct

wiffbi (approved):
Works as intended!

PR #2811, opened 2022-09-21
Waiting list: Allow transfer to other subevent

Bentrex95:


src/pretix/control/views/waitinglist.py
```diff
@@ -366,3 +427,87 @@ def get_success_url(self) -> str:
...
+        )
+    def get_object(self, queryset=None) -> WaitingListEntry:
+        try:
+            return self.request.event.waitinglistentries.get(
+                id=self.kwargs['entry'],
+                event=self.request.event,
+                voucher__isnull=True,
+            )
+        except WaitingListEntry.DoesNotExist:
+            raise Http404(_("The requested entry does not exist."))
```
wiffbi, line 455: Have a look at `get_object_or_404`, it shortens this code.

src/pretix/control/views/waitinglist.py
```diff
@@ -366,3 +427,87 @@ def get_success_url(self) -> str:
...
+        return reverse('control:event.orders.waitinglist', kwargs={
+            'event': self.request.event.slug,
+            'organizer': self.request.event.organizer.slug
+        })
+
+    @transaction.atomic
+    def update(self, request, *args, **kwargs):
+        self.object = self.get_object()
+        success_url = self.get_success_url()
+        self.object.log_action('pretix.event.orders.waitinglist.transfered', user=self.request.user)
```
wiffbi, line 467:
```suggestion
        self.object.log_action('pretix.event.orders.waitinglist.transferred', user=self.request.user)
```

src/pretix/control/views/waitinglist.py
```diff
@@ -366,3 +427,87 @@ def get_success_url(self) -> str:
...
+            raise Http404(_("The requested entry does not exist."))
+
+    def get_success_url(self) -> str:
+        return reverse('control:event.orders.waitinglist', kwargs={
+            'event': self.request.event.slug,
+            'organizer': self.request.event.organizer.slug
+        })
+
+    @transaction.atomic
+    def update(self, request, *args, **kwargs):
```
wiffbi, line 464: Where does this `update` method come from? I do not think this works/is called at all. I might be wrong, but shouldn’t that be `form_valid`? I haven’t looked to deep into this yet.

wiffbi (commented):
I just looked at updating a single waitinglist entry. Works although the UpdateView does not seem to actually call the `update` method, see my comment. Other than that, minor tweaks. Also run flake8 and isort.

src/pretix/control/views/waitinglist.py
wiffbi: Look at e.g. `DiscountUpdate` https://github.com/pretix/pretix/blob/master/src/pretix/control/views/discounts.py#L96-L136
There is also a `form_invalid` method.

src/pretix/control/views/waitinglist.py
```diff
@@ -366,3 +427,87 @@ def get_success_url(self) -> str:
...
+        self.object.log_action('pretix.event.orders.waitinglist.transfered', user=self.request.user)
+        messages.success(self.request, _('The selected entry has been transfered.'))
+        if "next" in self.request.GET and is_safe_url(self.request.GET.get("next"), allowed_hosts=None):
+            return redirect(self.request.GET.get("next"))
+        return HttpResponseRedirect(success_url)
+
+    def get_form_kwargs(self):
+        kwargs = super().get_form_kwargs()
+        kwargs['event'] = self.request.event
+        return kwargs
```
wiffbi, line 476: This is already done by `EventBasedFormMixin`, which `UpdateView` inherits from.

```suggestion
```

src/pretix/control/views/waitinglist.py
```diff
@@ -366,3 +427,87 @@ def get_success_url(self) -> str:
...
+
+
+
+    #def dispatch(self, request, *args, **kwargs):
+        #self.fields['subevent'] = WaitingListEntry.objects.filter(event=self.request.event)
+    #    return super().dispatch(request, *args, **kwargs)
+    def get_queryset(self):
+        return WaitingListEntry.objects.filter(
+            event=self.request.event,
+        )
```
wiffbi, line 446: Not needed, you already filter in `get_object`
```suggestion
```

src/pretix/control/views/waitinglist.py
```diff
@@ -47,16 +47,19 @@
 from django.utils.timezone import now
 from django.utils.translation import gettext_lazy as _, pgettext
 from django.views import View
-from django.views.generic import ListView
-from django.views.generic.edit import DeleteView
-
+from django.views.generic import ListView, FormView
+from django.views.generic.edit import DeleteView, UpdateView
```
wiffbi, line 51: Do not import UpdateView from django, import it from . like https://github.com/pretix/pretix/blob/master/src/pretix/control/views/discounts.py#L48 – that way you get `get_form_kwargs` through inheritance as described further down.

src/pretix/control/views/waitinglist.py
Bentrex95: I think this exists for a reason - leaving it out results in 'NoneType' object has no attribute 'subevents' in form class WaitingListEntryEditForm for "self.event.subevents.all()" . The standart method doesnt seem to hand over self.event

wiffbi:
Two things to note:

- I did not find a common way to block access to `/<id>/update`-URL if no subevents are available. The link is not shown through UI, but still accessible (similar to when accessing `./subevents/` on an event without subevents) – not super bad, but maybe we should think about adding a redirect or something.
- On the edit-form page in the SQL-console I get a warning about two similar SQL-queries regarding `SELECT … FROM "pretixbase_subevent"`. One is from setting the queryset-field in `WaitingListEntryEditForm`. I do not know where the other is coming from. Maybe some `fetch_related`?

src/pretix/control/forms/waitinglist.py
```diff
@@ -0,0 +1,22 @@
...
+        self.fields['subevent'].empty_label = None
+        self.fields['subevent'].queryset = self.event.subevents.filter(active=True)
+
+    class Meta:
+        model = WaitingListEntry
+        fields = [
+            'subevent',
+        ]
+        field_classes = {
+            'subevent': SafeModelChoiceField,
```
raphaelm, line 21: We always use select2 for subevent choices, we should do so here as well

src/pretix/control/templates/pretixcontrol/waitinglist/index.html
```diff
@@ -261,9 +264,17 @@ <h1>{% trans "Waiting list" %}</h1>
...
             <div class="batch-select-actions">
                 <button type="submit" class="btn btn-danger btn-save" name="action" value="delete">
-                    <i class="fa fa-trash"></i>{% trans "Delete selected" %}
+                    <i class="fa fa-trash"></i>
+                    {% trans "Delete selected" %}
+                </button>
+                {% if request.event.has_subevents %}
+                <button type="submit" class="btn btn-primary btn-save" name="action" value="update">
+                    <i class="fa fa-share-square-o" aria-hidden="true"></i>
+                    {% trans "Transfer selected" %}
```
raphaelm, line 273: The wording in this PR is very unclear. Is it an "update" or a "transfer" what we're doing here?

wiffbi: Sorry, this button should not be here as this was part of Ben’s attempt for batch-transfer. I will remove it. But yes, action should have been „transfer“.

wiffbi: Regarding `update` vs. `transfer` – should the URL have the parameter called /transfer/ instead of /update/ as well? And probably the View-Class should be called `EntryTransfer` instead of `EntryUpdate`?

src/pretix/control/forms/waitinglist.py
wiffbi: I will add select2. In `typeahead.subevent_select2` I do not see a way to filter only active subevents – only options seems to filter by name and location. IMHO that is okay to also offer inactive events when transfering to subevents, as maybe one wants to offer a special subevent for waitinglist-users that is not active (yet).

src/pretix/control/templates/pretixcontrol/waitinglist/index.html
wiffbi: Changed wording (as well as ClassNames and URLs) from „update“ to „transfer“. Makes way more sense. Thanks for pointing that out!

PR #2813, opened 2022-09-22
Improve voucher redeem if no products are available (Z#23107372)

wiffbi:
In an event series when trying to redeem a voucher on a date where the associated productis not available, this would result in an empty product list and an inactive add-to-cart button. This PR adds a message to describe that no products are available at the moment.

Feel free to improve the wording!

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/voucher.html
+++ b/src/pretix/presale/templates/pretixpresale/event/voucher.html
@@ -74,7 +74,15 @@
 
     <p>
     {% if options == 0 %}
-        You entered a voucher code that does not match any products.
+        {% if request.event.has_subevents and not voucher.subevent %}
+            {% blocktrans trimmed %}
+                For the selected date, there are currently no products available that can be bought with this voucher. Please try a different date or a different voucher.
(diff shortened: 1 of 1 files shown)


src/pretix/presale/templates/pretixpresale/event/voucher.html
```diff
@@ -73,9 +73,13 @@ <h3>{{ subevent.name }}</h3>
     {% endif %}
 
     <p>
+    {% if options == 0 %}
+        You entered a voucher code that does not match any products.
```
raphaelm, line 77: maybe like this?
```suggestion
        {% if request.event.has_subevents and not voucher.subevent %}
            {% blocktrans trimmed %}
                For the selected date, there are currently no products available that can be bought with this voucher. Please try a different date or a different voucher.
            {% endblocktrans %}
        {% else %}
            {% blocktrans trimmed %}
                There are currently no products available that can be bought with this voucher.
            {% endblocktrans %}
        {% endif %}
```

wiffbi: Way better, thank you very much!

PR #2821, opened 2022-09-28
PPv2: Fix continue button behaviour

julianrother:
PPv2 disables the continue button and reenables it after loading the sdk if paypal is the only option (and thus is preselected) or when one of the payment options is selected. This is inconsistent with other payment methods: Usually the continue button is enabled even if no payment option is selected. This also breaks the checkout process in one case:

If 1) paypal is enabled as a payment method, 2) paypal is not available (because it is hidden, limited to certain countries, ...) and 3) there is exactly one payment method available, PPv2 disables the continue button with no way for the user to enable it:
127 0 0 1_8000_test_test2_checkout_payment_
The user cannot generate a `change` event, since the sole radio button is preselected and cannot be unselected.

This change fixes the inconsistent behaviour and the broken checkout process. It also prevents unnecessary loading of the sdk, if paypal is not avilable.

### DIFF
--- a/src/pretix/plugins/paypal2/static/pretixplugins/paypal2/pretix-paypal.js
+++ b/src/pretix/plugins/paypal2/static/pretixplugins/paypal2/pretix-paypal.js
@@ -321,6 +321,23 @@ var pretixpaypal = {
 };
 
 $(function () {
+    // This script is always loaded if paypal is enabled as a payment method, regardless of
+    // whether it is available (it could e.g. be hidden or limited to certain countries).
+    // We do not want to unnecessarily load the sdk.
+    let sdk_needed = false
+    // If we are on a payment method selection page (event.checkout "payment" step or
+    // event.order.pay.change) and paypal/paypal_apm is one of the options, we need the sdk.
+    if ($("input[name=payment][value^='paypal']").length) {
+        sdk_needed = true
+    }
+    // If we are on the (APM) PayView, we need the sdk.
+    if ($('#paypal-button-container').data('paypage')) {
+        sdk_needed = true
+    }
+    if (!sdk_needed) {
+        return
+    }
+
     pretixpaypal.load();
 
     (async() => {
--- a/src/pretix/plugins/paypal2/static/pretixplugins/paypal2/pretix-paypal.js
+++ b/src/pretix/plugins/paypal2/static/pretixplugins/paypal2/pretix-paypal.js
@@ -85,7 +85,7 @@ var pretixpaypal = {
             pretixpaypal.restore();
         });
 
-        // If paypal is the only payment option, it is preselected and we must disable the continue button until the sdk is ready
+        // If paypal is pre-selected, we must disable the continue button and handle it after SDK is loaded
         if ($("input[name=payment][value^='paypal']").is(':checked')) {
             pretixpaypal.continue_button.prop("disabled", true);
         }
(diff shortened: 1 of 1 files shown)


raphaelm:
Thanks! The patch might break things on the other pages the script is used, such as when retrying a payment for an existing order, so that would need some special testing.

julianrother:
After some more testing I noticed that this change causes problems if paypal is selected while loading the sdk (e.g. if the internet connection is very slow). That is probably, why the continue button was disabled before loading the sdk regardless of whether paypal is selected. I think that is easy enough to fix.

julianrother:
You were right, the inital change broke the APM page (PayView). I've now tested:
- Selecting the payment method during checkout
- Changing payment method after checkout
- APM page (after checkout)

I don't know how to test retrying paypal payments.

raphaelm:
@pc-coholic @julianrother Can one of you check if this change is still relevant or if it was fixed in the meantime by another change, such as e.g. 0b95f8988?

raphaelm:
Yes, I think we'd still be interested in these changes :)

julianrother:
The rebased commits essentially undo 0b95f89 (and 7d2dd72) and restore what my original patch did.

I think I should explain the continue button changes a bit more: To get consistent continue button behaviour, we cannot just always disable the continue button in `pretixpaypal.load`. Instead we only disable the button in `pretixpaypal.load` if the paypal/paypal_apm radio button is pre-selected. Otherwise we disable the continue button in the onchange handler if the SDK is not yet initialized. To do that we have to move the onchange handler registration from `pretixpaypal.ready` (after the SDK was initialized) to `pretixpaypal.load` (before the SDK is initialized).

I've manually tested the cases that I've listed earlier again. The CI failure seems to be unrelated to my changes.

raphaelm (commented):
The changes look sensible to me. As our resident PayPal expert @pc-coholic is on vacation, maybe @wiffbi can give another look as well?

src/pretix/plugins/paypal2/static/pretixplugins/paypal2/pretix-paypal.js
```diff
@@ -321,6 +322,23 @@ var pretixpaypal = {
...
+        sdk_needed = true
+    }
+    // If we are on the (APM) PayView, we need the sdk.
+    if ($('#paypal-button-container').data('paypage')) {
+        sdk_needed = true
+    }
+    if (!sdk_needed) {
+        return
+    }
+
```
wiffbi, line 341: What about shortening that bit to the following if statement, no need for a temp var

```suggestion
    // If no paypal/paypal_apm payment option is present and we are not on
    // the (APM) PayView, then we do not need the SDK.
    if (!$("input[name=payment][value^='paypal']").length && !$('#paypal-button-container').data('paypage')) {
        return
    }

```

src/pretix/plugins/paypal2/static/pretixplugins/paypal2/pretix-paypal.js
```diff
@@ -73,12 +73,21 @@ var pretixpaypal = {
...
+            } else {
                 pretixpaypal.continue_button.prop("disabled", true);
             }
+        });
+
+        $("input[name=payment]").not("[value^='paypal']").change(function () {
+            pretixpaypal.restore();
+        });
+
+        // If paypal is the only payment option, it is preselected and we must disable the continue button until the sdk is ready
```
wiffbi, line 88: I think this comment is midleading as the selector also matches if there are multiple payment options and paypal is preselected – not sure, when this could happen, but maybe in some case of navigating between checkout-steps. The code itself is fine though as we should always disable the continue-button if paypal is selected. Also the bit about „until the sdk is ready“ could be a bit irritating when reading the code, as it kind of implies that once the SDK is ready, the button should be enabled again – which is not the case as it is only enabled in certain cases in renderButton and renderAPMs. Maybe update the comment to something like:

```suggestion
        // If paypal is pre-selected, we must disable the continue button and handle it after SDK is loaded
```

wiffbi (commented):
Looks good to me, just a minor code/if optimization and a suggestion regarding a comment to better follow the code what is happening when and where.

raphaelm (approved):

PR #2822, opened 2022-09-29
Introduce a hidden setting to toggle between "common" and "official" country names

raphaelm:
django-countries has a list of "common" names for some countries that override the ISO list, e.g.

Russian Federation → Russia
Taiwan (Province of China) → Taiwan
Korea (Republic of) → South Korea

This is usually useful, but can be offensive in some situations, because "being a country" is somehow not binary.

pretix will continue to use the "common" names by default, but this introduces a hidden setting to revert to the official ones if needed.

It does not replace *all* occurrences, but most. Some would take much effort (e.g. description of attendee profiles in the customer account), and I don't think they're as important. I'm mostly concerned about all places where the field is explicitly named "Country". In a postal address, it could be anything. :wink:

### DIFF
--- a/src/pretix/base/forms/questions.py
+++ b/src/pretix/base/forms/questions.py
@@ -87,7 +87,7 @@ from pretix.control.forms import (
     ExtFileField, ExtValidationMixin, SizeValidationMixin, SplitDateTimeField,
 )
 from pretix.helpers.countries import (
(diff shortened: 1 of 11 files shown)


raphaelm:
We'll merge this if a customer ever really needs it too bad

PR #2824, opened 2022-09-29
PDF/Badges: Improve performance/reduce filesize when creating multiple badges

wiffbi:
Depending on the background-PDF, handling background-PDFs when creating badges could be slow due to PyPDF handling the merging and creating larger filesizes as necessary. This PR splits creating the badge PDF and imposing multiple badges on one page for printing into two separate steps. That way one can use the `render_background` method of the renderer, which by default supports merging backgrounds with pdftk, which is faster and sometimes creates way smaller files (in my tests 3.5MB instead of 13MB).

### DIFF
--- a/src/pretix/plugins/badges/exporters.py
+++ b/src/pretix/plugins/badges/exporters.py
@@ -32,7 +32,6 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations under the License.
 
-import copy
 import json
 from collections import OrderedDict
 from datetime import datetime, time, timedelta
@@ -166,8 +165,7 @@ OPTIONS = OrderedDict([
 
 
 def render_pdf(event, positions, opt):
-    from PyPDF2 import PdfMerger, PdfReader, PdfWriter, Transformation
-    from PyPDF2.generic import RectangleObject
+    from PyPDF2 import PdfMerger, PdfReader, PdfWriter
     Renderer._register_fonts()
 
     renderermap = {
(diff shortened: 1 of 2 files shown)


src/pretix/plugins/badges/exporters.py
```diff
@@ -167,7 +165,8 @@ def _renderer(event, layout):
 
 
 def render_pdf(event, positions, opt):
-    from PyPDF2 import PdfReader, PdfWriter
+    from PyPDF2 import PdfMerger, PdfReader, PdfWriter, Transformation
+    from PyPDF2.generic import RectangleObject
```
raphaelm, line 169: I think we can import this all at top level? Not sure why it was done like this before.

src/pretix/plugins/badges/exporters.py
```diff
@@ -178,71 +177,65 @@ def render_pdf(event, positions, opt):
...
-        if len(pagebuffer) == npp:
-            render_page(pagebuffer)
-            pagebuffer.clear()
+    merger.write(outbuffer)
+    outbuffer.seek(0)
 
-    if pagebuffer:
-        render_page(pagebuffer)
+    badges_per_page = opt['cols'] * opt['rows']
+    if (badges_per_page == 1):
```
raphaelm, line 208:
```suggestion
    if badges_per_page == 1:
```

src/pretix/plugins/badges/exporters.py
```diff
@@ -178,71 +177,65 @@ def render_pdf(event, positions, opt):
...
-        '/Creator': 'pretix',
-    })
-    output_pdf_writer.write(outbuffer)
+    # place n-up badges/pages per page
+    badges_pdf = PdfReader(outbuffer)
+    nup_pdf = PdfWriter()
+    nup_page = None
+    for i, page in enumerate(badges_pdf.pages):
+        di = i % badges_per_page
+        if (di == 0):
```
raphaelm, line 218:
```suggestion
        if di == 0:
```

src/pretix/plugins/badges/exporters.py
wiffbi: Yeah, me neither, but thought it was on purpose. I will change accordingly.

src/pretix/plugins/badges/exporters.py
wiffbi: thanks fixed.

raphaelm (approved):

PR #2844, opened 2022-10-07
Clarify cart order

raphaelm:
Cart order in pretix was confusing:

- In situations where we can group positions, we sorted by order of categories, items, …
- In situations where we can't group positions, we sorted by order of addition to the cart

With this PR, we'll always sort by order of categories/items/… to be consistent.

### old text of this PR

A customer complained that by going back and forth in the checkout process and changing the number of add-on products, you can create very weird groupings in the cart such as this:

image006

This PR makes the cart view completely ignore "chronological" order and strictly sort by the sorting attributes of the products and variations involved, with one exception: Bundled products are always sorted before add-on products.

The same sorting is applied when submitting the order and computing the ``positionid`` attributes to avoid bad grouping and sorting in the order later on.

There's one downside to all this, which is that the ``position`` attribute of ``ItemAddOn`` is now completely ignored in the cart, only the ``position`` attribute of the ``ItemCategory`` itself is respected. The order of ``ItemAddOn``s is only relevant for the display in the add-on step it self now. People will complain about this as well, but I think it's a fair trade-off.

### DIFF
--- a/src/pretix/base/models/orders.py
+++ b/src/pretix/base/models/orders.py
@@ -2671,7 +2671,7 @@ class CartPosition(AbstractPosition):
         if self.addon_to_id:
             return self.addon_to.sort_key + (1 if self.is_bundled else 2,) + sort_key
         else:
-            return sort_key + (0,) + sort_key
+            return sort_key
 
     def update_listed_price_and_voucher(self, voucher_only=False, max_discount=None):
         from pretix.base.services.pricing import (
--- a/src/pretix/base/models/orders.py
+++ b/src/pretix/base/models/orders.py
@@ -2237,7 +2237,7 @@ class OrderPosition(AbstractPosition):
 
     @cached_property
     def sort_key(self):
-        return self.addon_to.positionid if self.addon_to else self.positionid, self.addon_to_id or 0
+        return self.addon_to.positionid if self.addon_to else self.positionid, self.addon_to_id or 0, self.positionid
 
     @property
     def checkins(self):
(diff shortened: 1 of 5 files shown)


src/pretix/base/models/orders.py
```diff
@@ -2659,6 +2659,20 @@ def tax_value(self):
...
+        category_key = (self.item.category.position, self.item.category.id) if self.item.category_id is not None else (0, 0)
+        item_key = self.item.position, self.item_id
+        variation_key = (self.variation.position, self.variation.id) if self.variation_id is not None else (0, 0)
+        line_key = (self.price, (self.voucher_id or 0), (self.seat.sorting_rank if self.seat_id else None), self.pk)
+        sort_key = subevent_key + category_key + item_key + variation_key + line_key
+
+        if self.addon_to_id:
+            return self.addon_to.sort_key + (1 if self.is_bundled else 2,) + sort_key
+        else:
+            return sort_key + (0,) + sort_key
```
wiffbi, line 2674: Is that even needed or would just `sortkey` be enough in this case? Because the sort_keys are anyways not the same length as `self.addon_to.sort_key + (1 if self.is_bundled else 2,) + sort_key` produces a longer sort_key than `sort_key + (0,) + sort_key`.

wiffbi (approved):
Looks good, just one minor comment regarding sort_key and its length.

src/pretix/base/models/orders.py
raphaelm: Uh, I did not think correctly there, yes. I was intending for them to always be the same length, but in fact python deals fine with comparing different-length tuples.

PR #2845, opened 2022-10-07
Allow to charge a cancellation fee on unpaid orders

raphaelm:
This PR adds the option to charge a cancellation fee on unpaid orders. It also changes the restriction of "cancellation fee must not be higher than sum of payments minus refunds" to "cancellation fee must not be higher than order total"

It does *not* implement cancellation requests or adjustable cancellation fees for unpaid orders, as I don't think these are relevant use cases.

This PR also adds one intentional regression: It removes any option for self-service cancellation for *partially paid* orders, because with the new options it becomes absolutely confusing of which rules should apply and there would be a ton of really weird edge-case situations. This might be a problem for some customers who use upgrade functionality maybe? Not sure if we can afford the regression, but I really don't want to deal with all those edge cases.

- [x] Settings API
- [x] Tests


pc-coholic (changes requested):
LGTM, except I think there might be an issue with order expiration.

- After order cancellation, the order will still be in the `pending` state, allowing on the one hand customers to still pay the outstanding fee but on the other hand also allowing the `periodic_task` to expire the order.
- If the cancelled order is expired, the customer cannot pay for it anymore.

Not sure, if this is something we can or want to easily remedy - potentially if would be enough to extend the `expire_orders()` to also ignore orders that contain a `OrderFee.FEE_TYPE_CANCELLATION`

pc-coholic (approved):
LGTM

Two possibly improvements (but I think we can also leave it as is):
- Redirect the user directly to the payment selections screen after the order is cancelled
- Extend the fragment_order_status.html to display something like `pending (canceled)`

PR #2875, opened 2022-11-01
Fix paypal disabling continue button (Z#23110784)

wiffbi:
During checkout when navigating back and forth between steps, the payment options can initially be selected. When PayPal was included, it disabled the continue-button even if a payment option was initially selected. This PR fixes this. I tested in console as I do not have a working PayPal setup. It would be great if someone with a PayPal-setup could test this. Thanks!


wiffbi:
isort error is on master also. Fixed it on master in the meantime, but will not do a rebase here.

pc-coholic (approved):

PR #2897, opened 2022-11-11
Update vue to 2.7.14

wiffbi:
The updated seating plan uses vue now in its current version 2.7.14. To match the versions and potentially only serve one vue, this PR updates vue in pretix-core to v2.7.14.

This PR also closes two TODOs for using vue.min.js instead of vue.js in production – which disables the developer hints in vue (and might improve performance?). This is done by regrouping some of the script-embeds for django-compressor. This makes for more HTTP-calls, but improves cacheability because it groups js-files that belong together, e.g. lib-files as d3, which will rarely be updated as well as plugin/feature-specific files such as webcheckin and ui/checkinrules.

Last but not least, this PR fixes a null-value error in checkinrules lookup-select2. It is not related to the version-update as it already exists with vue2.6, but as it is a tiny fix and was discovered during testing for updating to 2.7. it was fixed along the way. Sorry, should have been its own PR.


wiffbi:
Tested with pretix-widget, pretix-button, checkin-rules ui and webcheckin. Should be good to go.

PR #2899, opened 2022-11-14
PPv2: Improve error handling

raphaelm:


pc-coholic (approved):
Tested with 
`capturereq.headers['PayPal-Mock-Response'] = 
 {'mock_application_codes':'INSTRUMENT_DECLINED'}"`

I think however, that we can replace the newly imported `HttpError` with the `IoError` as used everywhere else.

PR #2906, opened 2022-11-15
Order list export: Add event meta data (Z#2397902)

wiffbi:
This PR optionally adds the event’s meta data to the order export and the subevent’s meta data to the order-position export. Should order-fees export should include event’s meta as well?
Meta data columns labels are prefixed with „Meta: “.

### DIFF
--- a/src/pretix/base/exporters/orderlist.py
+++ b/src/pretix/base/exporters/orderlist.py
@@ -271,10 +271,6 @@ class OrderListExporter(MultiSheetListExporter):
 
         headers = [
             _('Event slug'),
-        ]
-        # get meta_data labels from first cached event
-        headers += [_('Meta: {}').format(l) for l in next(iter(self.event_object_cache.values())).meta_data.keys()]
-        headers += [
             _('Order code'), _('Order total'), _('Status'), _('Email'), _('Phone number'),
             _('Order date'), _('Order time'), _('Company'), _('Name'),
         ]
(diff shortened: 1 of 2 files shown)


wiffbi:
P.S.: Another checkbox …
Bildschirmfoto 2022-11-15 um 10 11 58

raphaelm:
Yes, I think it should be included in the fees table as well.

Why do you feel we need a check box for this and don't just activate it all the time?

wiffbi:
I will add it to the fees as well. Regarding checkbox, I thought if you have several meta-data it „pollutes“ the export. I am happy to remove the checkbox. :)

raphaelm:
I just now saw that you put them right after the event slug. Yeah, that's in fact gonna be annoying if there's lots of meta data exports. Can we just move them to the very last columns as they're a special-case thing? I think we already have them there for other exports as well. And I don't think we need the Meta: prefix for the column headers.

wiffbi:
I will move the columns to the end. In the OrderPositions-export I have access to the event’s meta data as well as the subevent’s (if any) meta data. Currently I only include the subevents meta data (that is why I moved it next to the subevent column). Depending on what one wants to achieve, it makes sense to include both. What do you think? Should I prefix the subevent’s mata data labels with „Subevent:“? Otherwise there will be two sets of columns with the same name.

raphaelm:
> I will move the columns to the end. In the OrderPositions-export I have access to the event’s meta data as well as the subevent’s (if any) meta data. Currently I only include the subevents meta data (that is why I moved it next to the subevent column). Depending on what one wants to achieve, it makes sense to include both. What do you think? Should I prefix the subevent’s mata data labels with „Subevent:“? Otherwise there will be two sets of columns with the same name.

It should be just one set of columns. Subevent meta-data inherits meta data, so it should be the subevent meta data if the subevent has a property set, or fall back to the event meta-data if there is no subevent or the subevent has an empty property (``SubEvent.meta_data`` already handles that fallback``)

wiffbi:
I moved the meta data fields to the end and used subevent and event meta data as discussed (subevent meta data in order positions export, orders and fees export with event meta data). Should be ready to merge.

PR #2907, opened 2022-11-15
WIP: add answer options through client-side csv-file processing

wiffbi:
This PR adds the option to „upload“ a CSV-file to add answer options when creating/editing a question of type „Choose one/multiple from a list“. This is done by „parsing“ a local file-input. Parsing is currently very limited, it is actually just a regex split on `[;,\t]` – so currently not more than a hack and just a proof of concept to do this client-side.

Should I investige further to e.g. improve CSV-parsing – or include something like https://www.papaparse.com? What about fails matching CSV-columns with number of language inputs? There currently is no way to mix&match csv-columns to inputs, etc. So overall, this feels very very basic and – again – very hacky.

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/items/question_edit.html
+++ b/src/pretix/control/templates/pretixcontrol/items/question_edit.html
@@ -118,6 +118,10 @@
                             <button type="button" class="btn btn-default" data-formset-add>
                                 <i class="fa fa-plus"></i> {% trans "Add a new option" %}</button>
                         </p>
+                        <p>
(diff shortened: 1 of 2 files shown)


raphaelm:
Hmm, I didn't even properly think about the parsing issue and that we need multiple languages. It makes me question whether this is worth the complexity it brings.

wiffbi:
I don’t know, I tend towards it not being worth it.

raphaelm:
Okay, let's hold off for now

PR #2912, opened 2022-11-17
Support for external gift cards

raphaelm:
So far, gift cards have been a special-case payment provider with some extra logic in the core system. This has one major disadvantage: We cannot build similar payment providers where the source of truth lies in an external system.

This PR now generalizes this by introducing two new flags for payment providers:

- `multi_use_supported = True` tells the system that the payment provider can be instantiated multiple times in the same session. This requires the payment provider to play by some extra rules. 
- `execute_payment_needs_user = False` tells the system that execute_payment requires neither the users session nor any redirect, but instead can be processed within e.g. a celery task

This PR then drops the ``cart_session["payment"]`` key entirely and replaces it with a new ``cart_session["payments"]`` list which can contain multiple payments, each with a minimum or maximum value. Only one entry might have ``multi_use_supported=False`` for it to make sense. On order submission, multiple OrderPayment objects are created and executed in order of selection.

Generally, I think this also is a nice cleanup of the code base since it removes special cases, yay!

There is *one* regression caused by the generalization: If you open two different browsers, add a product in each, then add the same gift card in each, you will now be able to create two orders – the second one will just have a failed payment. Previously, the second order would have been blocked entirely. I'm not sure if that's better or worse? But I think it's okay?

### my todo list

- [x] Payment provider API
- [x] Rewrite checkout step handling
- [x] Give users option to discard payments
- [x] Show balance on order confirmation page
- [x] Actually perform charges
- [x] Remove all references to cart_session[payment]
- [x] Check all get_fees calls
- [x] Migrate our gift card provider
- [x] Test existing orders
- [x] Test edge cases with minimum amounts
- [x] Test edge cases with implicit providers
- [x] Test edge cases with payment fees
- [x] Test edge cases with service fees
- [x] Unit tests
- [x] Plugins needing adjustment
  - [x] resellers
  - [x] servicefees → https://github.com/pretix/pretix-servicefees/pull/26


pc-coholic (commented):
- `src/pretix/control/forms/item.py` seems to be unrelated to this PR - but looks good, too 
- `Your payment method has been applied, but xx,xx € still need to be paid. Please select a payment method for the remainder.` message looks good to me, but I expected the outstanding balance to be displayed closer to the actual payment... So perhaps adding a line "remaining balance" underneath the list of already added payments/giftcards?
- Might be related to `fake_giftcard` and not core: Paying the order in full, refunding 10 Euro and then only paying 5 through the giftcard will still deduct 10 Euro from the giftcard.

raphaelm:
> src/pretix/control/forms/item.py seems to be unrelated to this PR - but looks good, too

commit was on the wrong branch, fixed by rebasing this PR

> Your payment method has been applied, but xx,xx € still need to be paid. Please select a payment method for the remainder. message looks good to me, but I expected the outstanding balance to be displayed closer to the actual payment... So perhaps adding a line "remaining balance" underneath the list of already added payments/giftcards?

Fixed:

Screenshot 2022-11-23 at 13-39-08 Schritt 3 von 4 Zahlung – Bestellung fortsetzen Vouchermesse

> Might be related to fake_giftcard and not core: Paying the order in full, refunding 10 Euro and then only paying 5 through the giftcard will still deduct 10 Euro from the giftcard.

That was just not implemented in fake_giftcard. Fixed now!

PR #2927, opened 2022-11-23
Bump version of the Stripe API

yvovandoorn:
With https://github.com/pretix/pretix/commit/8c8e8031fc39eff6dfcb201242ca3c65d7ee1bd1, which (`stripe v5`) has an expectation to use `2022-11-15`, the version Stripe plugin uses to communicate should be updated as well.

### DIFF
--- a/src/pretix/plugins/stripe/payment.py
+++ b/src/pretix/plugins/stripe/payment.py
@@ -429,7 +429,7 @@ class StripeMethod(BasePaymentProvider):
         return kwargs
 
     def _init_api(self):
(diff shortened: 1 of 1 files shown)


raphaelm:
> With https://github.com/pretix/pretix/commit/8c8e8031fc39eff6dfcb201242ca3c65d7ee1bd1, which (stripe v5) has an expectation to use 2022-11-15, the version Stripe plugin uses to communicate should be updated as well.

Is that so? I don't read the v5.0.0 release logs this way

We'd first need to verify that none of the Stripe API chagnes affect us.

PR #2931, opened 2022-11-28
Make str.format_map with untrusted input safer

raphaelm:
We currently use ``str.format_map`` for email string interpolation, but we don't want users to use Python's full micro-language that allows nasty shenanigans like ``{foo:_<10}``. We already prevent users from doing this with form validation, but there are still some gaps, e.g. during preview formatting, that occasionally lead to accidental exceptions. This replaces this with a subclassed Formatter that should properly protect against this.

### DIFF
--- a/src/pretix/base/services/mail.py
+++ b/src/pretix/base/services/mail.py
@@ -111,7 +111,7 @@ def mail(email: Union[str, Sequence[str]], subject: str, template: Union[str, La
 
     :param template: The filename of a template to be used. It will be rendered with the locale given in the locale
         argument and the context given in the next argument. Alternatively, you can pass a LazyI18nString and
-        ``context`` will be used as the argument to a  Python ``.format_map()`` call on the template.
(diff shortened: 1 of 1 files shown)


src/pretix/helpers/format.py
```diff
@@ -0,0 +1,34 @@
...
+            return ('{' + str(field_name) + '}', field_name)
+        return super().get_field(field_name, args, kwargs)
+
+    def get_value(self, key, args, kwargs):
+        if key not in self.context:
+            return '{' + str(key) + '}'
+        return self.context[key]
+
+    def format_field(self, value, format_spec):
+        # Ignore format _spec
```
wiffbi, line 27: Should we just generally ignore the format_spec or ignore the whole field when it has a format_spec (if so, we probably need to overwrite `parse`)? I think it is okay to silently ignore `format_spec` as we do not advertise it and only want basic placeholder-replacement anyways.

wiffbi (approved):
Looks good. As in the comment, it is okay for me to silently ignore format_spec.

PR #2934, opened 2022-11-30
Add lifetime spending to customer details

ser8phin:
Add an entry for lifetime spending in a customer account's details section, so organizers can easily see how much a customer has spent on all of their events. It gets split for currency, as money exchange courses are not set in stone.

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/organizers/customer.html
+++ b/src/pretix/control/templates/pretixcontrol/organizers/customer.html
@@ -25,7 +25,6 @@
                     <form action="" method="post" class="row">
                         {% csrf_token %}
                         <dl class="dl-horizontal col-lg-6 col-sm-12">
-                            <div>
                                 <dt>{% trans "Customer ID" %}</dt>
                                 <dd>#{{ customer.identifier }}</dd>
                                 {% if customer.provider %}
@@ -72,24 +71,21 @@
                                 <dt>{% trans "Notes" %}</dt>
                                 <dd>{{ customer.notes|linebreaks }}</dd>
                             {% endif %}
-                            </div>
                         </dl>
+                        {% if lifetime_spending %}
                         <dl class="col-lg-6 col-sm-12 text-right">
-                            <div>
-                            {% if lifetime_spending %}
-                                <dt class="text-muted"
-                                    data-toggle="tooltip"
-                                    title="{% trans "This includes all paid orders by this customer across all your events." %}">
-                                    {% trans "Lifetime spending" %}
-                                </dt>
-                                <dd class="text-success helper-font-size-24">
-                                    {% for s in lifetime_spending %}
-                                            {{ s.spending|money:s.currency }} <br>
-                                    {% endfor %}
-                                </dd>
-                            {% endif %}
-                            </div>
+                            <dt class="text-muted"
+                                data-toggle="tooltip"
+                                title="{% trans "This includes all paid orders by this customer across all your events." %}">
+                                {% trans "Lifetime spending" %}
+                            </dt>
+                            <dd class="text-success helper-font-size-24">
+                                {% for s in lifetime_spending %}
+                                        {{ s.spending|money:s.currency }} <br>
+                                {% endfor %}
+                            </dd>
                         </dl>
+                        {% endif %}
                     </form>
                     <div class="text-right">
                         <a href="{% url "control:organizer.customer.edit" organizer=request.organizer.slug customer=customer.identifier %}"
--- a/src/pretix/control/templates/pretixcontrol/organizers/customer.html
+++ b/src/pretix/control/templates/pretixcontrol/organizers/customer.html
@@ -22,10 +22,10 @@
                     </h3>
                 </div>
                 <div class="panel-body">
-                <div class="row">
-                    <form action="" method="post" class="col-lg-8 col-md-6 col-sm-12">
+                    <form action="" method="post" class="row">
                         {% csrf_token %}
                         <dl class="dl-horizontal">
+                        <div class="col-lg-6 col-sm-12">
                             <dt>{% trans "Customer ID" %}</dt>
                             <dd>#{{ customer.identifier }}</dd>
                             {% if customer.provider %}
@@ -72,22 +72,23 @@
                             <dt>{% trans "Notes" %}</dt>
                             <dd>{{ customer.notes|linebreaks }}</dd>
                         {% endif %}
+                        </div>
+                        <div class="col-lg-6 col-sm-12 text-right">
+                        {% if lifetime_spending %}
+                            <dt class="text-muted text-right"
+                                data-toggle="tooltip"
+                                title="{% trans "This includes all paid orders by this customer across all your events." %}">
+                                {% trans "Lifetime spending" %}
+                            </dt>
+                            <dd class="text-success h3">
+                                {% for s in lifetime_spending %}
+                                        {{ s.spending|money:s.currency }}
+                                {% endfor %}
+                            </dd>
+                        {% endif %}
+                        </div>
                         </dl>
                     </form>
-                    {% if lifetime_spending %}
-                        <div class="text-right col-lg-4 col-md-6 col-sm-12"
-                                data-toggle="tooltip"
-                                title="{% trans "This includes all paid orders by this customer across all your events." %}"
-                        >
-                            <p class="text-muted">{% trans "Lifetime spending" %}</p>
-                                <h3 class="text-success">
-                                    {% for s in lifetime_spending %}
-                                            {{ s.spending|money:s.currency }}
-                                    {% endfor %}
-                                </h3>
-                        </div>
-                    {% endif %}
-                </div>
                     <div class="text-right">
                         <a href="{% url "control:organizer.customer.edit" organizer=request.organizer.slug customer=customer.identifier %}"
                                 class="btn btn-default">
--- a/src/pretix/control/templates/pretixcontrol/organizers/customer.html
+++ b/src/pretix/control/templates/pretixcontrol/organizers/customer.html
@@ -72,20 +72,22 @@
                                 <dd>{{ customer.notes|linebreaks }}</dd>
                             {% endif %}
                         </dl>
-                        {% if lifetime_spending %}
                         <dl class="col-lg-6 col-sm-12 text-right">
                             <dt class="text-muted"
                                 data-toggle="tooltip"
                                 title="{% trans "This includes all paid orders by this customer across all your events." %}">
                                 {% trans "Lifetime spending" %}
                             </dt>
+                        {% if lifetime_spending %}
                             <dd class="text-success helper-font-size-24">
                                 {% for s in lifetime_spending %}
                                         {{ s.spending|money:s.currency }} <br>
                                 {% endfor %}
                             </dd>
-                        </dl>
+                        {% else %}
+                            <dd class="helper-font-size-24">–</dd>
                         {% endif %}
+                        </dl>
                     </form>
                     <div class="text-right">
                         <a href="{% url "control:organizer.customer.edit" organizer=request.organizer.slug customer=customer.identifier %}"
--- a/src/pretix/control/templates/pretixcontrol/organizers/customer.html
+++ b/src/pretix/control/templates/pretixcontrol/organizers/customer.html
@@ -71,6 +71,10 @@
                             <dt>{% trans "Notes" %}</dt>
                             <dd>{{ customer.notes|linebreaks }}</dd>
                         {% endif %}
+                        <dt>{% trans "Lifetime spending" %}</dt>
+                        {% for currency, total in lifetime_spending.items %}
+                            <dd>{{ total|money:currency }}</dd>
+                        {% endfor %}
                         </dl>
                     </form>
                     <div class="text-right">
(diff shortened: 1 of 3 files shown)


src/pretix/control/views/organizer.py
```diff
@@ -2310,6 +2311,15 @@ def get_context_data(self, **kwargs):
             o.computed_payment_refund_sum = annotated.get(o.pk)['computed_payment_refund_sum']
             o.icnt = annotated.get(o.pk)['icnt']
             o.sales_channel_obj = scs[o.sales_channel]
+            currency = o.event.currency
+            if currency in spending:
+                total = spending[currency]
+                total += o.total
+                spending[currency] = total
```
wiffbi, line 2318: Könnte man kürzen
```suggestion
                spending[currency] += o.total
```

ser8phin: good improvement, though it's completely reworked now :smile:

src/pretix/control/views/organizer.py
```diff
@@ -2311,6 +2311,14 @@ def get_context_data(self, **kwargs):
             o.icnt = annotated.get(o.pk)['icnt']
             o.sales_channel_obj = scs[o.sales_channel]
 
+        ctx["lifetime_spending"] = (
+            self.get_queryset()
+            .filter(status__in=[Order.STATUS_PAID, Order.STATUS_PENDING])
```
ser8phin, line 2316: This could either include pending orders or not, depending on expected behavior. Would someone want to see spending that is yet to come, but will probably soon be paid, or would it be better to show only the numbers for confirmed paid orders?

src/pretix/control/templates/pretixcontrol/organizers/customer.html
```diff
@@ -71,6 +71,10 @@ <h3 class="panel-title">
                             <dt>{% trans "Notes" %}</dt>
                             <dd>{{ customer.notes|linebreaks }}</dd>
                         {% endif %}
+                        <dt>{% trans "Lifetime spending" %}</dt>
+                        {% for s in lifetime_spending %}
+                            <dd>{{ s.spending|money:s.currency }}</dd>
+                        {% endfor %}
```
raphaelm, line 77: Currently this looks a little weird if the customer has not bought anything. I think we should either drop the line entirely in this case, or show a zero (without currency? a little weird,  but how would we determine a currency?)

src/pretix/control/views/organizer.py
raphaelm: There's really no good way to determine this. Some of our customers would expect one, some the other, but I really don't want a switch and my vote goes for "just the paid ones".

raphaelm: We could add a ``data-toggle="tooltip" title="{% trans "This includes all paid orders by this customer across all your events." %}"`` to at least make it understandable

raphaelm (commented):
This works and is correctly implemented. I added some very minor notes below.

I also way thinking, do we maybe want to make this more prominent than just a row in that table? Showing the lifetime spending of a customer is something that I feel can be "celebrated" because it will make people happy opening customer accounts and seeing large numbers there. Here's a very quick mockup:



What do you (and others, @pc-coholic @wiffbi) think?

wiffbi:
> This works and is correctly implemented. I added some very minor notes below.
> 
> I also way thinking, do we maybe want to make this more prominent than just a row in that table? Showing the lifetime spending of a customer is something that I feel can be "celebrated" because it will make people happy opening customer accounts and seeing large numbers there. Here's a very quick mockup:
> 
> 
> 
> What do you (and others, @pc-coholic @wiffbi) think?

I think this is a great idea!

src/pretix/control/templates/pretixcontrol/organizers/customer.html
```diff
@@ -71,12 +72,22 @@ <h3 class="panel-title">
...
+                        <div class="text-right col-lg-4 col-md-6 col-sm-12"
+                                data-toggle="tooltip"
+                                title="{% trans "This includes all paid orders by this customer across all your events." %}"
+                        >
+                            <p class="text-muted">{% trans "Lifetime spending" %}</p>
+                                <h3 class="text-success">
+                                    {% for s in lifetime_spending %}
+                                            {{ s.spending|money:s.currency }}
+                                    {% endfor %}
+                                </h3>
```
wiffbi, line 87: `<h3>` has semantic meaning which is not fitting here, for a bigger font-size better use bootstraps `.h3` css-class. Probably semantically the most correct way – and aligned with the other infos marked up as `<dl><dt>` – would be to use a `<dl>`. So:

```suggestion
                            <dl class="horizontal">
                                <dt class="text-muted">{% trans "Lifetime spending" %}</dt>
                                <dd class="text-success h3">
                                    {% for s in lifetime_spending %}
                                            {{ s.spending|money:s.currency }}
                                    {% endfor %}
                                </dd>
                            </dl>
```

You could even go further (better!) and put everything inside the already existing `<dl>` and use `<dl><div class="col …"><dt>…</dt><dd>…</dd></div><div class="col …"><dt>…</dt><dd>…</dd></div></dl>` to group them into columns – `<div>`s are allowed in `<dl>`. That way the `<form>` can hold the obligatory `.row` css class for columns to correctly work (contain the float based columns).

src/pretix/control/templates/pretixcontrol/organizers/customer.html
```diff
@@ -22,7 +22,8 @@ <h3 class="panel-title">
                     </h3>
                 </div>
                 <div class="panel-body">
-                    <form action="" method="post">
+                <div class="row">
+                    <form action="" method="post" class="col-lg-8 col-md-6 col-sm-12">
```
wiffbi, line 26: See comment below to set `row` class on form for columns to work properly

src/pretix/control/templates/pretixcontrol/organizers/customer.html
ser8phin: I tried the version with `<div>`s, but in the end, I went with a separate `<dl>` with no `dl-horizontal` class to ensure no weird formatting occured.

src/pretix/control/templates/pretixcontrol/organizers/customer.html
ser8phin: I tried different ways of handling it, but in the end I decided on using a dash, the same as is already in use for "last login" in the same area

src/pretix/static/pretixcontrol/scss/main.scss
```diff
@@ -160,6 +160,9 @@ p.bigger {
 .helper-space-below {
     margin-bottom: 10px;
 }
+.helper-font-size-24 {
+    font-size: 24px;
+}
```
wiffbi, line 166: I don’t think using the absolut size in the utility classes name is a good idea. Boostrap3 already offers font-sizing classes .h1 to .h6 – problem is, that they also manipulate the margin. What about adding utility classes that are use the same naming scheme as already available? Like so:

```suggestion
.text-h1 { font-size: $font-size-h1; }
.text-h2 { font-size: $font-size-h2; }
.text-h3 { font-size: $font-size-h3; }
.text-h4 { font-size: $font-size-h4; }
.text-h5 { font-size: $font-size-h5; }
.text-h6 { font-size: $font-size-h6; }

```

In your case, you would need `.text-h3` which produces 24px font-size.

src/pretix/control/templates/pretixcontrol/organizers/customer.html
```diff
@@ -22,54 +22,70 @@ <h3 class="panel-title">
                     </h3>
                 </div>
                 <div class="panel-body">
-                    <form action="" method="post">
+                    <form action="" method="post" class="row">
```
wiffbi, line 25: I know we talked about adding the row-class to the form, but later we talked about leaving the form alone and using one `<dl>` element to contain all the info and use two `<div>` to create the two columns – `<div>` is allowed inside `<dl>` to group pairs of `<dt><dd>`. So:
```suggestion
                    <form action="" method="post">
```

src/pretix/control/templates/pretixcontrol/organizers/customer.html
```diff
@@ -22,54 +22,70 @@ <h3 class="panel-title">
...
-                            {% endif %}
-                            <dt>{% trans "Status" %}</dt>
-                            <dd>
-                                {% if not customer.is_active %}
-                                    {% trans "disabled" %}
-                                {% elif not customer.is_verified %}
-                                    {% trans "not yet activated" %}
-                                {% else %}
-                                    {% trans "active" %}
+                        <dl class="dl-horizontal col-lg-6 col-sm-12">
```
wiffbi, line 27: See comment above for one big dl, which is semantically better and better for screenreaders too.
```suggestion
                        <dl class="dl-horizontal row">
                            <div class="col-lg-6 col-sm-12">
```

src/pretix/control/templates/pretixcontrol/organizers/customer.html
```diff
@@ -22,54 +22,70 @@ <h3 class="panel-title">
...
+                                <dd>{{ customer.date_joined|date:"SHORT_DATETIME_FORMAT" }}</dd>
+                                <dt>{% trans "Last login" %}</dt>
+                                <dd>{% if customer.last_login %}{{ customer.last_login|date:"SHORT_DATETIME_FORMAT" }}{% else %}
+                                    –{% endif %}</dd>
+                            {% if customer.notes %}
+                                <dt>{% trans "Notes" %}</dt>
+                                <dd>{{ customer.notes|linebreaks }}</dd>
+                            {% endif %}
+                        </dl>
+                        <dl class="col-lg-6 col-sm-12 text-right">
```
wiffbi, line 75: See comment regarding one dl
```suggestion
                        </div>
                        <div class="col-lg-6 col-sm-12 text-right">
```

src/pretix/control/templates/pretixcontrol/organizers/customer.html
```diff
@@ -22,54 +22,70 @@ <h3 class="panel-title">
...
+                            <dt class="text-muted"
+                                data-toggle="tooltip"
+                                title="{% trans "This includes all paid orders by this customer across all your events." %}">
+                                {% trans "Lifetime spending" %}
+                            </dt>
+                        {% if lifetime_spending %}
+                            <dd class="text-success helper-font-size-24">
```
wiffbi, line 82: See comment below regarding naming the CSS utility class.

By the way, can we be sure, totals will always be positive? Probably in 99,9999% of the time … otherwise we should use `.text-error` for negative s.spending.

```suggestion
                            <dd class="text-success text-h3">
```

src/pretix/control/templates/pretixcontrol/organizers/customer.html
```diff
@@ -22,54 +22,70 @@ <h3 class="panel-title">
...
-                            <dt>{% trans "Registration date" %}</dt>
-                            <dd>{{ customer.date_joined|date:"SHORT_DATETIME_FORMAT" }}</dd>
-                            <dt>{% trans "Last login" %}</dt>
-                            <dd>{% if customer.last_login %}{{ customer.last_login|date:"SHORT_DATETIME_FORMAT" }}{% else %}
-                                –{% endif %}</dd>
-                        {% if customer.notes %}
-                            <dt>{% trans "Notes" %}</dt>
-                            <dd>{{ customer.notes|linebreaks }}</dd>
+                        {% else %}
+                            <dd class="helper-font-size-24">–</dd>
```
wiffbi, line 88: Hm, not sure this looks good having a big, black dash there. MAybe make it `.text-success` as well. Or just statically 0,00 without any currency? I do not know.

wiffbi (commented):
Looks good, just an open question regarding naming CSS utility classes.

src/pretix/control/templates/pretixcontrol/organizers/customer.html
wiffbi: Ah, just saw this now after I submitted my review. I guess regarding the black dash it is not easy. As I suggested in my review, I like the idea of 0,00 in green `.text-success` without any currency.

src/pretix/control/templates/pretixcontrol/organizers/customer.html
wiffbi: Ah, sorry, my bad. I did not see your comment regarding the `<dl class="dl-horizontal">`. Sorry, ignore this comment.

src/pretix/static/pretixcontrol/scss/main.scss
ser8phin: This is absolutely better, good suggestion :+1:

src/pretix/control/templates/pretixcontrol/organizers/customer.html
ser8phin: Good suggestions! This has been cause for a lot of thoughts… I will play around with it some more.

src/pretix/control/views/organizer.py
raphaelm, line 2316: The tooltip above says paid, but this still includes unpaid. I think we settled on just the paid ones :)

raphaelm (changes requested):

src/pretix/control/views/organizer.py
ser8phin: Did this somehow get lost in all my commits? Well, good catch!

src/pretix/control/templates/pretixcontrol/organizers/customer.html
ser8phin: It's now grey and 0,00 :+1:

raphaelm (approved):

PR #2936, opened 2022-12-01
Fix regression in handling gift card payments

raphaelm:
In #2912, I introduced two regressions I did not notice at the time:

- Previously, an order that is fully paid with gift cards has received the "free order" email text. Now it receives both the "order placed" and "order paid" emails.
- Previously, gift cards have been handled before the invoice was created, so the gift card was shown on the invoice even if invoices are created for all new orders immediately.

This PR solves those two regressions. Unfortunately, this means that the native gift card provider again gains a "special" place that cannot be fully emulated by a plugin. However, I think at this point in time this is warranted by the risks of a more flexible solution (see comment inline). We can always still "open up" more.


ser8phin:
lgtm :smiley:

PR #2944, opened 2022-12-02
Checkout: copy previous answers instead of first

wiffbi:
Originally when copying answer the first cartitem/set of answers was used. This PR changes behaviour so the answers from the previous cartitem are used instead of the first cartitem. This also changes some `const/let` to `var` for better backwards compatability. Furthermore file-inputs cannot be copied so are skipped.


ser8phin:
lgtm :+1:

PR #2965, opened 2022-12-09
Allow variations to override item meta data

raphaelm:


src/pretix/api/serializers/item.py
```diff
@@ -47,43 +47,102 @@
...
-                  'sales_channels', 'hide_without_voucher',)
+                  'sales_channels', 'hide_without_voucher', 'meta_data')
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields['require_membership_types'].queryset = lazy(lambda: self.context['event'].organizer.membership_types.all(), QuerySet)
 
+    def validate_meta_data(self, value):
+        for key in value['meta_data'].keys():
+            if key not in self.parent.parent.item_meta_properties:
```
wiffbi, line 73: I stumbled upon this `parent.parent` relationship. Is this the access to the parent `ListSerializer`? So this would access InlineItemVariationSerializer=>ItemVariationSerializer=>EventSerializer? I could not find much info in the [django rest docs](https://www.django-rest-framework.org/api-guide/serializers/#inspecting-a-modelserializer). Isn’t the self.parent an `ItemVariationSerializer` which itself got a `@cached_property item_meta_properties` introduced in this PR?

src/pretix/api/serializers/order.py
```diff
@@ -371,10 +371,19 @@ def to_representation(self, instance: OrderPosition):
...
-            if not hasattr(instance.item, '_cached_meta_data'):
-                instance.item._cached_meta_data = instance.item.meta_data
-            for k, v in instance.item._cached_meta_data.items():
-                res['itemmeta:' + k] = v
+            if instance.variation_id:
+                print(instance, instance.variation, instance.variation_id, instance.item)
+                if not hasattr(instance.variation, '_cached_meta_data'):
+                    instance.variation.item = instance.item  # saves some database lookups
+                    instance.variation._cached_meta_data = instance.variation.meta_data
+                print(instance.variation._cached_meta_data.items())
```
wiffbi, line 379:
```suggestion
```

src/pretix/api/serializers/order.py
```diff
@@ -371,10 +371,19 @@ def to_representation(self, instance: OrderPosition):
             for k, v in ev._cached_meta_data.items():
                 res['meta:' + k] = v
 
-            if not hasattr(instance.item, '_cached_meta_data'):
-                instance.item._cached_meta_data = instance.item.meta_data
-            for k, v in instance.item._cached_meta_data.items():
-                res['itemmeta:' + k] = v
+            if instance.variation_id:
+                print(instance, instance.variation, instance.variation_id, instance.item)
```
wiffbi, line 375:
```suggestion
```

src/pretix/base/models/event.py
```diff
@@ -804,12 +804,18 @@ def copy_data_from(self, other):
...
+        for imv in ItemMetaValue.objects.filter(item__event=other).prefetch_related('property'):
             imv.pk = None
-            imv.property = item_meta_properties_map[imv.property.pk]
+            imv.property = item_meta_properties_map[imv.property_id]
             imv.item = item_map[imv.item.pk]
             imv.save(force_insert=True)
 
+        for imv in ItemVariationMetaValue.objects.filter(variation__item__event=other).prefetch_related('property'):
+            imv.pk = None
+            imv.property = item_meta_properties_map[imv.property.pk]
```
wiffbi, line 815: Sometimes it is `imv.property_id`, sometimes `imv.property.pk` – is there a difference? `property_id` reads more like knowing the internals whereas `property.pk` requires an object to be instantiated?

wiffbi (commented):
Looks good, only minor comments/questions. Tried it locally and works as expected.

src/pretix/base/models/event.py
raphaelm: Yes, there is a difference. property_id does not require the property object to be loaded from the database. We can save a SQL query here by using property_id and removing the prefetch related :+1:

src/pretix/api/serializers/item.py
wiffbi: Ah, thank you!

PR #2980, opened 2022-12-15
Editor: fix sample text when key missing

wiffbi:
There seems to be a way to create a PDF-Layout with an element having no content-attribute. This PR currently changes the editor to not fail, but ignore such elements. This creates an element, that once selected in the editor has no option in the „Content“-dropdown selected. And once one selects e.g. „Other“ the text (Sample text?) gets replaced with an empty string. Do we want this or should we instead do `if (!d.content) d.content = "other";`.

### DIFF
--- a/src/pretix/static/pretixcontrol/js/ui/editor.js
+++ b/src/pretix/static/pretixcontrol/js/ui/editor.js
@@ -256,7 +256,7 @@ var editor = {
             } else if (d.content === "other_i18n") {
                 o.text_i18n = d.text_i18n
                 o.setText(d.text_i18n[Object.keys(d.text_i18n)[0]]);
-            } else {
+            } else if (d.content) {
                 o.setText(editor._get_text_sample(d.content));
             }
             if (d.locale) {


ser8phin:
Is there a sample text set for sure, if nothing is selected in the drop down? If so, that actually seems best for unset content attributes, as a sample text tells a user implicitly, there is need for some further action. Otherwise, I believe setting "other" would create a better user experience.

wiffbi:
One significant difference happens when creating an actual PDF. If we keep `d.content` as `null` and just ignore it, when previewing the layout, the string „(error)“ gets rendered. If `d.content` is set to `"other"` the existing text-string will be rendered as-is. Not sure, what to prefer, but I trend to the current way where „(error)“ gets rendered – in the actual bug-report the placeholder was „Sample event name“ and I guess it is easier to assume that the preview only shows a placeholder that will be replaced in real life – which is not the case, as its a literal string.

raphaelm (approved):

PR #2989, opened 2022-12-20
Bump Django to 4.1.*

raphaelm:
- [x] wait for https://github.com/zostera/django-bootstrap3/pull/859 or patch it
- [x] Deal with timezones
  - [x] Use deprecation shim
  - [x] Remove all usages of is_dst
  - [x] Remove all usages of localize
  - [x] Remove all usages of normalize
  - [x] Remove all usages of pytz.timezone
  - [x] Remove all usages of make_aware
  - [x] Check for use of get_current_timezone in plugins
- [x] Deprecation warning for 4.1 for test class setup
- [x] Deal with DeleteView change
  - [x] also check plugins


src/pretix/base/management/commands/export.py
```diff
@@ -60,7 +60,7 @@ def handle(self, *args, **options):
             sys.exit(1)
 
         locale = options.get("locale", None)
-        timezone = pytz.timezone(options['timezone']) if options.get('timezone') else None
+        timezone = pytz_deprecation_shim.timezone(options['timezone']) if options.get('timezone') else None
```
wiffbi, line 63: Is this deprectation_shim here because there might be plugins with exports that expect pytz-timezones?

wiffbi (commented):
I am not yet fully through my review, but have a question: the use of `pytz_deprecation_shim` is due to public facing properties on event and/or organizer to make sure externals/plugins get a warning when using e.g. localize? Should we add a test to make sure, that event.timezone and organizer.timezone produce these warnings?

PR #2990, opened 2022-12-21
Separate personalization from admission

raphaelm:
If there's one thing that caused confusion with customers for me *most often* in the last months, it's how "admission products" and "ticket personalization" work together. Countless times, I recommended customers to *wrongfully* mark their tickets "non-admission" just to get rid of the attendee name field without disabling it for all products.

With this PR, I propose splitting these two things and making ``personalized`` it's own property:

Screenshot 2022-12-21 at 10-21-31 Product pretix eu

As this PR is implemented right now, there are only three combinations allowed:
- not admission and not personalized
- admission and personalized
- admission and not personalized

The combination of "not admission and personalized" is currently forbidden. This is because I first allowed it and found the UX when creating new products to be horrible, since it's really hard to communicate to the user what the *usually recommended* option for personalization is if that depends on the admission option. I do see use cases for "not admission and personalized" (mostly when selling memberships), but I don't think they warrant horrible UX.

After merge:
- [ ] Adjust pretixPOS
- [ ] Adjust plugins (altares, corusoft, ticketswap)


wiffbi (commented):
Looks good. I only pushed some minot text-updates on help-texts regarding admission vs. personlized products.

Maybe we need a warning/hint on products’ settings page, when no attendee data is asked, but the product is personalized? I know, this is described in the help text with „according to your event settings.“ but might not be understood so easily? Or at least link to the corresponding setting from the help-text?

Also could you elaborate regarding the „horrible“ UX when creating personalized, non-admission products? The only thing I stumbled upon was the wording in event settings regarding „attendee data“, which of course for non-admission products are not „attendees“ anymore. But maybe a more general wording could be used, e.g. „user data (once per order)“ and „user data (once per personalized product)“?

raphaelm:
> Looks good. I only pushed some minot text-updates on help-texts regarding admission vs. personlized products.

Thanks!

> Maybe we need a warning/hint on products’ settings page, when no attendee data is asked, but the product is personalized? I know, this is described in the help text with „according to your event settings.“ but might not be understood so easily? Or at least link to the corresponding setting from the help-text?

Hmm, yeah, that makes sense. I pushed a warning text.

> Also could you elaborate regarding the „horrible“ UX when creating personalized, non-admission products?

Sure. Currently, the `personalized` field has a display dependency on `admission`. Therefore, when creating a new product, we can have `admission=True, personalized=True` as the default values of the form. If a user decides they want to create a non-admission product, the `personalized` field just hides and doesn't bother any more.

If we'd allow for personalized non-admission products, we'd need to remove the display dependency and show both fields. The question now is: How do we set the defaults in a useful way that does the right thing for new users?

We can't really set `personalized=True` as the default any more, since then people who are new to the system would create personalized non-admission products *all the time* even though they probably don't want their non-admission products to be personalized most of the time. We also can't really set `personalized=False` as the default, since the primary use case of pretix is tickets, which should be personalized by default. I also don't really want a fancy JavaScript that switches the default while working on it, because that would be very surprising behaviour for a web form.

wiffbi:
Ah, I see, thanks for pointing that out regarding UX on creating non-admission products. I did not really see it as a problem to always have personalization set with default to True when creating a product – for me it was just another settings to decide on independet from admission or non-admission. I actually stumbled upon/was irritaded it being hidden, when changing the admission-setting. For me it feels like it unnecessarily complicates things, but I also see your point in what is the best most of the time. It is better to start with how you set it up and maybe later add the option to have non-admission products with personalisation instead of later removing that option (when there are maybe setups with such non-admission personalized products).

wiffbi (approved):

PR #2991, opened 2022-12-21
Allow the automatic creation of customer accounts for guest users

raphaelm:
This PR adds the option to transparently create customer accounts for people when they make guest purchases. This is expected in some industries like museums or theaters (especially in North America, but sometimes also here) where the "customer account" feature is used more like a CRM and is expected to have a record of all customers ever interacting.

The user can still "claim" the customer account by verifying their email address if they sign up for one later.


raphaelm:
Apparently not a priority right now

PR #2994, opened 2022-12-21
Sendmail: Abstract away to allow more types of recipients

raphaelm:
This PR introduces a layer of abstraction into the mechanism of the "sendmail" plugin. This way, the same mechanic and basic UI can not only be used to send emails to ticket buyers, but also waiting list entries or even plugin-defined entities such as exhibitors:

2022-12-21T17:04:01+01:00

### DIFF
--- a/src/pretix/base/email.py
+++ b/src/pretix/base/email.py
@@ -520,20 +520,20 @@ def base_placeholders(sender, **kwargs):
             lambda event: (event if not event.has_subevents or not event.subevents.exists() else event.subevents.first()).get_date_from_display()
         ),
         SimpleFunctionalMailTextPlaceholder(
-            'url_remove', ['waiting_list_entry', 'event'],
-            lambda waiting_list_entry, event: build_absolute_uri(
+            'url_remove', ['waiting_list_voucher', 'event'],
+            lambda waiting_list_voucher, event: build_absolute_uri(
                 event, 'presale:event.waitinglist.remove'
-            ) + '?voucher=' + waiting_list_entry.voucher.code,
+            ) + '?voucher=' + waiting_list_voucher.voucher.code,
             lambda event: build_absolute_uri(
                 event,
                 'presale:event.waitinglist.remove',
             ) + '?voucher=68CYU2H6ZTP3WLK5',
         ),
         SimpleFunctionalMailTextPlaceholder(
-            'url', ['waiting_list_entry', 'event'],
-            lambda waiting_list_entry, event: build_absolute_uri(
+            'url', ['waiting_list_voucher', 'event'],
+            lambda waiting_list_voucher, event: build_absolute_uri(
                 event, 'presale:event.redeem'
-            ) + '?voucher=' + waiting_list_entry.voucher.code,
+            ) + '?voucher=' + waiting_list_voucher.code,
             lambda event: build_absolute_uri(
                 event,
                 'presale:event.redeem',
--- a/src/pretix/plugins/sendmail/forms.py
+++ b/src/pretix/plugins/sendmail/forms.py
@@ -70,14 +70,7 @@ class FormPlaceholderMixin:
         )
 
 
-class MailForm(FormPlaceholderMixin, forms.Form):
-    recipients = forms.ChoiceField(
-        label=_('Send email to'),
-        widget=forms.RadioSelect,
-        initial='orders',
-        choices=[]
-    )
-    sendto = forms.MultipleChoiceField()  # overridden later
+class BaseMailForm(FormPlaceholderMixin, forms.Form):
     subject = forms.CharField(label=_("Subject"))
     message = forms.CharField(label=_("Message"))
     attachment = CachedFileField(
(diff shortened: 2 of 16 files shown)


src/pretix/plugins/sendmail/views.py
```diff
@@ -126,6 +175,170 @@ def form_invalid(self, form):
...
+
+class OrderSendView(BaseSenderView):
+    form_class = forms.OrderMailForm
+    form_fragment_name = "pretixplugins/sendmail/send_form_fragment_orders.html"
+    context_parameters = ['event', 'order', 'position_or_address']
+    task = send_mails_to_orders
+
+    ACTION_TYPE = 'pretix.plugins.sendmail.sent'
+    TITLE = _("Orders or attendees")
+    DESCRIPTION = _("Send an email to every customer who purchased a ticket, to every person a ticket has been "
```
wiffbi, line 251: I think we need to be careful with the word „ticket“ here. AFAICT emails to customers are sent to any customer, even ones who only bought non-admission products. The second part „every person a ticket has been purchased for“ still stands currently, but if we decide to change the limitation that only admission products can be personalized, then the word „ticket“ does not work in this context as well. Maybe just shorten it?

```suggestion
    DESCRIPTION = _("Send an email to every customer, or to every person a ticket has been "
```

src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/index.html
```diff
@@ -0,0 +1,16 @@
+{% extends "pretixcontrol/event/base.html" %}
+{% load i18n %}
+{% block title %}{% trans "Send out emails" %}{% endblock %}
+{% block content %}
+    <h1>{% trans "Send out emails" %}</h1>
+    <ul class="list-group large-link-group">
+        {% for view in views %}
+            <a class="list-group-item" href="{{ view.url }}">
```
wiffbi, line 8: On a side note: it hurts a bit, that a `<li>` is missing, but adding it messes up bootstrap’s styling. Welp …

src/pretix/base/email.py
```diff
@@ -520,20 +520,20 @@ def base_placeholders(sender, **kwargs):
             lambda event: (event if not event.has_subevents or not event.subevents.exists() else event.subevents.first()).get_date_from_display()
         ),
         SimpleFunctionalMailTextPlaceholder(
-            'url_remove', ['waiting_list_entry', 'event'],
-            lambda waiting_list_entry, event: build_absolute_uri(
+            'url_remove', ['waiting_list_voucher', 'event'],
+            lambda waiting_list_voucher, event: build_absolute_uri(
                 event, 'presale:event.waitinglist.remove'
-            ) + '?voucher=' + waiting_list_entry.voucher.code,
+            ) + '?voucher=' + waiting_list_voucher.voucher.code,
```
wiffbi, line 526:
```suggestion
            ) + '?voucher=' + waiting_list_voucher.code,
```

src/pretix/plugins/sendmail/forms.py
```diff
@@ -91,7 +84,95 @@ class MailForm(FormPlaceholderMixin, forms.Form):
...
+        widget=forms.CheckboxSelectMultiple(
+            attrs={'class': 'scrolling-multiple-choice'}
+        ),
+        label=_('Only send to people who are waiting for'),
+        required=True,
+        queryset=Item.objects.none()
+    )
+    subevent = forms.ModelChoiceField(
+        SubEvent.objects.none(),
+        label=_('Only send to customers of'),
```
wiffbi, line 118: IMHO the wording „customer“ in the context of a waiting list is a bit strange. I do not have a better wording though maybe one can shorten all of the labels to not include anything like „customer“ or „person“ (and change the order so the product-filter is first – that feels more natural to me) like so:

Bildschirmfoto 2023-01-10 um 10 34 53

The sendmail orders page with subevents would need similiar labels, if we decide to go with this approach.

src/pretix/plugins/sendmail/views.py
```diff
@@ -189,88 +402,111 @@ def form_valid(self, form):
...
+class WaitinglistSendView(BaseSenderView):
+    form_class = forms.WaitinglistMailForm
+    form_fragment_name = "pretixplugins/sendmail/send_form_fragment_waitinglist.html"
+    context_parameters = ['event', 'waiting_list_entry', 'event_or_subevent']
+    task = send_mails_to_waitinglist
+
+    ACTION_TYPE = 'pretix.plugins.sendmail.sent.waitinglist'
+    TITLE = _("Waiting list")
+    DESCRIPTION = _("Send an email to every customer who is currently waiting to receive a ticket through the waiting "
+                    "list feature.")
```
wiffbi, line 438: „Customer“ and „ticket“ are words I would be careful in the context of waitinglists. What about something like:

```suggestion
    DESCRIPTION = _("Send an email to everyone who is currently on the waiting list (ie. has not yet received a voucher).")
```

wiffbi (commented):
Looks good, just a minor typo bug-fix added. And a few questions around wording regarding „customer“ and „ticket“.

src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/index.html
raphaelm: oh, that was not on purpose. I'll make the outer element a div instead of a list so it's not horribly invalid

src/pretix/plugins/sendmail/views.py
raphaelm: I rephrased it

src/pretix/plugins/sendmail/forms.py
wiffbi: Yes, the and/or part is indeed confusing/hard to grasp. :)

wiffbi (commented):
I really like the new wording, thank you! On litte thing I stumbled upon on the order-mail page (nothing new in this PR, but maybe worth a look): „Everyone who created a ticket order“ in sendmail/forms.py::235 https://github.com/pretix/pretix/pull/2994/files#diff-246c1cab8723510c02a7767178bb1e08f34ce3322895a690886c504b9f75d636L154

Should this maybe read „Everyone who created an order“?

raphaelm:
> Should this maybe read „Everyone who created an order“?

Good catch, gonna fix that on master

PR #2996, opened 2022-12-22
Device list: Hide revoked devices by default

raphaelm:
This has been requested by a large number of people by now :)


wiffbi (approved):
Works as intended. For those that are used to seeing all devices it might be irritating on how to show all devices as it is not super clear that a filter is active. But probably 99.9% of the time just showing the active ones or even being able to switch between active and revoked instead of showing all is what is used. So, PR is good to go!

PR #3003, opened 2023-01-02
Validate image size in pixels at upload time

raphaelm:
Uploading images that are below our 10MB barrier but have millions of pixels can eat up lots of memory, which leads to problems during thumbnail generation. We should already validate for that at upload time.

### DIFF
--- a/src/pretix/control/forms/__init__.py
+++ b/src/pretix/control/forms/__init__.py
@@ -51,6 +51,9 @@ from django_scopes.forms import SafeModelMultipleChoiceField
 from pretix.helpers.hierarkey import clean_filename
 
 from ...base.forms import I18nModelForm
+from ...helpers.images import (
+    IMAGE_EXTS, validate_uploaded_file_for_valid_image,
+)
 
 # Import for backwards compatibility with okd import paths
 from ...base.forms.widgets import (  # noqa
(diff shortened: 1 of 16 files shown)


src/pretix/control/forms/__init__.py
```diff
@@ -220,6 +223,10 @@ def clean(self, *args, **kwargs):
             ext = ext.lower()
             if ext not in self.ext_whitelist:
                 raise forms.ValidationError(_("Filetype not allowed!"))
+
+            if ext in IMAGE_EXTS:
+                validate_uploaded_file_for_valid_image(data)
+
```
wiffbi, line 229: Is `ExtValidationMixin` the _correct_ place to check for a valid image? Kind of yes, if you test, whether it can be parsed as an image (although „Ext“ kind of implies just to test whether the file-type/-extension is allowed at all). It could also go into `SizeValidationMixin` as we also check for pixel dimensions (which is also kind of size-related) and parsing the image for validity there is kind of a by-product when checkin for dimensions … not sure, what is best, just stumbled upon it. `SizeValidationMixin` seems to be used anywhere `ExtValidationMixin` is used, so should not be a problem swapping around. I am undecided though whether it needs to be swapped at all.

raphaelm:
Pushed more tests :)

src/pretix/control/forms/__init__.py
wiffbi: Thanks for pointing that out!

wiffbi:
Is this commit by accident? Probably not worth the effort to pull out though … https://github.com/pretix/pretix/pull/3003/commits/9437314a9c5ad853077b7e510daf22d7af039c1f

PR #3004, opened 2023-01-02
Writable API for ticket layouts

raphaelm:
This introduces writable API endpoints for creating and updating ticket layouts. It does not yet allow to write the assignment between layouts and products.

Exposing this API in a safe way required formalizing the format we use for layouts for a first time. I have therefore added a JSON schema that can be used to validate the JSON layout and will in the future also help us maintain compatibility between pretix and e.g. pretixPRINT. This schema is used to validate all API endpoints.

I have ran the schema on *all* ~60k layouts in the pretix.eu database. Almost all of them validate correctly. A very small number (~10) fails due to text areas with ``fontfamily=null`` (which is in fact invalid and breaks stuff) or some unsuccessful past exploitation attempts.

I have therefore also added the schema validation to the frontend editor when pasting in custom JSON layouts. This will prevent some human errors in the future. I have *not* added validation to the server-side code of the backend editor. This means we still can have invalid layouts end up in the database like before, but it allows to fail more softly with the invalid ones we already have. We can always still make this more strict in a separate PR.


robbi5 (changes requested):

src/pretix/static/pretixcontrol/js/ui/editor.js
```diff
@@ -988,8 +989,26 @@ var editor = {
...
     _source_save: function () {
-        editor.load(JSON.parse($("#source-textarea").val()));
-        $("#source-container").hide();
+        try {
+            var Ajv = window.ajv2020
+            var ajv = new Ajv()
+            var validate = ajv.compile(editor.schema)
+            var data = JSON.parse($("#source-textarea").val())
+            var valid = validate(data)
+            editor.load(data);
```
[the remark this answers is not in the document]
raphaelm: Correct, fix pushed

doc/plugins/ticketoutputpdf.rst
```diff
@@ -147,3 +151,122 @@ Endpoints
...
+
+      POST /api/v1/organizers/bigevents/events/sampleconf/ticketlayouts/ HTTP/1.1
+      Host: pretix.eu
+      Accept: application/json, text/javascript
+      Content-Type: application/json
+
+      {
+        "name": "Default layout",
+        "default": true,
+        "layout": {…},
```
wiffbi, line 171:
```suggestion
        "layout": […],
```

doc/plugins/ticketoutputpdf.rst
```diff
@@ -147,3 +151,122 @@ Endpoints
...
+        "background": {},
```
wiffbi, line 172: This will probably work as a „no background“ request, but would it make more sense to set it to something like empty string `""` or `false` or `null`?

doc/plugins/ticketoutputpdf.rst
```diff
@@ -147,3 +151,122 @@ Endpoints
...
+
+      HTTP/1.1 201 Created
+      Vary: Accept
+      Content-Type: application/json
+
+      {
+        "id": 1,
+        "name": "Default layout",
+        "default": true,
+        "layout": {…},
```
wiffbi, line 188:
```suggestion
        "layout": […],
```

src/tests/plugins/ticketoutputpdf/test_api.py
```diff
@@ -57,32 +70,138 @@ def env():
...
+        {
+            "name": "Bar",
+            "background": file_id_png,
+        },
+        format='json'
+    )
+    assert resp.status_code == 200
+    env[1].refresh_from_db()
+    assert env[1].name == "Bar"
+
```
wiffbi, line 185: Should the background be checked as well like in `test_api_create`? Also, should we test whether an update to the layout works?

doc/plugins/ticketoutputpdf.rst
```diff
@@ -147,3 +151,122 @@ Endpoints
...
+
+      HTTP/1.1 200 OK
+      Vary: Accept
+      Content-Type: application/json
+
+      {
+        "id": 1,
+        "name": "Default layout",
+        "default": true,
+        "layout": {…},
```
wiffbi, line 232:
```suggestion
        "layout": […],
```

wiffbi (commented):
Looks good, just some minor comments regarding doc/code examples for the layout-parameter being a list, not an object.

src/tests/plugins/ticketoutputpdf/test_api.py
raphaelm: sure

src/pretix/static/schema/pdf-layout.schema.json
```diff
@@ -0,0 +1,371 @@
...
+                "type": "number"
+              },
+              {
+                "type": "string",
+                "pattern": "^[0-9]+(\\.[0-9]+)?$"
+              }
+            ]
+          },
+          "content": {
+            "description": "Name of a variable to use. The available values depend on event configuration and installed plugins.",
```
robbi5, line 347: Not really for the powered by logo style, the values are fixed :)

PR #3006, opened 2023-01-04
Add setting presale end show date

Ket3r:
This option is similar to `presale_start_show_date`. 

It enables the organizer to inform customers about the end of the presale period.

### DIFF
--- a/src/pretix/api/serializers/event.py
+++ b/src/pretix/api/serializers/event.py
@@ -663,6 +663,7 @@ class EventSettingsSerializer(SettingsSerializer):
         'show_items_outside_presale_period',
         'display_net_prices',
         'presale_start_show_date',
+        'presale_end_show_date',
         'locales',
         'locale',
         'region',
--- a/src/pretix/base/settings.py
+++ b/src/pretix/base/settings.py
@@ -782,6 +782,17 @@ DEFAULTS = {
             widget=forms.CheckboxInput,
         )
     },
+    'presale_end_show_date': {
+        'default': 'False',
+        'type': bool,
+        'form_class': forms.BooleanField,
+        'serializer_class': serializers.BooleanField,
+        'form_kwargs': dict(
+            label=_("Show end date"),
+            help_text=_("Show the presale end date during active presale."),
+            widget=forms.CheckboxInput,
+        )
+    },
     'tax_rate_default': {
         'default': None,
         'type': TaxRule
--- a/src/pretix/control/forms/event.py
+++ b/src/pretix/control/forms/event.py
@@ -483,6 +483,7 @@ class EventSettingsForm(SettingsForm):
         'show_items_outside_presale_period',
         'display_net_prices',
         'presale_start_show_date',
+        'presale_end_show_date',
         'locales',
         'locale',
         'region',
--- a/src/pretix/control/templates/pretixcontrol/event/settings.html
+++ b/src/pretix/control/templates/pretixcontrol/event/settings.html
@@ -222,6 +222,7 @@
                 <legend>{% trans "Timeline" %}</legend>
                 {% bootstrap_field form.presale_start layout="control" %}
                 {% bootstrap_field sform.presale_start_show_date layout="control" %}
+                {% bootstrap_field sform.presale_end_show_date layout="control" %}
                 {% bootstrap_field form.presale_end layout="control" %}
                 {% bootstrap_field sform.show_items_outside_presale_period layout="control" %}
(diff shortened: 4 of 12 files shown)


raphaelm:
@cla-bot[bot] check

raphaelm:
Thanks a lot for your first contribution to pretix :sparkles: 

(Generally, we highly recommend pitching any ideas through a discussion first to see if its something that should be added, before putting effort into how it's added.)

I see that this might be useful. We are very cautious about introducing new settings (pretix has too many already), but I think it makes sense. However, I think I don't agree with how the setting is used:

- Showing it in the calendar as "Book until {{ date }}" with "SHORT_DATE_FORMAT" is not going to be good enough.
  - When thinking about a conference and a presale that ends a week before, showing a date is sufficient. However, a very common use case would be an event series with hourly events, think e.g. time slots to access a sauna. Often, presale_end is set to e.g. 1h before the start, in which case showing a *date* is useless, we'd need to show a *time*. In other cases, it is set to 48h before the start, in which case we'd likely need both.
  - Going further, I think the calendar is too crowded with information to put this in. I don't think this information is vital enough that it should be included in the calendar at all, it just distracts from more important information.
- I don't think an `alert-info` box is the right way to show it on the event page either. If we make this info available, it should still be much less prominent. An alert box is something we want everyone to read. We don't want every visitor to read this and think "oh, I have weeks to buy". We mostly want them to purchase a ticket, not think about rescheduling that.

I'd be happy to hear some thoughts about this from @wiffbi as well, but he's on vacation this week :)

wiffbi:
I like that you made it behave similar to presale_start, but I have the same remarks as @raphaelm. I do not think it helps showing it in the calender generally, as probably most of the time it is far in the future and just makes the UI harder to grasp/parse from a user’s perpective. I think it might be useful, if presale_end is on the same day as the current day (as in current day/now where the user is) though. What about we generally show a „Book now“ in the calender and only if presale_end is on the current day (or a few hours away), we show a „Book until {{ TIME_FORMAT }}“? To complex?

On the event’s page we might use a similar logic: if presale_end is on the current day (or a few hours away), we could show the info-box – IMHO it is a valid info-alert when time is running out to book … and generally we could show presale_end just below the general event’s start/end/admission dates?

Just a minor thing: I’d move the `bootstrap_field sform.presale_end_show_date` below the date-input for presale_end.

raphaelm:
> What about we generally show a „Book now“ in the calender and only if presale_end is on the current day (or a few hours away), we show a „Book until {{ TIME_FORMAT }}“? To complex?

It sounds nice and I don't think it's too complex, but there's one caveat: If we have e.g. `presale_end = date_from - 48h`, then "Book until 12pm" in a calendar tile might not be intuitive to understand whether that's 12pm on the day of the event or 12pm today.

wiffbi:
> It sounds nice and I don't think it's too complex, but there's one caveat: If we have e.g. `presale_end = date_from - 48h`, then "Book until 12pm" in a calendar tile might not be intuitive to understand whether that's 12pm on the day of the event or 12pm today.

I think, if we go with the timerange option when deciding on showing the „Book until …“, then this timerange should be less than 24 hours (or maybe even just 16 or 12 hours, just to make sure it is not to close to the current time of the day, if presale_end is almot 24h away). So maybe just skip the whole „on the same day“ logic and just compare if (sorry, pseudo code) `0 < presale_end - now(tz=users-timezone?) < 16 hours`.

Hm, I think your problem still persists in calendar view, as one could think that „book until {{time}}“ means on the day of the event (especially if presale_end is a week before the event or so). So if we limit this to less than 24 hours, we could make sure, that this will only need „book until today, {{time}}“ or „book until tomorrow, {{time}}“. Hm, not that great … also, I think, this „book until“ probably needs to be cleverer, something like either less than 24 hours away or „now“ is less time away from presale_end than presale_end is from date_from (`presale_end - now < date_from - presale_end`)? That way presales ending like a week before the event (probably indicating a long presale-period) will be shown a week before presale ending. Problem is though, that presale_end then often needs a date (if presale_end is not on the same day as date_from) and a time (if given in settings or if not given, but presale_end and date_from is on the same day). Puh, sorry, I guess, this makes it rather complicated.

Maybe just go super easy: if presale_end is on the same day as now(), then just show „book until {{time}} today“. Forthermore, if presale_end and date_from are on the same day (and on the same day as now()), then a „book until {{time}}“ should be sufficient. This does not cater for all needs, but probably the most urgent ones, where there really is not much time left to book?

raphaelm:
Closing for inactivity

PR #3025, opened 2023-01-11
Reorganize UI for exporters

raphaelm:
With more and more exporters being added to pretix, the export page has become very crowded and I noticed that new users have a hard time discovering which are the actual useful ones for their use case.

So I propose the following changes:

- Exporters have a description now
- Exporters are sorted into categories
- Our most recommended exporters that are useful for most users are now marked as "featured"
- Selecting an exporter and filling out its parameters are now two separate steps. This is especially relevant since I'm planning on adding another feature that will make the forms more … extensive

Screenshot 2023-01-11 at 10-51-24 Data export pretix eu

### DIFF
--- a/src/pretix/base/exporters/events.py
+++ b/src/pretix/base/exporters/events.py
@@ -35,7 +35,7 @@
 from django.dispatch import receiver
 from django.utils.formats import date_format
 from django.utils.functional import cached_property
-from django.utils.translation import gettext_lazy as _
+from django.utils.translation import gettext_lazy as _, pgettext_lazy
 
 from ...control.forms.filter import get_all_payment_providers
 from ..exporter import ListExporter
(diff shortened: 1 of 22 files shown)


src/pretix/base/exporters/events.py
```diff
@@ -45,6 +45,8 @@
 class EventDataExporter(ListExporter):
     identifier = 'eventdata'
     verbose_name = _('Event data')
+    category = pgettext_lazy('export_category', 'Event data')
+    description = _('Download a spread sheet with information on all events in this organizer account.')
```
wiffbi, line 49:
```suggestion
    description = _('Download a spreadsheet with information on all events in this organizer account.')
```

wiffbi (approved):
Looks good, just a minor typo

PR #3027, opened 2023-01-12
Exports: Add predefined timeframes

raphaelm:
Many of our exports have "date from" and "date until" fields. These are useful, but sometimes annoying to fill, but mostly they have one major problem: We soon want to create a feature that allows people to "save" and "schedule" exports, i.e. "send me a report every monday at 8am". For this, we need to bring all export configuration forms into a format where it is useful to store the format and then reuse it later.

Now, saving a report configuration with "from 2023-01-01 to 2023-01-31" is **much less useful** than saving a report configuration with just "this month", which can automatically be re-evaluated to be the correct timeframe whenever the export runs!

So this PR collapses the "date from" and "date until" fields of all exporters into one "date range" field that allows to **either* select one of a number of pre-defined time frames **or** enter a range on your own.

The only downside is that after this PR, you can no longer fill only **one** of the fields, i.e. only a start date without an end date. I think that's an acceptible compromise to keep UI and code simple because you can always enter a very early or very late date.

## After

2023-01-12-115708_1276x1401_scrot

### Plugins requiring adjustment

- [ ] pretix fees
- [ ] shipping
- [ ] reports
- [ ] datev
- [ ] accountingtools
- [ ] posbackend

### DIFF
--- a/src/pretix/base/exporters/dekodi.py
+++ b/src/pretix/base/exporters/dekodi.py
@@ -23,10 +23,9 @@ import json
 from collections import OrderedDict
 from decimal import Decimal
 
-import dateutil
-from django import forms
 from django.core.serializers.json import DjangoJSONEncoder
 from django.dispatch import receiver
+from django.utils.timezone import now
 from django.utils.translation import gettext, gettext_lazy, pgettext_lazy
 
 from pretix.base.i18n import language
@@ -34,6 +33,7 @@ from pretix.base.models import Invoice, OrderPayment
 
 from ..exporter import BaseExporter
 from ..signals import register_data_exporters
+from ..timeframes import DateFrameField, resolve_timeframe_to_dates_inclusive
 
 
 class DekodiNREIExporter(BaseExporter):
@@ -115,7 +115,7 @@ class DekodiNREIExporter(BaseExporter):
                         'PTNo14': p.info_data.get('reference') or '',
                         'PTNo15': p.full_id or '',
                     })
-            elif p.provider.startswith('stripe'):
+            elif p.provider and p.provider.startswith('stripe'):
                 src = p.info_data.get("source", p.info_data)
                 payments.append({
                     'PTID': '81',
@@ -194,17 +194,9 @@ class DekodiNREIExporter(BaseExporter):
     def render(self, form_data):
         qs = self.event.invoices.select_related('order').prefetch_related('lines', 'lines__subevent')
 
-        if form_data.get('date_from'):
-            date_value = form_data.get('date_from')
-            if isinstance(date_value, str):
-                date_value = dateutil.parser.parse(date_value).date()
-            qs = qs.filter(date__gte=date_value)
-
-        if form_data.get('date_to'):
-            date_value = form_data.get('date_to')
-            if isinstance(date_value, str):
-                date_value = dateutil.parser.parse(date_value).date()
-            qs = qs.filter(date__lte=date_value)
+        if form_data.get('date_range'):
+            d_start, d_end = resolve_timeframe_to_dates_inclusive(now(), form_data['date_range'], self.timezone)
+            qs = qs.filter(date__gte=d_start, date__lte=d_end)
 
         jo = {
             'Format': 'NREI',
@@ -220,21 +212,13 @@ class DekodiNREIExporter(BaseExporter):
     def export_form_fields(self):
         return OrderedDict(
             [
-                ('date_from',
-                 forms.DateField(
-                     label=gettext_lazy('Start date'),
-                     widget=forms.DateInput(attrs={'class': 'datepickerfield'}),
+                ('date_range',
+                 DateFrameField(
+                     label=gettext_lazy('Date range'),
+                     include_future_frames=False,
                      required=False,
-                     help_text=gettext_lazy('Only include invoices issued on or after this date. Note that the invoice date does '
+                     help_text=gettext_lazy('Only include invoices issued in this time frame. Note that the invoice date does '
                                             'not always correspond to the order or payment date.')
-                 )),
-                ('date_to',
-                 forms.DateField(
-                     label=gettext_lazy('End date'),
-                     widget=forms.DateInput(attrs={'class': 'datepickerfield'}),
-                     required=False,
-                     help_text=gettext_lazy('Only include invoices issued on or before this date. Note that the invoice date '
-                                            'does not always correspond to the order or payment date.')
                  )),
             ]
         )
--- a/src/pretix/base/exporters/dekodi.py
+++ b/src/pretix/base/exporters/dekodi.py
@@ -196,7 +196,10 @@ class DekodiNREIExporter(BaseExporter):
 
         if form_data.get('date_range'):
             d_start, d_end = resolve_timeframe_to_dates_inclusive(now(), form_data['date_range'], self.timezone)
-            qs = qs.filter(date__gte=d_start, date__lte=d_end)
+            if d_start:
+                qs = qs.filter(date__gte=d_start)
+            if d_end:
+                qs = qs.filter(date__lte=d_end)
 
         jo = {
             'Format': 'NREI',
--- a/src/pretix/base/exporters/orderlist.py
+++ b/src/pretix/base/exporters/orderlist.py
@@ -33,10 +33,8 @@
 # License for the specific language governing permissions and limitations under the License.
 
 from collections import OrderedDict
-from datetime import date, datetime, time
 from decimal import Decimal
 
-import dateutil
 import pytz
 from django import forms
 from django.db.models import (
@@ -46,7 +44,7 @@ from django.db.models import (
 from django.db.models.functions import Coalesce
 from django.dispatch import receiver
 from django.utils.functional import cached_property
-from django.utils.timezone import get_current_timezone, make_aware, now
+from django.utils.timezone import get_current_timezone, now
 from django.utils.translation import (
     gettext as _, gettext_lazy, pgettext, pgettext_lazy,
 )
@@ -67,6 +65,10 @@ from ..exporter import (
 )
 from ..signals import (
     register_data_exporters, register_multievent_data_exporters,
+)
+from ..timeframes import (
+    DateFrameField,
+    resolve_timeframe_to_datetime_start_inclusive_end_exclusive,
 )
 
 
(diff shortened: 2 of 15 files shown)


raphaelm:
Updated screenshot:
2023-01-12-124427_1276x1401_scrot

src/pretix/base/timeframes.py
```diff
@@ -0,0 +1,388 @@
...
+    def has_changed(self, initial, data):
+        if initial is None:
+            initial = self.widget.decompress(initial)
+        return super().has_changed(initial, data)
+
+    def clean(self, value):
+        if value[0] == 'custom':
+            if not value[1] or not value[2]:
+                raise ValidationError(self.error_messages['incomplete'])
+            # todo validate start<end
```
wiffbi, line 342: Would be an interesting hack to use this to create reports excluding a certain timerange in between two dates. :)

src/pretix/base/timeframes.py
```diff
@@ -0,0 +1,388 @@
...
+        return []
+
+    def get_context(self, name, value, attrs):
+        ctx = super().get_context(name, value, attrs)
+        ctx['required'] = self.timeframe_choices[0][0] == 'unset'
+        return ctx
+
+
+def _describe_timeframe(label, start, end, future, describe):
+    details = f'{date_format(start(now()), "SHORT_DATE_FORMAT")} - {date_format(end(now(), start(now())), "SHORT_DATE_FORMAT")}'
```
wiffbi, line 277: Hm, I think this could be made more readable, especially when start and end are the same. What about using the existing daterange-helper? This IMHO creates more human-readable date ranges.

```suggestion
    from pretix.helpers.daterange import daterange # todo: probably import at top
    d_start = start(now())
    d_end = end(now(), d_start)
    details = daterange(d_start, d_end)
```

Generally some describe + date-range labels could be optimized, but that would mean a custom label-method for every date-range – although I think it could be possible creating readable describe-lambdas with `d_end` and the use of the daterange-helper.

Actually, we could make out lives even easier: I do not think at least all „by month“, „by quarter“ and „by year“ do not need any daterange-labels at all (so just the existing describe-lambda). Even the day-ranges probably do not need them. For the week-range I find it helpful though.

src/pretix/base/exporters/dekodi.py
```diff
@@ -115,7 +115,7 @@ def _encode_invoice(self, invoice: Invoice):
                         'PTNo14': p.info_data.get('reference') or '',
                         'PTNo15': p.full_id or '',
                     })
-            elif p.provider.startswith('stripe'):
+            elif p.provider and p.provider.startswith('stripe'):
```
wiffbi, line 118: That is probably not meant in this PR, but a good fix nonetheless.

src/pretix/base/exporters/orderlist.py
```diff
@@ -186,48 +172,27 @@ def event_object_cache(self):
         return {e.pk: e for e in self.events}
 
     def _date_filter(self, qs, form_data, rel):
+        print(form_data)
```
wiffbi, line 175:
```suggestion
```

src/pretix/base/timeframes.py
```diff
@@ -0,0 +1,388 @@
...
+
+    def has_changed(self, initial, data):
+        if initial is None:
+            initial = self.widget.decompress(initial)
+        return super().has_changed(initial, data)
+
+    def clean(self, value):
+        if value[0] == 'custom':
+            if not value[1] or not value[2]:
+                raise ValidationError(self.error_messages['incomplete'])
```
wiffbi, line 341: Would it make sense to allow incomplete date-ranges? That way one could generate reports for everything up until a certain date or only beginning on a certain date. I would not auto-create a valid start-date or end-date if none is given, but patch time-range parsing. See my other comment.

src/pretix/base/timeframes.py
```diff
@@ -0,0 +1,388 @@
...
+    """
+    Given a serialized timeframe, evaluate it relative to `ref_dt` and return a tuple of dates
+    where the first element ist the first possible date value within the timeframe and the second
+    element is the last possible date value in the tiemframe.
+    """
+    if isinstance(ref_dt, datetime):
+        ref_dt = ref_dt.astimezone(timezone).date()
+    if "/" in frame:
+        start, end = frame.split("/", 1)
+        return date.fromisoformat(start), date.fromisoformat(end)
```
wiffbi, line 356: What about allowing incomplete dateranges? Might be helpful for certain reports. Would need some changes in other parts, but should be doable? I could try and add it, if you want.

```suggestion
        return date.fromisoformat(start) if start else None, date.fromisoformat(end) if end else None
```

src/pretix/plugins/reports/exporters.py
```diff
@@ -606,48 +620,28 @@ def export_form_fields(self):
...
-            if date_until:
-                qs = qs.filter(order__datetime__lt=date_until)
-        elif date_filter == 'last_payment_date':
+
+        if date_range:
+            dt_start, dt_end = resolve_timeframe_to_datetime_start_inclusive_end_exclusive(now(), date_range, self.timezone)
+
+        if date_filter == 'order_date' and date_range:
+            qs = qs.filter(order__datetime__gte=dt_start)
+            qs = qs.filter(order__datetime__lt=dt_end)
```
wiffbi, line 643: This needs changes if we decide on incomplete dateranges.

src/pretix/plugins/reports/exporters.py
```diff
@@ -656,10 +650,8 @@ def filter_qs(self, qs, form_data):
                 m=Max('payment_date')
             ).values('m').order_by()
             qs = qs.annotate(payment_date=Subquery(p_date, output_field=DateTimeField()))
-            if date_from:
-                qs = qs.filter(payment_date__gte=date_from)
-            if date_until:
-                qs = qs.filter(payment_date__lt=date_until)
+            qs = qs.filter(payment_date__gte=dt_start)
+            qs = qs.filter(payment_date__lt=dt_end)
```
wiffbi, line 654: This needs changes if we decide on incomplete dateranges.

wiffbi (commented):
Looks good! I have some comments regarding formatting the options’ labels. Also a comment, whether we maybe allow for incomplete custom date-ranges to make reports even more flexible.

And I know we already talked about it, but starting all options’ labels with uppercase, but having the optgroup labels start lowercase still feels a bit akward to me. Maybe give it another thought?

src/pretix/base/timeframes.py
raphaelm: I think for "last 7 days" it makes sense, and it also describes phrases like "quarter to date" to people who are not used to the phrase, even though it is quite common

src/pretix/base/timeframes.py
raphaelm: yeah… but no :D

src/pretix/base/timeframes.py
raphaelm: I changed it to be explicitly changeable for every option, what do you think?

raphaelm:
> Looks good! I have some comments regarding formatting the options’ labels. Also a comment, whether we maybe allow for incomplete custom date-ranges to make reports even more flexible.

Yeah, I reluctantly agree, I made the change ;)

> And I know we already talked about it, but starting all options’ labels with uppercase, but having the optgroup labels start lowercase still feels a bit akward to me. Maybe give it another thought?

The reasoning was that there's a verb in front, in some places explicitly, in some places implicitly. So it's read as "Filter by week", "Filter by month".

In German, I'd also write it as
- nach Tag
  - heute
  - morgen
- nach Woche
	- 	…

src/pretix/base/timeframes.py
wiffbi: Perfect! Thanks!

wiffbi:
Thanks for pointing out the reasoning behing the lowercase „by“ – I totally agree! I will do some moe testing and will approve if ready.

src/pretix/base/timeframes.py
```diff
@@ -0,0 +1,404 @@
...
+        **forms.MultiValueField.default_error_messages,
+        'inconsistent': gettext_lazy('The end date must be after the start date.'),
+    }
+
+    def __init__(self, *args, **kwargs):
+        include_future_frames = kwargs.pop('include_future_frames')
+
+        top_choices = [('custom', gettext_lazy('Custom timeframe'))]
+        if not kwargs.get('required', True):
+            top_choices.insert(0, ('unset', pgettext_lazy('reporting_timeframe', 'All time')))
```
wiffbi, line 314: Not super keen on the wording „All time“. I do not know something better though, maybe „no limit“ or „do not limit“ or just „–––“ or maybe „None“?

src/pretix/base/timeframes.py
```diff
@@ -0,0 +1,404 @@
...
+    (
+        'days_today',
+        pgettext_lazy('reporting_timeframe', 'Today'),
+        lambda ref_d: ref_d,
+        lambda ref_d, start_d: start_d,
+        False,
+        pgettext_lazy('reporting_timeframe', 'by day'),
+        daterange
+    ),
+    (
```
wiffbi, line 55: What about 'yesterday'? Might be interesting every morning to report what happened the day before …

src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -355,7 +355,16 @@ var form_handlers = function (el) {
         var dependent = $(this),
             dependency = $($(this).attr("data-display-dependency")),
             update = function (ev) {
-                var enabled = dependency.toArray().some(function(d) {return (d.type === 'checkbox' || d.type === 'radio') ? d.checked : (!!d.value && !d.value.match(/^0\.?0*$/g))});
+                var enabled = dependency.toArray().some(function(d) {
+                    if (d.type === 'checkbox' || d.type === 'radio') {
+                        return d.checked;
+                    } else if (d.type === 'select-one') {
+                        return d.value === dependent.attr("data-display-dependency-value");
```
wiffbi, line 362: I do not see where this is in use. Also an input-element of type `select-one` is not really common. Could you elaborate on this? Thanks! :)

wiffbi (commented):
Sorry, did not make my „review“ a review, but posted single comments.

src/pretix/base/timeframes.py
raphaelm: Fair, will add

src/pretix/static/pretixcontrol/js/ui/main.js
raphaelm: This is the code that makes the from/to fields show up when selecting "other".

`"select-one"` is the `DOMElement.type` value of a `<select>` without the `multiple` attribute. Wasn't my idea, and I also learned it yesterday: https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/type

src/pretix/base/timeframes.py
raphaelm: Hm, no, I disagree.

"––" or "None" might make the user think that they need to select something, but they don't, it's a valid option to keep empty.

"Date range: [nothing]" also feels less intuitive than "Date range: all time" to me.

wiffbi: Okay, I agree!

wiffbi (approved):

src/pretix/static/pretixcontrol/js/ui/main.js
wiffbi: Wow, thanks, TIL!

PR #3029, opened 2023-01-12
Checkout: remove Indonesia from zipcode-validation

wiffbi:
Indonesia uses a 4 digit zip-code with an additional province-select in localflavor. Pretix currently does not support province, so a common fallback seems to be a 5 digit zip-code, which fails in localflavor. This PR removes Indonesia from zipcode-validation through localflavor.

Should I add a custom validator that at least does a basic „is 5 digits“ validation?


raphaelm (approved):

raphaelm:
I think this is good enough for now

PR #3033, opened 2023-01-13
Scheduled exports

raphaelm:
Screenshot 2023-01-18 at 15-01-30 Data export pretix eu

Screenshot 2023-01-18 at 15-01-42 Data export pretix eu

Screenshot 2023-01-18 at 15-02-04 Data export pretix eu

- [x] Data model
  - [x] migration
- [x] Creation form
  - [x] Default texts
  - [x] Actually store schedule
- [x] Allow to manage stored reports
  - [x] update
  - [x] delete
  - [x] run
- [x] Actually execute schedule
  - [x] Error handling
- [x] Organizer level
  - [x] deal with newly created events?
  - [x] timezones
- [x] logdisplay
- [ ] Tests
  - [x] exporters
  - [ ] permissions
  - [ ] views
- [x] Spam protection?


pc-coholic (changes requested):
- `error_counter` seems only to be used to display the "will be disabled"-message, but won't actually disable sending the messages (side effect: all mails will still include the soft-fail warning)
- If an export is not available any longer (for example: plugin that provided the export is disabled), the edit-button just won't do anything.
- Running defective exports (for whatever reason) manually won't increase the failure counter. Perhaps by design?
- There is no way to force the sending of a scheduled report right away/at the next possible moment. If it is now 18.01. 13:00 and I set the first run to 18.01. 12:00, the next run won't be "asap" but rather 19.01. 12:00. I'm also fine with seeing this "as designed", but it might cause confusion.
- I kinda expected the "Play"-Button on the scheduled export to run the export *and* send it to the recipients. Perhaps we could add a secondary "rund and send"-button?

raphaelm:
> error_counter seems only to be used to display the "will be disabled"-message, but won't actually disable sending the messages (side effect: all mails will still include the soft-fail warning)

Fix pushed.

> If an export is not available any longer (for example: plugin that provided the export is disabled), the edit-button just won't do anything.

Fixed, better error message and only delete button-

> Running defective exports (for whatever reason) manually won't increase the failure counter. Perhaps by design?

Yeah, manual running shouldn't interfere here.

> There is no way to force the sending of a scheduled report right away/at the next possible moment. If it is now 18.01. 13:00 and I set the first run to 18.01. 12:00, the next run won't be "asap" but rather 19.01. 12:00. I'm also fine with seeing this "as designed", but it might cause confusion.

Hmm, yeah, I get that, but I also think that accepting runtimes that are already in the past (a) makes the algorithm for finding the next time much more confusing, and when scheduling e.g. a 04:00 runtime you probably don't expect it before "next night", so I'm leaning towards "as designed". I also don't want to give a min_date to the "start date" option since that would be very messy when editing existing schedules.

> I kinda expected the "Play"-Button on the scheduled export to run the export and send it to the recipients. Perhaps we could add a secondary "rund and send"-button?

I'd be happy to talk about that for a later PR, but I'd skip it right now for time reasons. Do you think there is another icon or tooltip we could use to make this clearer? More of a download symbol than a play symbol?

pc-coholic (changes requested):
- Can't delete exports (neither organizer nor event level): `NoReverseMatch at /control/organizer/bigevents/export/1/delete`

Rest looks good to me.

PR #3038, opened 2023-01-17
Product list: Fix add-to-cart-button being shown on seating-only event

wiffbi:
Currently the add-to-cart button is also shown, if all items do require a seat. This PR changes it, so the button is only shown, if there are items available, which do not require a seat.

UPDATE: at least pretix_seating uses `get_grouped_items` with expected result being `items, display_add_to_cart`. Should I try to update all plugins code or undo the change to `get_grouped_items` regarding `display_add_to_cart`? Hm.

### DIFF
--- a/src/pretix/presale/checkoutflow.py
+++ b/src/pretix/presale/checkoutflow.py
@@ -526,7 +526,7 @@ class AddOnsStep(CartMixin, AsyncAction, TemplateFlowStep):
 
                 if ckey not in item_cache:
                     # Get all items to possibly show
-                    items, _btn = get_grouped_items(
+                    items = get_grouped_items(
                         self.request.event,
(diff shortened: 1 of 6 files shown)


raphaelm:
I believe seating is the only public plugin to use `get_grouped_items` but I think reverting the change is safer? It also makes sense to keep the logic in a central place to not have it duplicated between e.g. event index page, voucher redemption page and widget

wiffbi:
I changed it back to set `display_add_to_cart` in `get_grouped_items`. It still feels a bit strange to have this function to return this as well, but it is safer to just leave the function’s return-value as is. This PR then basically just additionally checks the item if it is bound to a seated to determine if the add-to-cart-button needs to be shown.

raphaelm (approved):

PR #3049, opened 2023-01-23
Clean up localization or error messages in cart

### LINKED ISSUE #2519: german error message in english shop
**Describe the bug**
If a customer using english tries to add two or more of the same product in the cart, when the product is limited to one, the error message is shown in german in the site and the title of the browser.

**Expected behavior**
If a customer uses the shop in english, this error message should be in english. 

**Screenshots**
error_german

**Additional context**
We use pretix 7.4.1 in Docker.
The product has a max limit of 1 but two variants.
Because of that it is possible to order more than one.

wiffbi:
Currently cart.py uses gettext to localize its error-messages on process-startup. This is not a problem as long as english is the default locale on startup as during code execution the error-messages get run through `gettext` again. If default locale is not english, this then creates static error messages in that locale for every language. This PR fixes #2519 

This PR actually fixes two things:

1. All pre-defined error messages in `services/cart.py` and `services/order.py` were translated during startup time, which is only a problem if you happen to have a default-locale other than `en` in your pretix.cfg. This PR now changes all pre-defined error message to use the lazy-versions of gettext.
2. These lazy-versions then introduce a different problem, when using pretix with celery (the usual dev-setup does not include celery, so not easy to catch at first): the lazy_string are handed to celery as is and then in celery are converted to proper strings, which causes the translation process to use the current language in celery (usually the default-locale) which very well might differ from the event’s locale. This does not happen on all error messages though as some error messages are converted to string before they are handed to celery by applying string-formatting – to be more precise, they are converted before they are even handed to `CartError`/`OrderError`. To make sure all error messages are converted to strings in the currently needed locale, `CartError` and `OrderError` force convert the `msg` to `str()`.

And last, but not least, this PR adds proper i18n pluralization to the error messages. :)

### DIFF
--- a/src/pretix/base/services/cart.py
+++ b/src/pretix/base/services/cart.py
@@ -43,7 +43,9 @@ from django.db import DatabaseError, transaction
 from django.db.models import Count, Exists, IntegerField, OuterRef, Q, Value
 from django.dispatch import receiver
 from django.utils.timezone import make_aware, now
-from django.utils.translation import gettext as _, gettext_lazy, ngettext_lazy, pgettext_lazy
+from django.utils.translation import (
+    gettext as _, gettext_lazy, ngettext_lazy, pgettext_lazy,
+)
 from django_scopes import scopes_disabled
 
 from pretix.base.channels import get_all_sales_channels
--- a/src/pretix/base/services/cart.py
+++ b/src/pretix/base/services/cart.py
@@ -43,7 +43,7 @@ from django.db import DatabaseError, transaction
 from django.db.models import Count, Exists, IntegerField, OuterRef, Q, Value
 from django.dispatch import receiver
 from django.utils.timezone import make_aware, now
-from django.utils.translation import gettext_lazy as _, pgettext_lazy
+from django.utils.translation import gettext as _, gettext_lazy, ngettext_lazy, pgettext_lazy
 from django_scopes import scopes_disabled
 
 from pretix.base.channels import get_all_sales_channels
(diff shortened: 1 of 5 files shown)


raphaelm:
I have this vague memory that I built it this way (using `gettext` and another call of `gettext` later again) on purpose, since we pass these error messages around through celery messages and they might need to be proper strings because lazy strings can't be serialized properly?

But it seems to date back to 2015 and maybe things changed since then. Have you tested this still works properly with celery enabled and error messages that pass through celery?

There's also one failing test which can be fixed by adjusting the test, I guess.

wiffbi:
Regarding celery IIRC it worked, but I will double check. Interestingly enough, one error message is already lazy: `'subevent_required': pgettext_lazy('subevent', 'No date was specified.'),` I think gettext would be fine if the error message were only run through gettext when the errors are raised – the problem is, not all error messages are run through gettext when raised. And these error messages are used in bunch of places ...

wiffbi:
It works with celery – I checked with cart and max-items. While searching I stumbled upon error_messages in `services/orders.py` that have the same issue as they follow the same pattern as in cart.py. I will update that as well. And while on it, I will fix the i18n pluralization as well.

raphaelm:
Okay. I'm a *little* cautious/worried, as I strongly remember playing around with this for hours at some point before everything worked, and it seems like it was a specific choice not to be lazy here, e.g. see https://github.com/pretix/pretix/commit/8ee606f0 

But if everything works, do what you want to do, then we merge it and if there are any issues we can always roll back the commit quickly. If any issues occur, they shouldn't have major consequences.

wiffbi:
I guess then I just make these error-messages use `gettext_lazy` explicitly instead of making `gettext_lazy as _`. Honestly, localization of these error messages is quite „distributed“: they run through gettext on definition, then sometimes when they are raised (handed to CartError as string) and then finally inside `CartError::__init__`. I will look into this and try to clean this up a bit. Regarding https://github.com/pretix/pretix/commit/8ee606f0 it makes sense to use `gettext` for the email subject here as with the lazy version, it would be converted to string/localized during the async mail-send, which probably has the default-locale active, which might not be the same as the event-locale.

wiffbi:
Small update: the „normal“ lazy messages already worked, I ran localegen and localecompile and translated some of the pluralization-messages and tested with those as well. Works as expected, so IMHO good to go. Happy for reviews nonetheless. :)

raphaelm (approved):

PR #3056, opened 2023-01-24
Changing orders: Default to not notifying the user

raphaelm:
A customer requested this and it's kinda a breaking change, but I'm inclined to agree it's a good change. It's so easy to accidentally keep this box checked if you really don't want to notify your customer.


pc-coholic (approved):
I am very okay with this change.

PR #3058, opened 2023-01-25
Always make explicit which tables to lock

raphaelm:


src/pretix/base/services/checkin.py
```diff
@@ -729,8 +730,11 @@ def perform_checkin(op: OrderPosition, clist: CheckinList, given_answers: dict,
         _save_answers(op, answers, given_answers)
 
     with transaction.atomic():
-        # Lock order positions
-        op = OrderPosition.all.select_for_update().get(pk=op.pk)
+        # Lock order positions, if it is an entry. We don't need it for exists, as a race condition wouldn't be problematic
```
wiffbi, line 733:
```suggestion
        # Lock order positions, if it is an entry. We don't need it for exits, as a race condition wouldn't be problematic
```

PR #3059, opened 2023-01-25
Shop header: Fix logo link outline to contain image (Z#23115320)

wiffbi:
Currently when navigating with tab or after a click on the logo-link, the focus-outline is just one line-height high due to being `display: inline`. THis PR fixes the outline to contain the logo image but at the same time keeping text-only logo/page-header to stay on the same line by using `display: inline-block` instead of `display: block`, which would cause the text-only logo to break the line before displaying the date.

There is one edge-case where the layout then differs from the current behaviour: when there is not header-image, but the event’s title is used and the title is veeeeery long and has a line break in itself, then the following date will be on a new line.

Old behaviour:

Bildschirm­foto 2023-01-25 um 15 19 20

New behaviour:
Bildschirm­foto 2023-01-25 um 15 19 05

On „normal“ length event titles it looks the same as before.
Bildschirm­foto 2023-01-25 um 15 20 40

Another approach might be to use the „new“ `:has` selector to only target `.page-header a:has(img)`. That way we always have the same look as before, but older browser (is in current Firefox!?) with no support for `:has` will still show the focus-outline too small. Not sure what is the best tradeoff here.

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/base.html
+++ b/src/pretix/presale/templates/pretixpresale/event/base.html
@@ -66,7 +66,7 @@
     </div>
 {% endblock %}
 {% block page %}
-    <div class="page-header {% if event_logo and event_logo_image_large %}logo-large{% endif %}">
+    <div class="page-header{% if event_logo %} pager-header-with-logo{% endif %}{% if event_logo and event_logo_image_large %} logo-large{% endif %}">
         <div class="{% if not event_logo or not event_logo_image_large %}pull-left flip{% endif %}">
             {% if event_logo and event_logo_image_large %}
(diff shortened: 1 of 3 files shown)


raphaelm:
> Another approach might be to use the „new“ :has selector to only target .page-header a:has(img). 

We could just set a class when there is an image?

I'd prefer a conditional solution that does not change other cases. The edge-case you are citing is a pretty common one, I think.

wiffbi:
Yes, probably the best idea to not only change the CSS but to add a class to the containing link if there is an image. I will add this!

wiffbi:
Changed it to an extra CSS-class and added on the organizer page as well.

raphaelm (approved):

PR #3061, opened 2023-01-25
Ensure total ordering of paginated lists

raphaelm:
As a "funny" coincidence, I have just learned about this topic a few days ago [on the django-developers mailing list](https://groups.google.com/g/django-developers/c/fVZhDuiVfm8) – and today a customer ran into this for the first (documented) time.

We have a lot of paginated lists, both in the backend as well as the API. For paginated lists, ordering of the list obviously matters a lot – and we're doing it wrong in many cases. Whenever we are ordering by a field that is not guaranteed to be unique within the list, we're not dealing with a [total order](https://en.wikipedia.org/wiki/Total_order) and pagination might lead to inconsistent results being returned. See also [this](https://github.com/carltongibson/django-filter/issues/862).

This is especially problematic with user-chosen ordering, but sometimes also with our default ordering (e.g. creation date of an order is *almost* unique with microsecond-precision, but not *guaranteed* to be unique).

This PR fixes a few default orderings, and implements the solution from django-admin for the user-defined orderings: Whenever the ordering is not known to be unique, we add an order by `pk`.

There is one **major** disadvantage: This might make a few things **considerably** slower, as index usage will change. As an example, I just tested the query that the "order list" in the backend is doing, for an event with 100k+ orders. Before the change, we have a query like

```
SELECT … FROM "pretixbase_order" WHERE … ORDER BY "pretixbase_order"."datetime" DESC LIMIT 25
```

and the query is really quick because it does an index scan on the datetime index:

```
                                                                                       QUERY PLAN                                                                                       
----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=1.28..192.04 rows=25 width=513) (actual time=0.111..1.620 rows=25 loops=1)
   ->  Nested Loop Left Join  (cost=1.28..659591.30 rows=86444 width=513) (actual time=0.110..1.616 rows=25 loops=1)
         ->  Nested Loop  (cost=0.85..249435.50 rows=86444 width=398) (actual time=0.104..1.529 rows=25 loops=1)
               ->  Index Scan Backward using pretix_addidx_order_event_date on pretixbase_order  (cost=0.56..248346.64 rows=86444 width=398) (actual time=0.094..1.503 rows=25 loops=1)
                     Index Cond: (event_id = X)
               ->  Materialize  (cost=0.29..8.31 rows=1 width=4) (actual time=0.000..0.001 rows=1 loops=25)
                     ->  Index Scan using pretixbase_event_pkey on pretixbase_event  (cost=0.29..8.31 rows=1 width=4) (actual time=0.008..0.009 rows=1 loops=1)
                           Index Cond: (id = X)
                           Filter: (organizer_id = Y)
         ->  Index Scan using pretixbase_invoiceaddress_order_id_key on pretixbase_invoiceaddress  (cost=0.43..4.74 rows=1 width=115) (actual time=0.003..0.003 rows=1 loops=25)
               Index Cond: (pretixbase_order.id = order_id)
 Planning Time: 0.488 ms
 Execution Time: 1.671 ms
(13 rows)
```

However, with this change the query would be

```
SELECT … FROM "pretixbase_order" WHERE … ORDER BY "pretixbase_order"."datetime" DESC, "pretixbase_order"."id" DESC LIMIT 25
```

which now is 200x slower as PostgreSQL is not using the index to sort the list:

```
                                                                                                 QUERY PLAN                                                                                                 
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Limit  (cost=358384.17..358384.23 rows=25 width=513) (actual time=234.713..301.227 rows=25 loops=1)
   ->  Sort  (cost=358384.17..358600.28 rows=86444 width=513) (actual time=234.713..301.224 rows=25 loops=1)
         Sort Key: pretixbase_order.datetime DESC, pretixbase_order.id DESC
         Sort Method: top-N heapsort  Memory: 102kB
         ->  Nested Loop  (cost=1001.28..355944.78 rows=86444 width=513) (actual time=0.323..254.863 rows=95028 loops=1)
               ->  Index Scan using pretixbase_event_pkey on pretixbase_event  (cost=0.29..8.31 rows=1 width=4) (actual time=0.009..0.011 rows=1 loops=1)
                     Index Cond: (id = X)
                     Filter: (organizer_id = Y)
               ->  Gather  (cost=1000.99..355072.03 rows=86444 width=513) (actual time=0.313..228.158 rows=95028 loops=1)
                     Workers Planned: 2
                     Workers Launched: 2
                     ->  Nested Loop Left Join  (cost=0.99..345427.63 rows=36018 width=513) (actual time=0.103..178.888 rows=31676 loops=3)
                           ->  Parallel Index Scan using pretixbase_order_event_id_fcc6f9c6 on pretixbase_order  (cost=0.56..174530.96 rows=36018 width=398) (actual time=0.036..91.564 rows=31676 loops=3)
                                 Index Cond: (event_id = X)
                           ->  Index Scan using pretixbase_invoiceaddress_order_id_key on pretixbase_invoiceaddress  (cost=0.43..4.74 rows=1 width=115) (actual time=0.002..0.002 rows=1 loops=95028)
                                 Index Cond: (pretixbase_order.id = order_id)
 Planning Time: 0.309 ms
 Execution Time: 301.275 ms
(18 rows)
```

Now, in this example, the result is still okayish, but not great. Of course, we can solve it by creating a new index on ``(datetime, id)`` – but that index will likely take 2GB+ of disk space today.

Still, I don't see a way around this, if we want to output correct data. But we might need to deploy this carefully and watch for performance issues.


src/pretix/control/views/item.py
```diff
@@ -820,7 +821,9 @@ def get_queryset(self):
         }
 
         if self.request.GET.get("ordering", "-date") in valid_orders:
-            qs = qs.order_by(*valid_orders[self.request.GET.get("ordering", "-date")])
+            qs = qs.order_by(*get_deterministic_ordering(Quota, valid_orders[self.request.GET.get("ordering", "-date")]))
```
wiffbi, line 824: Should we instead of using `get_deterministic_ordering` just define the proper order-tuples in `valid_orders` just above this code section? I.e.:

```
        valid_orders = {
            '-date': ('-subevent__date_from', 'name', 'pk'),
            'date': ('subevent__date_from', '-name', 'pk'),
            'size': ('size', 'name', 'pk'),
            '-size': ('-size', '-name', 'pk'),
            'name': ('name', 'pk'),
            '-name': ('-name', 'pk'),
        }
```

src/pretix/control/views/organizer.py
```diff
@@ -2016,6 +2016,8 @@ class LogView(OrganizerPermissionRequiredMixin, PaginationMixin, ListView):
     context_object_name = 'logs'
 
     def get_queryset(self):
+        # technically, we'd also need to sort by pk since this is a paginated list, but in this case we just can't
+        # bear the performance cost
```
wiffbi, line 2020: What about replacing the datetime-index with (datetime, id)-index? Would that improve performance? AFAIK a pure datetime-based access would/could still use the (datetime, id)-index.

src/pretix/control/views/event.py
```diff
@@ -1040,6 +1040,8 @@ class EventLog(EventPermissionRequiredMixin, PaginationMixin, ListView):
     context_object_name = 'logs'
 
     def get_queryset(self):
+        # technically, we'd also need to sort by pk since this is a paginated list, but in this case we just can't
+        # bear the performance cost
```
wiffbi, line 1044: See comment below regarding replacing the datetime-index with a (datetime, id)-index.

wiffbi (commented):
Looks good! I found `filter_backends` for `WaitingListViewSet` in api/views/waitinglist.py to still use the normal `OrderingFilter`. Is this on purpose? Ordering is by `created`, so it might be affected as well …

What about the changes to indexes in the DB? We talked about changing datetime-indexes to be changed to (datetime, pk). Is this something this PR should address as well?

raphaelm:
Updated according to review. We need to deploy this carefully in multiple steps on our production system since it might take a long time to create the new indexes.

SQL Before deploy:

```
CREATE INDEX CONCURRENTLY IF NOT EXISTS "pretixbase_logentry_datetime_id_5a819491_idx" ON "pretixbase_logentry" ("datetime", "id");
CREATE INDEX CONCURRENTLY IF NOT EXISTS "pretixbase_order_datetime_id_fdbdc55e_idx" ON "pretixbase_order" ("datetime", "id");
CREATE INDEX CONCURRENTLY IF NOT EXISTS "pretixbase_order_last_modified_id_0a993acf_idx" ON "pretixbase_order" ("last_modified", "id");
CREATE INDEX CONCURRENTLY IF NOT EXISTS "pretixbase_transaction_datetime_id_bde8c131_idx" ON "pretixbase_transaction" ("datetime", "id");
INSERT INTO django_migrations (app, name, applied) VALUES ('pretixbase', '0229_auto_20230201_1058', NOW());
```

SQL after deploy, carefully

```
BEGIN;
DROP INDEX IF EXISTS "pretixbase_logentry_datetime_60ff2d52";
DROP INDEX IF EXISTS "pretixbase_order_datetime_b588c1be";
DROP INDEX IF EXISTS "pretixbase_order_last_modified_30e4a7b2";
DROP INDEX IF EXISTS "pretixbase_transaction_datetime_311e3dd5";
COMMIT;
```

PR #3082, opened 2023-02-02
Export: Use YYYY-MM-DD format for check-in CSV-export (Z#23115722)

wiffbi:


wiffbi:
I like the idea of using a native date format in excel! Regarding localized dates in CSV, I think machine readable dates would also help here, as CSV lends itself to automated machine processing even more than xls. So I think I am more in the camp „no localized dates at all“ in data exports (except maybe PDF?). I think if we have localized dates with native date format in excel, that would make it the readable export and CSV with YYYY-MM-DD would make it easier to process with automation.

Regarding this PR: currently in this export, order date is formatted as YYYY-MM-DD – so differs from check-in, check-out and subevent.date.

PR #3091, opened 2023-02-06
Use local name in control variation list heading

pajowu:
Currently the first translation of the variation is shown as the variation name in the variation list. This commits changes this to prefer the variation name in the language that the control interface is set to.


raphaelm:
@cla-bot[bot] check

raphaelm (approved):

raphaelm:
Thanks! :)

PR #3096, opened 2023-02-07
Docs: Add comment to webhook API

ser8phin:
Extend the API documentation with info and examples for the new webhooks comment field


pc-coholic (changes requested):
Plain JSON does not like trailing commas, so the new lines should all be `"comment": null`.

pc-coholic (approved):

PR #3098, opened 2023-02-07
Cart: More compact spacing

wiffbi:
As discussed, this PR makes the product-listing in cart more compact.


raphaelm:
As discussed in Slack, we'll hold off for now

PR #3104, opened 2023-02-08
Attendee ticket page: Allow to hide prices

raphaelm:


src/pretix/static/pretixpresale/scss/_cart.scss
```diff
@@ -103,7 +109,7 @@
 .cart .firstchild-in-panel .cart-row:first-child {
     padding-top: 0;
 }
-.cart [role=rowgroup]:last-child {
+.cart .rowgroup-total:last-child {
```
wiffbi, line 112: I think this an improvement to the previous solution which applied a border to any last rowgroup (of course, it would have needed an update to make sure the last rowgroup is not the first one as well). But what we really want, is a line before the total. I just suggest a slight name change.

```suggestion
.cart .cart-rowgroup-total {
```

wiffbi (changes requested):
Looks good, just a minor typo regarding role="rowgroup-total" not being a role but a CSS-class.

PR #3022, opened 2023-01-10
New check-in features

raphaelm:
- [x] Data model
- [x] REST API
  - [x] variation.checkin_attention
  - [x] order.valid_if_pending
  - [x] orderposition.blocked
  - [x] orderposition.valid_from
  - [x] orderposition.valid_until
- [x] Show state in order view
  - [x] frontend
	  - [x] orderposition.blocked
	  - [x] orderposition.valid_from
	  - [x] orderposition.valid_until
  - [x] backend
	  - [x] order.valid_if_pending
	  - [x] orderposition.blocked
	  - [x] orderposition.valid_from
	  - [x] orderposition.valid_until
- [x] Allow to modify state when changing orders
  - [x] order.valid_if_pending
  - [x] orderposition.blocked
  - [x] orderposition.valid_from
  - [x] orderposition.valid_until
- [x] Support new fields in order import
  - [x] orderposition.valid_from
  - [x] orderposition.valid_until
- [x] Add new fields to any exports
  - [x] orderposition.blocked
  - [x] orderposition.valid_from
  - [x] orderposition.valid_until
- [x] Update offline barcodes
- [x] Allow plugins to provide a reason for a block
- [x] Update revocation mechanism
- [x] Adjust check-in logic
  - [x] variation.checkin_attention
  - [x] order.valid_if_pending
  - [x] orderposition.blocked
  - [x] orderposition.valid_from
  - [x] orderposition.valid_until
  - [x] tests
- [x] Web-checkin queries
- [x] Check-in algorithm documentation
  - [x] order.valid_if_pending
  - [x] orderposition.blocked
  - [x] orderposition.valid_from
  - [x] orderposition.valid_until
- [ ] Find other places that need to respect valid_if_pending
  - [x] block expiry of orders
  - [x] checkin list in backend
  - [x] ticket downloads
  - [x] emails?
  - [x] expiry warnings?
  - [ ] plugins?
	  - [ ] expoip
	  - [ ] venueless
	  - [ ] scoocs
	  - [ ] digital
- [x] new logdisplay lines
- [ ] Adjust SCAN apps
  - [x] Android
  - [ ] iOS
  - [x] Desktop
  - [x] Proxy

PR #3105, opened 2023-02-08
Allow to define ticket validity through a product

raphaelm:
- [x] Data model
- [x] Rebase against master after merge of #3022
- [x] Config UI
- [x] Date-time logic
	- [x] Tests
- [x] API
  - [x] Items
    - [x] Tests
    - [x] Docs
  - [x] Order creation
    - [x] Tests
    - [x] Docs
- [x] Order creation
- [x] Customer-specified start date
	- [x] presale
		- [x] tests
	- [x] POS?


pc-coholic (changes requested):
- While the Date-part of the validity-picker correctly inhibits setting an earlier `valid_to` than `valid_from`, the same does not apply to the Time-Picker. Setting a validity of "15.02.2023 14:30:00 - 15.02.2023 14:00:00" is possible.
- Duplicating a product will not duplicate validity.
- I might be miscalculating; but just to be sure: Dynamic validity, +1 month, +2 days, +3 hours, +4 minutes.
  - Start: 13.02.2023 12:20
  - +1 month: 13.03.2023 12:20
  - +2 days: 14.03.2023 00:00 --> 15.03.2023 00:00
  - +3 hours: 15.03.2023 03:00
  - +4 minutes: 15.03.2023 03:04
  - Actual result in pretix: 16.03.2023 03:03
  - I suspect that I miscalculated the "rollover to next midnight"...
  - The missing minute is apparently a precision issue since in the order overview we get minute precision but it's actually handled with second precision (It's actually saved as 03:03:59). Perhaps we might need some rounding for minute precision displays?
- The `current value` column in the order change backend displays the `valid_from` for both `from` and `until`.
- The `current value` column in the order change backend does not display the `valid_until` time if only a `valid_until` is set.
- The `current value` column in the order change backend displays the validity times only with minute precision.

raphaelm:
> While the Date-part of the validity-picker correctly inhibits setting an earlier valid_to than valid_from, the same does not apply to the Time-Picker. Setting a validity of "15.02.2023 14:30:00 - 15.02.2023 14:00:00" is possible.

Yeah, same as for any date pickers in pretix (such as event start and end date). Just to hard to fix without ending up with real weird UX. I've added a server-side validation now.

> Duplicating a product will not duplicate validity.

Good catch! Fixed.

>    I might be miscalculating; but just to be sure: Dynamic validity, +1 month, +2 days, +3 hours, +4 minutes.
>        Start: 13.02.2023 12:20
>        +1 month: 13.03.2023 12:20
>        +2 days: 14.03.2023 00:00 --> 15.03.2023 00:00
>        +3 hours: 15.03.2023 03:00
>        +4 minutes: 15.03.2023 03:04
>        Actual result in pretix: 16.03.2023 03:03
>        I suspect that I miscalculated the "rollover to next midnight"...

The misunderstanding is in how "+1 month" or "+1 days" works. Every calculation that involves days or months will always end at midnight. So "13.02.2023 12:00 + 1 month" will be "13.03.2023 23:59", and "13.02.2023 12:00 + 1 day" will be "13.02.2023 23:59".

The reasoning behind that logic is that most existing "day passes" or "month passes" work on full days. And if you really need a 24h pass or a 72h pass, you can do that using the hours field.

In your example, the calculation is therefore:

- Start: 13.02.2023 12:20
- +1 month: 13.03.2023 **23:59**
- +2 days: 15.03.2023 **23:59**
- +3 hours: **16.03.2023** **02:59**
- +4 minutes: **16.03.2023** **03:03**

>        The missing minute is apparently a precision issue since in the order overview we get minute precision but it's actually handled with second precision (It's actually saved as 03:03:59). Perhaps we might need some rounding for minute precision displays?

I actually think it's best this way for real-world use cases:

1. I think showing only minute precision makes sense, because second precision is just too much information to be useful.
2. You only end up with the :59 times if you use day- or month-based math. So you'd usually end up with a validity like "13.02.2023 13:37 – 23:59". I believe this will be pretty well-understood by humans as "until end of day", even if we don't spell "23:59:59" out. If we round up, it would show "13.02.2023 13:37 – 14.02.2023 00:00", which, while mathematically correct, would be harder to read and understand.
3. The only (sensible) use case to *combine* day/month-based math with minute/hour-based math that I can come up with is something like "we're a night club, our day rollover is at 06:00 instead of midnight". In this case, showing 05:59 will still be absolutely understandable for everyone to mean "until just before 06:00".

Happy to be convinced otherwise if you have a good real-world use case where it is important that we show 03:04 instead of 03:03.

>    The current value column in the order change backend displays the valid_from for both from and until.
>    The current value column in the order change backend does not display the valid_until time if only a valid_until is set.

Fixed. (It's the same bug)

>    The current value column in the order change backend displays the validity times only with minute precision.

See above.

PR #3113, opened 2023-02-10
Doc: Upgrade to Sphinx 6

raphaelm:


pc-coholic (approved):
LGTM.

Only thing that looks very different now seems to be the big comparison-table in `/user/events/ticket_secrets.html`. But I can live with that.

PR #3119, opened 2023-02-14
Add OrderPosition.ignore_from_quota_while_blocked

raphaelm:


src/pretix/base/services/orders.py
```diff
@@ -2278,7 +2280,9 @@ def _perform_operations(self):
                     op.position.blocked = [b for b in op.position.blocked if b != op.block_name]
                     if not op.position.blocked:
                         op.position.blocked = None
-                    op.position.save(update_fields=['blocked'])
+                    if op.ignore_from_quota_while_blocked:
+                        op.position.ignore_from_quota_while_blocked = op.ignore_from_quota_while_blocked
```
wiffbi, line 2284: As we remove the block anyways, does that make any sense here? The question is, whether to (always?) set `op.position.ignore_from_quota_while_blocked = False` when removing a block? Otherwise, what if we block it with ignore_quota, then unblock it again and then want to block it, but without ignore_quota (very theoretical problem, but nonetheless)?

raphaelm: Did you notice the difference between op.position.ignore... and op.ignore...? `op` here, confusingly, is short for operation, not orderposition. Callers of remove_block() can opt in to this behaviour.

wiffbi: Yes, I stumbled upon this `op` being operation and not orderposition and thought of commenting on it – but the whole huge code-block uses op so I left it intouched. The question though is, how do I set the `ignore_from_quota_while_blocked` on an orderposition once it has been set to True? If I pass `op.ignore_from_quota_while_blocked` as `False`, it will never be set on the orderposition. So again, if I once set the block with `op.ignore_from_quota_while_blocked` to `True`, then – as e.g. the ticketswap somehow later failed – remove the block with `op.ignore_from_quota_while_blocked == False`, `op.position.ignore_from_quota_while_blocked` will still be `True`. So if something fishy is happening later to the ticket and just want to block it generally, I add_block with `op.ignore_from_quota_while_blocked` to `False` (because I maybe still want the ticket to count against the quota), then `op.position.ignore_from_quota_while_blocked` will still be `True` – which I would not expect. I know, this is rather theoretical as one usually wants to block and not count it to quota, but if that feature/option is there, it should work as expected.

wiffbi: So TLDR;: AFAICT from the code, I can remove a block, but I can never set `orderposition.ignore_from_quota_while_blocked` to `False` once it has been set to `True`.

raphaelm: Okay, got it, will likely replace it with a `is None` check

PR #3121, opened 2023-02-14
API: Expose organizer and event URL (Z#23116269)

pc-coholic:


raphaelm (commented):
Should we maybe name the field `public_url` instead of URL to not confuse it with backend or API URLs?

That said, looks fine, modulo docs and tests.

raphaelm (approved):

PR #3135, opened 2023-02-27
Product list: Indent variations, improve alignment of price and quantity

wiffbi:


src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
```diff
@@ -17,7 +17,7 @@ <h3 id="category-{{ tup.0.id }}">{{ tup.0.name }}</h3>
             {% if item.has_variations %}
                 <article aria-labelledby="item-{{ item.pk }}-legend"{% if item.description %} aria-describedby="item-{{ item.pk }}-description"{% endif %} class="item-with-variations{% if event.settings.show_variations_expanded %} details-open{% endif %}" id="item-{{ item.pk }}">
                     <div class="row product-row headline">
-                        <div class="col-md-8 col-xs-12">
```
raphaelm, line 20: Do we need the same changes from fragment_product_list.html in voucher.html as well? Probably.

raphaelm (changes requested):
I think you need to check what happens if a variation has e.g. a description, it currently looks like this:

Screenshot 2023-03-02 at 14-12-49 Democon mit langem Eventnamen

It probably should look like this:

Screenshot 2023-03-02 at 14-13-51 Democon mit langem Eventnamen_2

At least on mobile, the price is intended:

Screenshot 2023-03-02 at 14-14-31 Democon mit langem Eventnamen_3

src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
wiffbi: Good catch, I fixed the indent for description, etc. and added `col-sm-` classes to voucher.html and fragment_availability.html. I also fixed an alignment issue when free price input was used.

PR #3142, opened 2023-03-02
Event series calendar: Allow to hide all past events

raphaelm:


wiffbi (commented):
Looks good. I still want to test it thoroughly though on Monday.

Do you think it might be good to instead of a boolean use a relative date? „Now“ might be a bit quick for a hard cutoff.

raphaelm:
> Do you think it might be good to instead of a boolean use a relative date? „Now“ might be a bit quick for a hard cutoff.

I don't think the additional configuration complexity is necessary / can't think of a use case where you'd really need this

raphaelm:
Yeah, let's not make it this complex for now ;) If you'd want the last 6 weeks to be visible, you could just… not use the new option and let the month calendar do its thing. You'll let me know once you're done testing :)

wiffbi (approved):
Looks good! I actually like, that it cuts the calendar off beginning at the current date. Only thing is the month-select dropdown, that it allows for past months to be selected, which then of course are rendered empty (even no calendar-table with empty cells). Not a big problem though, so not sure, that changing it is worth the effort.

PR #3143, opened 2023-03-02
Voucher bulk creation: Add markdown preview for email field

raphaelm:


pc-coholic (changes requested):
- When pasting merge-fields, that are body-exclusive (such as `{voucher_list}` or `{voucher_url_list}`), the preview does render those properly.
- However the render of the subject then also suddenly contains newlines
- But it doesn't probably matter, since the form won't queue the messages with these merge-fields.
- However: Once the Subject form-fields is error-marked after submit, the previews (Subject and Body) won't display anything anymore.

raphaelm:
Fixed!

PR #3146, opened 2023-03-03
Allow to set privacy policy URL per language

raphaelm:


wiffbi (approved):
Looks goood and works as intended. One minor super edgy edge case is for deployment: when I edit my settings with the old single-language URL input, in the background pretix is updated to multi-lang privacy-URL, then hit save, the URL will not be saved. I think that is negligible though.

If we introduce multi-lang URLFields, what about the imprint-URL and footer-links?

raphaelm:
> Looks goood and works as intended. One minor super edgy edge case is for deployment: when I edit my settings with the old single-language URL input, in the background pretix is updated to multi-lang privacy-URL, then hit save, the URL will not be saved. I think that is negligible though.

Agreed

> If we introduce multi-lang URLFields, what about the imprint-URL and footer-links?

One at a time ;)

PR #3150, opened 2023-03-06
Allow attendees to change selected add-ons of same price

raphaelm:


src/pretix/base/models/orders.py
```diff
@@ -2555,6 +2555,27 @@ def resend_link(self, user=None, auth=None):
...
+    @property
+    @scopes_disabled()
+    def user_change_allowed(self) -> bool:
+        """
+        Returns whether or not this order can be changed by the attendee.
+        """
+        from .items import ItemAddOn
+
+        if not self.event.settings.change_allow_attendee or not self.order.user_change_allowed:
+            return False
```
wiffbi, line 2567: I am undecided regarding the name. Shouldn’t it be called `attendee_change_allowed`? Currently it does not make a difference as this is only used on the attendee-order-page, but what if – at some point – on the order-overview page we realize, that we need to decide on a per OrderPosition-basis, that an OrderPosition can be changed or not even by the buyer? We would need `user_change_allowed` on an OrderPosition to have context whether it is on an attendee-page or the buyers order-page.

Wouldn’t it make sense to make two `@property` like so:

```suggestion
    @property
    @scopes_disabled()
    def attendee_change_allowed(self) -> bool:
        if not self.event.settings.change_allow_attendee:
            return false
        return self.user_change_allowed

    @property
    @scopes_disabled()
    def user_change_allowed(self) -> bool:
        """
        Returns whether or not this order can be changed by the attendee.
        """
        from .items import ItemAddOn

        if not self.order.user_change_allowed:
            return False
```

src/pretix/presale/views/order.py
```diff
@@ -371,6 +371,7 @@ def get_context_data(self, **kwargs):
             order=self.order
         )
         ctx['tickets_with_download'] = [p for p in ctx['cart']['positions'] if p.generate_ticket]
+        ctx['user_change_allowed'] = self.position.user_change_allowed
```
wiffbi, line 374: See comment above. As `OrderPositionDetails` is used on a per-ticket context of a user being an attendee, I’d change this to 

```suggestion
        ctx['user_change_allowed'] = self.position.attendee_change_allowed
```

One could keep `ctx['user_change_allowed']` as that way, the template `pretixpresale/event/position.html` could be reused.

wiffbi (commented):
Looks good, other than a minor naming thing I stumbled upon that on an OrderPosition the prop is called `user_change_allowed` (the same as on an order), but basically does a check whether an attendee can change that OrderPosition. Of course, „user“ is a broad term and can mean customer/buyer and/or attendee, but that then requires to know the context, in which `user_change_allowed` is called on an OrderPosition, which currently is only in the context of an attendee-page. What to expect when in the future we want to check each OrderPosition on the buyers page, whether it could be changed?

ser8phin (approved):
I concur with the suggestion by @wiffbi that attendee_change_allowed is a bit more fitting here, otherwise :+1:

raphaelm:
I renamed the property, you are right, but I did the change a bit differently because I don't think we should have `OrderPosition.user_change_allowed` in that case since it would not be used anywhere and could only cause confusion

PR #3160, opened 2023-03-14
Order-exports: localize salutation

wiffbi:


raphaelm (approved):

raphaelm:
Yeah, that makes sense. We should make sure it's till importable from email.py, since plugins (like pretix-exhibitors) use it, but that should be no problem as email.py needs to import it anyways.

PR #3174, opened 2023-03-21
Widget: add role=button to close element 

### LINKED ISSUE #3171: Widget: close button should be a <button> not a link
### Problem and impact

The close button in the widget is an `<a>` element, which is semantically wrong and not ideal especially regarding screen-readers.

### Expected behaviour

Semantically the close button in the widget should be a `<button>` instead of an `<a>` element.

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Link

_No response_

_No response_

_No response_

### Version

_No response_

wiffbi:
Closes #3171

### DIFF
--- a/src/pretix/static/pretixpresale/js/widget/widget.js
+++ b/src/pretix/static/pretixpresale/js/widget/widget.js
@@ -705,9 +705,9 @@ var shared_iframe_fragment = (
     + '        referrerpolicy="origin">'
     + 'Please enable frames in your browser!'
     + '</iframe>'
-    + '<div class="pretix-widget-frame-close"><button type="button" @click.prevent.stop="close" aria-label="'+strings.close+'">'
+    + '<div class="pretix-widget-frame-close"><a href="#" @click.prevent.stop="close">'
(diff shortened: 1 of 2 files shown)


raphaelm:
I hope we don't break third-party styling of the customers with this, but I see no way to prevent it, really.

wiffbi:
That is what I hope to mitigate by setting most/all usual CSS properties, so the close button is basically . Not sure I forgot some though. And hopefully nobody listens to a click event on an `<a>`-element. A safer alternative might be to just set `role="button"` on the `<a>`-element, but that felt like a hack. But maybe the _right_ hack in this case?

raphaelm:
> That is what I hope to mitigate by setting most/all usual CSS properties

That's not what I'm worried about, I'm worried about a customer who deliberately applied custom styling to the close button that will now no longer match. So if role=button is good enough from an a11y point of view, it would be preferable from a compatbility point ov view

wiffbi:
Well, I guess it is not the best way, but probably all things considered the pragmatic approch. So role=button it will be.

PR #3179, opened 2023-03-23
[WIP] Widget: add labels to inputs

wiffbi:
This PR tries to improve a11y of the widget by adding aria-labels and without changing any HTML structure.


src/pretix/static/pretixpresale/js/widget/widget.js
```diff
@@ -468,6 +482,24 @@ Vue.component('item', {
             expanded: this.$root.show_variations_expanded
         };
     },
+    watch: {
+        expanded: function (newValue) {
+            if (newValue) {
+                this.$refs.variations.hidden = false;
+                this.$refs.variations.style.maxHeight = this.$refs.variations.scrollHeight + 'px';
```
raphaelm, line 489: Do we really need this change? Did the 1000px have any negative effect?

This new method is more exact, but also does not account for changes in e.g. the window size that cause a reflow of the variations, which can cause variations to be hidden after you make your window smaller.

wiffbi: The 10.000px caused the max-height animation to be very fast when opening and have a lag when closing (as when closing, it started animating at 10.000 down to 0, which had the effect that the start was not visible).

Good catch regarding resize and reflow. Best would be to remove the maxHeight altogether on transitionsend, that way reflow/resize/DOM-changes to variations would not be an issue.

raphaelm: Is that possible?

wiffbi: Yup, just added it. The only „complicated“ thing is, one needs to set the current scrollHeight as a starting value for the closing animation – and then with some delay set the maxHeight to 0. Vue.nextTick does not work here as it is a direct change on style, not DOM. But a simple window.setTimeout does the trick.

PR #3185, opened 2023-03-27
PDF editor: Small UX improvements

raphaelm:


src/pretix/static/pretixcontrol/js/ui/editor.js
```diff
@@ -943,22 +948,40 @@ var editor = {
...
+
+    _update_save_button() {
+        if ($("#editor-save span").prop("disabled")) {
+            // Currently saving
+            return;
+        }
+        if (editor.dirty || !editor._ever_saved) {
+            $("#editor-save").removeClass("btn-success").addClass("btn-primary").find(".fa").attr("class", "fa fa-fw fa-save");
+        } else {
+            $("#editor-save").addClass("btn-success").removeClass("btn-primary").find(".fa").attr("class", "fa fa-fw fa-check");
```
wiffbi, line 963: Should we update the headline as well? Note this code would update the headline only if the editor was not changed while saving and uses the current value of name … a proper way would be to pass `payload.name` to `editor._update_save_button()` and always set it instead of only setting it when editor is not dirty.
```suggestion
            $("#editor-save").addClass("btn-success").removeClass("btn-primary").find(".fa").attr("class", "fa fa-fw fa-check");
            $("h1 small").text($("#pdf-info-name").val())
```

src/pretix/plugins/ticketoutputpdf/views.py
```diff
@@ -252,8 +252,11 @@ def title(self):
     def save_layout(self):
         self.layout.layout = self.request.POST.get("data")
         self.layout.save(update_fields=['layout'])
+        if "name" in self.request.POST:
+            self.layout.name = self.request.POST.get("name")
+            self.layout.save(update_fields=['name'])
```
wiffbi, line 257: Probably not worth the effort to just call `save` once …
```suggestion
        update_fields = ['layout']
        if "name" in self.request.POST:
            self.layout.name = self.request.POST.get("name")
            update_fields.append('name')
        self.layout.save(update_fields=update_fields)
```

src/pretix/static/pretixcontrol/js/ui/editor.js
raphaelm: I thought about this, the problem is that the template does not just show the ``name`` in that ``<small>`` element, it shows ``title``, which is a property that plugins can override to do something like ``Badge-Layout {name}`` or ``Ticket-Layout {name}``, to show which type of layout this is. I don't think it's worth it replicating that logic (with it's extensibility) on the client side.

src/pretix/plugins/ticketoutputpdf/views.py
raphaelm: yeah that's cleaner

PR #3187, opened 2023-03-27
Check-in API: Extend reach of "force" flag

raphaelm:
When a canceled ticket is scanned offline, and the scanner does not know the ticket is canceled, the scanner will show a green result. If the scan is later uploaded, it is shown as a "failed scan" in the backend. This is incorrect, the scan was successful, even though it should not have been. This PR tries to fix that :)


pc-coholic (approved):
LGTM

PR #3202, opened 2023-04-01
implement customer_accounts_required setting

stapelberg:
Currently, when customers are not logged in, they can still see events, for example on which date(s) the event is scheduled.

When the customer_accounts_required setting is enabled, customers always need to login to view any page of this shop.

This is useful for shops with closed user groups and prevents confusion: We have had a customer not understand how to order, because the “add to cart” button was missing when the customer is not logged in. With this setting, we enforce a login, and hence a functional event view.

### DIFF
--- a/src/pretix/base/settings.py
+++ b/src/pretix/base/settings.py
@@ -147,6 +147,16 @@ DEFAULTS = {
                         "advanced features like memberships.")
         )
     },
+    'customer_accounts_required': {
+        'default': 'False',
+        'type': bool,
+        'form_class': forms.BooleanField,
+        'serializer_class': serializers.BooleanField,
+        'form_kwargs': dict(
+            label=_("Require a customer account to use this shop"),
+            help_text=_("If enabled, all requests are redirected to the login form.")
+        )
+    },
     'customer_accounts_native': {
         'default': 'True',
         'type': bool,
--- a/src/pretix/control/forms/organizer.py
+++ b/src/pretix/control/forms/organizer.py
@@ -358,6 +358,7 @@ class OrganizerSettingsForm(SettingsForm):
     auto_fields = [
         'allowed_restricted_plugins',
         'customer_accounts',
+        'customer_accounts_required',
         'customer_accounts_native',
         'customer_accounts_link_by_email',
         'invoice_regenerate_allowed',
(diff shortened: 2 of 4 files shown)


raphaelm:
Hmm, I'm unsure of we actually want this feature and if so, under what naming.

-  "Customer accounts required" could also mean "you can see the shop but need to create/select an account while creating an order", which is an even more often requested feature which we declined so far since we believe it's not user-friendly and there are little good reasons for it.
- Also, if we want this, we'd possibly want to coordinate it with the settings from #2991 in some way. 
- This implementation is simple, but will break quite a number of things. Quite a lot of payment methods would stop working because they involve requests to event-level pages that don't carry the user's session. Some other plugins would behave very unexpectedly as well.
- With this patch, viewing the details of an existing order will require login, which signals additional security, but it's fake security, since you can be logged in to any customer account, not just the one used to make the order (which is something we want to have af some point probably).

I think a better solution to the described problem would be just showing a more clever error message about why no products are visible and a link to log in?

raphaelm:
Ah I think a simple solution might be to move #2991 forward, setting the guest policy to "forbidden", and adding a message to EventIndex with a suggestion to log in when no products are available and guest orders are forbidden.

stapelberg:
Thanks for taking a look!

> * "Customer accounts required" could also mean "you can see the shop but need to create/select an account while creating an order", which is an even more often requested feature which we declined so far since we believe it's not user-friendly and there are little good reasons for it.

For context: we’re conducting a scientific study. There are numerous requirements from various layers of government (federal law, cantonal ethics board, city project leadership). Our shop is supposed to only be accessible to participants in the study. The fewer public views we have, the better — ideally every view would require login.

> * This implementation is simple, but will break quite a number of things. Quite a lot of payment methods would stop working because they involve requests to event-level pages that don't carry the user's session. Some other plugins would behave very unexpectedly as well.

Thanks for pointing that out. I verified that my payment plugin still works in our installation, even with the patch.

> * With this patch, viewing the details of an existing order will require login, which signals additional security, but it's fake security, since you can be logged in to any customer account, not just the one used to make the order (which is something we want to have af some point probably).

True. There are other places like that, for example the “re-send order link” feature. It currently asks for an email address, even when you’re logged in, and you can enter any user’s email address.

> Ah I think a simple solution might be to move #2991 forward, setting the guest policy to "forbidden", and adding a message to EventIndex with a suggestion to log in when no products are available and guest orders are forbidden.

Yeah, that would also be good. 

But, personally, I’d still feel a lot more comfortable if the entire shop could be restricted to logged-in customers, though.

raphaelm:
I think something like this will happen some day, not sure what yet, but closing this PR because we are drowning a little in PRs. It's still around if we need it.

PR #3211, opened 2023-04-05
ical: add organizer field (displayed by Gmail)

stapelberg:
Without this change, Gmail displays “Unknown Organizer” as the organizer based on the attached calendar event:

before


raphaelm:
We rejected this previously: If we add this, some calendar apps will send "your invitation was accepted" emails to the organizer every time someone adds this to their calendar, which is annoying and not worth it :(

stapelberg:
Thanks for letting me know. We haven’t seen these so far, but I’ll keep an eye out and set up a filter.

Maybe add a comment to the code so that others don’t go down the same rabbithole, though? :)

raphaelm:
Good idea, done!

PR #3214, opened 2023-04-05
Question: Allow limit of string length

raphaelm:


wiffbi (commented):
Looks good, just a comment regarding naming that var/setting. And a few files seem to have slipped into this PR.

PR #3215, opened 2023-04-05
Add option to copy shop URL and generate QR code

raphaelm:
Non-technical users often struggle with finding the URL of their shop or generating a QR code for posters etc.

This adds a simple helper to make it easier:

Screenshot 2023-04-05 at 15-55-26 Vouchermesse pretix eu

Of course, we could make this really complicated with options for the QR code colors and sizes but I think with this small effort we can already do a lot.

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/event/index.html
+++ b/src/pretix/control/templates/pretixcontrol/event/index.html
@@ -16,6 +16,42 @@
             {% endif %}
         </small>
     </h1>
+    <div class="helper-space-below">
+        {% trans "Shop URL:" %}
+        <span id="shop_url" class="text-muted">{% abseventurl request.event "presale:event.index" %}</span>
+        <button type="button" class="btn btn-default btn-xs btn-clipboard" data-clipboard-target="#shop_url">
+            <i class="fa fa-clipboard" aria-hidden="true"></i>
+            <span class="sr-only">{% trans "Copy to clipboard" %}</span>
+        </button>
+        <div class="btn-group helper-display-inline-block">
+            <button type="button" class="btn btn-default btn-xs dropdown-toggle" data-toggle="dropdown" title="{% trans "Create QR code" %}" aria-haspopup="true" aria-expanded="false">
+                <i class="fa fa-qrcode" aria-hidden="true"></i>
+            </button>
+            <ul class="dropdown-menu">
+                <li>
+                    <a href="{% url "control:event.qrcode" event=request.event.slug organizer=request.organizer.slug filetype="png" %}" target="_blank" download>
+                        {% blocktrans with filetype="PNG" %}Download QR-code as {{ filetype }} image{% endblocktrans %}
+                    </a>
+                </li>
+                <li>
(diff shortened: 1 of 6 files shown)


src/pretix/control/templates/pretixcontrol/event/index.html
```diff
@@ -16,6 +16,42 @@ <h1>
             {% endif %}
         </small>
     </h1>
+    <div class="helper-space-below">
+        {% trans "Shop URL:" %}
+        <span id="shop_url" class="text-muted">{% abseventurl request.event "presale:event.index" %}</span>
+        <button type="button" class="btn btn-default btn-xs btn-clipboard" data-clipboard-target="#shop_url">
```
wiffbi, line 22: As this is dependent on JS, add `js-only`?
```suggestion
        <button type="button" class="btn btn-default btn-xs btn-clipboard js-only" data-clipboard-target="#shop_url">
```

src/pretix/control/templates/pretixcontrol/event/index.html
```diff
@@ -16,6 +16,42 @@ <h1>
...
+            <i class="fa fa-clipboard" aria-hidden="true"></i>
+            <span class="sr-only">{% trans "Copy to clipboard" %}</span>
```
wiffbi, line 24: Might be overkill, but probably better to understand.
```suggestion
            <span class="sr-only">{% trans "Copy shop URL to clipboard" %}</span>
```

src/pretix/control/views/event.py
```diff
@@ -1504,3 +1506,35 @@ def formset(self):
...
+            version=1,
+            error_correction=qrcode.constants.ERROR_CORRECT_M,
+            box_size=10,
+            border=4,
+        )
+        qr.add_data(url)
+        qr.make(fit=True)
+
+        if filetype == 'svg':
+            factory = qrcode.image.svg.SvgPathImage
```
wiffbi, line 1527: I do not know why, but importing `qrcode.image.svg` fails for me locally with error 

AttributeError: module 'qrcode.image' has no attribute 'svg'

If I `import qrcode.image.svg` at the top, it works. :/

wiffbi (approved):
Looks good. Just minor comments. 

Two things not part of this PR:

1. On smaller/medium sized screens, the dropdown-menu could cause horizontal overflow of the page when opened.
2. When I hit the copy to clipboard button more than once, after the first time the confirmation title-thingie disappears far quicker.

Also I had an import issue with SVG, which I could not work out why. I update to latest py-qrcode and it works when imported at the top. See comment. But if it works on you machine, maybe something is wrong with my setup.

src/pretix/control/templates/pretixcontrol/event/index.html
raphaelm: Do you think it is worth it? with my text, the same text is used in other places. I tend to try to avoid introducing new strings that can be avoided because translating pretix becomes more impossible every day

wiffbi: Yeah, I do not think it is worth it. That’s why I added the „overkill“-part. Same thought, less strings to translate. But just wanted to make sure. So, leave it like that. :)

PR #3216, opened 2023-04-05
Add webhook for pretix.event.order.payment.confirmed

raphaelm:


wiffbi (approved):
Looks good, `order.log_action` in `OrderPayment.confirm` does the work here.

PR #3222, opened 2023-04-08
Fix for #3130 -- OIDC with Azure AD issues

raphaelm:


vb-03:
It works now!
Tested with regular Azure.
Testing next with Azure B2C which is more costumer-focused, however I'll need a bit more time due to the set-up of the UserInfo endpoint there (in B2C isn't supported by default)

PR #3223, opened 2023-04-10
sendmail/orders: supply event_or_subevent, add event_date_from

stapelberg:
This allows sending emails to customers to remind them of the specific date of the time slot (subevent) they booked, which is very useful when using pretix to schedule appointments.

### DIFF
--- a/src/pretix/base/email.py
+++ b/src/pretix/base/email.py
@@ -505,6 +505,17 @@ def base_placeholders(sender, **kwargs):
             lambda event: str(event.location or ''),
         ),
         SimpleFunctionalMailTextPlaceholder(
+            'event_date_from', ['event_or_subevent'],
+            lambda event_or_subevent: str(date_format(
+                event_or_subevent.date_from,
+                'SHORT_DATETIME_FORMAT'
+            )),
+            lambda event: str(date_format(
+                event.date_from,
+                'SHORT_DATETIME_FORMAT'
+            )),
+        ),
+        SimpleFunctionalMailTextPlaceholder(
             'event_admission_time', ['event_or_subevent'],
             lambda event_or_subevent:
                 date_format(event_or_subevent.date_admission.astimezone(event_or_subevent.timezone), 'TIME_FORMAT')
--- a/src/pretix/control/views/orders.py
+++ b/src/pretix/control/views/orders.py
@@ -2052,7 +2052,11 @@ class OrderSendMail(EventPermissionRequiredMixin, OrderViewMixin, FormView):
         )
         self.preview_output = {}
         with language(order.locale, self.request.event.settings.region):
-            email_context = get_email_context(event=order.event, order=order)
+            email_context = get_email_context(
+                event=order.event,
+                order=order,
+                event_or_subevent=order.subevent or order.event,
+            )
         email_template = LazyI18nString(form.cleaned_data['message'])
         email_subject = format_map(str(form.cleaned_data['subject']), email_context)
         email_content = render_mail(email_template, email_context)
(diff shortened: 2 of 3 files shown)


raphaelm:
Hmm have you tested this? order.subevent does not exist and can't exist as an order can have multiple subevents in it

stapelberg:
> Hmm have you tested this? order.subevent does not exist and can't exist as an order can have multiple subevents in it

I only tested it in the preview view, where it seemed to work (text “komm zu event={event} um {event_date_from}”):

2023-04-10-pretix-placeholder

Now that you ask, I set up the dev SMTP server. When I send out the actual emails, I notice that they don’t have the `{event_date_from}` placeholder replaced :-/

When I change the code in email.py like so:

```patch
diff --git i/src/pretix/base/email.py w/src/pretix/base/email.py
index 2913a8070..bb6619425 100644
--- i/src/pretix/base/email.py
+++ w/src/pretix/base/email.py
@@ -505,9 +505,9 @@ def base_placeholders(sender, **kwargs):
             lambda event: str(event.location or ''),
         ),
-            'event_date_from', ['event_or_subevent'],
-            lambda event_or_subevent: str(date_format(
-                event_or_subevent.date_from,
+            'event_date_from', ['event'],
+            lambda event: str(date_format(
+                event.date_from,
                 'SHORT_DATETIME_FORMAT'
             )),
             lambda event: str(date_format(
```

…then I do see the placeholders replaced, but with the start date of the event itself, not of the subevent. So, I do think that I need the subevent here after all?

We use quotas and memberships to ensure that one shop customer can only book at most one subevent of this event series.

Do you have a tip for how to model what I need? Thanks!

stapelberg:
> `event_or_subevent` will always resolve to `event` for an email that is sent in context of an _order_. It only resolves to `subevent` for an email that is sent specifically in context of an _order position_.

Aha, thanks!

> That's also why every email has a short order summary _including the dates_ at the bottom of the configurable text, so the info is always there.

Yes, I know that, but we still have study participants who are confused and get the time wrong. Maybe the text is too small at the bottom of the email. Or maybe they don’t read that far. Heck, who knows if they open up the email at all.

We want to make it really in-your-face and put the time in the reminder email subject (and body).

> Sure, you can the shop to to "one ticket per order" which is somewhat of a protection, but orders with multiple different subevents could still exist (from before the setting was done, from the API, imports, …). However, with the assumption that there is only ever one subevent per order, you could do:

Thanks for providing the example. I have updated it to use the event’s timezone (otherwise the times are incorrect) and include not just `{event_date_from}`, but also `{event_time_from}`: https://github.com/stapelberg/pretix/commit/801145d2e9cbcb2fcbcbb1b62536c7556479b670

> The placeholder you are looking for does not exist (and likely will not) since there's no way to guaranteed prevent an order from including multiple different subevents (that's kinda why subevents/series exist in the first place). And if that happens, what would the placeholder do? List all of them comma-separated? It will be hard to come up with a solution that does not confuse/surprise users and still allows to write readable texts. We've been thinking about this a lot since this feature is requested all the time, but have found no great solution. 

I agree that listing multiple subevents comma-separated is not a good idea :)

How about something like `{soonest_event_date_from}`, which would go through all orders, sort by subevent datetime, and then return the one happening the soonest? That would perfectly fit the reminder use-case without any ambiguous behavior.

raphaelm:
> How about something like {soonest_event_date_from}, which would go through all orders, sort by subevent datetime, and then return the one happening the soonest? That would perfectly fit the reminder use-case without any ambiguous behavior.

Yup, that would work, it would at least be well-defined. But hard to explain to most of our users, we'd probably need to implement a mechanism for help texts for placeholders before this makes sense :D There's also enough use cases where you'd rather want "soonest_event_date_that_is_not_yet_in_the_past" or "event_date_of_ticket_that_triggered_this_rule" or "latest_event_date_from". I think we wont open this can of worms in core for now :(

PR #3224, opened 2023-04-10
sendmail: add option to attach calendar invites

stapelberg:
This is helpful for sending out reminders.


src/pretix/plugins/sendmail/forms.py
```diff
@@ -218,6 +218,11 @@ class OrderMailForm(BaseMailForm):
         help_text=_("Will be ignored if tickets exceed a given size limit to ensure email deliverability."),
         required=False
     )
+    attach_ical = forms.BooleanField(
+        label=_("Attach calendar files"),
+        help_text=_("Will be ignored if tickets exceed a given size limit to ensure email deliverability."),
```
raphaelm, line 223:
```suggestion
```
not the case here :)

src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/send_form.html
```diff
@@ -57,6 +60,12 @@ <h1>
                                     {% trans "Tickets" %}
                                 </p>
                             {% endif %}
+                            {% if form.cleaned_data.attach_ical %}
+                                <p>
+                                    <span class="fa fa-file" aria-hidden="true"></span>
+                                    {% trans "Calendar invites" %}
```
raphaelm, line 66: maybe we could even stick with the same label
```suggestion
                                    {% trans "Attach calendar files" %}
```

raphaelm (commented):
Looks good on first glance, will test in more detail when I'm back from vacation (or someone else might)

src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/send_form.html
stapelberg: Ah, thanks, I couldn’t find that when I created the PR.

stapelberg:
Updated per your suggestions :)

stapelberg:
Updated once more to also add the attach_ical field to email rules (not just manually sent emails).

raphaelm (approved):

PR #3226, opened 2023-04-12
Add checkin requires attention to advanced order search

robbi5:
_Checkin requires attention_ is a flag that can be set on orders, but couldn't be searched/filtered for in the advanced order search.

### DIFF
--- a/src/pretix/control/forms/filter.py
+++ b/src/pretix/control/forms/filter.py
@@ -573,6 +573,7 @@ class EventOrderExpertFilterForm(EventOrderFilterForm):
         required=False,
         widget=FilterNullBooleanSelect,
         label=_('Requires special attention'),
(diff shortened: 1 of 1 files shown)


src/pretix/control/forms/filter.py
```diff
@@ -569,6 +569,11 @@ class EventOrderExpertFilterForm(EventOrderFilterForm):
         label=_('Sales channel'),
         required=False,
     )
+    checkin_attention = forms.NullBooleanField(
+        required=False,
+        widget=FilterNullBooleanSelect,
+        label=_('Requires special attention'),
```
raphaelm, line 575:
```suggestion
        label=_('Requires special attention'),
        help_text=_('Only matches orders with the attention checkbox set directly for the order, not based on the product.'),
```

raphaelm (approved):

PR #3228, opened 2023-04-14
Docs: Note on periodic tasks in dev setup

stapelberg:
It took a little bit of searching to figure out that in the dev environment, by default, celery tasks are run synchronously, but periodic tasks are not run at all.


raphaelm (approved):

raphaelm:
Good point!

PR #3235, opened 2023-04-17
Provide hidpi versions of logos and product pictures

stapelberg:
The idea is to provide a 200% and 300% version for hi-dpi mobile devices, tablets and laptops.

### DIFF
--- a/src/pretix/base/templates/pretixbase/forms/widgets/thumbnailed_file_input.html
+++ b/src/pretix/base/templates/pretixbase/forms/widgets/thumbnailed_file_input.html
@@ -1,7 +1,7 @@
 {% load thumb %}
 {% if widget.is_initial %}{{ widget.initial_text }}: <a href="{{ widget.value.url }}">{{ widget.value }}</a>{% if not widget.required %}
     <input type="checkbox" name="{{ widget.checkbox_name }}" id="{{ widget.checkbox_id }}">
-    <label for="{{ widget.checkbox_id }}">{{ widget.clear_checkbox_label }}</label>{% endif %}{% if widget.value.is_img %}<br><a href="{{ widget.value.url }}" data-lightbox="{{ widget.name }}"><img src="{{ widget.value|thumb:"200x100" }}" /></a>{% endif %}<br>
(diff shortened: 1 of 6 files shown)


raphaelm:
Hmm. hidpi is nice of course and I'd probably merge this for images on the web, but I'm not sure it's worth it for emails. Our email stack will convert these pictures to attachments, so this – especially if the logo contains a photo – can lead to *significantly* larger emails which is bad for deliverability. I'm not sure that's worth the effect.

stapelberg:
> Our email stack will convert these pictures to attachments, so this – especially if the logo contains a photo – can lead to _significantly_ larger emails which is bad for deliverability. I'm not sure that's worth the effect.

Fair point. Maybe using an SVG logo is a better choice for emails. 

Either way, I have removed the email srcset change from this PR. Please take another look :)

raphaelm:
> Fair point. Maybe using an SVG logo is a better choice for emails.

Yeah… but email is such a broken ecosystem, I don't wanna bet on the SVG implementations of some outlook version still used in large companies.

> Either way, I have removed the email srcset change from this PR. Please take another look :)

I updated the PR and added `srcset` to some some more places for consistency :)

ser8phin (approved):
Looks good. I agree with mail being a bit… difficult.
`srcset` is widely supported and I additionally tested it out with different settings and devices, so :+1:

raphaelm:
I needed to revert this for now, unfortunately. We have many organizers who uploaded a picture with a size of exactly 1170px, since that is what we recommended so far. Since our thumbnailer only scales down, not up, the 2x image is still 1170px wide. At least Firefox (I did not test other browsers) at 2x resolution will however not check that and always scale down the picture by factor 2. After that, it is displayed to small.

If we still want this feature, we'd need to do it conditionally on the size of the available source picture in some way? Sounds quite tedious.

stapelberg:
Thanks for letting me know. It was worth a shot. I might just go the SVG route for our case instead.

wiffbi:
I’d love to add „better“ sized images. One problem with fixed px-values for sizing though also would be, that one could not upload high-res images that are to be displayed only as smaller images – e.g. I uploaded an event_logo in 400 x 100 px size. If I upload this as high-res in e.g. 1200 x 300 px, then currently the event logo would be scaled to a width of 1170px (if full_width setting is used, otherwise it would be scaled to 120 height). Not what I intended. Due to this, uploading high-res images would only work for full-width images – not ideal, but IMHO a okayish tradeoff.

As a side-note: especially event-logo is super tricky. When using a background other than pure white, it is shown as 1170 and page-content is set to 1120px width. If a pure white background is used, the image is scaled to 1140px and the page-content is set to match 1140px. The event-logo in 1140px on white background still uses the 1170px-thumb. Not ideal.

Not sure how to proceed. Maybe some kind of thumb-operator to force upscale as well? E.g. like imagemagick with the `<`, so `event_logo|thumb:'2340x5000<'`.

PR #3241, opened 2023-04-18
Event settings: Extend product metadata (Z#23116647)

ser8phin:
Extend product metadata model and forms by adding the fields `required` and `allowed_values` in addition to the already existing `name` and `default` fields. Create additional API endpoints, as the earlier method only allowed to enter a key:value pair, and tests for those. Add logging to both the back-end forms and the API endpoints for product metadata changes.


src/pretix/base/migrations/0239_auto_20230510_1017.py
```diff
@@ -0,0 +1,23 @@
...
+
+    dependencies = [
+        ('pretixbase', '0238_giftcard_owner_ticket'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='itemmetaproperty',
+            name='allowed_values',
+            field=models.TextField(null=True),
```
wiffbi, line 16: See below, as this is auto-generated, probably do not update by hand. But for the sake of completeness.
```suggestion
            field=models.TextField(),
```

wiffbi (approved):
Looks good, just some minor typos in the docs and one comment regarding nullable field in DB. One thing I stumbled upon is empty lines in between allowed_values – they are currently not filtered. Combined with required, an empty line in allowed_values does not make sense. Also having the same value multiple times is allowed. Not super bad, but IMHO it would be better to filter empty lines (or output a warning?) and unify multiple same values. I do not know what to do though, when through the API I add e.g. 
```data={
            "name": "Color",
            "default": "Red",
            "required": true,
            "allowed_values": ["Red", "", "Green", "Blue"]
        }
```
If that value is required, then an empty value in the allowed_values can never be selected. Should we silently drop the empty value or report an 400 error?

I know the questions above also affect the event meta properties, so maybe that is out of scope for this PR.

doc/api/resources/item_meta_properties.rst
```diff
@@ -0,0 +1,209 @@
+Item Meta Properties
+====================
+
+Resource description
+--------------------
+
+The Item Meta Properties resource contains the following public fields:
```
raphaelm, line 7: Could you maybe add 1-2 sentences what an item meta property **is**?

doc/api/resources/item_meta_properties.rst
```diff
@@ -0,0 +1,209 @@
...
+
+===================================== ========================== =======================================================
+Field                                 Type                       Description
+===================================== ========================== =======================================================
+id                                    integer                    Unique ID for this property
+name                                  string                     Name of the property
+default                               string                     Value of the default option
+required                              boolean                    If ``true``, this property will have to be assigned a
+                                                                 value in all items of the related event
+allowed_values                        list                       Optional list of all permitted values for this property
```
raphaelm, line 19: Add: "or ``null`` for no limitation."

doc/api/resources/item_meta_properties.rst
```diff
@@ -0,0 +1,209 @@
...
+            "default": "red",
+            "required": true,
+            "allowed_values": ["red", "green", "blue"]
+          }
+        ]
+      }
+
+   :param organizer: The ``slug`` field of the organizer to modify
+   :param event: The ``slug`` field of the event to modify
+   :param id: The ``id`` field of the item meta property to modify
```
raphaelm, line 62:
```suggestion
```

doc/api/resources/item_meta_properties.rst
```diff
@@ -0,0 +1,209 @@
...
+        "id": 1,
+        "name": "Color",
+        "default": "red",
+        "required": true,
+        "allowed_values": ["red", "green", "blue"]
+      }
+
+   :param organizer: The ``slug`` field of the organizer to modify
+   :param event: The ``slug`` field of the event to modify
+   :param id: The ``id`` field of the item meta property to modify
```
raphaelm, line 93:
```suggestion
   :param id: The ``id`` field of the item meta property to retrieve
```

src/pretix/base/models/items.py
```diff
@@ -2001,6 +2001,15 @@ class ItemMetaProperty(LoggedModel):
         verbose_name=_("Name"),
     )
     default = models.TextField(blank=True)
+    required = models.BooleanField(
+        default=False, verbose_name=_("Required for products"),
+        help_text=_("If checked, this property must be set in each product. Does not apply if a default value is set.")
+    )
+    allowed_values = models.TextField(
+        null=True, blank=True,
```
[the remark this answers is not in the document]
ser8phin: I checked and it seems to only be set to `null` for old database entries, new ones created using the form get a blank entry automatically. And I also checked the docs: ["the Django convention is to use the empty string, not NULL"](https://docs.djangoproject.com/en/4.2/ref/models/fields/#null)
But, it seems I made it impossible to not set anything, besides an empty list if so wanted, using the API, so I'll have to fix that.

src/pretix/base/migrations/0239_auto_20230510_1017.py
ser8phin: As discussed, it will continue to be nullable :)

PR #3251, opened 2023-04-22
urlize event location (and linebreaksbr in email)

stapelberg:
This makes URLs in event locations a little more convenient, for example when saying “at <eventlocation>.ch”, or when linking to a detailed description for where the entrance is.


raphaelm:
Sorry, this somehow fell between the cracks. I don't think we use |urlize ever, and using it here will have slightly different (and less secure) behaviour than links in our |rich_text filter. At the same time, I don't really want rich-text locations because the location field is also commonly used in places (eg through the API) where only plain text is supported.

PR #3259, opened 2023-04-25
Cart: Add sneak-peek preview

wiffbi:
This PR adds an button „show cart“ and shows parts of the active cart during checkout. This should make it more obvious that the cart can be opened.

Bildschirmfoto 2023-04-25 um 12 46 41


raphaelm:
Sorry, closed by accident.

I like the last version with the darkened gradient more.

Personally, I would move the button a bit further down below, so it is "one normal padding unit" (15px) above the bottom of the panel. And maybe "Show full cart" as a text?

raphaelm (approved):
LGTM, let's merge right after the release

PR #3270, opened 2023-04-26
Update jQuery to 3.6.4

raphaelm:


wiffbi:
Looks good so far, could not find any issues. Not sure whether jquery-migrate might still be worth a try, as maybe I did not realize if something ran at all …
Regarding jquery-plugins, do we update them as well? E.g. jquery.fileupload.js seems to have a newer version (at least based on line-count). jquery.are-you-sure.js still uses bind/unbind (which are deprecated, but still there – so not yet a problem), but there is no updated version.
Regarding pretix-plugins, a quick search through my plugins only revealed pretix_digital to have a deprecated `$(document).ready`, pretix_nov***t and pretix_res***s to use `.load` on an iframe.

raphaelm:
> Regarding jquery-plugins, do we update them as well? E.g. jquery.fileupload.js seems to have a newer version (at least based on line-count). jquery.are-you-sure.js still uses bind/unbind (which are deprecated, but still there – so not yet a problem), but there is no updated version.

I checked most of them if there are updated versions and most of the time couldn't find any, so I don't think it's a problem for right now.

> Regarding pretix-plugins, a quick search through my plugins only revealed pretix_digital to have a deprecated $(document).ready, pretix_nov***t and pretix_res***s to use .load on an iframe.

I have already updated these

Issue #39, opened 2015-03-21
Better integration of permissions


raphaelm:
- [x] Hide links the users has no permission for
- [x] Add an editor for event permissions
- [x] Add an editor for organizer permissions
- [x] Add a _nice_ editor for teams/permissions
- [x] Restructure permissions such that plugins can define them

raphaelm:
possible duplicate of #123

fabian-marquardt:
Hi @raphaelm, are there currently any plans to implement custom permissions which plugins can define and require? This would be a very useful feature if the functionality provided by a plugin should only be available to a subset of the event organizers. Currently plugin views can only depend on the default set of permissions, which is not sufficient in some cases.

raphaelm:
No, there are currently no such plans for the forseeable future.

We know that a more fine-grained permission system is something lots of people want. However, we haven't yet been able to come up with a convincing design.

Currently, there are 12 permissions in the system. The number has grown only by 1-2 per year so far. If we open the floodgates of making this more fine-grained and allowing plugins to define more permissions, we'd **very likely and quickly** not end up with ~25 in a typical installation, but rather >500 permissions, making it much harder for users to configure them and much easier to get it wrong. Even if we build a great UI to configure them, we're adding new useful features to pretix every month. If we'd do permissions on a fine level, we'd also need to add new permissions every month, causing the new features to not show up for users until they manually re-configure the permissions (since we can hardly guess what they want).

fabian-marquardt:
Thank you for the swift and detailed response. I totally agree that such a redesigned permission system needs to be based on a well thought out concept and this will certainly not be done easily!

PR #3278, opened 2023-04-28
Support for IDNA domain names

raphaelm:
New approach at #2743

- [ ] test stuff
- [ ] harden against homograph attacks, see e.g. https://wiki.mozilla.org/IDN_Display_Algorithm

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -62,6 +62,7 @@ dependencies = [
   "drf_ujson2==1.7.*",
   "geoip2==4.*",
   "importlib_metadata==6.6.*",  # Polyfill, we can probably drop this once we require Python 3.10+
+  "idna",
   "isoweek",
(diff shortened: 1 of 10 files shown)


raphaelm:
Abandoning this for now. There's no good python implementation to protect against homoglyph attacks, which would even be required to fulfill GMail sender requirements:

https://support.google.com/mail/answer/81126#zippy=%2Crequirements-for-all-senders%2Crequirements-for-sending-or-more-messages-per-day

> Format these international domains according to the [Highly Restrictive guidelines in section 5.2 of Unicode Technical Standard #39](http://www.unicode.org/reports/tr39/#Restriction_Level_Detection):
>    Authenticating domain
>    Envelope from domain
>    **Payload domain**
>    Reply-to domain
>    Sender domain

(emphasis mine)

PR #3292, opened 2023-05-04
Fix shell_scoped without shell_plus

rixx:


pc-coholic:
Thank you so much <3

I think, I stumbled over this already at least 10 times but somehow never bothered researching the issue and rather just installed shell_plus 🙈

But this solution is a lot better 😁

rixx:
It turns out, all it takes is stumbling over it when you're not able or willing to install packages! :laughing:

PR #3296, opened 2023-05-08
Open ID: validate requested claims only if provider lists claims_supported

wiffbi:
In OIDC `claims_supported` are only recommended, so it happens, that an Open ID provider does not output them. We only check `claims_supported` during OIDC if they are listed in the provider’s config.

Note: `scopes_supported` is recommended only, but seems better supported. Should we also add an `if` or do we always check for it?

Closes #3294


raphaelm (approved):

raphaelm:
> Note: scopes_supported is recommended only, but seems better supported. Should we also add an if or do we always check for it?

Let's wait if we also run into actual issues there?

PR #3308, opened 2023-05-10
GiftCard: Add info field to transactions

raphaelm:


ser8phin (approved):
Looks good as far as I can tell, though I'm not completely sure if all the template changes work

PR #3309, opened 2023-05-11
Allow to highlight order code on invoice layouts

raphaelm:


ser8phin (approved):
Lgtm :+1: 
Checked different settings, and how it's displayed and it all works and looks nice; it's initially enabled in new non-copied events as intended; as the helptext says, it doesn't work in all renderers, so that's expected behavior :smiley:

PR #3315, opened 2023-05-15
Order list: Remove sums that cause lots of confusion

raphaelm:


pc-coholic (changes requested):
The translations are both broken.
image

Also, I am really not a big fan of this dirty yellow in the warning-icon... ;-)

pc-coholic (approved):

wiffbi:
Oh sorry for the noise, I just realized that @raphaelm already updated the icon to `.text-muted`. That works!

PR #3316, opened 2023-05-15
Fix Order._can_be_paid checks

julianrother:
`Order._can_be_paid` returns `True` if an order can be paid and a `str` with an error message otherwise. The result of `_can_be_paid` should thus be check somewhat like `order._can_be_paid() is not True` instead of `not order._can_be_paid()`. In a few cases this check in done incorrectly.


raphaelm (approved):
Good catch, thanks! I added a few regression tests as well.

PR #3319, opened 2023-05-16
Self-service order change: Respect Item.max/min_per_order (Z#23122195)

raphaelm:


src/pretix/base/services/orders.py
```diff
@@ -1929,6 +1945,27 @@ def set_addons(self, addons, limit_main_positions=None):
                                 }
                             )
                         self.cancel(a)
+                        item_counts[a.item] += 1
```
wiffbi, line 1948: Shouldn’t this decrease the items_count as the addon is canceled?
```suggestion
                        item_counts[a.item] -= 1
```

PR #3320, opened 2023-05-16
Show name including saluation in some places (Z#23121817)

raphaelm:
I went with really the minimum amount of places to change for now since I still feel without the salutations it is more natural in all kinds of lists, etc, but it's useful to see in the situation where you are e.g. just about to call someone.

### DIFF
--- a/src/pretix/helpers/names.py
+++ b/src/pretix/helpers/names.py
@@ -24,7 +24,7 @@ from pretix.base.settings import PERSON_NAME_SCHEMES
 
 def build_name(parts, concatenation=None, fallback_scheme=None):
     if not parts:
-        return ""
+        return None
     if "_legacy" in parts:
         return parts["_legacy"]
     if "_scheme" in parts:
--- a/src/pretix/helpers/names.py
+++ b/src/pretix/helpers/names.py
@@ -1,0 +1,15 @@
+from pretix.base.settings import PERSON_NAME_SCHEMES
+
+
+def build_name(parts, concatenation=None):
+    if not parts:
+        return ""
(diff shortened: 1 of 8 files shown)


wiffbi (commented):
Looks good. As all of these are very similar, what about a helper-method in base/settings.py that takes a name_parts and optionally a name_scheme as a fallback and only throws if no _scheme and fallback is provided? Would only work if we had the same return (e.g. `""`) `if not name_parts:` – although I am not sure why there is a distinction between `""` and `None` …

src/pretix/helpers/names.py
```diff
@@ -0,0 +1,15 @@
...
+        return ""
+    if "_legacy" in parts:
+        return parts["_legacy"]
+    if "_scheme" in parts:
+        scheme = PERSON_NAME_SCHEMES[parts["_scheme"]]
+    else:
+        raise TypeError("Invalid name given.")
+    if concatenation:
+        return scheme.get(concatenation, scheme["concatenation"])(parts).strip()
+    return scheme["concatenation"](parts).strip()
```
wiffbi, line 15: For me, this is easier to read. But probably YMMV.

```suggestion
    if not concatenation or concatenation not in scheme:
        concatenation = "concatenation"
    return scheme[concatenation](parts).strip()
```

wiffbi (approved):
Looks good. I have a minor comment regarding concatenation. Also, this sometimes now raises a TypeError when it previously it used `self.event.settings.name_scheme` and `self.customer.organizer.settings.name_scheme`. I guess you are aware of this and it is not a problem. So, good to go!

PR #3322, opened 2023-05-16
Vouchers: Allow to set all addons or bundles as included

raphaelm:
- [x] POS support


src/pretix/base/services/cart.py
```diff
@@ -512,7 +512,10 @@ def extend_expired_positions(self):
             if cp.is_bundled:
                 bundle = cp.addon_to.item.bundles.filter(bundled_item=cp.item, bundled_variation=cp.variation).first()
                 if bundle:
-                    listed_price = bundle.designated_price or Decimal('0.00')
+                    if cp.addon_to.voucher_id and cp.addon_to.voucher.all_bundles_included:
+                        listed_price = Decimal('0.00')
+                    else:
+                        listed_price = bundle.designated_price or Decimal('0.00')
```
wiffbi, line 518: Is that `or`-part still needed? `designated_price` has a default of `Decimal('0.00')` so I guess this is not needed – if it is still needed, `bundle.designated_price` is used in other places where it should have an `or Decimal('0.00')`.

```suggestion
                        listed_price = bundle.designated_price
```

src/pretix/base/models/vouchers.py
```diff
@@ -296,6 +296,14 @@ class Voucher(LoggedModel):
         verbose_name=_("Shows hidden products that match this voucher"),
         default=True
     )
+    all_addons_included = models.BooleanField(
+        verbose_name=_("Offer all add-on products for free when redeeming this voucher"),
+        default=False
+    )
+    all_bundles_included = models.BooleanField(
+        verbose_name=_("Include all bundled products without a designated price when redeeming this voucher"),
```
wiffbi, line 304: That help-text is a bit ambigious – it could mean, only bundled products without a designated price are included, when this voucher is redeemed. Does not really make sense if one thinks about it, but IMHO the following is easier to understand:

```suggestion
        verbose_name=_("Include all bundled products for free when redeeming this voucher"),
```

src/pretix/control/templates/pretixcontrol/vouchers/bulk.html
```diff
@@ -78,6 +78,8 @@ <h1>{% trans "Create multiple vouchers" %}</h1>
             {% bootstrap_field form.tag layout="control" %}
             {% bootstrap_field form.comment layout="control" %}
             {% bootstrap_field form.show_hidden_items layout="control" %}
+            {% bootstrap_field form.all_addons_included layout="control" %}
+            {% bootstrap_field form.all_bundles_included layout="control" %}
```
wiffbi, line 82: Somehow it feels more natural to me to place this somewhere around price change, e.g. before `{% if form.subevent %}`. Although one needs to remove the margin-bottom from the `div.alert`

Bildschirmfoto 2023-05-16 um 22 33 41

src/pretix/control/templates/pretixcontrol/vouchers/detail.html
```diff
@@ -90,6 +90,8 @@ <h1>{% trans "Voucher" %}</h1>
                     {% bootstrap_field form.tag layout="control" %}
                     {% bootstrap_field form.comment layout="control" %}
                     {% bootstrap_field form.show_hidden_items layout="control" %}
+                    {% bootstrap_field form.all_addons_included layout="control" %}
+                    {% bootstrap_field form.all_bundles_included layout="control" %}
```
wiffbi, line 94: See above regarding placement near price change.

wiffbi (approved):
Looks good, again just minor comments.

wiffbi:
One thing I just stumbled upon, which probably is out of scope for this PR: addons with free price input are included for free without any price input, when using a voucher that has `all_addons_included=True`. Should we still show the free pice input, but with a default-value of 0.00?

src/pretix/base/services/cart.py
raphaelm: I was unable to trace that back, the or-part existed since we added bundles. I don't see why we need it either.

PR #3323, opened 2023-05-16
Widget: Fix waiting list integration of seated events

raphaelm:


wiffbi (approved):
Looks good and works!

PR #3343, opened 2023-05-22
Invoices: Support font choice and Arabic text

raphaelm:


wiffbi (commented):
Looks good, just a minor typo/mixup of order for Paragraph/_normalize. See comments.

Also, `_draw_testmode` (invoice.py::512) is not normalized, so might create an error when translated to arabic? The same with to the formatted strings in `_get_intro` (lines 551, 560, 568, and 575).

Line 654, 655, 669 and a few other places sets a `money_filter`. Do these need to be normalized as well? Not sure about arabic localization of currency letters/symbols, but I guess they do not need to be normalized …

Line 818 adds tax-rate `name` to `tdata` – I think the name be run through `self._normalize` as it could be set to something arabic?

raphaelm:
> Also, _draw_testmode (invoice.py::512) is not normalized, so might create an error when translated to arabic? The same with to the formatted strings in _get_intro (lines 551, 560, 568, and 575).

Good catch, fixed.

> Line 654, 655, 669 and a few other places sets a money_filter. Do these need to be normalized as well? Not sure about arabic localization of currency letters/symbols, but I guess they do not need to be normalized …

As far as I tested, it is not required, at least not for Arabic.

> Line 818 adds tax-rate name to tdata – I think the name be run through self._normalize as it could be set to something arabic?

Yup, will fix!

PR #3347, opened 2023-05-23
Invoice: Improve handling of special characters in file names

raphaelm:


src/pretix/base/services/mail.py
```diff
@@ -466,9 +466,17 @@ def mail_send_task(self, *args, to: List[str], subject: str, body: str, html: st
...
                 if inv.file:
                     try:
+                        # We try to give the invoice a more human-readable name, e.g. "Invoice_ABC-123.pdf" instead of
+                        # just "ABC-123.pdf", but we only do so if our currently selected language allows to do this
+                        # as ASCII text. For example, we would not want a "فاتورة_" prefix for our filename since this
+                        # has shown to cause deliverability problems of the email and deliverability wins.
+                        filename = pgettext('invoice', 'Invoice {num}').format(num=inv.number).replace(' ', '_') + '.pdf'
+                        if not re.match("^[a-zA-Z0-9-_%./,&:# ]+$", filename):
+                            filename = inv.number.replace(' ', '_') + '.pdf'
+                        filename = re.sub("[^a-zA-Z0-9-_.]", "_", filename)
```
wiffbi, line 476: Should multiple occurrences in a row only be replaced by a single underscore? Could be done to the spaces-replacement as well, but there it is probably overkill.

```suggestion
                        filename = re.sub("[^a-zA-Z0-9-_.]+", "_", filename)
```

src/pretix/presale/views/order.py
```diff
@@ -1235,7 +1235,7 @@ def get(self, request, *args, **kwargs):
         except FileNotFoundError:
             invoice_pdf_task.apply(args=(invoice.pk,))
             return self.get(request, *args, **kwargs)
-        resp['Content-Disposition'] = 'inline; filename="{}.pdf"'.format(invoice.number)
+        resp['Content-Disposition'] = 'inline; filename="{}.pdf"'.format(re.sub("[^a-zA-Z0-9-_.]", "_", invoice.number))
```
wiffbi, line 1238: See above
```suggestion
        resp['Content-Disposition'] = 'inline; filename="{}.pdf"'.format(re.sub("[^a-zA-Z0-9-_.]+", "_", invoice.number))
```

wiffbi (approved):
Looks good, first I thought of adding a `@property` to `Invoice`, but how do you name that? – and it is to special anyways.
Just one minor comment regarding replacing mulitple special chars in a row with just one underscore.

PR #3351, opened 2023-05-24
Event calendar: Respect voucher for availability

raphaelm:


src/pretix/base/models/event.py
```diff
@@ -1454,10 +1467,10 @@ def _seats(self, ignore_voucher=None):
         return qs_annotated
 
     @classmethod
-    def annotated(cls, qs, channel='web'):
+    def annotated(cls, qs, channel='web', voucher=None):
         from .items import SubEventItem, SubEventItemVariation
 
-        qs = super().annotated(qs, channel)
+        qs = super().annotated(qs, channel, voucher=voucher)
```
wiffbi, line 1473: It works named, but to make it more consistent:
```suggestion
        qs = super().annotated(qs, channel, voucher)
```

wiffbi (approved):
Looks good, just a minor comment regarding consistency.

Should we respect the voucher when showing the subevents-overview in the widget?

raphaelm:
The widget currently does not allow passing a voucher to an event list according to it's documentation, so I'd hold off for now

src/pretix/base/models/event.py
raphaelm: Yeah, but using more keyword arguments generally makes the code easier to maintain in the future

PR #3357, opened 2023-05-26
Export: fix timezone in event-data export (Z#23122901)

wiffbi:


src/pretix/base/exporters/events.py
```diff
@@ -73,18 +74,19 @@ def iterate_list(self, form_data):
             header.append(p.name)
         yield header
 
+        tz = get_current_timezone()
```
raphaelm, line 77: 1. There is ``self.timezone`` available in all exporters which we should probably use here
2. What if the events have different time zones configured? Should we output all of them in the same (organizer/user default) zone or should we use the one of the individual event (and maybe add that as a column)?

raphaelm (changes requested):

src/pretix/base/exporters/events.py
wiffbi: Good catch, I just copied that bit from CustomerExport, which is organizer-level as well. `self.timezone` is definitly better. Regarding 2: good question, I think best would be to export to event’s timezone and add a column for the timezone. Even if we at some point change datetimes in CSV-files to ISO-formatted (and xls to xls-native date format), xls probably needs timezone anyways. I will change accordingly.

raphaelm (approved):

PR #3358, opened 2023-05-26
Checkout: Hide empty add-on forms and show seat above add-on form (Z#23122801)

wiffbi:
When some products have add-ons, during checkout, we currently show all products in the cart – even the ones without addons. This PR removes all products that have no add-ons from the add-ons checkout-step. This simplifies otherwise long forms, where only a few products have add-ons. We loose the strong coupling of all products in cart-order though. But IMHO that is not much of an issue. As YMMV feel free to test with other events.

To improve matching cart-item and add-on when ordering multiple seated tickets, this PR also adds seating-info to each addon-section (seat-info breaking line if not enough space).

Bildschirmfoto 2023-05-30 um 12 06 14

Bildschirmfoto 2023-05-30 um 12 06 30

P.S.: If you look closely, this uses a refined version of the seat-icon, which I already updated inside the pretix-seating plugin. We should update all other occurrences of the seat-icon in pretix-core as well.

P.P.S.: Current/old seat-icon for reference
Bildschirmfoto 2023-05-30 um 12 10 14


raphaelm:
I'm torn, maybe let's try it and roll back if it confuses people.

But other than the seat, should we also show the selected subevent?

wiffbi:
Good idea, I will add the subevent info!

wiffbi:
Subevent-info was already shown inside the panel. Showing both seat and subevent in the panel’s title was a bit too much, so I moved the seat-info inside the panel as well.

One thing I stumbled upon: the algorithm, which decides whether to show the Add-On-step in checkout does not take quotas into account. This can cause an „empty“ add-on step which just says „For some of the products in your cart, you can choose additional options before you continue.“ – and no options show as the quotas ran out. Should we change the checkout-step-algorithm to take quotas into account (heavier on the db) and could lead to organizers wondering why their checkout-step Addons is not shown. Or should we add a message that all Addons are sold out? Although this could clash the setting `hide_sold_out`.

I am not sure, what to do best here: hide add-on-step when there is no quota left and `hide_sold_out` is `True`? Otherwise show add-on-step and even show sold-out add-ons as sold-out? Or the easiest leave as is and only replace the text with e.g. „There are currently no additional options available for the products in your cart.“

PR #3367, opened 2023-05-30
Update translations

pretix-translations:


pc-coholic:
This is just a remnant of weblate doing weird things after a painful translation merge/rebase. There should be nothing in this PR, that hasn't already been merged minutes before.

PR #3372, opened 2023-06-01
[WIP] Presale: respect item.order_min when changing quantity (Z#23122112)

wiffbi:
This PR adds support for item.order_min on spinner buttons. Furthermore it adds an onblur-handler that set the input to the order_min if not 0 – as this happens without any further warning, I am not sure whether that is a good idea, although IMHO it is not bad either. As it is a separate commit we can remove that behaviour easily from this PR.


raphaelm:
Not sure this makes sensevon first glance, what if you already have the min amount in your cart and want to add one more?

wiffbi:
That is a good catch! I would rather use the +/- buttons in the cart, but of course the way you describe should be possible as well. Well, then we drop that PR. Or I could try to match the `item.id` from that input to the cart (I could add the `item.id` to e.g. the quantity-output in cart). But probably a bit to complex for that small feature? Although would be a nice touch.

wiffbi:
I improved the onchange-handling (which is triggered by the custom spinner-buttons as well, so no code duplication). I also consider existing items in the cart in the calculation (based on item.id).

wiffbi:
You are right, it even does not work for products with variants.

PR #3380, opened 2023-06-04
Add check-in simulator

raphaelm:
Here's a handy tool to figure out why exactly a check-in rule succeeds or fails:
image

### DIFF
--- a/src/pretix/control/views/checkin.py
+++ b/src/pretix/control/views/checkin.py
@@ -19,6 +19,7 @@
 # You should have received a copy of the GNU Affero General Public License along with this program.  If not, see
 # <https://www.gnu.org/licenses/>.
 #
+
 # This file is based on an earlier version of pretix which was released under the Apache License 2.0. The full text of
 # the Apache License 2.0 can be obtained at <http://www.apache.org/licenses/LICENSE-2.0>.
 #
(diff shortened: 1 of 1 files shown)


src/pretix/control/templates/pretixcontrol/checkin/simulator.html
```diff
@@ -0,0 +1,130 @@
+{% extends "pretixcontrol/items/base.html" %}
+{% load i18n %}
+{% load bootstrap3 %}
+{% load escapejson %}
+{% load getitem %}
+{% load static %}
+{% load compress %}
+{% block title %}{% trans "Check-in simulator" %}{% endblock %}
+{% block inside %}
+    <h1>{% trans "Check-in simulator" %}</h1>
```
pc-coholic, line 10: I feel like the simulator is a little bit "disconnected" from the selected checkin-list.

Perhaps we can try to have the headline the same or similar to the checkinlist positions- oder edit-view; at least with the name of the list, possibly even with buttons to edit the list/return to its main view.

pc-coholic (changes requested):

src/pretix/control/forms/checkin.py
```diff
@@ -177,3 +178,26 @@ class Meta:
             'subevent': SafeModelChoiceField,
             'gates': SafeModelMultipleChoiceField,
         }
+
+
+class CheckinListSimulatorForm(forms.Form):
+    raw_barcode = forms.CharField(
+        label=_("Barcode"),
```
robbi5, line 185: Hmm, at the other places in the pretix backend, we call this `ticket code`.

pc-coholic: True... But I would assume that people who use the simulator do indeed scan a barcode to check the system...

I see this one valid one or the other way :-)

PR #3390, opened 2023-06-06
Fix: use format_lazy for formatted translation in settings

wiffbi:
When trying to access something in pretix-control, that the currently logged in user has no access-right to, the error message was not localised correctly.
In settings there were some formatted strings used incorrectly with translation – some where translated before being fed into format_lazy, one was using gettext_lazy, but directly using Python’s `.format`.

### DIFF
--- a/src/pretix/base/settings.py
+++ b/src/pretix/base/settings.py
@@ -2119,11 +2119,14 @@ Your {event} team"""))
             label=_('Attachment for new orders'),
             ext_whitelist=(".pdf",),
             max_size=settings.FILE_UPLOAD_MAX_SIZE_EMAIL_AUTO_ATTACHMENT,
-            help_text=format_lazy(_('This file will be attached to the first email that we send for every new order. Therefore it will be '
-                        'combined with the "Placed order", "Free order", or "Received order" texts from above. It will be sent '
-                        'to both order contacts and attendees. You can use this e.g. to send your terms of service. Do not use '
-                        'it to send non-public information as this file might be sent before payment is confirmed or the order '
-                        'is approved. To avoid this vital email going to spam, you can only upload PDF files of up to {size} MB.'),
+            help_text=format_lazy(
+                _(
+                    'This file will be attached to the first email that we send for every new order. Therefore it will be '
+                    'combined with the "Placed order", "Free order", or "Received order" texts from above. It will be sent '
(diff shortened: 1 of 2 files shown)


src/pretix/base/settings.py
```diff
@@ -546,8 +546,8 @@ def unserialize(cls, s):
         'form_kwargs': dict(
             label=_("Ask for VAT ID"),
             help_text=format_lazy(
-                _("Only works if an invoice address is asked for. VAT ID is never required and only requested from "
-                  "business customers in the following countries: {countries}"),
+                "Only works if an invoice address is asked for. VAT ID is never required and only requested from "
+                "business customers in the following countries: {countries}",
```
raphaelm, line 550: I believe this is wrong. If you take the ``_`` away, `makemessages` will no longer find this string as a translatable one *at all*, so it will no longer be included in .po files if you run makemessages and will no longer get translated, or am I missing something I don't know about gettext?

src/pretix/control/middleware.py
```diff
@@ -41,7 +41,7 @@
 from django.template.response import TemplateResponse
 from django.urls import get_script_prefix, resolve, reverse
 from django.utils.encoding import force_str
-from django.utils.translation import gettext as _
+from django.utils.translation import gettext_lazy as _
```
raphaelm, line 44: I don't get how this is related to the issue and it might not be the correct fix. The call to `_()` down below in the file is not executed at import time (which is why we usually need gettext_lazy), but only on view time.

I believe the reason we are seeing flaky translation on the error message is a different one: `LocaleMiddleware`, which decides which language to use for the request, is run **after** `PermissoinMiddleware`, so we'd need to swap the order of these two middlewares.

raphaelm (changes requested):

src/pretix/base/settings.py
wiffbi: Hm, django-docs recommend using format_lazy. When using _() before, then when having a default locale other en, this will break.

wiffbi: Fixed.

src/pretix/control/middleware.py
wiffbi: Sorry, removed.

raphaelm (approved):

PR #3409, opened 2023-06-14
Docs: update requirements and links

fabm3n:
Just updated the requirements and links.

Is Postgres 11+ still the current requirement?


raphaelm (approved):

raphaelm:
Thanks! Yes, PostgreSQL 11 is still correct until we upgrade to Django 4.2

fabm3n:
> Thanks! Yes, PostgreSQL 11 is still correct until we upgrade to Django 4.2

Ok, thanks for the information :)

PR #3413, opened 2023-06-16
Bank transfer: Do not use <pre> for bank details in emails

raphaelm:
The main reason for using ``<pre>`` here previously was that we did not have a good way to do force new lines otherwise, but I fixed that in 3df64a46e.

### DIFF
--- a/src/pretix/plugins/banktransfer/payment.py
+++ b/src/pretix/plugins/banktransfer/payment.py
@@ -33,7 +33,6 @@
 # License for the specific language governing permissions and limitations under the License.
 
 import json
-import textwrap
 from collections import OrderedDict
 from decimal import Decimal
 
@@ -56,6 +55,7 @@ from pretix.base.i18n import language
 from pretix.base.models import InvoiceAddress, Order, OrderPayment, OrderRefund
 from pretix.base.payment import BasePaymentProvider
 from pretix.base.services.mail import SendMailException, mail, render_mail
+from pretix.base.templatetags.money import money_filter
 from pretix.helpers.format import format_map
 from pretix.plugins.banktransfer.templatetags.ibanformat import ibanformat
 from pretix.presale.views.cart import cart_session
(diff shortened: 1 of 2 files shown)


src/pretix/plugins/banktransfer/payment.py
```diff
@@ -454,26 +454,25 @@ def checkout_confirm_render(self, request, order=None):
...
-        ctx = {
-            'event': self.event,
-            'order': order,
-            'code': self._code(order),
-            'amount': payment.amount,
-            'details': textwrap.indent(''.join(str(i) for i in bankdetails), '    '),
-        }
-        return template.render(ctx)
+
+        t += ''.join(str(i) for i in bankdetails)
```
wiffbi, line 474: Works, but for me it would be easier to read, if we separate content and formatting like so. Also, IMHO explicitly joining with a variable called e.g. `mk_nl2br` makes the intent of this `"  \n"` clearer. Sorry for this huge suggestion (could not format as suggestion, due to deleted lines, but anyways probably better to test again locally):

```
        t = gettext("Please transfer the full amount to the following bank account:")
        t += "\n\n"

        md_nl2br = "  \n"
        if self.settings.get('bank_details_type') == 'sepa':
            bankdetails = (
                (_("Reference"), self._code(order)),
                (_("Amount"), money_filter(payment.amount, self.event.currency)),
                (_("Account holder"), self.settings.get('bank_details_sepa_name')),
                (_("IBAN"), ibanformat(self.settings.get('bank_details_sepa_iban'))),
                (_("BIC"), self.settings.get('bank_details_sepa_bic')),
                (_("Bank"), self.settings.get('bank_details_sepa_bank')),
            )
            t += md_nl2br.join([f"**{k}:** {v}" for k, v in bankdetails])
            if self.settings.get('bank_details', as_type=LazyI18nString):
                t += md_nl2br
        t += str(self.settings.get('bank_details', as_type=LazyI18nString))
        return t
```

wiffbi (changes requested):
I added a bigger code suggestion that IMHO makes it more readable – if you disagree feel free to ignore as it already works now.

PR #3417, opened 2023-06-17
add rudimentary img support in email messages

yvovandoorn:
As I proposed in https://github.com/pretix/pretix/discussions/3416, I was looking for a way to include an image when sending mails to my attendees. Specifically the image was a pickup map on where a boat would pick them up. 

By modifying the files changed, I was able to get Pretix to send out mails with images.

### DIFF
--- a/doc/user/markdown.rst
+++ b/doc/user/markdown.rst
@@ -145,7 +145,7 @@ to get a better plain text representation of your text. Note however, that for
 security reasons you can only use the following HTML elements::
 
     a, abbr, acronym, b, br, code, div, em, h1, h2,
-    h3, h4, h5, h6, hr, i, li, ol, p, pre, span, strong,
+    h3, h4, h5, h6, hr, i, img, li, ol, p, pre, span, strong,
     table, tbody, td, thead, tr, ul
 
 Additionally, only the following attributes are allowed on them::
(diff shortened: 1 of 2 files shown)


raphaelm:
Thanks! For markdown embedded in the website, this won't work due to our content security policy (and we don't really want it to work). For emails, this would technically work, but could be a security hazard in a shared-hosting environment, especially since all images in mails are currently inlined server-side and attached to the email. We'll need to look into the tradeoffs in doing this and see if we want to go down that path.

Note that on a selfhosted installation you can always roll your own email renderer plugin that does whatever you like.

yvovandoorn:
> Thanks! For markdown embedded in the website, this won't work due to our content security policy (and we don't really want it to work). For emails, this would technically work, but could be a security hazard in a shared-hosting environment, especially since all images in mails are currently inlined server-side and attached to the email. We'll need to look into the tradeoffs in doing this and see if we want to go down that path.

I suspected that there was a security implication around it! With Pretix also being a SaaS offering, this policy obviously prevents the nefarious from doing such :-) 

> Note that on a selfhosted installation you can always roll your own email renderer plugin that does whatever you like.

For that, I will admit, I lack the necessary coding skills to code up. While I was able to determine where to make the modifications to allow me to send out mails with images, doing this as a plugin is above my skill level.

raphaelm:
I'm closing this PR as we'll not merge this at this time, I'm afraid :(

PR #3434, opened 2023-06-26
Add OS name and version to stored device information

robbi5:
This may help when debugging pretix device issues.

TODO: 
- [x] check if fields are really optional enough, so that older clients don't get errors when submitting device info without OS


doc/api/resources/devices.rst
```diff
@@ -123,6 +127,8 @@ Device endpoints
         "security_profile": "full",
         "hardware_brand": "Zebra",
         "hardware_model": "TC25",
+        "os_name": "Android",
+         "os_version": "8.1.0",
```
raphaelm, line 131:
```suggestion
        "os_version": "8.1.0",
```

raphaelm (approved):

PR #3439, opened 2023-06-28
Widget: add lightbox for product images (Z#23123811)

wiffbi:
This PR adds a lightbox for fullsize product images to the widget. It reuses the styling of other overlays in the widget (ie. light background, close button top right), but adds its own classes, as positioning and sizing work slightly different on images vs. e.g. iframes.

Note: Styling on customized widget implementations (e.g. the close button) will be broken when lightbox is used/shown.


src/pretix/static/pretixpresale/scss/widget.scss
```diff
@@ -768,6 +768,99 @@
...
+    margin: 0;
+    padding: 0;
+  }
+  .pretix-widget-lightbox-image img {
+    max-width: 80vw;
+    max-height: 80vh;
+    object-fit: scale-down;
+  }
+  .pretix-widget-lightbox-image figcaption {
+    margin: 0.5em 0;
```
raphaelm, line 832: I personally like it better with
```suggestion
    margin: 0.5em 0 0;
    text-align: center;
```
What do you think?

raphaelm (approved):

src/pretix/static/pretixpresale/scss/widget.scss
wiffbi: Yup, I like it! I also added a safe-guard max-width and max-height, when e.g. the text is very long. Both text and image are now centered, as it could happen, that the image even in fullsize is rather small, but the text is very long.

raphaelm (approved):

PR #3444, opened 2023-06-30
PProv: Implement detection of wallets such as Google Pay and Apple Pay

pc-coholic:
Until now, we have never actively advertised card tokenization wallet-payment methods such as Apple or Google Pay in the payment provider selection screen.

This is based on the fact, that there are a lot of factors involved in deciding if the wallet payment method is actually offered to the customer: browser, browser settings (incognito/regular), cards added to the wallet, payment-gateway support, merchant enrollment, etc.

So we ended up only displaying "Credit Card" on the selection screen and leaving it up to the payment provider to render one or more wallet payment buttons if they deemed it appropriate.

Understandably, some event organizers do not like this, since their customers do not realize that Apple Pay/Google Pay is just a glorified credit card transaction and wouldn't think of clicking on the Credit Card payment option.

This PR is an attempt at remedying this situation.

- A payment provider can now optionally declare a list of `walletqueries`. These are added to the payment method selection accordions on the checkout and order change pages.
- It is the job of the payment provider to only request those wallets, which - to it's best knowledge - are available to the user. So if payment provider knows that a certain wallet is not available to the merchant (Example for Stripe: disabled in the Stripe Dashboard), it should not query for that wallet.
- Client-side, a JS will aggregate the requested wallets and then attempt a best-effort check of the availability of said wallets.
- If the requested wallet is available, the JS will (similar to PayPal2) add the name of the wallet to the public_name of the payment method.

The biggest caveat of this solution is that it cannot be 100% precise:
- For Apple Pay, we only call `canMakePayments()`. In order to get a more precise result, we would need to call `canMakePaymentsWithActiveCard(merchantIdentifier)`. But since in 99% of all cases, we do not have the `merchantIdentifier`, this is impossible.
- For Google Pay, we only check the availability using `test` parameters. As with Apple Pay, one would need the real/live gateway type, merchantId and gatewayMerchantId. In addition, a proper PaymentRequest would need to be constructed, passing the desired card networks, amount, etc. - something we do not have most of the time. By checking against the test-gateway, we only check if the device is capable of doing a Google Pay transaction in theory, which hopefully is enough.
- In order to check for Google Pay availability, we need to include the Google Pay JS-SDK.

In theory, we could provide facilities for payment providers to expose their merchantIdentifiers and query with them - but seeing how none of our currently implemented payment providers allow for this, I think we can skip over this.

I do not know how precise this detection is (especially for Google Pay) - so I think we should merge it but perhaps declare it as experimental?

### DIFF
--- a/src/pretix/base/middleware.py
+++ b/src/pretix/base/middleware.py
@@ -249,7 +249,7 @@ class SecurityMiddleware(MiddlewareMixin):
 
         h = {
             'default-src': ["{static}"],
-            'script-src': ['{static}', 'https://checkout.stripe.com', 'https://js.stripe.com'],
+            'script-src': ['{static}', 'https://checkout.stripe.com', 'https://js.stripe.com', 'https://pay.google.com'],
             'object-src': ["'none'"],
             'frame-src': ['{static}', 'https://checkout.stripe.com', 'https://js.stripe.com'],
             'style-src': ["{static}", "{media}"],
--- a/src/pretix/base/payment.py
+++ b/src/pretix/base/payment.py
@@ -449,10 +449,13 @@ class BasePaymentProvider:
     @property
     def walletqueries(self):
         """
+        .. warning:: This property is considered **experimental**. It might change or get removed at any time without
+                     prior notice.
+
         A list of wallet payment methods that should be dynamically joined to the public name of the payment method,
         if they are available to the user.
         The detection is made on a best effort basis with no guarantees of correctness and actual availability.
-        Wallets that pretix can check for are exposed through pretix.base.payment.WalletQueries.
+        Wallets that pretix can check for are exposed through ``pretix.base.payment.WalletQueries``.
         """
         return []
 
--- a/src/pretix/base/payment.py
+++ b/src/pretix/base/payment.py
@@ -78,6 +78,16 @@ from pretix.presale.views.cart import cart_session, get_or_create_cart_id
 logger = logging.getLogger(__name__)
 
 
+class WalletQueries:
+    APPLEPAY = 'applepay'
+    GOOGLEPAY = 'googlepay'
+
+    WALLETS = (
+        (APPLEPAY, pgettext_lazy('payment', 'Apple Pay')),
+        (GOOGLEPAY, pgettext_lazy('payment', 'Google Pay')),
+    )
+
+
 class PaymentProviderForm(Form):
     def clean(self):
         cleaned_data = super().clean()
@@ -435,6 +445,16 @@ class BasePaymentProvider:
         d['_restricted_countries']._as_type = list
         d['_restrict_to_sales_channels']._as_type = list
         return d
+
+    @property
+    def walletqueries(self):
+        """
+        A list of wallet payment methods that should be dynamically joined to the public name of the payment method,
+        if they are available to the user.
+        The detection is made on a best effort basis with no guarantees of correctness and actual availability.
+        Wallets that pretix can check for are exposed through pretix.base.payment.WalletQueries.
+        """
+        return []
 
     def settings_form_clean(self, cleaned_data):
         """
--- a/src/pretix/plugins/stripe/payment.py
+++ b/src/pretix/plugins/stripe/payment.py
@@ -59,7 +59,7 @@ from pretix import __version__
 from pretix.base.decimal import round_decimal
 from pretix.base.forms import SecretKeySettingsField
 from pretix.base.models import Event, OrderPayment, OrderRefund, Quota
-from pretix.base.payment import BasePaymentProvider, PaymentException
+from pretix.base.payment import BasePaymentProvider, PaymentException, WalletQueries
 from pretix.base.plugins import get_all_plugins
 from pretix.base.services.mail import SendMailException
 from pretix.base.settings import SettingsSandbox
@@ -746,6 +746,11 @@ class StripeCC(StripeMethod):
     verbose_name = _('Credit card via Stripe')
     public_name = _('Credit card')
     method = 'cc'
+
+    @property
+    def walletqueries(self):
+        # ToDo: Check against Stripe API, if ApplePay and GooglePay are even activated/available
+        return [WalletQueries.APPLEPAY, WalletQueries.GOOGLEPAY]
 
     def payment_form_render(self, request, total) -> str:
         account = get_stripe_account_key(self)
--- a/src/pretix/plugins/stripe/payment.py
+++ b/src/pretix/plugins/stripe/payment.py
@@ -59,7 +59,9 @@ from pretix import __version__
 from pretix.base.decimal import round_decimal
 from pretix.base.forms import SecretKeySettingsField
 from pretix.base.models import Event, OrderPayment, OrderRefund, Quota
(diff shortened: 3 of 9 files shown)


src/pretix/static/pretixpresale/js/walletdetection.js
```diff
@@ -0,0 +1,120 @@
+'use strict';
+
+var walletdetection = {
+    applepay: function () {
+        // This is a weak check for Apple Pay -  in order to do a proper check, we would need to also call
+        // canMakePaymentsWithActiveCard(merchantIdentifier)
+
+        return !!(window.ApplePaySession && window.ApplePaySession.canMakePayments());
+    },
+    googlepay: function () {
```
wiffbi, line 10: See below, I think this needs `aysnc`.

src/pretix/static/pretixpresale/js/walletdetection.js
```diff
@@ -0,0 +1,120 @@
...
+    // Filtering out any doubles
+    requestedWallets = new Set(requestedWallets);
+
+    // Perform the actual check *only* on the requested wallets, if they are supported by the browser
+    let availableWallets = Array();
+    requestedWallets.forEach(function (it) {
+        if (walletdetection[it]()) {
+            availableWallets.push(it);
+        }
+    });
```
wiffbi, line 86: I have not tested this, but wouldn’t `walletdetection[it]()` with `it == 'googlepay'` always return true, as $.ajax without `async/await` returns a jqXHR-object?

With `async/await` (see comments above on async applepay/googlepay) I’d shorten this to:

```suggestion
    let paymentMethods = $('[data-wallets]');
    let availableWallets = paymentMethods
        .map(pm => pm.getAttribute("data-wallets").split("|"))
        .flat()
        .filter(function(item, pos, self) {
            return item && self.indexOf(item) == pos;
        })
        .filter(function(item) {
            return await walletdetection[item]();
        })
```

src/pretix/static/pretixpresale/js/walletdetection.js
```diff
@@ -0,0 +1,120 @@
...
+        textselector.fadeOut(300, function () {
+            wallets.unshift(textselector.text());
+            textselector.text(wallets.join(", "));
+            textselector.fadeIn(300);
+        });
+        textselector2.fadeOut(300, function () {
+            wallets.unshift(textselector2.text());
+            textselector2.text(wallets.join(", "));
+            textselector2.fadeIn(300);
+        });
```
wiffbi, line 118: As you run the same code on both elements, you could just combine both selectors:

```suggestion
        $('+ label strong, + strong').fadeOut(300, function () {
            $this = $(this);
            wallets.unshift($this.text());
            $this.text(wallets.join(", "));
            $this.fadeIn(300);
        });
```

wiffbi (changes requested):
Looks good, just some questions regarding aysnc/await. I know when using Promises and async/await, we can use let const – just to make sure, as in other JS-code we do not (yet) use it. Should we somehow guard this, when embedding this script – e.g. `<script type="module">`?

wiffbi:
As discussed I changed the walletdetection to a Promise-based approach. Furthermore I changed it so it does not block and waits for loading/checking all wallets, but async-checks each wallet and adds it as soon as the availability is known for each wallet individually. It does so by adding spinning loader `.fa-cog` to the labels and replaces and fades in or removes them as soon availability is known. To not show mulitple loading indicators, CSS hides all adjacent indicators.

pc-coholic:
Love it, thank you :-)

src/pretix/base/middleware.py
```diff
@@ -249,7 +249,7 @@ def process_response(self, request, resp):
 
         h = {
             'default-src': ["{static}"],
-            'script-src': ['{static}', 'https://checkout.stripe.com', 'https://js.stripe.com'],
+            'script-src': ['{static}', 'https://checkout.stripe.com', 'https://js.stripe.com', 'https://pay.google.com'],
```
raphaelm, line 252: LGTM, but I wonder… why do we even have stripe hardcoded here in a core module? :thinking:

src/pretix/base/payment.py
```diff
@@ -78,6 +78,16 @@
 logger = logging.getLogger(__name__)
 
 
+class WalletQueries:
```
raphaelm, line 81: This looks almost like an enum or almost like a models.TextChoices but isn't – is that on purpose?

src/pretix/plugins/stripe/payment.py
```diff
@@ -747,6 +749,13 @@ class StripeCC(StripeMethod):
     public_name = _('Credit card')
     method = 'cc'
 
+    @property
+    def walletqueries(self):
+        # ToDo: Check against Stripe API, if ApplePay and GooglePay are even activated/available
```
raphaelm, line 754: Should we offer a checkbox in pretix payment settings in the meantime, like we do for other payment methods as well?

src/pretix/static/pretixpresale/js/walletdetection.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+}
+
+$(function () {
+    const wallets = $('[data-wallets]')
+        .map(function(index, pm) {
+            return pm.getAttribute("data-wallets").split("|");
+        })
+        .get()
+        .flat()
```
raphaelm, line 49: Do we need another call to make this list unique, so we don't load the Google Pay SDK multiple times in the worst case?

src/pretix/base/middleware.py
pc-coholic: We can absolutely move that to the stripe signal - I'm assuming this is a leftover of when we didn't implement custom CSPs that much...

src/pretix/plugins/stripe/payment.py
pc-coholic: So just a "check for Google Pay/Apple Pay availability"-checkbox?

src/pretix/base/payment.py
pc-coholic: That was freestyled after PAYMENT_STATES... We can change this to make it look different or even drop it completely: It was just meant as a fixture to make it easier for plugins to use the correct string when requesting wallets...

src/pretix/static/pretixpresale/js/walletdetection.js
pc-coholic: Isn't that happening in the next statement (`.filter()`) is doing?

raphaelm: oh, i get it now. nevermind

src/pretix/plugins/stripe/payment.py
raphaelm: yeah

src/pretix/base/payment.py
raphaelm: I'd like a enum.StrEnum more, but that only appeard in py3.11, so we can also keep it like this for now

src/pretix/base/middleware.py
pc-coholic: Will do.

But good news: I think we can drop checkout.stripe.com - as this seems to be a leftover of long forgotten times, when we actually offered Stripe Checkout :)

PR #3454, opened 2023-07-05
banktransfer: show swiss QR bill code for non-CH IBANs, too

stapelberg:
I have bought tickets to two events that use pretix recently, and both times I could not do the bank transfer with my online banking app, because the swiss QR bill code was not displayed.

The current check does not make sense: A swiss QR bill code should be presented to guests with a swiss bank account, not just for events that have a swiss bank account.

### DIFF
--- a/src/pretix/plugins/banktransfer/payment.py
+++ b/src/pretix/plugins/banktransfer/payment.py
@@ -479,7 +479,7 @@ class BankTransfer(BasePaymentProvider):
         return t
 
     def swiss_qrbill(self, payment):
-        if not self.settings.get('bank_details_sepa_iban') or not self.settings.get('bank_details_sepa_iban')[:2] in ('CH', 'LI'):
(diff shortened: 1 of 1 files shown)


raphaelm:
I remember the QR bill specification to explicitly forbid non-Swiss IBANs to be encoded, but I'd need to check that again after my vacation

raphaelm:
https://www.six-group.com/dam/download/banking-services/standardization/qr-bill/ig-qr-bill-v2.2-de.pdf

page 27 says

> nur IBANs mit CH- oder LI-Landescode zulässig.

stapelberg:
Oh! You’re right. Sorry for the noise. I verified in the ZKB and PostFinance apps, and indeed, codes with a german IBAN are detected as “invalid code”. Neither app supports EPC-QR or bezahlcode unfortunately. I’ll take it up with the banks :)

PR #3470, opened 2023-07-11
Add signals for customer account creation and sign in

kiancross:
Adds signals for customer account creation and sign in.

https://github.com/pretix/pretix/discussions/3455#discussioncomment-6419843

### DIFF
--- a/src/pretix/base/signals.py
+++ b/src/pretix/base/signals.py
@@ -787,3 +787,19 @@ return a dictionary mapping names of attributes in the settings store to DRF ser
 
 As with all event-plugin signals, the ``sender`` keyword argument will contain the event.
 """
+
+customer_created = GlobalSignal()
+"""
+Arguments: ``customer``
+
+This signal is sent out every time a customer account is created. The ``customer``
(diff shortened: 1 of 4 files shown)


src/pretix/base/signals.py
```diff
@@ -787,3 +787,19 @@ class and point to a valid JPEG or PNG file. If no image is available, ``evaluat
 
 As with all event-plugin signals, the ``sender`` keyword argument will contain the event.
 """
+
+customer_created = GlobalSignal()
+"""
+Arguments: ``customer``
+
+This signal is sent out every time a customer account is created. The ``customer``
+object is given as the first argument.
```
raphaelm, line 796: We should document what the ``sender`` will be, and I'd argue we should send the ``Organizer``. That way, we have a path towards future compatibility if we ever add better support for organizer-level plugins

raphaelm:
Thanks a lot!

raphaelm:
@cla-bot[bot] check

raphaelm (approved):

PR #3472, opened 2023-07-12
Badges: reduce memory usage when placing multiple per page (Z#23125583)

wiffbi:
This PR tries to optimize memory usage when placing multiple badges on one page. Previously the whole single-page PDF was held in memory as well as the NUP-PDF. This PR changes it so the NUP-PDF is split into several tempfiles (currently every 100 pages, not sure what a good value for this is though), then releases the single-page PDF from memory and joins the NUP-tempfiles into one big PDF.

### DIFF
--- a/src/pretix/plugins/badges/exporters.py
+++ b/src/pretix/plugins/badges/exporters.py
@@ -63,7 +63,7 @@ from reportlab.pdfgen import canvas
 from pretix.base.exporter import BaseExporter
 from pretix.base.i18n import language
 from pretix.base.models import Order, OrderPosition, Question, QuestionAnswer
-from pretix.base.pdf import Renderer
+from pretix.base.pdf import merge_background, Renderer
 from pretix.base.services.export import ExportError
 from pretix.base.settings import PERSON_NAME_SCHEMES
 from pretix.helpers.templatetags.jsonfield import JSONExtract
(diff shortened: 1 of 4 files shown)


src/pretix/plugins/badges/exporters.py
```diff
@@ -225,10 +227,43 @@ def render_pdf(event, positions, opt):
     # place n-up badges/pages per page
     badges_pdf = PdfReader(outbuffer)
     nup_pdf = PdfWriter()
+    nup_pdf.add_metadata({
+        '/Title': 'Badges',
+        '/Creator': 'pretix',
+    })
     nup_page = None
+    max_nup_pages = 100
```
raphaelm, line 235: I'd actually rather err on the lower end, maybe 20. If the customer uploads a background layout with images, a page can be already quite large. Or do you see a big downside to that?

wiffbi: Yeah, you are right, for large backgrounds this gets big very quickly. I thought about how to better handle these cases instead of a static number, maybe call `sys.getsizeof(nup_pdf)` and split on a certain size (but which?). But probably 20 pages is enough as well. I just did not want to shift too much from memory-heavy to disk-io-heavy.

PR #3473, opened 2023-07-12
Send mail on payment failure [Z#23122835]

ser8phin:
Todo:

- [x] Write mail template
- [x] Call send
- [x] Add template to settings
- [x] Add logging
- [x] Write tests

### DIFF
--- a/src/pretix/base/settings.py
+++ b/src/pretix/base/settings.py
@@ -2304,6 +2304,22 @@ You can view the payment information and the status of your order at
 Best regards,  
 Your {event} team"""))  # noqa: W291
     },
+    'mail_subject_order_payment_failed': {
+        'type': LazyI18nString,
+        'default': LazyI18nString.from_gettext(gettext_noop("Payment failed: {code}")),
+    },
+    'mail_text_order_payment_failed': {
+        'type': LazyI18nString,
+        'default': LazyI18nString.from_gettext(gettext_noop("""Hello,
+
+the initiated payment for your order for {event} failed.
+
+You can retry the payment and view the status of your order at
+{url}
+
+Best regards,  
+Your {event} team"""))  # noqa: W291
+    },
     'mail_subject_waiting_list': {
         'type': LazyI18nString,
         'default': LazyI18nString.from_gettext(gettext_noop("You have been selected from the waitinglist for {event}")),
--- a/src/pretix/base/settings.py
+++ b/src/pretix/base/settings.py
@@ -2306,7 +2306,7 @@ Your {event} team"""))  # noqa: W291
     },
     'mail_subject_order_payment_failed': {
         'type': LazyI18nString,
-        'default': LazyI18nString.from_gettext(gettext_noop("Payment failed: {code}")),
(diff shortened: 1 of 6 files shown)


src/tests/base/test_orders.py
```diff
@@ -704,6 +704,47 @@ def test_sent_not_immediately_after_purchase(self):
...
+                price=Decimal("23.00"), attendee_name_parts={'full_name': "Peter"}, positionid=1
+            )
+            djmail.outbox = []
+
+    @classscope(attr='o')
+    def test_send_payment_fail_mail(self):
+        payment = self.order.payments.create(state=OrderPayment.PAYMENT_STATE_PENDING, amount=self.order.total)
+        payment.save()
+        payment.fail()
+        assert len(djmail.outbox) == 1
```
wiffbi, line 737: Would it make sense to save the len of outbox before failing and then compare whether another email was sent? That way the test should not fail, if we decide to add another email when creating a payment. So:

```suggestion
        nr_of_mails = len(djmail.outbox)
        payment.fail()
        assert len(djmail.outbox) - nr_of_mails == 1
```

src/tests/base/test_orders.py
```diff
@@ -704,6 +704,47 @@ def test_sent_not_immediately_after_purchase(self):
...
+        assert "fail" in djmail.outbox[0].subject
+
+    @classscope(attr='o')
+    def test_no_payment_fail_mail_setting(self):
+        payment = self.order.payments.create(state=OrderPayment.PAYMENT_STATE_PENDING, amount=self.order.total)
+        payment.save()
+        payment.fail(send_mail=False)
```
wiffbi, line 744: See above regarding counting mails before fail.

wiffbi (approved):
Looks good, just a minor comment regarding tests and counting mails.

src/pretix/base/settings.py
```diff
@@ -2301,6 +2301,22 @@ def unserialize(cls, s):
...
+        'default': LazyI18nString.from_gettext(gettext_noop("Payment failed for your order: {code}")),
+    },
+    'mail_text_order_payment_failed': {
+        'type': LazyI18nString,
+        'default': LazyI18nString.from_gettext(gettext_noop("""Hello,
+
+the initiated payment for your order for {event} failed.
+
+You can retry the payment and view the status of your order at
+{url}
```
raphaelm, line 2318:
```suggestion
your payment attempt for your order for {event} has failed.

Your order is still valid and you can try to pay again using the same or a different payment method. Please complete your payment before {expire_date}.

You can retry the payment and view the status of your order at
{url}
```

src/tests/base/test_orders.py
raphaelm: There is an ``djmail.outbox = []`` in ``setUp`` which I think should be good enough, it's very unlikely we ever send email on model creation

raphaelm (approved):

src/tests/base/test_orders.py
ser8phin: I think that will be good enough too :)

PR #3489, opened 2023-07-21
Sendmail rules: Extend filter by check-in state

ser8phin:
- [x] Add options
- [x] Fix naming
- [x] Extend restriction settings
- [x] Check/Test


src/pretix/plugins/sendmail/models.py
```diff
@@ -137,13 +157,15 @@ def send(self):
             status_q |= Q(status=Order.STATUS_PENDING, require_approval=False, valid_if_pending=False,
                           expires__lt=now())
 
+        orders = orders.filter(self.filter_checkins(limit_products))
```
wiffbi, line 160: See above regarding working on the queryset directly.

```suggestion
        orders = self.filter_checkins(orders)
```

src/pretix/plugins/sendmail/models.py
```diff
@@ -97,14 +98,33 @@ def recompute(self):
...
+
+        checkin_q = Q()
+        if self.rule.checked_in_status == "no_checkin":
+            checkin_q = ~Exists(
+                Checkin.objects.filter(position__order_id=OuterRef('pk'),
+                                       position_id__in=OrderPosition.objects.filter(filter_positions)))
+        if self.rule.checked_in_status == "checked_in":
+            checkin_q = Exists(Checkin.all.filter(position__order_id=OuterRef('pk'),
+                                                  position_id__in=OrderPosition.objects.filter(filter_positions)))
+        return checkin_q
```
wiffbi, line 126: This would need rewriting to filter directly on `orders` queryset, e.g. like so (also I think we do not need the filter on OrderPosition as that should already be filtered by line 138 ff and 145 ff?):

```
       if self.rule.checked_in_status == "no_checkin": 
           orders = orders.filter(
                Exists(Checkin.objects.filter(position__order_id=OuterRef('pk'))
            )
        elif self.rule.checked_in_status …
            …
        return orders
```

doc/api/resources/sendmail_rules.rst
```diff
@@ -27,6 +27,10 @@ restrict_to_status                    list                       List of order s
                                                                  ``n__valid_if_pending`` for payment pending but already confirmed,
                                                                  and ``n__pending_overdue`` for pending with payment overdue.
                                                                  The default is ``["p", "n__valid_if_pending"]``.
+checked_in_status                     string                     Checked in states that can be used to filter recipients. Valid strings
+                                                                 are: ``all`` (no filtering, default), ``checked_in`` for
+                                                                 sending mails to checked in customers, and ``no_checkin`` to send mails
+                                                                 to all customers who have not checked in.
```
wiffbi, line 33: Not really an issue, I would make the wording and formatting (use of paretheses) more consistent.

```suggestion
checked_in_status                     string                     Checked in status to restrict recipients to. Valid strings
                                                                 are: ``all`` for no filtering (default), ``checked_in`` for limiting to
                                                                 attendees that are or have been checked-in, and ``no_checkin`` for
                                                                 limiting to attendees who have not checked-in.
```

wiffbi (commented):
Looks good! I have two naming/style suggestions: one regarding naming a var and another somewhat bigger style-change regarding filtering directly on the orders-queryset rather than creating and returning a Q-object.

src/pretix/plugins/sendmail/models.py
```diff
@@ -97,14 +98,33 @@ def recompute(self):
             )
             if not datetime_exists(self.computed_datetime):
                 self.computed_datetime = make_aware(
-                    datetime.combine(d, time(hour=st.hour, minute=st.minute, second=st.second, microsecond=0)) + timedelta(hours=1),
+                    datetime.combine(d,
+                                     time(hour=st.hour, minute=st.minute, second=st.second, microsecond=0)) + timedelta(
+                        hours=1),
```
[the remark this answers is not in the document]
ser8phin: Ah, I thought I reverted all the auto-formatted changes, this must've slipped through. Thanks for noticing!

wiffbi:
I think the tests fail due to orders in tests not having any `OrderPosition`s. But instead of changing the tests to make orders have order-positions, you could try to not filter orders on `pk__in=op_qs.values_list …` when no subevent, products-filter or checking-filter is active – that gives the additional benefit of a performance optimization. :)

doc/api/resources/sendmail_rules.rst
```diff
@@ -27,6 +27,10 @@ restrict_to_status                    list                       List of order s
                                                                  ``n__valid_if_pending`` for payment pending but already confirmed,
                                                                  and ``n__pending_overdue`` for pending with payment overdue.
                                                                  The default is ``["p", "n__valid_if_pending"]``.
+checked_in_status                     string                     Checked in status to restrict recipients to. Valid strings
+                                                                 are: ``all`` for no filtering (default), ``checked_in`` for limiting to
+                                                                 attendees that are or have been checked-in, and ``no_checkin`` for
+                                                                 limiting to attendees who have not checked-in.
```
raphaelm, line 33:
```suggestion
checked_in_status                     string                     Check-in status to restrict recipients to. Valid strings
                                                                 are: ``all`` for no filtering (default), ``checked_in`` for limiting to
                                                                 attendees that are or have been checked in, and ``no_checkin`` for
                                                                 limiting to attendees who have not checked in.
```

src/pretix/plugins/sendmail/api.py
```diff
@@ -88,6 +90,11 @@ def validate(self, data):
                 ]:
                     raise ValidationError(f'status {s} not allowed: restrict_to_status may only include valid states')
 
+        checked_in_status = full_data.get('checked_in_status')
+        if checked_in_status not in [state[0] for state in Rule.CHECK_IN_STATUS_CHOICES]:
+            raise ValidationError(
```
raphaelm, line 95: Is this not already validated by REST framework for a field with choices?

raphaelm (commented):
Looking quite good, small cosmetic ideas

doc/api/resources/sendmail_rules.rst
```diff
@@ -27,6 +27,10 @@ restrict_to_status                    list                       List of order s
                                                                  ``n__valid_if_pending`` for payment pending but already confirmed,
                                                                  and ``n__pending_overdue`` for pending with payment overdue.
                                                                  The default is ``["p", "n__valid_if_pending"]``.
+checked_in_status                     string                     Checked in status to restrict recipients to. Valid strings
+                                                                 are: ``all`` for no filtering (default), ``checked_in`` for limiting to
```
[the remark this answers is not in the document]
wiffbi: I thought about that too, although I thought of `"ignore"` instead of `null`. I do like `null` better though – this is more similar to just leaving the query-param out completely (which works as well).

src/pretix/plugins/sendmail/api.py
ser8phin: Yes, yes it does. I now deleted these lines.

doc/api/resources/sendmail_rules.rst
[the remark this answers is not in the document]
ser8phin: This was a bigger change than anticipated, but I agree with you both and changed it to accept `null`

src/pretix/plugins/sendmail/migrations/0005_rule_checked_in_status.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('sendmail', '0004_rule_restrict_to_status'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='rule',
+            name='checked_in_status',
+            field=models.CharField(default='all', max_length=10, null=True),
```
raphaelm, line 16:
```suggestion
            field=models.CharField(max_length=10, null=True),
```

src/pretix/plugins/sendmail/models.py
```diff
@@ -205,6 +219,12 @@ class Rule(models.Model, LoggingMixin):
         (BOTH, _('Both (all order contact addresses and all attendee email addresses)'))
     ]
 
+    CHECK_IN_STATUS_CHOICES = [
+        ("", _("Everyone")),
```
raphaelm, line 223: Does this need to be
```suggestion
        (None, _("Everyone")),
```
as well so None is stored when I select it in the form?

src/pretix/plugins/sendmail/models.py
```diff
@@ -219,6 +239,14 @@ class Rule(models.Model, LoggingMixin):
         default=['p', 'n__valid_if_pending'],
     )
 
+    checked_in_status = models.CharField(
+        verbose_name=_("Restrict to check-in status"),
+        default=None,
+        choices=CHECK_IN_STATUS_CHOICES,
+        max_length=10,
+        null=True,
```
raphaelm, line 247: Do we not need blank=True so the empty option can be selected?

src/pretix/plugins/sendmail/models.py
ser8phin: Yes, it does

raphaelm (approved):

PR #3490, opened 2023-07-21
PDF: Deduplicate list of add-ons

raphaelm:
This is technically not backwards-compatibile, but I'm not able to come up with a situation in which it is not wanted.

### DIFF
--- a/src/pretix/base/pdf.py
+++ b/src/pretix/base/pdf.py
@@ -43,7 +43,7 @@ import subprocess
 import tempfile
 import unicodedata
 import uuid
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 from functools import partial
 from io import BytesIO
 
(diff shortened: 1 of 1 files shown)


ser8phin (approved):
Looks good. Maybe it could be improved by using the × character just as during checkout instead of the letter x? But that's just a tiny thing, all in all this PR seems fine as it is :)

image
image

raphaelm:
Yeah, I did not risk the × character since it might not be supported by some fonts :(

wiffbi:
I know it makes it more complex, but would it make sense to show „1x“ as well, but only when other addons with higher quantites are in the list as well? So either:

Add-On - Tea
Add-On - Bagel

Or, if e.g. more Bagels have been ordered:

1x Add-On - Tea
2x Add-On - Bagel

PR #3492, opened 2023-07-23
Recognise `title` and `template` attributes on `item_forms` signal

kiancross:
The forms returned by the `item_forms` signal can now set the `title` and (optional) `template` attributes, which behave in the same way as `item_formsets`.

This was proposed in https://github.com/pretix/pretix/discussions/3480.

Currently, there are no automated tests for this feature, which is not ideal. I'm happy to work on some tests if you think it's necessary.

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/item/index.html
+++ b/src/pretix/control/templates/pretixcontrol/item/index.html
@@ -187,7 +187,11 @@
                         {% endif %}
                         {% for f in plugin_forms %}
                             {% if f.is_layouts and not f.title %}
-                                {% bootstrap_form f layout="control" %}
+                                {% if f.template %}
+                                    {% include f.template with form=f %}
+                                {% else %}
+                                    {% bootstrap_form f layout="control" %}
+                                {% endif %}
                             {% endif %}
                         {% endfor %}
                     </fieldset>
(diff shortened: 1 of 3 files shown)


src/pretix/control/templates/pretixcontrol/item/index.html
```diff
@@ -186,7 +186,7 @@
                             {% bootstrap_field form.media_type layout="control" %}
                         {% endif %}
                         {% for f in plugin_forms %}
-                            {% if f.is_layouts %}
+                            {% if f.is_layouts and not f.title %}
```
raphaelm, line 189: Hmm, a form with ``f.is_layouts and f.title`` would now not be rendered anywhere, do I see that correctly?

raphaelm (commented):
Thanks! One remark attached. I don't think we need automated tests for this.

src/pretix/control/templates/pretixcontrol/item/index.html
kiancross: That is correct: there are some invalid combinations.

However, rather than fail silently and not render the form, an exception will be thrown for such forms, with a helpful error message:

https://github.com/pretix/pretix/blob/bf40956a2d1f8cbb0efe43eea250e2b7b5bc1313/src/pretix/control/views/item.py#L1302-L1306

kiancross: > However, I believe "template but no title" could and should work with only marginal effort?

I suppose if no `title` is provided, then the form should not have it's own tab, but

```
{% bootstrap_form f layout="control" %}
```

could be replaced with

```
{% include f.template with form=f %}
```

Is this what you have in mind?

raphaelm: Yes, exactly!

kiancross: Done!

raphaelm (approved):

PR #3502, opened 2023-08-02
Sidebar dropdown: remove menu load delay

luelista:
This forces a query of the typeahead menu to immediately occur once the sidebar dropdown is opened by the user, instead of waiting 250ms. 

The delay causes UX problems as users might see that the menu is empty and immediately close it again, not noticing that options would be available after a delay.


ser8phin (approved):
Lgtm :)

PR #3505, opened 2023-08-04
Import: Allow to import "False"-value (Z#23127414)

pc-coholic:
As of right now, it is only possible to import `True` values for Yes/Now-Questions, because `if value` on a bool is always skipped if the value happens to be `False`.

This should remedy the situation.

However with this change, there are only two ways to get a non-`False` value into the database:
- Verbatim `True` --> `True`
- `None` --> `None`/Not answered

Everything else will be `False`; without this change, everything else would have been `None`/Not answered.

### DIFF
--- a/src/pretix/base/orderimport.py
+++ b/src/pretix/base/orderimport.py
@@ -805,7 +805,7 @@ class QuestionColumn(ImportColumn):
                 return self.q.clean_answer(value)
 
     def assign(self, value, order, position, invoice_address, **kwargs):
-        if value:
+        if value is not None:
             if not hasattr(order, '_answers'):
                 order._answers = []
             if isinstance(value, QuestionOption):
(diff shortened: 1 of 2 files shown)


src/pretix/base/orderimport.py
```diff
@@ -805,7 +805,7 @@ def clean(self, value, previous_values):
                 return self.q.clean_answer(value)
 
     def assign(self, value, order, position, invoice_address, **kwargs):
-        if value:
+        if value is not None:
```
raphaelm, line 808: This is the type of change where an accompanying unit test would not only be easy to write, but also be extremely helpful to prevent this issue from surfacing again if we ever refactor the order import :)

raphaelm (changes requested):

raphaelm (approved):

Issue #1089, opened 2018-11-15
Support for SVG images


j-herden:
It would be nice, if I could upload SVG images too.

raphaelm:
I love SVG, but we'd need to carefully evaluate the security implications of this. SVGs can contain JavaScript which could potentially read sensitive data, such as session data of other events running on the same domain. Unless someone knows an established way to sanitize SVGs, this is not likely to be something we could deploy on pretix Hosted.

The [internet](https://security.stackexchange.com/questions/177412/restrict-js-in-svgs) says ``<img>`` tags are pretty safe, but doesn't cite a credible source on it. Also, direct access of the SVG file would still be a problem (that is solved only with using a separate domain for user content, which we do on pretix.eu, but which barely any self-hosted installation does).

j-herden:
I see your difficulties.
I found this page: 
https://stackoverflow.com/questions/4881364/sanitizing-an-svg-document-by-whitelisting-elements

As I understand it, you can remove the script tag from SVG with XSLT.

pc-coholic:
Thats good news - thanks for the information!

Well look into this asap :)

PR #3506, opened 2023-08-04
Allow svg shop header

eltorio:
Since (nearly) all browsers have native support to svg, it might be useful to allow svg from header logo in the presale


pc-coholic:
:eyes: https://github.com/pretix/pretix/issues/1089

pc-coholic:
@cla-bot check

eltorio:
try with this image on a csp enable site and open it directly from your local disk
xss-image

eltorio:
See SVG in Gecko 
https://developer.mozilla.org/en-US/docs/Web/SVG/SVG_as_an_Image
Not found for Chromium based nor webkit based, but tested on Edge 115, Safari 16.5 without xss exploit

raphaelm:
There's still quite a lot to consider here that makes me hesitant. For example, we also use the header image e.g. as a fallback for the social (open graph) image, which will no longer work with SVG.

eltorio:
Because I'm a bit stubborn 🙈,
I changed some things…
* thumbnails of svg are generated with cairosvg (new python dependency 😔). 
* a new templatetag is added "thumbnotsvg" for using the legacy thumb tag if file is not svg otherwize return the original svg file
* all thumb tags return a png sized of the svg image
* the thumbnotsvg tag is only applied in the event and the organizer templates

raphaelm:
Oh wow, I'll appreciate the effort! We'll look into the security risks of running untrusted files through cairo as well as the licensing risk involved with using a LGPL code (I think we usually avoid that) and will get back to you!

johan12345:
So neither SVG nor HiDPI images (#3235) are a viable option at the moment? At least one of these would be great to have, as event logos do still look quite blurry in pretix on HiDPI screens, especially if they contain text or vector graphics.

I found a blogpost about how another Django-based project added SVG support and (claim to) have addressed the security issues: https://wagtail.org/blog/how-we-added-svg-support-to-wagtail-50/

PR #3508, opened 2023-08-04
Typeahead: Add load indicator

luelista:
Adds a loading indicator to the typeahead dropdowns. It is only displayed if loading takes longer than a predefined timeout.

Also some refactoring of duplications in the typeahead code.

### Screenshots 

grafik

grafik


luelista:
(I forgot that you can throttle the connection using dev tools and had the line below temporarily in my code for testing 😁)

grafik

raphaelm (approved):

PR #3511, opened 2023-08-07
WIP: Bank transfer payment: allow customer to send latest invoice via email (Z#207218)

luelista:


src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/pending.html
```diff
@@ -103,3 +104,14 @@
 {% if swiss_qrbill %}
 <link rel="stylesheet" href="{% static "pretixplugins/banktransfer/swisscross.css" %}">
 {% endif %}
+
+<hr>
+<form class="form-inline helper-display-inline" method="post"
+        action="{% eventurl event "plugins:banktransfer:mail_invoice" order=order.code secret=order.secret %}">
+    {% csrf_token %}
+    <input type="email" name="email" class="form-control" value=""/>
```
raphaelm, line 112: This input field could probably do with
- a `required` attribute
- a `<label>`, even if you hide it with `class="sr-only"`, to make screenreader users happy

src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/pending.html
```diff
@@ -103,3 +104,14 @@
...
+    <button class="btn btn-default">
+        {% trans "Send invoice via E-Mail" %}
```
raphaelm, line 114: We're not as consistent with this as we should be, but we tend to use:

```suggestion
        {% trans "Send invoice via email" %}
```

src/pretix/plugins/banktransfer/views.py
```diff
@@ -886,3 +891,23 @@ def get_object(self, *args, **kwargs):
...
+        if not self.order:
+            raise Http404(_('Unknown order code or not authorized to access this order.'))
+
+        last_payment = self.order.payments.last()
+        last_invoice = self.order.invoices.last()
+        if not last_payment or not last_invoice:
+            messages.error(request, _('No payment or invoice found, please request an invoice first.'))
+            return redirect(self.get_order_url())
+        provider = last_payment.payment_provider
+        provider.send_invoice_to_alternate_email(self.order, last_invoice, request.POST['email'])
```
raphaelm, line 908: There is no validation going on here that the email address is in fact a valid email address. We should at least use `EmailValidator()` from Django, or (maybe even more "clean", maybe overkill, depending on taste) use a django form class.

luelista: I added a validation using validate_email. 

I'd say a Django Form is not necessary as it is only one field and it is also validated by type=email client-side, so a low probability that a user will experience a server-side validation error. 

(Also, I'm not sure how a Django Form could be implemented here, because form and validation are in different views. I guess we could store the validation errors in the session, redirect back to the order view, and display them there. I think this would also unnecessarily complicate the payment_pending_render method, because we'd need to instantiate the Form there as well.)

raphaelm: > I'd say a Django Form is not necessary as it is only one field and it is also validated by type=email client-side, so a low probability that a user will experience a server-side validation error.

I'm totally okay with that.

> (Also, I'm not sure how a Django Form could be implemented here, because form and validation are in different views. I guess we could store the validation errors in the session, redirect back to the order view, and display them there? However, I think this would unnecessarily complicate the payment_pending_render method, because we'd need to instantiate the Form there as well.)

Uh yeah, that's maybe too complicated, yes.

src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/pending.html
```diff
@@ -103,3 +104,20 @@
 {% if swiss_qrbill %}
 <link rel="stylesheet" href="{% static "pretixplugins/banktransfer/swisscross.css" %}">
 {% endif %}
+
+{% if has_invoices %}
```
raphaelm, line 108: Should we maybe also only show this if the `invoice_email` setting on the payment provider has been set? At least to me it makes sense if the email option during checkout and the email option after checkout are controlled by the same setting.

src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/pending.html
```diff
@@ -103,3 +104,20 @@
...
+        {% csrf_token %}
+        <p>
+            {% blocktrans trimmed %}
+                To send the invoice directly to your accounting department, please enter their email address:
+            {% endblocktrans %}
+        </p>
+        <label for="mail_invoice_email" class="sr-only">{% trans "Invoice recipient email" %}:</label>
+        <input type="email" name="email" id="mail_invoice_email" class="form-control" value="" required />
+        <button class="btn btn-default">{% trans "Send invoice via email" %}</button>
+    </form>
```
raphaelm, line 121: Here's a suggestion that I think looks a little cleaner, but it's probably a matter of taste – take it or leave it :)

```suggestion
    <form method="post" action="{% eventurl event "plugins:banktransfer:mail_invoice" order=order.code secret=order.secret %}">
        {% csrf_token %}
        <p>
            {% blocktrans trimmed %}
                To send the invoice directly to your accounting department, please enter their email address:
            {% endblocktrans %}
        </p>
        <div class="row">
            <div class="col-md-9 col-xs-12">
                <label for="mail_invoice_email" class="sr-only">{% trans "Invoice recipient email" %}:</label>
                <input type="email" name="email" id="mail_invoice_email" class="form-control" value="" required
                    placeholder="{% trans "Email address" %}" />
            </div>
            <div class="col-md-3 col-xs-12">
                <button class="btn btn-default btn-block">
                    <span class="fa fa-envelope-o" aria-hidden="true"></span>
                    {% trans "Send invoice via email" %}
                </button>
            </div>
        </div>
    </form>
```

src/pretix/plugins/banktransfer/views.py
```diff
@@ -886,3 +892,27 @@ def get_object(self, *args, **kwargs):
...
+            raise Http404(_('Unknown order code or not authorized to access this order.'))
+        try:
+            validate_email(request.POST['email'])
+        except ValidationError:
+            messages.error(request, _('Please enter a valid email address.'))
+            return redirect(self.get_order_url())
+
+        last_payment = self.order.payments.last()
+        last_invoice = self.order.invoices.last()
+        if not last_payment or not last_invoice:
```
raphaelm, line 910: Technically, we should probably validate that the payment is (a) a bank transfer and (b) in pending state. (And if we put this behind the setting, as suggested above, we'd also need to check that here.)

raphaelm (commented):
This looks very neat already and almost done! :partying_face: I have added some minor suggestions where I think we can make it more consistent with the rest of the system, but generally I'm very happy about the implementation :)

src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/pending.html
luelista: I think this way we can even get rid of one of the `<hr>`s

grafik

raphaelm: Yep, agreed, the hr is not necessary then

src/pretix/plugins/banktransfer/views.py
luelista: How do you usually handle these error cases, which in theory should never be visible to users, because they only would be triggered if somebody manually called the endpoint despite the form being hidden? E.g., do you use translated strings for the error messages?

raphaelm: We'd usually use a `messages.error` + redirect as well, in some cases a `Http404`/`PermissionDenied`.

On one hand, it's additional work translating a string for them, on the other hand, they are not really **never** shown to users. In this case, they might e.g. be shown if someone has an outdated tab open in their browser and then submits the form from there :)

luelista: Good point, then I'll use translated strings there as well

src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/pending.html
luelista: Sounds reasonable, especially as the config options for the mail text are only visible if invoice_email is set.

raphaelm (approved):

raphaelm:
Looks good, the test needs to be adjusted with the correct payment state, I think

PR #3512, opened 2023-08-07
Cart: Disable sneak peek on very small carts

raphaelm:
Recently, we have introduced a "almost collapsed" state for the cart:

image

This however feels odd if the fully expanded cart is only a few pixels larger than the collapsed one. Therefore this PR adds a heuristic to only collapse when the full cart would take more than 200px


pc-coholic:
On mobile (Chrome devtools with mobile view) , this will still cause the same effect, as the box is 214.xx px. So perhaps increasing to 215 might be a good middle ground?

Or leave at 200 but move the "Items in your cart are reserved for xx Minutes" out of the collapsing section?

PR #3517, opened 2023-08-09
Allow PDF variables to provide a bulk evaluation method

raphaelm:


pc-coholic (approved):
LGTM; works with pretix_exhibitors

wiffbi:
As discussed, I reverted this PR/commit on master. It raises an JSON-not-serializable error when scanning a ticket with a voucher that is connected with an exihibitor. See error 85bee2b4c91d43f3800078e2cc2789ba.

PR #3518, opened 2023-08-10
QR code generator for voucher URLs and general URLs

raphaelm:


wiffbi (approved):
Looks good, I will swap hostname with netloc and merge. Maybe we want to add an optional parameter `name` that gets added to the downloaded filename? Might be easier to recognize when one downloads multiple qr-codes for e.g. different vouchers. But probably not worth the effort.

PR #3519, opened 2023-08-10
Voucher redemption: Display event title in some cases (Z#23127871)

raphaelm:
If you have an event logo that does not include the event name, and have therefore enabled the "show event title anyways" option in the design settings, users who enter the shop through the voucher redemption page will not know what event they are buying for. This PR tries to fix that in the least invasive way possible.


wiffbi (approved):
Looks good. Adds the subevent’s location that was previously not there, when a voucher on a subevent was redeemed. I think that is okay!

wiffbi:
I will merge tomorrow,

PR #3520, opened 2023-08-11
Tax rules: add custom rules for country subdivision (e.g. state) (Z#23111850)

luelista:
Example:

grafik

Still looking for a better idea for the select box than Unicode box drawing characters ;)

grafik

Todo:
- [x] Tests
- [x] Maybe better select UI
- [x] Fix translation


wiffbi:
Regarding the UI I like that it is a straight-forward approach and it gets the meaning across as well as allows to select the parent (which using optgroups would not allow). The only problem I see is that when a state is selected, one does not see the country. Although I do not like it, maybe it’s not a problem though, as the events that need state-level taxes anyway only operate in one country and then it is mostly only a list of states?

Otherwise instead of adding UTF-8 characters, prefixing every state with its parent country would be way better regarding accessibility – although it does not look as elegant. Also once the dropdown is closed, IMHO it would help to recognise and group with the parent country visible.

Bildschirmfoto 2023-08-14 um 10 54 40

Bildschirmfoto 2023-08-14 um 10 54 51

One could also build a two-dropdown solution – one country, one state if available. Instead of loading states dynamically, I would go with a properly optgrouped select, that gets converted into two selects with JavaScript. But that is more complicated as needed and IMHO does not look better. So I’d go with prefixing the country to every state-option.

src/pretix/control/forms/event.py
```diff
@@ -1428,9 +1430,22 @@ class CountriesAndEU(CachedCountries):
     cache_subkey = 'with_any_or_eu'
 
 
+class CountriesAndEUAndStates(CountriesAndEU):
+    cache_subkey = 'with_any_or_eu_or_states'
+
+    def __iter__(self):
```
wiffbi, line 1436: I do not think caching properly works here – we set the cache_subkey, but the cached result is the same as `CountriesAndEU` and subdivisions are added dynamically. I am not sure whether we have the same performance issues here when sorting the subdivisions, so do we need to add caching here? See https://github.com/pretix/pretix/blob/master/src/pretix/helpers/countries.py#L36

src/pretix/control/forms/event.py
```diff
@@ -1428,9 +1430,22 @@ class CountriesAndEU(CachedCountries):
...
+        for country_code, country_name in super().__iter__():
+            yield country_code, country_name
+            if country_code in COUNTRIES_WITH_STATE_IN_ADDRESS:
+                types, form = COUNTRIES_WITH_STATE_IN_ADDRESS[country_code]
+                yield from sorted(((state.code, country_name + " - " + state.name)
```
wiffbi, line 1441: Not sure we can do much about it, but one thing I stumpled upon: in german translations there is a „USA – Sonstige kleine Inseln“ (United States Minor Outlying Islands) and „Vereinigte Staaten von Amerika“ with each state being „Vereinigte Staaten von Amerika – Alabama“. The thing I stumbled upon is the dash „-“ – with this „USA“ looks like a translation error. On the other hand, if it weren't called „USA – …“ it could be mistaken for being a subdivision of „Vereinigte Staaten von Amerika“ and by that one could assume „Vereinigte Staaten von Amerika“ would apply to the minor outlying islands as well – which it doesn’t as they do not have the same country-code.

TLDR; I played around with different separators, but dash just feels and looks the best. So we just live with „USA – Sonstige kleine Inseln“ looking like a subdivision, but being a „country“ of itself.

wiffbi (commented):
Looks good, not sure about proper caching though. Could you have a look at it?

src/pretix/control/forms/event.py
wiffbi: Hm, I looked at it again and I am unsure, whether we should not translate the states. I am fine with saying states/districts do not need to be translated and most of the time it does not even make a difference, but a few differences especially around US occur, e.g. „Vereinigte Staaten von Amerika - Commonwealth der Nördlichen Marianen“ and „Vereinigte Staaten von Amerika - USA - Sonstige Kleine Inseln“ – the last one being a super strange special case, as we have two entries for the same „thing“: UM and US-UM. As you noted, it practically has no meaning as it won’t be used anyway, but just to make sure we know about it.

Regarding translation and caching I would love to wait for @raphaelm’s feedback. Otherwise I think it is ready to merge. :)

raphaelm: > I think caching is not needed here, because the subdivisions are not translated, and the performance problems only occurred when sorting strings in foreign character sets, right?

I agree, caching should not be required here.

For context, the performance problem **does** occur with English as well, as it occurs whenever `pyuca` is called a lot, which happens in this function:

https://github.com/SmileyChris/django-countries/blob/f24637d32f08beed4941d452d3fef744f1594256/django_countries/__init__.py#L40

This is used in the ``__iter__`` method of ``Countries``:

https://github.com/SmileyChris/django-countries/blob/f24637d32f08beed4941d452d3fef744f1594256/django_countries/__init__.py#L324

In the implementation of this PR we call ``super().__iter__()``, which will already be a cached result, so we are fine.

The performance problem would occur again if we sorted the *states* with `pyuca` support as well, which *would* be more correct if we translate them – but still not as bad as with the countries, since the lists are shorter (although not *that* much).

I think translating state names is not important for now. Yes, of course, we could argue that a German user should be shown "Kalifornien" instead of "California" for US-CA, but given that this is a very advanced backend feature only used by someone familiar with local law of the country, using the "official" name seems to be fine. The "official" name is trivially also the English name for AU, US, and CA, but I checked the lists for BR, MY, and MX and I they also seem to use the "official" name. For example, MX lists "Ciudad de México " and not "Mexico City", even in the English locale. So I guess we're fine for now.

src/pretix/static/schema/tax-rules-custom.schema.json
```diff
@@ -8,8 +8,8 @@
...
...X-VER", "MX-YUC", "MX-ZAC", "FM", "MD", "MC", "MN", "ME", "MS", "MZ", "MM", "NA", "NR", "NP", "NC", "NZ", "NI", "NL", "NE", "NG", "NU", "KP", "NF", "NO", "OM", "AT", "TL", "PK", "PS", "PW", "PA", "PG", "PY", "PE", "PH", "PN", "PL", "PT", "PR", "RE", "RW", "RO", "RU", "BL", "PM", "SB", "ZM", "WS", "SM", "ST", "SA", "SE", "CH", "SN", "RS", "SC", "SL", "ZW", "SG", "SX", "SK", "SI", "SO", "ES", "SJ", "LK", "SH", "KN", "LC", "MF", "VC", "ZA", "SD", "GS", "KR", "SS", "SR", "SZ", "SY", "TJ", "TW", "TZ", "TH", "TG", "TK", "TO", "TT", "TD", "CZ", "TN", "TR", "TM", "TC", "TV", "UG", "UA", "HU", "UY", "UM", "UZ", "VU", "VA", "VE", "AE", "US", "US-AL", "US-AK", "US-AS", "US-AZ", "US-AR", "US-CA", "US-CO", "US-CT", "US-DE", "US-DC", "US-FL", "US-GA", "US-GU", "US-HI", "US-ID", "US-IL", "US-IN", "US-IA", "US-KS", "US-KY", "US-LA", "US-ME", "US-MD", "US-MA", "US-MI", "US-MN", "US-MS", "US-MO", "US-MT", "US-NE", "US-NV", "US-NH", "US-NJ", "US-NM", "US-NY", "US-NC", "US-ND", "US-MP", "US-OH", "US-OK", "US-OR", "US-PA", "US-PR", "US-RI", "US-SC", "US-SD", "US-TN", "US-TX", "US-UM", "US-UT", "US-VT", "US-VI", "US-VA", "US-WA", "US-WV", "US-WI", "US-WY", "GB", "VN", "WF", "CX", "BY", "EH", "CF", "CY"]
```
raphaelm, line 12: Now that this list is less trivial, should we add a comment somewhere (unfortunately JSON doesn't support comments…) on how to re-generate this list if it needs updating?

raphaelm (approved):

src/pretix/static/schema/tax-rules-custom.schema.json
luelista: The list can be generated by going to the tax rule edit page in the browser, making sure at least one custom rule line is visible, and running the following one-liner in the dev tools:

    console.log(JSON.stringify(Array.prototype.map.call($("#id_form-0-country")[0].options,x=>x.value)))

PR #3525, opened 2023-08-12
Add warning about configuration of Celery in development mode to docs

kiancross:
I spent a long time debugging why certain things were not working during development. It ended up being because Celery was configured to use Redis in my `pretix.cfg`. This PR adds a warning in the docs about this.

### DIFF
--- a/doc/development/setup.rst
+++ b/doc/development/setup.rst
@@ -103,6 +103,13 @@ in ``pretix.cfg``. i.e., you should remove anything such as::
     backend=redis://redis:6379/2
     broker=redis://redis:6379/2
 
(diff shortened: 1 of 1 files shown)


raphaelm:
Hmm. But it should work just fine in development, you just need to start a `celery worker` process. (We could document that, if it's currently not in there.)

kiancross:
I don't think that is currently documented. I guess the documentation should say to either remove the Celery config, _or_ start a Celery worker process.

kiancross:
I've added a comment about starting a worker process. However, I don't think this auto-reloads code changes like Django does.

PR #3527, opened 2023-08-14
Cart: make disabling sneak-peek more robust

wiffbi:
This PR improves #3512 in three ways:

1. It removes the doubling of the `$elements` selector
2. It uses `outerHeight()`, which takes margin into account
3. It makes height-calculation more robust as technically details/summary allow for more than one child other than summary, so it uses array-reduce to sum all outerHeights.


wiffbi:
Yes, 2. and 3. are currently not needed, but for future-proofing.
`.show()` does indeed feel somewhat smelly, but that is probably partly due to jQuery’s chaining and the coding style that goes along with it? I actually do not even realize it, it is just select, show, then filter to the ones I actually need and only save the filtered result. It’s just less local vars to keep in mind.

PR #3533, opened 2023-08-17
Feature: Prefill InvoiceAddress type heuristically (Z#23126061)

luelista:
Set the default value for the "Individual" / "Business" field in the invoice address based on which one is more common in the overall orders of that organizer.

### DIFF
--- a/src/pretix/presale/checkoutflow.py
+++ b/src/pretix/presale/checkoutflow.py
@@ -707,7 +707,10 @@ def business_heuristic(organizer):
     if cached_result is None:
         result = InvoiceAddress.objects.filter(order__event__organizer=organizer).aggregate(
             total=Count('*'), business=Sum(Cast('is_business', output_field=models.IntegerField())))
-        is_business = result['business'] / result['total'] >= 0.6
+        if result['business'] and result['total']:
+            is_business = result['business'] / result['total'] >= 0.6
+        else:
+            is_business = False
         organizer.cache.set('checkout_heuristic_is_business', str(is_business), timeout=3600)
         return is_business
     else:
--- a/src/pretix/presale/checkoutflow.py
+++ b/src/pretix/presale/checkoutflow.py
@@ -708,7 +708,7 @@ def business_heuristic(organizer):
         result = InvoiceAddress.objects.filter(order__event__organizer=organizer).aggregate(
             total=Count('*'), business=Sum(Cast('is_business', output_field=models.IntegerField())))
         is_business = result['business'] / result['total'] >= 0.6
-        organizer.cache.set('checkout_heuristic_is_business', str(is_business), timeout=15)
+        organizer.cache.set('checkout_heuristic_is_business', str(is_business), timeout=3600)
         return is_business
     else:
         return cached_result == 'True'
--- a/src/pretix/presale/checkoutflow.py
+++ b/src/pretix/presale/checkoutflow.py
@@ -42,7 +42,9 @@ from django.contrib import messages
 from django.core.exceptions import ImproperlyConfigured, ValidationError
 from django.core.signing import BadSignature, loads
 from django.core.validators import EmailValidator
-from django.db.models import F, Q
+from django.db import models
+from django.db.models import Count, F, Q, Sum
+from django.db.models.functions import Cast
 from django.http import HttpResponseNotAllowed, JsonResponse
 from django.shortcuts import redirect
 from django.utils import translation
@@ -700,6 +702,14 @@ class AddOnsStep(CartMixin, AsyncAction, TemplateFlowStep):
                        sales_channel=request.sales_channel.identifier)
 
 
+def business_heuristic(organizer):
+    print(repr(organizer))
+    result = InvoiceAddress.objects.filter(order__event__organizer=organizer).aggregate(
+        total=Count('*'), business=Sum(Cast('is_business', output_field=models.IntegerField())))
+    print("business_heuristic", organizer.slug, result)
+    return result['business'] / result['total'] >= 0.6
+
+
 class QuestionsStep(QuestionsViewMixin, CartMixin, TemplateFlowStep):
     priority = 50
     identifier = "questions"
@@ -802,7 +812,9 @@ class QuestionsStep(QuestionsViewMixin, CartMixin, TemplateFlowStep):
     @cached_property
     def invoice_form(self):
         wd = self.cart_session.get('widget_data', {})
-        if not self.invoice_address.pk:
+        if self.invoice_address.pk:
+            wd_initial = {}
+        elif wd:
             wd_initial = {
                 'name_parts': {
                     k[21:].replace('-', '_'): v
@@ -817,7 +829,9 @@ class QuestionsStep(QuestionsViewMixin, CartMixin, TemplateFlowStep):
                 'country': wd.get('invoice-address-country', ''),
             }
         else:
-            wd_initial = {}
+            wd_initial = {
+                'is_business': business_heuristic(self.request.organizer),
+            }
         initial = dict(wd_initial)
 
         if self.cart_customer:
--- a/src/pretix/presale/checkoutflow.py
+++ b/src/pretix/presale/checkoutflow.py
@@ -707,7 +707,6 @@ def business_heuristic(organizer):
     if cached_result is None:
         result = InvoiceAddress.objects.filter(order__event__organizer=organizer).aggregate(
             total=Count('*'), business=Sum(Cast('is_business', output_field=models.IntegerField())))
-        print("business_heuristic", organizer.slug, result)
         is_business = result['business'] / result['total'] >= 0.6
         organizer.cache.set('checkout_heuristic_is_business', str(is_business), timeout=15)
         return is_business
--- a/src/pretix/presale/checkoutflow.py
+++ b/src/pretix/presale/checkoutflow.py
@@ -714,7 +714,7 @@ def business_heuristic(event):
             is_business = result['business'] / result['total'] >= 0.6
         else:
             is_business = False
-        event.cache.set('checkout_heuristic_is_business', is_business, timeout=12 * 3600) # 12 hours
+        event.cache.set('checkout_heuristic_is_business', is_business, timeout=12 * 3600)  # 12 hours
         return is_business
     else:
         return cached_result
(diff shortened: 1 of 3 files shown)


src/pretix/presale/checkoutflow.py
```diff
@@ -700,6 +702,18 @@ def post(self, request, *args, **kwargs):
                        sales_channel=request.sales_channel.identifier)
 
 
+def business_heuristic(organizer):
+    cached_result = organizer.cache.get('checkout_heuristic_is_business')
+    if cached_result is None:
+        result = InvoiceAddress.objects.filter(order__event__organizer=organizer).aggregate(
+            total=Count('*'), business=Sum(Cast('is_business', output_field=models.IntegerField())))
```
wiffbi, line 709: Do we really have to count all invoices ever created for that organizer? What about we just look at e.g. the last 50 orders?

wiffbi (commented):
Looks good already! Just a few ideas regarding adding the event to the heuristic and maybe not look at all invoices created ever, but just the last X.

One thing I did not check: do InvoiceAddress exist for every order or are they only created, when an address is given? Because I would assume that adding an invoice address is probably something that is done more often for businesses than private customers, so that heruistic might be wrong?

src/pretix/presale/checkoutflow.py
```diff
@@ -700,6 +702,18 @@ def post(self, request, *args, **kwargs):
...
+        is_business = result['business'] / result['total'] >= 0.6
+        organizer.cache.set('checkout_heuristic_is_business', str(is_business), timeout=15)
```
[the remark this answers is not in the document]
luelista: String is because the cache API seems to only accept strings.

The 15s was from testing and then forgotten, I'll change it to ... maybe an hour?

src/pretix/presale/checkoutflow.py
```diff
@@ -700,6 +702,18 @@ def post(self, request, *args, **kwargs):
                        sales_channel=request.sales_channel.identifier)
 
 
+def business_heuristic(organizer):
```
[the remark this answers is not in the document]
luelista: I also thought about it and I'm not sure which is better, therefore I just kept it the way Raphael did in his example. Advantage of using the organizer is that if the organizer has previous events, there is data immediately for new events. One could add a condition to only use the organizer's data if the event has too few orders.

src/pretix/presale/checkoutflow.py
luelista: Depending on how the indices are set, it might actually be faster to count all of them than doing date comparison? But I don't have a Pretix installation with enough data in it to perform meaningful tests here. 

And if it's not extremely slow, I would rather set a longer cache timeout to fix performance hits.

wiffbi: Apart from perfomance, which a cache might help, my reasoning that also regards including the event applies here as well: if we look at all orders ever created, what if the organizer had one large event for businesses a long time ago, but switched to doing smaller events for a personal audience? I wouldn’t do a date-comparison, but just a `limit`, ordered desc by primary key (not sure about the indices though, but wouldn’t the Cast and Sum mean the DB has to look at every row anyways? So looking at 50 would be better than all of them – but you are right, cache would help, but then my case regarding big followed by small events applies).

I am still thinking about a clever way to prefer orders for the event over all other orders of that organizer. I just don’t get a better solution than a union, which seems a bit too complicated and not worth the effort – usually I’d expect an organizer to sell only few events at the same time, whereas all of them are targeted to the same audience. I am not sure what is better: to keep it the way it currently is (is_business=False), when we do not have enough orders for the event or always base it on any past order for the organizer.

src/pretix/presale/checkoutflow.py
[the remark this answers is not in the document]
wiffbi: Yep, that might be the right middle-way. Use the event’s orders if we have enough (whatever that is), else use any orders for the organizer (whether all or just the last X – see discussion below).

src/pretix/presale/checkoutflow.py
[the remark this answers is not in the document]
wiffbi: Ah, sorry, it is not django cache, but a custom one for isolation.

An hour sound reasonable. But @raphaelm probably has input on it as well. :)

src/pretix/presale/checkoutflow.py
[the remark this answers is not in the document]
raphaelm: Exactly, I suggested organizer so we don't have confusion because the logic resets everytime a new event is created. But yes, there are some organizers who have separate events that are heavily business or heavily personal.

So yeah, maybe "use event data if there are more than 100 orders on the event, otherwise use organizer data"?

src/pretix/presale/checkoutflow.py
raphaelm: > Depending on how the indices are set, it might actually be faster to count all of them than doing date comparison?

I just ran the query on an organizer with 500k invoice addresses, with the following result:

```
In [10]: %timeit InvoiceAddress.objects.filter(order__event__organizer=organizer).aggregate(total=Count('*'), busi
    ...: ness=Sum(Cast('is_business', output_field=models.IntegerField())))
511 ms ± 20.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

Half a second is a significant time to spend during a request, but if we cache it for long enough, I think it's okay? If we go with the approach from the other comment of checking on event level first, here's the data for an event with 43k addresses within that organizer:

```
In [17]: %timeit InvoiceAddress.objects.filter(order__event=event).aggregate(total=Count('*'), business=Sum(Cast('
    ...: is_business', output_field=models.IntegerField())))
169 ms ± 2.2 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
```

And only if that number were low, we would use the query across the full organizer.

It's still not great, but I don't see direct optimization opportunities in the execution plan without changing the schema (a direct invoiceaddress→event key might help, but probably not worth it):

```
                                                                                          QUERY PLAN                                                                                           
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
 Finalize Aggregate  (cost=157690.67..157690.68 rows=1 width=16) (actual time=523.750..526.649 rows=1 loops=1)
   ->  Gather  (cost=157690.45..157690.66 rows=2 width=16) (actual time=457.932..526.632 rows=3 loops=1)
         Workers Planned: 2
         Workers Launched: 2
         ->  Partial Aggregate  (cost=156690.45..156690.46 rows=1 width=16) (actual time=471.633..471.634 rows=1 loops=3)
               ->  Nested Loop  (cost=0.86..156666.30 rows=3220 width=1) (actual time=6.372..465.820 rows=176604 loops=3)
                     ->  Nested Loop  (cost=0.43..153101.09 rows=7266 width=4) (actual time=5.730..146.842 rows=269305 loops=3)
                           ->  Parallel Seq Scan on pretixbase_event  (cost=0.00..5776.73 rows=53 width=4) (actual time=5.712..11.123 rows=42 loops=3)
                                 Filter: (organizer_id = REDACTED)
                                 Rows Removed by Filter: 20614
                           ->  Index Scan using pretixbase_order_event_id_fcc6f9c6 on pretixbase_order  (cost=0.43..2768.18 rows=1152 width=8) (actual time=0.005..2.953 rows=6463 loops=125)
                                 Index Cond: (event_id = pretixbase_event.id)
                     ->  Index Scan using pretixbase_invoiceaddress_order_id_key on pretixbase_invoiceaddress  (cost=0.43..0.49 rows=1 width=5) (actual time=0.001..0.001 rows=1 loops=807916)
                           Index Cond: (order_id = pretixbase_order.id)
 Planning Time: 0.291 ms
 JIT:
   Functions: 44
   Options: Inlining false, Optimization false, Expressions true, Deforming true
   Timing: Generation 1.317 ms, Inlining 0.000 ms, Optimization 0.852 ms, Emission 14.836 ms, Total 17.005 ms
 Execution Time: 527.173 ms
(20 rows)
```

> And if it's not extremely slow, I would rather set a longer cache timeout to fix performance hits.

Agreed. I think a 8h or 12h timeout would even be fine here.

src/pretix/presale/checkoutflow.py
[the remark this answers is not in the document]
raphaelm: I think the cache would accept booleans :thinking: But not worth diving too deep into it for this.

For cache timeout, see other conversation.

src/pretix/presale/checkoutflow.py
luelista: Thanks for running the numbers :) Do you have a script to generate such large datasets for testing purposes, or did you just try it on pretix.eu? I'm asking because I'd like to profile some more (e.g. is it faster to query only the last X orders)

src/pretix/presale/checkoutflow.py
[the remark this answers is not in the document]
luelista: The custom cache has a type annotation that says otherwise, https://github.com/pretix/pretix/blob/master/src/pretix/base/cache.py#L65 

However this annotation seems to be incorrect, as it relies on Django cache internally, which pickles everything, so bool should be okay,

src/pretix/presale/checkoutflow.py
wiffbi: Regarding the data-set: do you have any numbers regarding is_business is True vs. False especially when giving an invoice is optional? I think if giving an invoice-address is optional, then most of the invoice-addresses given, will be is_business==True. That „bias“ will influence the heuristic. One could argue that of course, if that is the case, why should is_business==True not be set initially. On the other hand that could also mean: why do all this fancy heuristic, when 99% of the time that heuristic is biased to is_business==True?

TLDR: maybe the heuristic will be biased and pre-filling is_business is more kind of an „image-setting“ – ie. showing the customer we are more a business-event or more a private event. So maybe it does make more sense to just add a setting? I don’t know.

wiffbi: > Or we should count all orders without an invoice-address as is_business=False?

Sorry, stupid idea. That kind of biases the heuristic in the wrong way, as I said especially on invoice-optional events, most invoices will probably be is_bussiness=True. Hm …

raphaelm: > Do you have a script to generate such large datasets for testing purposes, or did you just try it on pretix.eu? 

I ran it on prod. I have written scripts to create thousands of orders in the past, but it's hard to come up with a realistic dataset for all kinds of use cases. I'll send you an (old) script via Slack.

luelista: We now run the detection logic in a background task, so the 100 or 500ms query delay shouldn't matter anymore. It's cached for a day. 

Running multiple tasks for the same event is prevented by putting a temporary result in the cache for ten seconds before starting the background task.

raphaelm (approved):

PR #3539, opened 2023-08-22
Fix quota cache mixup

raphaelm:
We allow computing quotas in different ways, e.g. with and without ignoring the waiting list. This is necessary since e.g. some backend actions take priority over the waiting list, but the shop does not.

So far, we have only **cached** the results for `count_waitinglist=True` and we have made sure never to read from the cache if `count_waitinglist=False` was requested. However, we have *written* to the cache when `count_waitinglist=False` was set and therefore wrong data was written to the cache.

A simple solution would be to never write to the cache with `count_waitinglist=False`, but this PR fixes this problem by caching both results, although with separate cache keys.

It also introduces a new dependency that was required to test this properly, as I don't want this error to creep back in.

### DIFF
--- a/src/pretix/base/services/quotas.py
+++ b/src/pretix/base/services/quotas.py
@@ -135,6 +135,7 @@ class QuotaAvailability:
 
                 for eventid, evquotas in quotas_by_event.items():
                     d = rc.hmget(f'quotas:{eventid}:availabilitycache{self._cache_key_suffix}', [str(q.pk) for q in evquotas])
+                    print(f'quotas:{eventid}:availabilitycache{self._cache_key_suffix}', [str(q.pk) for q in evquotas], d)
                     for redisval, q in zip(d, evquotas):
                         if redisval is not None:
                             data = [rv for rv in redisval.decode().split(',')]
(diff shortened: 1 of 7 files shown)


wiffbi (commented):
Looks good so far. I think `rebuild_cache` in [items.py](https://github.com/pretix/pretix/blob/3e122e027004455365f4aa31df5eae29952eedb7/src/pretix/base/models/items.py#L1914) need to be adapted to the new cache-keys.

raphaelm:
Good catch! Pushed.

src/pretix/base/services/quotas.py
```diff
@@ -121,17 +127,15 @@ def compute(self, now_dt=None, allow_cache=False, allow_cache_stale=False):
...
-                rc = get_redis_connection("redis")
+                rc = django_redis.get_redis_connection("redis")
                 quotas_by_event = defaultdict(list)
                 for q in [_q for _q in self._queue if _q.id in quota_ids_set]:
                     quotas_by_event[q.event_id].append(q)
 
                 for eventid, evquotas in quotas_by_event.items():
-                    d = rc.hmget(f'quotas:{eventid}:availabilitycache', [str(q.pk) for q in evquotas])
+                    d = rc.hmget(f'quotas:{eventid}:availabilitycache{self._cache_key_suffix}', [str(q.pk) for q in evquotas])
+                    print(f'quotas:{eventid}:availabilitycache{self._cache_key_suffix}', [str(q.pk) for q in evquotas], d)
```
wiffbi, line 138: I guess this came through accidentally with the last commit regarding cache invalidation. :)
```suggestion
```

PR #3540, opened 2023-08-22
Event dashboard: Fix incorrect optimization introduced in 8e9f0f07a

raphaelm:


wiffbi (approved):
Not part of this PR, but wording in the widget might be a bit misleading. `happy` counts all waiting-list-entries, that can be fulfilled. Wording is „available to give to people on waiting list“, which kind of feels wrong when e.g. a product is stocked up again and has a huge quota of e.g. 2000, but only 100 on the waiting-list. Maybe we could change the wording more to the likes of „100 entries on the waiting-list can be fullfilled“- Hm, not really better, but maybe you can come up with something better … It does the right thing now, only wording could be a bit misleading, but that is not part of this PR.

pc-coholic (approved):
To add to Richard's comment: Perhaps we could use something like "x Wartelisten-Einträge können bedient werden"

PR #3542, opened 2023-08-23
Order data export: Add product/variation ID

raphaelm:
Someone will complain about the column order changing, but I'd argue it's worth it – this makes a lot of sense for processing the file


ser8phin (approved):
It is logical to add IDs directly after the names, so, yes, some may complain, but I think this is good :+1:

PR #3543, opened 2023-08-23
Discounts: Allow "buy X to get Y" with different product sets for X and Y

raphaelm:
Sample scenario: You have "student tickets" and "teacher tickets" at different price point. For every 10 students, one teacher should get in for free. This PR allows this :)


src/tests/base/test_pricing_discount.py
```diff
@@ -1012,3 +1012,228 @@ def test_available_until(event, item):
...
+        (item2.pk, None, Decimal('100.00'), False, False, Decimal('0.00')),
+        (item2.pk, None, Decimal('100.00'), False, False, Decimal('0.00')),
+        (item2.pk, None, Decimal('100.00'), False, False, Decimal('0.00')),
+        (item2.pk, None, Decimal('100.00'), False, False, Decimal('0.00')),
+        (item2.pk, None, Decimal('100.00'), False, False, Decimal('0.00')),
+        (item2.pk, None, Decimal('100.00'), False, False, Decimal('0.00')),
+        (item2.pk, None, Decimal('100.00'), False, False, Decimal('0.00')),
+        (item2.pk, None, Decimal('100.00'), False, False, Decimal('0.00')),
+        (item2.pk, None, Decimal('100.00'), False, False, Decimal('0.00')),
+        (item2.pk, None, Decimal('100.00'), False, False, Decimal('0.00')),
```
wiffbi, line 1102: I think it might be better to remove two `item2` to make sure the `condition_limit_products` really applies to both `item` and `item2` – so we have 11 `item2`, which would result in only 2 `item` being reduced, but as `item` is in condition_limit_products as well, in total we have 14 items the condition applies to and so we can reduce 3.
```suggestion
```

wiffbi: Also, I’d raise the price of `item` to something higher than 100, so the discount gets applied due to `benefit_limit_products` and not because of ` benefit_only_apply_to_cheapest_n_matches`.

src/tests/base/test_pricing_discount.py
```diff
@@ -1012,3 +1012,228 @@ def test_available_until(event, item):
...
+        (item.pk, None, Decimal('90.00'), False, False, Decimal('0.00')),
+        (item.pk, None, Decimal('90.00'), False, False, Decimal('0.00')),
+        (item.pk, None, Decimal('90.00'), False, False, Decimal('0.00')),
```
wiffbi, line 1105: Also, I’d raise the price of `item` to something higher than 100, so the discount gets applied due to `benefit_limit_products` and not because of ` benefit_only_apply_to_cheapest_n_matches`.

```suggestion
        (item.pk, None, Decimal('110.00'), False, False, Decimal('0.00')),
        (item.pk, None, Decimal('110.00'), False, False, Decimal('0.00')),
        (item.pk, None, Decimal('110.00'), False, False, Decimal('0.00')),
```

wiffbi: Ah sorry, the `benefit_limit_products` will be tested here properly as `benefit_only_apply_to_cheapest_n_matches` is 2, which would result in far more discounted items if it were not limited to item.

src/tests/base/test_pricing_discount.py
wiffbi: Hm, the more I think about it, we probably need 2 more `item` in position, if discounted items do not count for the condition to match.

```suggestion
        (item.pk, None, Decimal('110.00'), False, False, Decimal('0.00')),
        (item.pk, None, Decimal('110.00'), False, False, Decimal('0.00')),
```

src/tests/base/test_pricing_discount.py
raphaelm: I've made the test clearer with a n=1 and some comments

src/pretix/base/models/discount.py
```diff
@@ -187,6 +197,12 @@ def validate_config(cls, data):
                   'on a minimum value.')
             )
 
+        if data.get('subevent_mode') == cls.SUBEVENT_MODE_DISTINCT and not data.get('benefit_same_products'):
+            raise ValidationError(
+                _('You cannot apply the discount only to bookings of different dates if you are applying the discount '
+                  'to a different set of products than the condition.')
+            )
+
```
wiffbi, line 205: Should we add an error when `benefit_same_products` is not set, but no products are selected in `benefit_limit_products`?

src/pretix/base/models/discount.py
```diff
@@ -286,11 +305,25 @@ def apply(self, positions: Dict[int, Tuple[int, Optional[int], Decimal, bool, De
...
+            benefit_candidates = list(condition_candidates)
+        else:
+            benefit_products = {p.pk for p in self.benefit_limit_products.all()}
+            benefit_candidates = [
+                idx
+                for idx, (item_id, subevent_id, line_price_gross, is_addon_to, voucher_discount) in positions.items()
+                if (
+                    item_id in benefit_products and
+                    (self.condition_apply_to_addons or not is_addon_to) and
+                    (not self.condition_ignore_voucher_discounted or voucher_discount is None or voucher_discount == Decimal('0.00'))
```
wiffbi, line 318: Do we need extra checkboxes for applying benefit to addons and ignoring voucher-discounted? As both checkboxes currently are located in the condition fieldset, maybe one can overlook, that this also applies to the benefits? Not sure though. Or move both checkboxes to the top? Might also need a rename in the code then as it is not `condition_XXX` but also `benefit_XXX`, so maybe rename to `ignore_voucher_discounted` and `apply_to_addons`? On the other hand probably to much hassle.

wiffbi (approved):
Looks good, just minor comments regarding a config-error/warning and regarding where to put the apply-to-addons and ignore-voucher-discounted in the form.

src/pretix/base/models/discount.py
raphaelm: No,

1. validation on many-to-many fields is hard and has pitfalls
2. this is annoying if one deletes the selected product and then naturally arrives in an "invalid" state

PR #3547, opened 2023-08-25
API: Allow organizer-level access of orders and invoices

raphaelm:
This allows to list all orders and invoices within an organizer account, which is very useful to e.g. sync data to a finanical system when you have many events.

For orders, I have disabled all write access to limit complexity. All writes and actions need to be performed on event level. I have not actively disabled detail access, but I have also not documented it, so we can remove it later if we see reason to. The canonical way to access the detail view should be the event-level version.

For invoices, I have not disabled the two write actions, since there was no reason to. If you think we should disable them for consistency reasons, I can do it, otherwise I'd also keep them as just uncodumented, and everything undocumented is not officially supported ;)


src/pretix/api/serializers/order.py
luelista, line 1: The new orders API fails if unpaid orders exist:
```
  File "/home/mira/Repos/pretix/pretix/src/pretix/api/serializers/order.py", line 617, in to_representation
    return build_absolute_uri(instance.event, 'presale:event.order.pay', kwargs={
                              ^^^^^^^^^^^^^^
AttributeError: 'OrderPayment' object has no attribute 'event'
```

Line 617 needs to be:
        return build_absolute_uri(instance.order.event, 'presale:event.order.pay', kwargs={

(Github seems to not support annotations to unchanged blocks, sadly...)

raphaelm: Fixed, thanks!

src/pretix/api/views/order.py
```diff
@@ -219,27 +213,32 @@ def get_queryset(self):
         if 'invoice_address' not in self.request.GET.getlist('exclude'):
             qs = qs.select_related('invoice_address')
 
-        qs = qs.prefetch_related(self._positions_prefetch(self.request))
+        qs = qs.prefetch_related(*self._positions_prefetch(self.request))
```
wiffbi, line 216: AFAICT currently there will allways only be one `Prefetch` returned from `self._positions_prefetch`. Is the list something you added for future updates or something you added, because you thought you need it and turned out you didn’t? I guess that is also where the new vars `items_base` and `subevents_base` come from – they are only used once. It does not hurt, I am just curious.

src/pretix/api/views/order.py
```diff
@@ -695,7 +733,7 @@ def create(self, request, *args, **kwargs):
                 invoice = generate_invoice(order, trigger_pdf=True)
 
             # Refresh serializer only after running signals
-            prefetch_related_objects([order], self._positions_prefetch(request))
+            prefetch_related_objects([order], *self._positions_prefetch(request))
```
wiffbi, line 736: See comment above regarding only one prefetch in that list.

wiffbi (approved):
Looks good, just a question regarding changing Prefetch to a list – that currently AFAICT will only ever contain one Prefetch-object.

src/pretix/api/views/order.py
raphaelm: Yup, these were part of my efforts to support ?pdf_data=true on the new endpoints, which… failed, and I did not roll them back fully, apparently. Will remove from the diff.

PR #3548, opened 2023-08-25
Add bulk operations for orders

raphaelm:
- [x] Tests


wiffbi:
Looks good so far. I think this button-dropdown is an interesting addition as this opens up the possibility to add more options easily. Not sure it is needed here already, but if we add the options to „mark as paid“, „extend payment period“, and „cancel“, it might be needed. Maybe change the wording on the button so something like „Select action …“. I think adding the number of selected items on the main button might be enough and the dropped-down buttons do not need the „(X)“ added – I can add the javascript to filter them, either through a data-attribute or generally. Furthermore, could you try to color just the icons?

When selecting the `__ALL` option, the results-filter does not get applied and the bulk-action will be applied to all results and not all filtered results accross multiple pages. Also, there are a lot of empty hidden inputs from the filter-form, maybe this confuses the filter-application as all filter-options are enable so practically no filter is applied? I did not look deeper into this though yet.

Probably a Firefox-bug: the order-code search-box is pre-focused, but when I click the checkbox in the bulk-select-area, the search-box stays focused and even opens a suggest-dropdown (native, so a Firefox-problem). The same happens, when I have focus on the „Search for…“ input. Not sure what we can do about it other than not autofocus the order-code search-input. Or maybe manually blur the input, if a click happens outside? I am hesitant to do so as that might introduce other issues when interacting with that form-input.

raphaelm:
> Looks good so far. I think this button-dropdown is an interesting addition as this opens up the possibility to add more options easily. Not sure it is needed here already, but if we add the options to „mark as paid“, „extend payment period“, and „cancel“, it might be needed.

Yes, I thought about these three as well, but:

- "Mark as paid" usually takes a payment amount and a payment date, which is not useful in bulk
- "Extend payment period" requires validation that I don't "shorten" the payment period of an order, which is hard in bulk, but probably possible
- "Cancel" felt to dangerous of a footgun, and gets complex if we want to solve the "how to refund" in the same step

> Maybe change the wording on the button so something like „Select action …“. I think adding the number of selected items on the main button might be enough and the dropped-down buttons do not need the „(X)“ added – I can add the javascript to filter them, either through a data-attribute or generally. Furthermore, could you try to color just the icons?

Sounds good.

> When selecting the __ALL option, the results-filter does not get applied and the bulk-action will be applied to all results and not all filtered results accross multiple pages.

That's bad, will double-check.

> Also, there are a lot of empty hidden inputs from the filter-form, maybe this confuses the filter-application as all filter-options are enable so practically no filter is applied? I did not look deeper into this though yet.
Not sure what you mean.

> Probably a Firefox-bug: the order-code search-box is pre-focused, but when I click the checkbox in the bulk-select-area, the search-box stays focused and even opens a suggest-dropdown (native, so a Firefox-problem). The same happens, when I have focus on the „Search for…“ input. Not sure what we can do about it other than not autofocus the order-code search-input. Or maybe manually blur the input, if a click happens outside? I am hesitant to do so as that might introduce other issues when interacting with that form-input.

:shrug:

wiffbi:
As discussed I will look into the frontend stuff and if you want the __ALL-vs-filter thing.

raphaelm:
I made the frontend changes because I had some spare time.

> When selecting the __ALL option, the results-filter does not get applied and the bulk-action will be applied to all results and not all filtered results accross multiple pages. Also, there are a lot of empty hidden inputs from the filter-form, maybe this confuses the filter-application as all filter-options are enable so practically no filter is applied? I did not look deeper into this though yet.

I cannot reproduce this. What filter are you using exactly?

raphaelm:
Ah I think I found it

wiffbi (approved):

PR #3551, opened 2023-08-29
Scheduled emails: Extend filter by subevents (Z#23122902)

ser8phin:
To create automated mail rules for specific subevent only, this adds a selection widget to choose which, only appearing if there are any subevents to select.

Todo:
- [x]  implementation
- [x] tests


src/pretix/plugins/sendmail/migrations/0006_rule_subevent_alter_rule_checked_in_status.py
```diff
@@ -0,0 +1,24 @@
...
+    operations = [
+        migrations.AddField(
+            model_name='rule',
+            name='subevent',
+            field=models.ForeignKey(null=True, on_delete=django.db.models.deletion.CASCADE, to='pretixbase.subevent'),
+        ),
+        migrations.AlterField(
+            model_name='rule',
+            name='checked_in_status',
+            field=models.CharField(default=None, max_length=10, null=True),
```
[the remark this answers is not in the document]
raphaelm: That probably points to a mistake in the last PR, but yes, it's okay to keep it here.

src/pretix/plugins/sendmail/models.py
```diff
@@ -227,6 +227,7 @@ class Rule(models.Model, LoggingMixin):
 
     id = models.BigAutoField(primary_key=True)
     event = models.ForeignKey(Event, on_delete=models.CASCADE, related_name='sendmail_rules')
+    subevent = models.ForeignKey(SubEvent, null=True, on_delete=models.CASCADE)
```
raphaelm, line 230: `CASCADE` feels a little dangerous here, although safer than `SET_NULL`. Still, it's bad if texts just vanish, I guess, so maybe `PROTECT`? But maybe it's also annoying, I'm not sure. Any opinions from other readers?

src/pretix/plugins/sendmail/models.py
```diff
@@ -280,9 +281,10 @@ def save(self, **kwargs):
         if self.event.has_subevents:
             for se in self.event.subevents.annotate(has_sm=Exists(ScheduledMail.objects.filter(
                     subevent=OuterRef('pk'), rule=self))).filter(has_sm=False):
-                sm = ScheduledMail(rule=self, subevent=se, event=self.event)
-                sm.recompute()
-                create_sms.append(sm)
+                if not self.subevent or self.subevent == se:
```
raphaelm, line 284: 1. This **is** a correct implementation, but a little inefficient – there's probably no reason to have a `for` loop over **all** subevents when we already know we only need one?
2. If ``self.subevent`` **changes**, we also need to clean up ScheduledMail instances for other subevents that have previously existed.

src/pretix/plugins/sendmail/models.py
raphaelm: SET_NULL is certainly wrong, yes ;)

src/pretix/plugins/sendmail/migrations/0006_rule_subevent_alter_rule_checked_in_status.py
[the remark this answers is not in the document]
ser8phin: Afraid so, yes. Then this will fix it at least.

src/pretix/plugins/sendmail/models.py
```diff
@@ -292,6 +294,11 @@ def save(self, **kwargs):
         if not is_creation:
             update_sms = []
             for sm in self.scheduledmail_set.prefetch_related('event').select_related('subevent'):
+
+                if self.subevent and self.subevent != sm.subevent:
+                    sm.delete()
```
raphaelm: Should we keep ScheduledMails around that have already been sent?

src/pretix/plugins/sendmail/models.py
```diff
@@ -278,16 +279,22 @@ def save(self, **kwargs):
...
+                    sm = ScheduledMail(rule=self, subevent=se, event=self.event)
+                    sm.recompute()
+                    create_sms.append(sm)
             ScheduledMail.objects.bulk_create(create_sms)
         else:
             ScheduledMail.objects.get_or_create(rule=self, event=self.event)
 
         if not is_creation:
+            if self.subevent:
+                ScheduledMail.objects.filter(rule=self).exclude(subevent=self.subevent).delete()
```
wiffbi, line 296: I move @raphaelm’s comment to this line:

> Should we keep ScheduledMails around that have already been sent?

Probably yes. So we need to extend `exclude` to keep failed and completed.

ser8phin: Sounds very reasonable, added those states to exclusion

src/pretix/plugins/sendmail/models.py
```diff
@@ -278,16 +279,27 @@ def save(self, **kwargs):
...
         else:
             ScheduledMail.objects.get_or_create(rule=self, event=self.event)
 
         if not is_creation:
+            if self.subevent:
+                keep_states = [ScheduledMail.STATE_COMPLETED,
+                               ScheduledMail.STATE_FAILED]  # we keep rules that have already been sent at least partially
+                ScheduledMail.objects.filter(rule=self).exclude(
+                    subevent=self.subevent,
+                    state__in=keep_states
```
wiffbi, line 300: Sorry, I did not catch this earlier, but I think this needs to be subevent=self.subevent OR state__in=keep_states. Should be doable with `Q`-objects. :)

src/pretix/plugins/sendmail/models.py
```diff
@@ -278,16 +279,26 @@ def save(self, **kwargs):
...
+                    sm.recompute()
+                    create_sms.append(sm)
             ScheduledMail.objects.bulk_create(create_sms)
         else:
             ScheduledMail.objects.get_or_create(rule=self, event=self.event)
 
         if not is_creation:
+            if self.subevent:
+                keep_states = [ScheduledMail.STATE_COMPLETED,
+                               ScheduledMail.STATE_FAILED]  # we keep rules that have already been sent at least partially
```
raphaelm, line 297: I just read through ``recompute()`` and I believe COMPLETED is the only state where we have no risk of the rule running ever again.

```suggestion
                keep_states = [ScheduledMail.STATE_COMPLETED]  # we keep rules that have already been sent
```

src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/rule_inspect.html
```diff
@@ -45,7 +45,7 @@ <h1>{% trans "Inspect Email Rule" %}</h1>
                     {% if request.event.has_subevents %}
                         <td>
                             <a href="{% url "control:event.subevent" organizer=request.event.organizer.slug event=request.event.slug subevent=sm.subevent.id %}?returnto={{ request.GET.urlencode|urlencode }}">
-                                {{ sm.subevent.name }}
+                                {{ sm.subevent.name }} - {{ sm.subevent.date_from }}
```
raphaelm, line 48: Outputting unformatted dates is not nice, plus this is redundant with a (broken) line below

```suggestion
                                {{ sm.subevent.name }}
```

src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/rule_inspect.html
```diff
@@ -45,7 +45,7 @@ <h1>{% trans "Inspect Email Rule" %}</h1>
...
                             </a><br>
                             {{ sm.get_date_range_display }}
```
raphaelm, line 50: should be sm.subevent.get_date_range_display :)

raphaelm (commented):
LGTM except for small comments inline

src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/rule_inspect.html
ser8phin: Oh, it was already intended to show, just broken, good then :)

src/pretix/plugins/sendmail/templates/pretixplugins/sendmail/rule_inspect.html
ser8phin: Ah, of course! :)

raphaelm (approved):

PR #3566, opened 2023-09-04
Control: add link to orders for each subevent in list of subevents (Z#23129436)

wiffbi:
This is the most naive approach to add a link to the orders overview to each element of the subevents list. Currently it does not even check the quotas if any orders are available for that subevent – maybe we could add that check and disable the link?

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/subevents/index.html
+++ b/src/pretix/control/templates/pretixcontrol/subevents/index.html
@@ -165,6 +165,8 @@
                                 {% endif %}
                             </td>
                             <td class="text-right flip">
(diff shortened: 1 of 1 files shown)


raphaelm (commented):
Somehow this PR got lost. I'm not sure it's worth it?

a) It's quite special-case, there's lots of other buttons we could add here and don't
b) It leads to a list of *orders* which might contain orders with mixed subevents causing to popular confusion

wiffbi:
Oh sorry, I totally forgot about that myself. I think we can drop it, it was just a test and I was not really happy with it. Sorry!

PR #3568, opened 2023-09-04
Allow to sort the list of check-in lists (Z#23128978)

raphaelm:


ser8phin (approved):
It does what it should, not sure if things could be approved because this is the first time I read deeper into ordering, but :+1:

PR #3570, opened 2023-09-04
Bank transfer: Hide email form if email was already sent (Z#23128941)

raphaelm:


pc-coholic (approved):

src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/pending.html
```diff
@@ -106,7 +106,20 @@
 {% endif %}
 
 {% if invoice_email_enabled and has_invoices %}
-    <form method="post" action="{% eventurl event "plugins:banktransfer:mail_invoice" order=order.code secret=order.secret %}">
+    {% if info.send_invoice_to %}
+        <p>
+            {% blocktrans trimmed with recipient=info.send_invoice_to %}
+                At your request, we are sending the invoice directly to {{ recipient }}.
```
luelista, line 112: Usually when this message is displayed, the invoice was already sent (only exception I can think of is if it's still in the task queue for actually sending the email), right? So we should better use past tense here

luelista, line 112:
```suggestion
                At your request, we sent the invoice directly to {{ recipient }}.
```

src/pretix/plugins/banktransfer/payment.py
```diff
@@ -530,6 +530,7 @@ def payment_pending_render(self, request: HttpRequest, payment: OrderPayment):
             'code': self._code(payment.order),
             'order': payment.order,
             'amount': payment.amount,
+            'info': payment.info_data,
```
luelista, line 533: I would recommend a somewhat less ambiguous variable name than `info`. Either `payment_info`, `info_data` or `payment_info_data`.
```suggestion
            'payment_info': payment.info_data,
```

luelista (approved):
Two remarks, apart from that LGTM :)

src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/pending.html
raphaelm: Yes. I was worried about the task not yet executed or the email not sent successfully, but let's not bother the custoemr with it.

PR #3573, opened 2023-09-05
Payment: Add setting to prevent reminder mails (Z#23123914)

ser8phin:
This adds a payment provider setting to prevent payment reminder mails to be sent if that provider is chosen for payment. The setting is accessed (via method that could be overwritten in specific payment providers if needed) during the periodic task sending out those reminder mails.

### DIFF
--- a/src/pretix/base/services/orders.py
+++ b/src/pretix/base/services/orders.py
@@ -1294,7 +1294,11 @@ def send_expiry_warnings(sender, **kwargs):
     ).only('pk', 'event_id', 'expires').order_by('event_id'):
 
         lp = o.payments.last()
-        if lp and lp.payment_provider.prevent_reminder_mail(o, lp):
+        if (
+                lp and
+                lp.state in [OrderPayment.PAYMENT_STATE_CREATED, OrderPayment.PAYMENT_STATE_PENDING] and
+                lp.payment_provider.prevent_reminder_mail(o, lp)
+        ):
             continue
 
         if event_id != o.event_id:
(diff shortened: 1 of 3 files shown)


src/pretix/base/services/orders.py
```diff
@@ -1289,9 +1289,14 @@ def send_expiry_warnings(sender, **kwargs):
...
     for o in Order.objects.filter(
-        expires__gte=today, expiry_reminder_sent=False, status=Order.STATUS_PENDING,
-        datetime__lte=now() - timedelta(hours=2), require_approval=False
+            expires__gte=today, expiry_reminder_sent=False, status=Order.STATUS_PENDING,
+            datetime__lte=now() - timedelta(hours=2), require_approval=False
     ).only('pk', 'event_id', 'expires').order_by('event_id'):
+
+        lp = o.payments.last()
+        if lp and lp.payment_provider.prevent_reminder_mail(o, lp):
+            continue
```
raphaelm, line 1298: I don't think we should prevent the reminder email if the last payment is e.g. in status `failed`/`canceled` (the user will need to do something to complete the payment) or in status `completed` (the amount of payment is not sufficient and the user needs to do another payment). I think this should only respect payments in status `created` and `pending`.

ser8phin: That makes sense. As a general condition that would better be handled here directly and not in the .prevent_reminder_mail function so that other payment providers could potentially say 'nope, don't care', right?

raphaelm: I was thinking about this as well, and I think it's better handled here. A payment provider could *still* implement their own distinction between created and pending, but we should exclude the "clear-cut cases" here in core

ser8phin: I changed it accordingly, but I want to look at the tests again, because new ones for this didn't work correctly, so I changed the pr into a draft again, please don't merge early

raphaelm (approved):

PR #3576, opened 2023-09-06
Check-in: Show more information

raphaelm:
This adds the necessary APIs to show additional information in the pretixSCAN apps during a checkin:

- A selection of questions through a new attribute ``show_during_checkin``
- A set of custom texts which can be configured on every level that we have `checkin_attention` on (item, variation, order) but that can also be used without `checkin_attention`

- [x] implementation
- [x] tests
- [x] support in web checkin
- [x] support in simulator
- [x] libpretixsync support https://github.com/pretix/libpretixsync/pull/31
  - [x] pretixscan android release https://github.com/pretix/pretixscan-android/pull/77
  - [x] pretixscan desktop release
  - [x] pretixscan proxy release
- [ ] pretixscan-ios support


src/pretix/base/exporters/json.py
```diff
@@ -110,6 +111,7 @@ def render(self, form_data):
                                 'description': str(variation.description),
                                 'position': variation.position,
                                 'checkin_attention': variation.checkin_attention,
+                                'checkin_Text': variation.checkin_text,
```
robbi5, line 114:
```suggestion
                                'checkin_text': variation.checkin_text,
```

robbi5:
Do we want to support the texts in the Search Result (`checkinrpc/search`) too? Then `CheckinListOrderPositionSerializer` needs `checkin_text`(`s`?), or? :)

robbi5:
When doing a checkin in online mode, pretixSCAN would need the `question`s too from the `redeem`-endpoint. Currently `positions.answers` is available, but the (id-referenced) `question`s are only set on status `incomplete`.

raphaelm:
> Do we want to support the texts in the Search Result (`checkinrpc/search`) too? Then `CheckinListOrderPositionSerializer` needs `checkin_text`(`s`?), or? :)

Looking at the intended uses of the texts, I'd argue for "no". It would also make the search results quite long

raphaelm:
> When doing a checkin in online mode, pretixSCAN would need the `questions` too from the `redeem`-endpoint. Currently `position.answers` is available, but the (id-referenced) `questions` are only set on status `incomplete`.

Hmm, yes, we could change that, or rely on them being synced (which they always should be in our apps, right?)

robbi5:
> > Do we want to support the texts in the Search Result (`checkinrpc/search`) too? Then `CheckinListOrderPositionSerializer` needs `checkin_text`(`s`?), or? :)
> 
> Looking at the intended uses of the texts, I'd argue for "no". It would also make the search results quite long

Ah, I misunderstood the usage of SearchResult in the pretixSCAN code, though that data is also used for the redeem result card. Thats not the case - so yeah, let's don't do that.

> > When doing a checkin in online mode, pretixSCAN would need the `questions` too from the `redeem`-endpoint. Currently `position.answers` is available, but the (id-referenced) `questions` are only set on status `incomplete`.
> 
> Hmm, yes, we could change that, or rely on them being synced (which they always should be in our apps, right?)

Well, as an api consumer (like that the web check in now also needs to pull data from two sources) I think it's cleaner that the api response delivers all the needed data that we want to show. The checkin list data, for example, is also returned in a (successful) checkin.

robbi5:
> > Do we want to support the texts in the Search Result (`checkinrpc/search`) too? Then `CheckinListOrderPositionSerializer` needs `checkin_text`(`s`?), or? :)
> 
> Looking at the intended uses of the texts, I'd argue for "no". It would also make the search results quite long

Ah, I misunderstood the usage of SearchResult in the pretixSCAN code, though that data is also used for the redeem result card. Thats not the case - so yeah, let's don't do that.

> > When doing a checkin in online mode, pretixSCAN would need the `questions` too from the `redeem`-endpoint. Currently `position.answers` is available, but the (id-referenced) `questions` are only set on status `incomplete`.
> 
> Hmm, yes, we could change that, or rely on them being synced (which they always should be in our apps, right?)

Well, as an api consumer (like that the web check in now also needs to pull data from two sources) I think it's cleaner that the api response delivers all the needed data that we want to show. The checkin list data, for example, is also returned in a (successful) checkin.

raphaelm:
Ok, will look into it if it's possible in a good way

robbi5:
locally I tried it temporarily by adding the following to the `return Response`-calls in `pretix/api/views/checkin.py`:
```py
'questions': [
   QuestionSerializer(q).data for q in op.item.questions.filter(show_during_checkin=True)
],
```

I still have to check what would be the best way to serve the questions when some have to still be answered on checkin but others already can be shown :)

src/pretix/control/templates/pretixcontrol/global_sysreport.html
pc-coholic, line 1: This file does not seem to belong to this PR

doc/api/resources/item_variations.rst
```diff
@@ -53,6 +55,10 @@ meta_data                             object                     Values set for
 
    The ``meta_data`` and ``checkin_attention`` attributes have been added.
 
+.. versionchanged:: 2023.8
```
pc-coholic, line 58:
```suggestion
.. versionchanged:: 2023.10
```

doc/api/resources/items.rst
```diff
@@ -196,6 +200,10 @@ meta_data                               object                     Values set fo
 
    The ``media_policy`` and ``media_type`` attributes have been added.
 
+.. versionchanged:: 2023.8
```
pc-coholic, line 203:
```suggestion
.. versionchanged:: 2023.10
```

doc/api/resources/orders.rst
```diff
@@ -133,7 +135,7 @@ last_modified                         datetime                   Last modificati
 
 .. versionchanged:: 2023.8
 
-   The ``event`` attribute has been added. The organizer-level endpoint has been added.
+   The ``event`` and ``checkin_text`` attributes has been added. The organizer-level endpoint has been added.
```
pc-coholic, line 138: This should probably a whole new `..versionchanged` that only includes the `checkin_text` (at 2023.10) and leave the `event` to the existing 2023.8.

doc/api/resources/questions.rst
```diff
@@ -77,6 +79,10 @@ dependency_value                      string                     An old version
                                                                  for one value. **Deprecated.**
 ===================================== ========================== =======================================================
 
+.. versionchanged:: 2023.8
```
pc-coholic, line 82:
```suggestion
.. versionchanged:: 2023.10
```

pc-coholic (changes requested):

pc-coholic:
Read through it only - but looks good to me from that POV

PR #3583, opened 2023-09-07
Allow to refund currencies with zero decimals

pc-coholic:
This is in relation to Z#23129921.

This fixes the immediate issue at hand, but we should probably comb through all templates and replace any `|floatformat:2` with `|money_numberfield:request.event.currency` where appropriate?

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/order/index.html
+++ b/src/pretix/control/templates/pretixcontrol/order/index.html
@@ -110,7 +110,7 @@
             <input type="hidden" name="start-action" value="do_nothing">
             <input type="hidden" name="start-mode" value="partial">
             {% localize off %}
-            <input type="hidden" name="start-partial_amount" value="{{ overpaid|floatformat:2 }}">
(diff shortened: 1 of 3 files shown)


raphaelm (approved):
> This fixes the immediate issue at hand, but we should probably comb through all templates and replace any |floatformat:2 with |money_numberfield:request.event.currency where appropriate?

Can you keep track of doing that?

pc-coholic:
> > This fixes the immediate issue at hand, but we should probably comb through all templates and replace any |floatformat:2 with |money_numberfield:request.event.currency where appropriate?
> 
> Can you keep track of doing that?

To move things along quickly and not block our customer, I would propose to merge this and then do a separate PR for all other instances. Would that work for you?

raphaelm:
> To move things along quickly and not block our customer, I would propose to merge this and then do a separate PR for all other instances. Would that work for you?

Yeah absolutely, I just want to make sure we don't each think the other one will do it and then it gets forgotten :D

PR #3586, opened 2023-09-08
Widget: make data-attributes reactive

wiffbi:
When using the widget with additional data (e.g. cookie consent) one needed to make sure, that all needed data-attributes are available when initializing the widgets. This PR adds the reactive data-attributes so one can change/add data-attributes on widgets/buttons which then get passed through to the reactive `widget_data`. This is e.g. useful to add `data-consent` which usually are available async.

Note: once the checkout process is started, changing data-attributes are not reflected onto the iframe (no reload ist done) to not interfere with checkout-UX.

### DIFF
--- a/doc/user/events/widget.rst
+++ b/doc/user/events/widget.rst
@@ -293,6 +293,11 @@ with that information::
     </pretix-widget>
 
 This works for the pretix Button as well, if you also specify a product.
+
+As data-attributes are reactive, you can change them with JavaScript as well. Please note, that once the user
+started the checkout process, we do not update the data-attributes in the existing checkout process to not
+interrupt the checkout UX.
+
 Currently, the following attributes are understood by pretix itself:
 
 * ``data-email`` will pre-fill the order email field as well as the attendee email field (if enabled).
--- a/doc/user/events/widget.rst
+++ b/doc/user/events/widget.rst
@@ -334,11 +334,12 @@ Hosted or pretix Enterprise are active, you can pass the following fields:
 * If you use the campaigns plugin, you can pass a campaign ID as a value to ``data-campaign``. This way, all orders
   made through this widget will be counted towards this campaign.
 
-* If you use the tracking plugin, you can enable cross-domain tracking. Use the HTML code to embed the widget and 
-  add one the following code snippets. Make sure to replace all occurrences of <MEASUREMENT_ID> with your Google 
-  Analytics MEASUREMENT_ID (G-XXXXXXXX)
+* If you use the tracking plugin, you can enable cross-domain tracking. Please note: when you run your pretix-shop on a
+  subdomain of your main tracking domain, then you do not need cross-domain tracking as tracking automatically works
+  across subdomains. See `Using a custom domain <https://docs.pretix.eu/en/latest/user/organizers/domain.html>` for
+  how to set this up.
 
-  Please also make sure to add the embedding website to your `Referral exclusions
+  Please make sure to add the embedding website to your `Referral exclusions
   <https://support.google.com/analytics/answer/2795830>`_ in your Google Analytics settings.
 
   Add Google Analytics as you normally would with all your `window.dataLayer` and `gtag` configurations. Also add the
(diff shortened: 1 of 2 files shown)


src/pretix/static/pretixpresale/js/widget/widget.js
```diff
@@ -1986,8 +2000,19 @@ var create_button = function (element) {
         }
     }
 
+    var observer = new MutationObserver((mutationList) => {
+        mutationList.forEach((mutation) => {
+            if (mutation.type == "attributes" && mutation.attributeName.startsWith("data-")) {
+                Vue.set(app.widget_data, mutation.attributeName, mutation.target.getAttribute(mutation.attributeName));
```
raphaelm, line 2006: Don't we need to strip the `data-` here? At least in create_widget, we do it:

```
    for (var i = 0; i < element.attributes.length; i++) {
        var attrib = element.attributes[i];
        if (attrib.name.match(/^data-.*$/)) {
            widget_data[attrib.name.replace(/^data-/, '')] = attrib.value;
        }
    }
```

wiffbi: Good catch!

doc/user/events/widget.rst
```diff
@@ -329,125 +334,70 @@ Hosted or pretix Enterprise are active, you can pass the following fields:
 * If you use the campaigns plugin, you can pass a campaign ID as a value to ``data-campaign``. This way, all orders
   made through this widget will be counted towards this campaign.
 
-* If you use the tracking plugin, you can enable cross-domain tracking. To do so, you need to initialize the 
-  pretix-widget manually. Use the html code to embed the widget and add one the following code snippets. Make sure to
-  replace all occurrences of <MEASUREMENT_ID> with your Google Analytics MEASUREMENT_ID (UA-XXXXXXX-X or G-XXXXXXXX)
+* If you use the tracking plugin, you can enable cross-domain tracking. Use the HTML code to embed the widget and
```
wiffbi, line 337: Should we add a paragraph that they do not need cross-domain tracking when they run pretix on a subdomain of the main tracking domain? Because then it’s not cross-domain in at least Google Analytics’ terms of cross-domain.

raphaelm: Yes, please!

doc/user/events/widget.rst
```diff
@@ -293,6 +293,11 @@ with that information::
     </pretix-widget>
 
 This works for the pretix Button as well, if you also specify a product.
+
+As data-attributes are reactive, you can change them with JavaScript as well. Please note, that once the user
```
raphaelm, line 297: Maybe make it sound less "obviously" to not suggest other attributes are reactive as well.
```suggestion
Data-attributes are reactive, therefore you can change them with JavaScript as well. Please note, that once the user
```

raphaelm: Also, should we add a note that the mutation needs to be done on the div.pretix-widget-wrapper since the pretix-widget element goes away (at least in firefox)?

raphaelm: Yes, a note would be great, thanks!

doc/user/events/widget.rst
```diff
@@ -334,19 +334,21 @@ Hosted or pretix Enterprise are active, you can pass the following fields:
 * If you use the campaigns plugin, you can pass a campaign ID as a value to ``data-campaign``. This way, all orders
   made through this widget will be counted towards this campaign.
 
-* If you use the tracking plugin, you can enable cross-domain tracking. Use the HTML code to embed the widget and 
-  add one the following code snippets. Make sure to replace all occurrences of <MEASUREMENT_ID> with your Google 
-  Analytics MEASUREMENT_ID (G-XXXXXXXX)
+* If you use the tracking plugin, you can enable cross-domain tracking. Please note: when you run your pretix-shop on a
+  subdomain of your main tracking domain, then you do not need cross-domain tracking as tracking automatically works
+  across subdomains. See `Using a custom domain <https://docs.pretix.eu/en/latest/user/organizers/domain.html>` for
```
raphaelm, line 344: That should be a sphinx crossref, not a fully qualified link:
https://docs.readthedocs.io/en/stable/guides/cross-referencing-with-sphinx.html

wiffbi: Thanks! Should be fixed. :)

PR #3594, opened 2023-09-12
Fix voucher and order views for users with read-only permissions

raphaelm:


src/pretix/control/views/vouchers.py
```diff
@@ -283,6 +283,12 @@ def get_form_class(self):
                 form_class = response
         return form_class
 
+    def get_form(self, form_class=None):
+        form = super().get_form(form_class)
+        for f in form.fields.values():
+            f.disabled = True
```
wiffbi, line 289: Doesn’t this disable all fields even if I have the permission `can_change_vouchers`? I did a quick test and cannot edit any voucher.

wiffbi (changes requested):
Looks good and works as expected. Only editing vouchers does not work anymore – the inputs are always disabled. :(

Not part of this PR, but I realized we do not have the `__ALL` option for batch selecting all vouchers.

src/pretix/control/views/vouchers.py
raphaelm: m) yes, oversight on my end

PR #3595, opened 2023-09-12
Placeholders: insert placeholder into text on click

wiffbi:
When editing email content, placeholders needed to be selected and copied over – to make things harder, a double-click did not select the surrounding {} of placeholders. This PR converts the placeholders into buttons, which insert the placeholder into the selected text input (or the first text input the placeholder-button „belongs“ to). It even helps when replacing existing placeholders by checking the previous and next characters if they are {} – if so, it replaces them as well. This makes it easier to replace placeholders by double-clicking the placeholder in the email content and then hitting the corresponsing button.

Should we go even further an just check if the cursor is inside a placeholder and replace the whole placeholder?

One downside: as the placeholders are buttons now, they are not text-selectable anymore. Even using CSS’s `user-select: text;` it is rather awkward and would currently trigger a click as well. So I kept the button-text unselectable.

### DIFF
--- a/src/pretix/base/forms/widgets.py
+++ b/src/pretix/base/forms/widgets.py
@@ -60,6 +60,16 @@ def replace_arabic_numbers(inp):
     return inp.translate(table)
 
 
+def format_placeholders_help_text(placeholders, event=None):
+    phs = [
+        '<button type="button" class="content-placeholder" title="%s">{%s}</button>' % (v.render_sample(event) if event else v, k)
+        for k, v in placeholders.items()
+    ]
+    return _('Available placeholders: {list}').format(
+        list=', '.join(phs)
+    )
+
+
 class DatePickerWidget(forms.DateInput):
     def __init__(self, attrs=None, date_format=None):
         attrs = attrs or {}
--- a/src/pretix/static/pretixcontrol/js/ui/main.js
+++ b/src/pretix/static/pretixcontrol/js/ui/main.js
@@ -699,6 +699,35 @@ var form_handlers = function (el) {
     el.find("input[name*=question], select[name*=question]").change(questions_toggle_dependent);
     questions_toggle_dependent();
     questions_init_photos(el);
+
+    var lastFocusedInput;
+    $(document).on('focusin', 'input, textarea', function(e) {
+        lastFocusedInput = e.target;
+    }).on("click", function(e) {
+        if (e.target.classList.contains('content-placeholder')) {
+            var container = e.target.parentNode.parentNode;
+            if (!lastFocusedInput || !container.contains(lastFocusedInput)) {
+                lastFocusedInput = container.querySelector("input, textarea");
+                //lastFocusedInput.selectionStart = lastFocusedInput.selectionEnd = lastFocusedInput.value.length;
+            }
+            if (lastFocusedInput) {
+                var start = lastFocusedInput.selectionStart;
+                var end = lastFocusedInput.selectionEnd;
+                var v = lastFocusedInput.value;
+                var p = e.target.textContent;
+                if (start !== end && v.substring(start, start+1) !== "{" && v.substring(start-1, start) == "{" &&
+                    v.substring(end-1, end) !== "}" && v.substring(end, end+1) == "}") {
+                    // double-click on word does not select surrounding {}, so help to replace placeholders
+                    start--;
+                    end++;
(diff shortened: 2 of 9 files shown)


src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -699,6 +699,35 @@ var form_handlers = function (el) {
     el.find("input[name*=question], select[name*=question]").change(questions_toggle_dependent);
     questions_toggle_dependent();
     questions_init_photos(el);
+
+    var lastFocusedInput;
+    $(document).on('focusin', 'input, textarea', function(e) {
+        lastFocusedInput = e.target;
+    }).on("click", function(e) {
+        if (e.target.classList.contains('content-placeholder')) {
+            var container = e.target.parentNode.parentNode;
```
raphaelm, line 708: $(e.target).closest(".form-group") might be more flexible, not sure

src/pretix/base/forms/widgets.py
```diff
@@ -60,6 +60,16 @@ def replace_arabic_numbers(inp):
     return inp.translate(table)
 
 
+def format_placeholders_help_text(placeholders, event=None):
+    phs = [
+        '<button type="button" class="content-placeholder" title="%s">{%s}</button>' % (v.render_sample(event) if event else v, k)
```
raphaelm, line 65: I think the `title` should contain something like "Sample: " in front of the sample to explain what that piece of information is. (Also, I notice that not all of them have samples, e.g. meta parameters with no default have an empty sample.)

src/pretix/static/pretixcontrol/scss/_forms.scss
```diff
@@ -2,6 +2,14 @@
   color: $text-muted;
 }
 
+.content-placeholder {
+  padding-left: 0;
+  padding-right: 0;
+  display: inline;
+  background: none;
+  border: 0;
```
raphaelm, line 10: Should we make them look more clickable? Currently it's hard to disover it. Either make them look more like buttons, or a "click to insert" (as part of the `title`?)

raphaelm (commented):
Works, some UI remarks inline.

> Should we go even further an just check if the cursor is inside a placeholder and replace the whole placeholder?

Sounds reasonable. `{eve{total}nt}` is never useful or valid :D

src/pretix/static/pretixcontrol/js/ui/main.js
wiffbi: Oh yes! I did not look that far up the DOM-tree for a closest-selector so chose the ugly parentNode.parentNode, but I initially hestitated to do so. Yours is way better!

src/pretix/static/pretixcontrol/scss/_forms.scss
wiffbi: I had them styled as buttons initially, but that was a wall of buttons with lots of grey lines (btn-default). So chose not to style as buttons. I tried `btn-link`, which was okay (except alignment issues with the text abseline of „Available placholders:“) What about a new style? Not sure about introducing a new style, but maybe this:

Bildschirmfoto 2023-09-12 um 20 18 03

src/pretix/base/forms/widgets.py
wiffbi: Good catch with the meta_values! Regarding „Sample“ – good idea, although for some it is not a sample, e.g. when using `{event}` within an event’s settings. But generally prefixing with „Sample: “ is better. Pushed.

wiffbi:
> > Should we go even further an just check if the cursor is inside a placeholder and replace the whole placeholder?

> Sounds reasonable. {eve{total}nt} is never useful or valid :D

Pushed.

src/pretix/static/pretixcontrol/scss/_forms.scss
raphaelm: I think this looks nice. It's not too distracting but still makes clear that these are special and you can do something with them.

wiffbi: I added that light button style. I also removed the comma in between each button element to A) reduce visual clutter and B) in certain conditions, the line break was happening before the comma and having a comma at the beginning of a line looks a bit unusual.

raphaelm (approved):

PR #3596, opened 2023-09-13
Widget: fix scoped Vue / multiple versions of Vue

wiffbi:
When embedding pretix’ widget in a page where an version of Vue older than 2.7 (e.g. 2.6), the widget had problems and wasn’t not working right – `$root.overlay` being undefined due to a `vm._scope` issue with older versions of Vue.

This PR removes vue-resize, which somehow caused the `vm._scope` issue when e.g. Vue 2.6 was loaded on the page. As the widget calculates the clientWidth on its own wrapper-object, it does not need the resize-element but a plain window.resize does the job (to be nice, I debounced the window.resize).

To test, just embed the widget in a page as one normally would, and load Vue2.6 before the widget.js – with current pretix widget, that should throw a js error. With this PR, it does not throw an error and works normally.


raphaelm (changes requested):
> As the widget calculates the clientWidth on its own wrapper-object, it does not need the resize-element but a plain window.resize does the job (to be nice, I debounced the window.resize).

Unfortunately, this does not seem to be the case. window.resize really only listens on a resize of the entire window. vue-resize also listens to resizes caused by a resizing of containing elements. I'm pretty sure this is behaviour at least a few customer websites rely on :(

wiffbi:
I added the shiny `ResizeObserver` (well supported) with a fallback to a normal window.resize for older browsers (IE).

raphaelm (approved):

PR #3603, opened 2023-09-18
Add language translation score to selector widget

raphaelm:
Before:

Screenshot 2023-09-18 at 12-24-07 General settings pretix eu

After:

image

### DIFF
--- a/src/pretix/control/forms/__init__.py
+++ b/src/pretix/control/forms/__init__.py
@@ -51,6 +51,7 @@ from django_scopes.forms import SafeModelMultipleChoiceField
 from pretix.helpers.hierarkey import clean_filename
 
 from ...base.forms import I18nModelForm
+from ...helpers.i18n import get_language_score
 from ...helpers.images import (
     IMAGE_EXTS, validate_uploaded_file_for_valid_image,
 )
@@ -325,6 +326,7 @@ class MultipleLanguagesWidget(forms.CheckboxSelectMultiple):
         opt = super().create_option(name, value, label, selected, index, subindex, attrs)
         opt['official'] = value in settings.LANGUAGES_OFFICIAL
         opt['incubating'] = value in settings.LANGUAGES_INCUBATING
+        opt['score'] = round(get_language_score(value) / get_language_score("de") * 100)
         return opt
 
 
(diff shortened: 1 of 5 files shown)


luelista (approved):

wiffbi:
Looks good! I’d probably reduce the „Community edition“ to just an icon with a tooltip and maybe even combine into one label – although then not the label, but the icons and the score need its tooltips. The divisor line (or just a gap – I tried both) is a try to group elements – works with a `:has` selector (which currently is not supported in Firefox, but it does not hurt adding it?).

Bildschirmfoto 2023-09-19 um 08 00 30


I got a bit carried away with the following, but I think although it does look interesting, it does not help with the actual task: selecting a language. It feels more like a gamification of „we need to translate this more!“

Bildschirmfoto 2023-09-19 um 08 53 56

So I’d go with the combined labels and added `label-warning`. What do you think?

raphaelm:
> Looks good! I’d probably reduce the „Community edition“ to just an icon with a tooltip and maybe even combine into one label – although then not the label, but the icons and the score need its tooltips.

I like it visually, but I'm not sure it's clear enough. It's very important to me to highlight the "community" part as a trust indicator. Customers need to know not only that the translation is incomplete (which the %-age shows), but also that *we* are not the ones doing it. I also feel that the texts of both current tooltips are important, but with combining them into one label, people would not discover that there are two distinct tooltips, and having just one would be very long, I believe?

wiffbi:
Would it then make sense to group these translations – e.g. something like this? I’d then probably even remove the group-icons from each label; less visual clutter …

Bildschirmfoto 2023-09-19 um 09 59 09

wiffbi:
I think with a readable help-text it is even better.

Bildschirmfoto 2023-09-19 um 10 03 05

src/pretix/control/forms/__init__.py
```diff
@@ -300,18 +301,40 @@ def get_context(self, name, value, attrs):
...
-        ))
+                '',
+                filter_and_sort(self.choices, settings.LANGUAGES_OFFICIAL)
+            ),
+            (
+                (
+                    _('Community translations'),
+                    _('These translations are not maintained by the pretix team. We cannot vouch for their correctness '
+                        'and new or recently changed features might not be translated and will show in English instead. '
+                        'You can <a href="{translate_url}" target="_blank">help translating</a>.'.format(translate_url='https://translate.pretix.eu')),
```
raphaelm, line 323: @wiffbi What's the purpose of the `.format` call? Should it make the URL not part of the translatable string? In that case it would need to be `).format()` instead of `.format())`

wiffbi: Oh yes, good catch! Fixed with django’s `lazy_format`

PR #3622, opened 2023-09-28
Calender/ics: always add dtend (Z#23131739)

wiffbi:
Calendar apps have problems, when dtend is omitted. This PR always adds a dtend – with a default of one hour if `show_times` is set, else one day.

### DIFF
--- a/src/pretix/presale/ical.py
+++ b/src/pretix/presale/ical.py
@@ -68,13 +68,16 @@ def get_public_ical(events):
         else:
             vevent.add('dtstart').value = ev.date_from.astimezone(tz).date()
 
-        if event.settings.show_date_to and ev.date_to:
-            if event.settings.show_times:
-                vevent.add('dtend').value = ev.date_to.astimezone(tz)
-            else:
-                # with full-day events date_to in pretix is included (e.g. last day)
-                # whereas dtend in vcalendar is non-inclusive => add one day for export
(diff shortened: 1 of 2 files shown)


src/pretix/presale/ical.py
```diff
@@ -68,13 +68,16 @@ def get_public_ical(events):
...
             vevent.add('dtstart').value = ev.date_from.astimezone(tz).date()
 
-        if event.settings.show_date_to and ev.date_to:
-            if event.settings.show_times:
-                vevent.add('dtend').value = ev.date_to.astimezone(tz)
-            else:
-                # with full-day events date_to in pretix is included (e.g. last day)
-                # whereas dtend in vcalendar is non-inclusive => add one day for export
-                vevent.add('dtend').value = ev.date_to.astimezone(tz).date() + datetime.timedelta(days=1)
+        # ignore event.settings.show_date_to as ics needs an end for an event
```
raphaelm, line 71: No, I don't think we should never ignore `show_date_to`. We currently never show the end date in this case, and I think it would be a big surprise for organizers if it suddenly does – in a pretty hidden place. :\

wiffbi: Okay, I changed it, so when `show_date_to == False` then `ev.date_from` is used, even when an `ev.date_to` is available.

raphaelm (approved):

PR #3633, opened 2023-10-06
Widget: add single-item-select shopping cart (Z#23131246)

wiffbi:
This PR adds the single item select button-checkbox to the widget. The display-style is controlled through an optional attribute `single-item-select` on `<pretix-widget>` with values `"checkbox|button"` – default being checkbox.

`checkbox` (as before)

Bildschirmfoto 2023-10-06 um 12 01 22

`button`

Bildschirmfoto 2023-10-06 um 12 02 09

Should we add this attribute to the generated widget-embed code in pretix-control with e.g. `single-item-select="button"`, so every newly created/embedded widget will have the new button-select-UI?


src/pretix/static/pretixpresale/js/widget/widget.js
```diff
@@ -214,7 +214,13 @@ Vue.component('availbox', {
         + '<a :href="waiting_list_url" target="_blank" @click="$root.open_link_in_frame">' + strings.waiting_list + '</a>'
         + '</div>'
         + '<div class="pretix-widget-availability-available" v-if="!require_voucher && avail[0] === 100">'
-        + '<label class="pretix-widget-item-count-single-label" v-if="order_max === 1">'
+        + '<label class="pretix-widget-item-count-single-label pretix-widget-btn-checkbox" v-if="order_max === 1 && $root.single_item_select == \'button\'">'
+        + '<input type="checkbox" value="1" :checked="!!amount_selected" @change="amount_selected = $event.target.checked" :name="input_name"'
+        + '       v-bind:aria-label="label_select_item"'
+        + '>'
+        + '<i class="pretix-widget-icon-cart" aria-hidden="true"></i> Select'
```
raphaelm, line 221: Are we purposefully using `<i>` and not span here? IIRC most people today advocate against using `<i>` for this these days

raphaelm: Also, the "Select" is not translated here, is it?

src/pretix/static/pretixpresale/js/widget/widget.js
```diff
@@ -1904,6 +1910,7 @@ var create_widget = function (element) {
     var items = element.attributes.items ? element.attributes.items.value : null;
     var variations = element.attributes.variations ? element.attributes.variations.value : null;
     var categories = element.attributes.categories ? element.attributes.categories.value : null;
+    var single_item_select = element.getAttribute("single-item-select") || "checkbox";
```
raphaelm, line 1913: A docs mention is missing, maybe even including a warning that the default might change in the future

raphaelm:
> Should we add this attribute to the generated widget-embed code in pretix-control with e.g. single-item-select="button", so every newly created/embedded widget will have the new button-select-UI?

I think so, yes

raphaelm (changes requested):

src/pretix/static/pretixpresale/js/widget/widget.js
wiffbi: Oh sorry, totally missed the translation part. Just copied the generated HTML from pretix-presale. I think `<i>` is taken from fontawesome docs, but they state themselves, that `<i>` is used for bevity, but `<span>` would be more semantically correct. I will change that.

src/pretix/static/pretixpresale/js/widget/widget.js
wiffbi: I added some docs with a note, that this default might change in the future. I also added a small screenshot to better illustrate what the docs are about.

raphaelm (approved):

PR #3636, opened 2023-10-09
Subevent editor: Allow plugins to declare form titles

pc-coholic:
Somewhat analog to #3492

The panel-`<div>`s might not be the best solution for this PR.

Another idea I had was the same `<fieldset>` and `<legend>`, that are already sectioning the page. For a cleaner style, this might be preferable.

However: in that case, all form-fields that are not part of a form with a title should be grouped together *and* moved within the `Additional Settings` block to the end of the page, else it could get confusing for the user when they see
- Legend: Additional settings
- random Formfield
- Legend: Plugin
- formfield plugin
- formfield plugin
- formfield plugin
- random formfield

(Excellent showcase for this, when enabling CineSend, Seating and Kulturpass)

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/subevents/bulk.html
+++ b/src/pretix/control/templates/pretixcontrol/subevents/bulk.html
@@ -599,26 +599,27 @@
                     </button>
                 </p>
         </fieldset>
+        {% for f in plugin_forms %}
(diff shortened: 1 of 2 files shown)


raphaelm:
I changed to the same style as with product forms, for consistency

PR #3646, opened 2023-10-16
Do not treat event as free if there are mandatory non-free addons

raphaelm:


wiffbi (approved):
Looks good. Looking at the code in widget.js for `buy_label` we could optimize those 3 loops by `return strings.buy;` instead of `all_free=false;break` – as break only breaks the inner loop. Or add `&& all_free` to each loops condition, although I do not think that is especially readable … This is not part of this PR, but maybe I should do another PR?

raphaelm:
> Looks good. Looking at the code in widget.js for `buy_label` we could optimize those 3 loops by `return strings.buy;` instead of `all_free=false;break` – as break only breaks the inner loop. Or add `&& all_free` to each loops condition, although I do not think that is especially readable … This is not part of this PR, but maybe I should do another PR?

Not sure it's worth it, but sure, we can do that.

PR #3649, opened 2023-10-17
PDF: fix position of rotated overflown textfields (Z#23133660)

wiffbi:
When editing a PDF-layout and not using the „downward“ option, overflown textfields grow from bottom to top. This creates problems with rotated textfields as the rotation base is offset due to the overflown textfield. This PR sets the rotation base always to one line-height – the default height of textfields.

Note: when using the editor and making the textfield very small, so that the preview value creates a line break, this makes the textbox grow to multiple lines, but as we do not have any info height, it can create weird effects – not limited to, but especially with rotation.


wiffbi:
As discussed, this is not a legacy-compatible change, so we will need to deprecate the default behaviour and add an „upwards“ option to textfields.

PR #3652, opened 2023-10-17
Check-in: Add export with all valid codes (Z#23133941)

raphaelm:


wiffbi (approved):
Looks good and works as intended with my checkin-lists. First stumbled over the change in exporters-py::481, but that just removes a filter that is already in CheckInListMixin, so good to go.

PR #3659, opened 2023-10-20
API: Add endpoints for scheduled exports

raphaelm:


pc-coholic (approved):
LGTM

The only thing that made me pause for a moment was the email-address of the owner.

As of right now, we do not expose pretix system user email-addresses or PKs anywhere. Even on `/api/v1/me` we only return the email-address.

But from a "stable API"-standpoint I think there might be a point to be made, that systems consuming this API should be able to rely on the owner-value to be static, if there is no way to change the owner of a report.

I am aware, that going for a User-ID instead of the eMail here would also mean that we need to add a /user-endpoint which is a whole different nightmare in terms of "who can see what"... So perhaps - as a middle ground - add a warning to the resource documentation that no system should rely on this never changing?

raphaelm:
> As of right now, we do not expose pretix system user email-addresses or PKs anywhere.

Not in the API, but we do expose them in the "Teams" page, or with logentries if the user has no display name set, so they're not considered "secret" within an org.

> So perhaps - as a middle ground - add a warning to the resource documentation that no system should rely on this never changing?

I'll add a note to the docs.

PR #3665, opened 2023-10-23
Improve help texts when deleting questions (Z#23134288)

wiffbi:
This PR is hard for me to write a proper commit-message for. It adds a few tiny text improvements to make clear that deleting a question deletes the answer as well – although this could maybe be misinterpreted as the answer-options, that are created when having a „Choose one from a list“-question.

- changes delete-button to „Delete question and all answers“
- makes warning text more prominent by moving it inside an div.alert.alert-warning
- adding a help-text with link to set question ot hidden instead
- Fix: use internal of item if any

Is hiding a question a common operation? If so, we might add a button to do so directly in the questions overview? Or should the checkbox „hidden product“ be moved to the top of the first tab like the „Active“-checkbox on products?

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/items/question_delete.html
+++ b/src/pretix/control/templates/pretixcontrol/items/question_delete.html
@@ -18,7 +18,7 @@
                 {% trans "Cancel" %}
             </a>
             <button type="submit" class="btn btn-danger btn-save">
-                {% trans "Delete" %}
+                {% trans "Delete question and all answers" %}
             </button>
 		</div>
 	</form>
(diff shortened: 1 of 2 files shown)


raphaelm:
> Is hiding a question a common operation? If so, we might add a button to do so directly in the questions overview? Or should the checkbox „hidden product“ be moved to the top of the first tab like the „Active“-checkbox on products?

-0.5 from me, since hiding a question is at best a workaround here. The intention of hiding a question was never to be a "soft-delete", it was to be a "backend-only data collection". It does work as a workaround though.

raphaelm (approved):

Issue #1196, opened 2019-02-27
Pay what you want Ticket: default value


navlisilvan-old:
I hope there will be no similar issue, i didn´t find one but i am not very familiar with github.

I would be really happy to have a default value for the "pay what you want"-Ticket. At the moment the lower threshold is the default and the customer has to click +++ or enter a number. 
I Would like do have the default value to be configured in the control panel, so that the customer has to click --- to select a lower price (or +++ for a higher price or enter a number).

e.g.:
Tickets will be sold from 10€. At The moment when you enter the ticket-shop there will be shown 10€ and the customers can enter more.
We would like to have it this way:
Tickets will be sold from 10€. Default Value is 20€. When you enter the Ticket-shop there will be written 20€ and you can click more or less until 10€.

navlisilvan:
We did a hack for this problem, we added  `add:'10'|` to to line 387 of src/pretix/presale/template/event/index.html

`value="{% if event.settings.display_net_prices %}{{item.display_price.net|add:'10'|money_numberfield:event.currency }}{% else %}{{ item.display_price.gross|add:'10'|money_numberfield:event.currency }}{% endif %}"`


:)

raphaelm:
We could re-use the `original_price` field for this, although it is a little  unclean to use it for two purposes. Anyone has thoughts on this?

raphaelm:
Fixed in #3666

PR #3666, opened 2023-10-25
Free price: Allow to suggest a different price than the minimum

raphaelm:
- [x] Data model
- [x] Backend
- [x] Product selection
- [x] Add-on selection
- [x] Add-on change dialog
- [x] Voucher redemption
- [x] Widget
- [x] API


wiffbi (approved):
Looks good.

One thing on variations I stumbled upon is: it uses the display_price for „starting at“. When opening the variations it might be irritating that that price is not shown as the suggestion is used to fill the input. Should we add some indicator for the min-value? We could e.g. extend the title-attribute used for the tooltip-thingie to „Modify price for <item>, at least <display_price>“.

wiffbi (approved):
Wonderful!

c0de-bender:
@raphaelm @wiffbi Great, super nice to see this feature coming!

I was looking for it. I only found #1196 and wanted to push this issue, because I assumed Pretix cannot do it. Then I saw the `free_price_suggestion` branch in the git log mentioning this pull request. Very nice!

Note: #1196 can be closed then, I guess.

Question: Is this already tested and release-ready? I'd like to cherry-pick it and deploy and use it right away, if it's "safe".

raphaelm:
> Is this already tested and "release-ready"? I'd like to cherry-pick it and deploy and use it right away, if it's "safe" to use.

It's tested and it's safe to use. However, I'd avoid cherry-picking commits that include database migrations. Applying them in the wrong order might need to chaos, if you're not very well-versed in django and know how to work around it. But it should be usually fine to run from the master branch in production for a while ;)

> Looking at the commit changes, I stumbled upon an addition in `doc/api/resources/item_variations.rst`:

Thanks!

c0de-bender:
> > Is this already tested and "release-ready"? I'd like to cherry-pick it and deploy and use it right away, if it's "safe" to use.
> 
> It's tested and it's safe to use. However, I'd avoid cherry-picking commits that include database migrations. Applying them in the wrong order might need to chaos, if you're not very well-versed in django and know how to work around it. But it should be usually fine to run from the master branch in production for a while ;)
> 

That's great! :)

Thanks for the advice, too. I compared the current master with the last version tag and the database migration script in this commit was the only one so far, so I thought it's safe to cherry-pick. However, you're right that it's not good practice, so I'll stick to the commit the master is pointing to.

PR #3673, opened 2023-10-31
Add public filters based on meta data

raphaelm:
- [x] Data model
- [x] UI
- [ ] Test in all combinations
	- [ ] org level list
		- [ ] filter on "any subevent"?
	- [ ] org level week
	- [ ] org level month
	- [ ] org level day
	- [ ] event level list
	- [ ] event level week
	- [ ] event level month
- [ ] Add support to widget
	- [ ] org level list
		- [ ] filter on "any subevent"?
	- [ ] org level week
	- [ ] org level month
	- [ ] org level day
	- [ ] event level list
	- [ ] event level week
	- [ ] event level month


src/pretix/presale/forms/organizer.py
```diff
@@ -0,0 +1,88 @@
...
+# <https://www.gnu.org/licenses/>.
+#
+from django import forms
+from django.conf import settings
+from i18nfield.strings import LazyI18nString
+
+from pretix.base.models import EventMetaValue, SubEventMetaValue
+
+
+def meta_filtersets(organizer, event=None):
```
wiffbi, line 29: In my test organizer, I do have lots of old events, that are live and accessible via the „old=1“ GET-param. Some of them have custom meta-data that newer events don’t. The filter shows these meta-data-values as options to filter for, but when I filter, I do not get any results as old events are not enabled to be shown. What would be the best way to handle this? Instead of automatically including old events if a filter does not yield any result, I would not add meta-data-values of events that happened in the past (if the old=1 GET param is not set).

src/pretix/base/models/event.py
```diff
@@ -1641,10 +1641,18 @@ class EventMetaProperty(LoggedModel):
...
                     "optional to set a value for individual dates")
     )
-    allowed_values = models.TextField(
+    allowed_values = models.JSONField(
         null=True, blank=True,
+        encoder=CustomJSONEncoder,
         verbose_name=_("Valid values"),
-        help_text=_("If you keep this empty, any value is allowed. Otherwise, enter one possible value per line.")
+    )
+    filter_public = models.BooleanField(
```
wiffbi, line 1649: Do we need a way to sort the MetaProperties? It might be overkill as I guess usually one will only filter by one or two properties. But then, when it is two, maybe one wants to change the order around?

src/pretix/static/pretixpresale/js/widget/widget.js
```diff
@@ -1093,6 +1093,46 @@ Vue.component('pretix-widget-event-form', {
     }
 });
 
+Vue.component('pretix-widget-event-list-filter-field', {
+    template: ('<div class="pretix-widget-event-list-filter-field">'
+        + '<label :for="id">{{ field.label }}</label>'
+        + '<select :id="id" :name="field.key" @change="onChange($event)" :value="currentValue">'
+        + '<option v-for="choice in field.choices" :value="choice[0]">{{ choice[1] }}</option>'
```
wiffbi, line 1100: TIL `:value` on `<select>` works in Vue!

wiffbi (commented):
I think it is ready to release code-wise, but I have a few comments regarding UX. 

- Would it make sense to make MetaProperties sortable?
- What about Meta-Property-values that only exist on old events – these values are included in the filter-dropdown, but when filtered, they do not display any events. IMHO we should respect the `old=1`-GET-param when reading all filter-values.
- As a side-node to the filter-values above: when filtering in the backend, filtering events on organizer-level does not include meta-values for subevents (as does presale with `match_subevents_with_conditions`. I know, this is complex as the event-series itself can have meta-values other than its subevents, so I am not sure how to handle this (maybe try to match both and return series even if only a subevent matched?). I just stumbled upon this as I wanted to find out where certain meta-values came from, but no results were yielded.

raphaelm:
>     * Would it make sense to make MetaProperties sortable?

Agreed. I'll add it.

>     * What about Meta-Property-values that only exist on old events – these values are included in the filter-dropdown, but when filtered, they do not display any events. IMHO we should respect the `old=1`-GET-param when reading all filter-values.

I thought about this for a while as well, and I'm still unsure. Imagine that you filter for Language=English and you switch between past and future events. Or you switch between calendar and list view. Odds are your current filter suddenly becomes an *invalid form input*. I don't think that's nice either? I'd rather know that there are no English results instead of suspecting a form input error.


>     * As a side-node to the filter-values above: when filtering in the backend, filtering events on organizer-level does not include meta-values for subevents (as does presale with `match_subevents_with_conditions`. I know, this is complex as the event-series itself can have meta-values other than its subevents, so I am not sure how to handle this (maybe try to match both and return series even if only a subevent matched?). I just stumbled upon this as I wanted to find out where certain meta-values came from, but no results were yielded.

Hmmm. Let's see if it actually confuse users and maybe change it. I see value in making it consistent, but I also see value in making the backend as clear and directly understandable as possible while having the new behaviour in the frontend to make it as helpful as possible.

PR #3674, opened 2023-10-31
Order list export: Expose same "extended status" as in backend

raphaelm:


src/pretix/base/models/orders.py
```diff
@@ -325,6 +325,16 @@ def gracefully_delete(self, user=None, auth=None):
     def email_confirm_hash(self):
         return hashlib.sha256(settings.SECRET_KEY.encode() + self.secret.encode()).hexdigest()[:9]
 
+    def get_extended_status_display(self):
```
luelista, line 328: maybe we should add a comment here, referencing the other place where the exact same logic is implemented (src/pretix/control/templates/pretixcontrol/orders/fragment_order_status.html), and also a reference the other way round, from the template to this place - so that in case one of them is changed it's obvious the other needs to be changed, too

src/pretix/base/models/orders.py
```diff
@@ -325,6 +325,16 @@ def gracefully_delete(self, user=None, auth=None):
...
+        if self.status == Order.STATUS_PENDING:
+            if self.require_approval:
+                return _("Approval pending")
+            elif self.valid_if_pending:
+                return pgettext_lazy("order state", "Pending (confirmed)")
+        elif self.status == Order.STATUS_PAID and self.count_positions == 0:
+            return _("Canceled (paid fee)")
+        return self.get_status_display()
```
luelista, line 336: this returns a string starting with lower case letter, the texts above are upper case. this looks a bit inconsistent in the excel sheet... we could do .title() here but I'm not sure on the impact on other translations. or just keep it as is and live with the inconsistency

grafik

luelista (approved):
two ideas in comments, works locally, LGTM

src/pretix/base/models/orders.py
luelista: e.g.
```suggestion
    def get_extended_status_display(self):
        # Changes in this method need to be replicated in pretixcontrol/orders/fragment_order_status.html
```

src/pretix/base/models/orders.py
raphaelm: This annoyed me, too, and since you've also noticed it, I'm just gonna go for real, lower-case strings for consistency. Even more translatable strings, but oh well, here we are

src/pretix/base/models/orders.py
raphaelm: Good point, added

PR #3682, opened 2023-11-02
Allow hidden payment methods on payment method change

raphaelm:


pc-coholic:
Should we perhaps also put the unlock-hashes into the order meta-data during the checkout process, so that the unlocked methods may be used without having to click the magic link again?

PR #3688, opened 2023-11-03
Improve check-in deletion

raphaelm:
- Add a confirmation step
- Don't write a log entry if nothing was done

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/checkin/bulk_revert_confirm.html
+++ b/src/pretix/control/templates/pretixcontrol/checkin/bulk_revert_confirm.html
@@ -11,7 +11,13 @@
                 <input type="hidden" name="{{ k }}" value="{{ v }}">
             {% endfor %}
         {% endfor %}
-		<p>{% blocktrans with count=cnt %}Are you sure you want to permanently delete the check-ins of <strong>{{ count }} tickets</strong>?{% endblocktrans %}</p>
+		<p>
+            {% blocktrans trimmed count count=cnt %}
+                Are you sure you want to permanently delete the check-ins of <strong>one ticket</strong>.
+            {% plural %}
+                Are you sure you want to permanently delete the check-ins of <strong>{{ count }} tickets</strong>?
+            {% endblocktrans %}
+        </p>
 		<div class="form-group submit-group">
             <a href="{% url "control:event.orders.checkinlists" organizer=request.event.organizer.slug event=request.event.slug %}"
                     class="btn btn-default btn-cancel">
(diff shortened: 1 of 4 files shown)


src/pretix/control/templates/pretixcontrol/checkin/bulk_revert_confirm.html
```diff
@@ -0,0 +1,25 @@
...
+{% block inside %}
+	<h1>{% trans "Delete check-ins" %}</h1>
+	<form action="{% url "control:event.orders.checkinlists.bulk_action" event=request.event.slug organizer=request.event.organizer.slug list=checkinlist.pk %}" method="post" class="form-horizontal" data-asynctask>
+		{% csrf_token %}
+        {% for k, l in request.POST.lists %}
+            {% for v in l %}
+                <input type="hidden" name="{{ k }}" value="{{ v }}">
+            {% endfor %}
+        {% endfor %}
+		<p>{% blocktrans with count=cnt %}Are you sure you want to permanently delete the check-ins of <strong>{{ count }} tickets</strong>?{% endblocktrans %}</p>
```
wiffbi, line 14: Not sure if it is a rare edge case, but what if count==1? Probably not worth the effort. And using „ticket(s)“ instead of „tickets“ does probably do more harm than help.

wiffbi (approved):
Looks good, just a minor comment regarding pluralization – AFAIK not easy in django templates, so probably not worth the effort.

Not part of this PR, but maybe worth thinking about: when I have not selected any check-ins, I can click any of the bulk-actions and it does execute them on all checkins. Not sure we want this, especially as we do have a checkbox to select all and also as check-in/out bulk-actions do not (yet?) have confirmation step – I think we could get away without a confirmation step on check-in/out if we not allow to check-in/out all without ticking the checkbox „Select all results on other pages as well“. I.e. disable the bulk-action buttons if no check-ins are selected.

raphaelm:
> Not part of this PR, but maybe worth thinking about: when I have not selected any check-ins, I can click any of the bulk-actions and it does execute them on all checkins. Not sure we want this, especially as we do have a checkbox to select all and also as check-in/out bulk-actions do not (yet?) have confirmation step – I think we could get away without a confirmation step on check-in/out if we not allow to check-in/out all without ticking the checkbox „Select all results on other pages as well“. I.e. disable the bulk-action buttons if no check-ins are selected.

I just added our `.batch-select-actions` class which helps with this :)

src/pretix/control/templates/pretixcontrol/checkin/bulk_revert_confirm.html
wiffbi: TIL `{% plural %}`! Probably had the wrong search terms around ngettext and django templates. Thanks!

PR #3689, opened 2023-11-03
Order change: Allow price reduction as long as no refund is required (Z#23135268)

raphaelm:
This adds a new strategy option for price changes in self-service order changes.

### DIFF
--- a/src/pretix/base/settings.py
+++ b/src/pretix/base/settings.py
@@ -1677,8 +1677,8 @@ DEFAULTS = {
                 ('gte', _('Only allow changes if the resulting price is higher or equal than the previous price.')),
                 ('gt', _('Only allow changes if the resulting price is higher than the previous price.')),
                 ('eq', _('Only allow changes if the resulting price is equal to the previous price.')),
-                ('gte_paid', _('Allow changes regardless of price, as long as no refund is required (i.e. not more '
-                               'than the resulting price has already been paid).')),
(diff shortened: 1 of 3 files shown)


src/pretix/base/settings.py
```diff
@@ -1677,6 +1677,8 @@ def unserialize(cls, s):
                 ('gte', _('Only allow changes if the resulting price is higher or equal than the previous price.')),
                 ('gt', _('Only allow changes if the resulting price is higher than the previous price.')),
                 ('eq', _('Only allow changes if the resulting price is equal to the previous price.')),
+                ('gte_paid', _('Allow changes regardless of price, as long as no refund is required (i.e. not more '
+                               'than the resulting price has already been paid).')),
```
wiffbi, line 1681: I read the „i.e.“ part multiple times and it is still hard for me to understand correctly and not think there is a word missing. If I translate it in german, it works for me, but in plain english I do not get the connection between „not more“ and „has already been paid“, but connect „the resulting price has already been paid“, which of course on its own does not make any sense. Maybe it’s me, but what about:

```suggestion
                ('gte_paid', _('Allow changes regardless of price, as long as no refund is required (i.e. the resulting '
                               'price is not lower than what has already been paid).')),
```

wiffbi (approved):
Looks good, just a suggestion regarding wording.

PR #3692, opened 2023-11-06
Bank transfer: Discourage payments before an order code exists (Z#23135042)

raphaelm:
- If there is only SEPA account information, the bank information will no longer be shown until the order is complete
- If there is "other information" available, this change would be backwards-incompatible for some users, so we go a different route:  The bank information is no longer copy-able and if you hover it, there is a warning. This will hopefully catch users who try to copy the IBAN to early ;)

### DIFF
--- a/src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/checkout_payment_form.html
+++ b/src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/checkout_payment_form.html
@@ -2,33 +2,41 @@
 {% load ibanformat %}
 {% load bootstrap3 %}
 
-<p>{% blocktrans trimmed %}
-    After completing your purchase, we will ask you to transfer the money to the following
-    bank account, using a personal reference code:
-{% endblocktrans %}</p>
+{% if details or code %}
+    <p>{% blocktrans trimmed %}
+        After completing your purchase, we will ask you to transfer the money to the following
+        bank account, using a personal reference code:
+    {% endblocktrans %}</p>
 
-{% if settings.bank_details_type == "sepa" %}
-    <dl class="dl-horizontal">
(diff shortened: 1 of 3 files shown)


src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/checkout_payment_form.html
```diff
@@ -2,33 +2,41 @@
...
-    <dl class="dl-horizontal">
-        <dt>{% trans "Account holder" %}: </dt><dd>{{ settings.bank_details_sepa_name }}</dd>
-        <dt>{% trans "IBAN" %}: </dt><dd>{{ settings.bank_details_sepa_iban|ibanformat }}</dd>
-        <dt>{% trans "BIC" %}: </dt><dd>{{ settings.bank_details_sepa_bic }}</dd>
-        <dt>{% trans "Bank" %}: </dt><dd>{{ settings.bank_details_sepa_bank }}</dd>
-    </dl>
-{% endif %}
+    {% if not code %}<div class="user-select-none">{% endif %}
+    {% if settings.bank_details_type == "sepa" %}
+        <dl class="dl-horizontal" data-toggle="tooltip" title="{% trans "Please do not yet start a payment. We'll assign you a personal reference code after you completed the order." %}">
```
luelista, line 13: The tooltip should be handled the same as in the other template:

```suggestion
    {% if not code %}<div class="user-select-none" data-toggle="tooltip" title="{% trans "Please do not yet start a payment. We'll assign you a personal reference code after you completed the order." %}">{% endif %}
    {% if settings.bank_details_type == "sepa" %}
        <dl class="dl-horizontal">
```

luelista (approved):
LGTM with one change

PR #3697, opened 2023-11-07
Paginator: Add options for advanced users in backend

raphaelm:
- Jump to last page (I don't really like the additional button visually, but was requested internally)
- Jump to specific page without cluttering the UI (was requested by a customer)


wiffbi (commented):
Looks good, I thought about using [step-backward icon](https://fontawesome.com/v4/icon/step-backward), but that does not work optically (although it would make more sense, as to me double-pointers is just a faster/bigger step). Other comments just regarding UI, making the page-jumper-link look more clickable and adding a hint to the max. page number.

PR #3698, opened 2023-11-07
Add autoscroll to widget after changing pages (Z#23135651)

raphaelm:
Question: Are we in danger of this breaking pages of customers who show the widget in a lightbox-style overlay?

### DIFF
--- a/src/pretix/static/pretixpresale/js/widget/widget.js
+++ b/src/pretix/static/pretixpresale/js/widget/widget.js
@@ -1649,6 +1649,7 @@ var shared_root_methods = {
             url = url + '&style=' + encodeURIComponent(this.$root.style);
         }
         var root = this.$root;
+        var oldView = root.view;
         api._getJSON(url, function (data, xhr) {
             if (typeof xhr.responseURL !== "undefined") {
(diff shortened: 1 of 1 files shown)


wiffbi (commented):
Instead of handling this in the reload-method, I would add a watcher to the Vue-app element in `create_widget` – assuming pretix-button basically never changes its view-property? Moving it to a watcher makes it more robust, when at some point we change root.view somewhere else and also more elegant to access the wrapper through `this.$el` instead of `root.$children[0].$refs.wrapper`. Furthermore I would add a check, if we need to scrollIntoView at all.

Should I make the changes regarding the watcher and the scrollIntoView-check?

raphaelm:
Yes, if you have good ideas on how to do it, please go ahead before I start researching :)

wiffbi:
> Question: Are we in danger of this breaking pages of customers who show the widget in a lightbox-style overlay?

I do not think so as getBoundingClientRect tells us where the element is inside the viewport indepent of any container. And scrollIntoView should do the right thing and scroll whatever element is appropriate to scroll to move the widget into view.

PR #3704, opened 2023-11-08
Add customizable terms of cancellation (Z#23135646)

wiffbi:
As discussed instead of showing a date, we add customizable text snippets for terms of cancellation for unpaid/free and paid orders. If terms are available, they replace pretix’ auto-generated terms.

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/order.html
+++ b/src/pretix/presale/templates/pretixpresale/event/order.html
@@ -345,13 +345,7 @@
         <div class="panel panel-default panel-cancellation">
             <div class="panel-heading">
                 <h3 class="panel-title">
-                    {% if order.user_cancel_deadline %}
-                        {% blocktrans trimmed context "action" with deadline=order.user_cancel_deadline %}
-                            Change or cancel your order until {{ deadline }}
-                        {% endblocktrans %}
-                    {% else %}
-                        {% trans "Change or cancel your order" context "action" %}
-                    {% endif %}
+                    {% trans "Change or cancel your order" context "action" %}
                 </h3>
             </div>
             <ul class="list-group">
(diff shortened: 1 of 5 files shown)


src/pretix/presale/templates/pretixpresale/event/order.html
```diff
@@ -445,45 +447,31 @@ <h3 class="panel-title">
...
                                 {% else %}
                                     {% blocktrans trimmed %}
                                         You can cancel this order using the following button.
                                     {% endblocktrans %}
+                                    {% trans "This will invalidate all tickets in this order." %}
                                 {% endif %}
-                                {% trans "This will invalidate all tickets in this order." %}
-                            </p>
-
-                            {% if order.user_cancel_deadline %}
```
raphaelm: Has this been removed intentionally? It seems useful?

wiffbi: I moved it to the automatically generated cancelation terms to lines 459 and 464 as I thought one might want to overwrite _everything_ when adding custom terms. But you are right, maybe this is nonetheless useful to note and might easily be forgotten in custom terms. I can move back in if you want.

raphaelm: Sorry, I was refering to the `{% if order.user_cancel_deadline %}` block. I agree that this should be overwritten by custom terms! But I only see it removed from the div, not added somewhere else?

wiffbi: Ah, I did not see the line you are referring to as changed. Are you sure you are on the latest commit on that branch? Super strange.

raphaelm (approved):

src/pretix/presale/templates/pretixpresale/event/order.html
raphaelm: Huh. It was showing it to me multiple times. GitHub :shrug: Now it's clean

PR #3707, opened 2023-11-09
Relative dates: Add UI to specify dates after reference date

raphaelm:


wiffbi (commented):
It’s a pity that default direction of reldate was/is before – code-wise now it reads a bit strange having e.g. `RelativeDate(days=2, …` mean two days _before_. It would be nicer, if it were `RelativeDate(days=-2, …`. That way we could even get rid of lots of those `if self.data.is_after:` as timedelta supports negative values as well. The form-select would need some extra care to convert negative integers to "before" in the dropdown. And to be backwards compatible we would need to change the `to_string`/`from_string` so that strings starting with 'RELDATE' are handled as before-dates and new strings are e.g. marked as 'TIMEDELTA/{}/{}/{}', which would support negative values such as `TIMEDELTA/-2/-/…'`

What do you think? I could take the PR over if you want.

raphaelm:
> It’s a pity that default direction of reldate was/is before – code-wise now it reads a bit strange having e.g. `RelativeDate(days=2, …` mean two days _before_. It would be nicer, if it were `RelativeDate(days=-2, …`. That way we could even get rid of lots of those `if self.data.is_after:` as timedelta supports negative values as well. The form-select would need some extra care to convert negative integers to "before" in the dropdown. And to be backwards compatible we would need to change the `to_string`/`from_string` so that strings starting with 'RELDATE' are handled as before-dates and new strings are e.g. marked as 'TIMEDELTA/{}/{}/{}', which would support negative values such as `TIMEDELTA/-2/-/…'`


Technically, even if we can still parse the old strings, that would be even more of a breaking change of our REST API. I'm not sure if any API consumer *parses* the values we output, though.

wiffbi:
Well, you are right, let’s keep it this way. Good enough. :)

src/pretix/base/reldate.py
```diff
@@ -143,17 +155,19 @@ def from_string(cls, input: str):
...
                     )
                 except ValueError:
                     data = RelativeDate(
-                        days_before=0,
+                        days=0,
                         base_date_name=parts[3],
                         time=time,
-                        minutes_before=None
+                        minutes=None,
+                        is_after=parts[4] == "after",
```
wiffbi, line 170:
```suggestion
                        is_after=len(parts) > 4 and parts[4] == "after",
```

src/pretix/base/reldate.py
```diff
@@ -143,17 +155,19 @@ def from_string(cls, input: str):
                     time = datetime.time(hour=int(timeparts[0]), minute=int(timeparts[1]), second=int(timeparts[2]))
                 try:
                     data = RelativeDate(
-                        days_before=int(parts[1] or 0),
+                        days=int(parts[1] or 0),
                         base_date_name=parts[3],
                         time=time,
-                        minutes_before=None
+                        minutes=None,
+                        is_after=parts[4] == "after",
```
wiffbi, line 162: For the sake of completeness as above.

```suggestion
                        is_after=len(parts) > 4 and parts[4] == "after",
```

wiffbi (approved):
Just two more safe-guards regarding `len(parts) > 4`

PR #3709, opened 2023-11-10
UI: Improve panel collapse visibility (Z#23132549)

wiffbi:
This PR tries to improve the visibility of the collapse-feature on panels by moving it closer to the actual label of the panel.

Instead of changing the existing HTML for .collapse-indicator in lots of places – as well as probably lots of plugins – this PR just hides `.collapse-indicator` and adds its own as a ::before.

Currently it uses FontAwesome similar to the .collapse-indicator, but does not get the fout-hiding from .fa-elements. So I think we need to switch to a svg-based background as a FOUT could happen.

collapse indicator

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/event/mail_settings_fragment.html
+++ b/src/pretix/control/templates/pretixcontrol/event/mail_settings_fragment.html
@@ -5,7 +5,6 @@
     <summary class="panel-heading">
         <h4 class="panel-title">
             <strong>{% trans title %}</strong>
-            <i class="fa fa-angle-down collapse-indicator"></i>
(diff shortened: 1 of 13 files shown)


wiffbi:
I removed the unnecessary collapse-indicators from panel-title HTML in presale as well as control. I still hide the indicators through CSS if e.g. a plugin would use them inside a `details>.panel-title`. I also unified the use in email-settings as well as product-variants (which used a special chevron not seen elsewhere). I also moved the collapse-indicator inside the toggle-variants-button in presale to the left.

raphaelm (approved):

PR #3715, opened 2023-11-13
Widget: label button in event-list with "More info“ when availability is unknown (Z#23135197)

wiffbi:


src/pretix/presale/views/widget.py
```diff
@@ -405,7 +405,7 @@ def _get_availability(self, ev, event, tz=None):
                 availability['reason'] = 'full'
             else:  # unknown / no product
                 availability['color'] = 'none'
-                availability['text'] = ''
+                availability['text'] = 'More info'
```
raphaelm, line 408:
```suggestion
                availability['text'] = gettext('More info')
```

raphaelm (changes requested):

src/pretix/presale/views/widget.py
wiffbi: Thanks!

PR #3716, opened 2023-11-13
Allow to postpone invoice creation on order changes

raphaelm:
We offer the following option in invoice settings:

image

A customer who sets this option will assume that (not counting the special payment methods), no invoice will ever be generated for an amount that has not been paid. However, this is not the case, as we currently have an implicit rule "when an order has at least one invoice, it will always receive a current invoice". Consider the following scenario as it happens now:

1. Customer purchases ticket for €20
    - No invoice generated yet
2. Customer pays by credit card
    - Invoice over €20 is now generated
3. Customer adds a €5 add-on to the order, the order goes back to pending state
    - A cancellation + a partially paid invoice over €25 is now generated
4. Customer pays €5 by credit card
    - Nothing happens as invoice already exists

Whereas the organizer might rightfully assume the selection in the screenshot to cause the following:

1. Customer purchases ticket for €20
    - No invoice generated yet
2. Customer pays by credit card
    - Invoice over €20 is now generated
3. Customer adds a €5 add-on to the order, the order goes back to pending state
    - No new invoice generated yet, as new amount is not paid
4. Customer pays €5 by credit card
    - A cancellation + a paid invoice over €25 is now generated

This PR changes this behaviour by introducing an internal "invoice_dirty" flag that says: Whenever this order next has a transition from "pending" to "paid" state, it needs a new invoice.


src/pretix/presale/views/order.py
```diff
@@ -487,9 +487,13 @@ def dispatch(self, request, *args, **kwargs):
 
     def post(self, request, *args, **kwargs):
         try:
-            if not self.order.invoices.exists() and invoice_qualified(self.order):
+            i = self.order.invoices.filter(is_cancellation=False).last()
+            has_active_invoice = i and not i.refered.exists()
```
wiffbi, line 491: See comment above

```suggestion
            has_active_invoice = i and not i.canceled
```

src/pretix/base/services/orders.py
```diff
@@ -2619,14 +2619,34 @@ def _payment_fee_diff(self):
...
+                    # Backwards-compatible behaviour
+                    self.event.settings.invoice_generate not in ("True", "paid") and
+                    i and
+                    not i.refered.exists()
+                )
+            )
+
+            if order_now_qualified:
+                if invoice_should_be_generated:
+                    if i and not i.refered.exists():
```
wiffbi, line 2641: I am not sure, the code comments read `Invoice.refered` as a general way to refer invoices. Shouldn’t this be explicitly about refered.is_cancelation, ie.:
```suggestion
                    if i and not i.canceled:
```
Otherwise this code IMHO assumes that every refered is a cancellation – might be the case in reality though as it was used this way in older code (e.g. line 348 or 2622) …

src/pretix/base/services/orders.py
```diff
@@ -2619,14 +2619,34 @@ def _payment_fee_diff(self):
...
+                        self._invoices.append(generate_cancellation(i))
+                    self._invoices.append(generate_invoice(self.order))
+                else:
+                    self.order.invoice_dirty = True
+                    self.order.save(update_fields=["invoice_dirty"])
+            else:
+                if i and not i.refered.exists():
```
wiffbi, line 2648: See comment above
```suggestion
                if i and not i.canceled:
```

src/pretix/base/services/orders.py
```diff
@@ -2619,14 +2619,34 @@ def _payment_fee_diff(self):
...
+                    self.open_payment is not None and
+                    self.open_payment.payment_provider.requires_invoice_immediately
+                ) or (
+                    self.event.settings.invoice_generate == "paid" and
+                    self.order.status == Order.STATUS_PAID
+                ) or (
+                    # Backwards-compatible behaviour
+                    self.event.settings.invoice_generate not in ("True", "paid") and
+                    i and
+                    not i.refered.exists()
```
wiffbi, line 2635: See comment below

```suggestion
                    not i.canceled
```

src/pretix/base/models/orders.py
```diff
@@ -1841,9 +1845,14 @@ def _mark_order_paid(self, count_waitinglist=True, send_mail=True, force=False,
...
             gen_invoice = (
                 (invoices == 0 and self.order.event.settings.get('invoice_generate') in ('True', 'paid')) or
-                0 < invoices <= cancellations
+                0 < invoices <= cancellations or
+                self.order.invoice_dirty
             )
             if gen_invoice:
+                if invoices:
+                    last_i = self.order.invoices.filter(is_cancellation=False).last()
+                    if not last_i.refered.exists():
```
wiffbi, line 1854: See comment below regarding refered being exclusively cancelations. Might be that case in reality, but code comments IMHO do not read that way. So maybe more robust

```suggestion
                    if not last_i.canceled:
```

wiffbi (approved):
Looks good. Just a minor comment regarding refered being regarded as always being a cancelation …

PR #3718, opened 2023-11-14
A11y: add date and time labels in MultiWidget (Z#23132744)

wiffbi:


raphaelm:
Anything still required here?

wiffbi:
Could probably be merged, was actually waiting for feedback from a client regarding other issues the client reported regarding a11y on the date/time selector widget, which I could not reproduce. I will look into the current changes to this PR and will mark it as ready for review then.

raphaelm (approved):

PR #3724, opened 2023-11-15
This voucher code can only be redeemed %d more times.

HansF:
Issue: To simulate: create a voucher that gives access to multiple tickets. 
Let's say 2, open the link generated by the code, add 5 tickets to your cart, click next. 

It will display the error message "This voucher code can only be redeemed 2 more times."
My expectation would be to display "This voucher code can only be redeemed 2 times."

So I did some find / replace stuff in the strings and updated the translations I was comfortable with,  the french translation was allready correct btw. I hope I didn't break anything. :-)

### DIFF
--- a/src/pretix/base/services/cart.py
+++ b/src/pretix/base/services/cart.py
@@ -162,7 +162,7 @@ error_messages = {
         'to redeem it before but did not complete the checkout process. You can try to use it '
         'again in %d minutes.'
     ),
-    'voucher_redeemed_partial': gettext_lazy('This voucher code can only be redeemed %d more times.'),
+    'voucher_redeemed_partial': gettext_lazy('This voucher code can only be redeemed %d times.'),
     'voucher_whole_cart_not_combined': gettext_lazy('Applying a voucher to the whole cart should not be combined with other operations.'),
     'voucher_double': gettext_lazy(
         'You already used this voucher code. Remove the associated line from your '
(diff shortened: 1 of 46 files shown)


raphaelm:
Thanks, but not sure it's that simple. With your change it would be confusing/wrong for a voucher that can be used 5 times but has been used 3 times already (which is where the "more" comes from)

HansF:
In that case you get a green box stating "The products have been successfully added to your cart." and not the aforementioned message.

raphaelm:
> In that case you get a green box stating "The products have been successfully added to your cart." and not the aforementioned message in red. 

No, I meant:

- Voucher redeemed 3 out of 5 times
- You try to redeem 3 more
- Error message says "can only be redeemed 2 more times", which is correct, whereas "can only be redeemed 2 times" (your suggestion) is incorrect.

> Also the number it returns is the total number it can be used, and not the amount of uses left.

This is incorrect.

```
                    err = err or (error_messages['voucher_redeemed_partial'] % voucher_available_count)
```

voucher_available_count is the number remaining

HansF:
Thanks for taking your time to reply to me.

Sorry,  I worded that wrong:
The number it returns is the total number it '_has been used in the last add to cart_', and not the amount of uses left.
I think changing it to "**could** only be redeemed 2 times" might solve it. 

https://www.youtube.com/watch?v=JLefTf_01ZI

raphaelm:
> Sorry, I worded that wrong: The number it returns is the total number it '_has been used in the last add to cart_', and not the amount of uses left. I think changing it to "**could** only be redeemed 2 times" might solve it.

No. It is the number of times it can still be used, not counting the operation which the error message is a reply to, and as such, it is correctly worded as it is.

PR #3725, opened 2023-11-16
Add notification signal (Z:#23127501)

ser8phin:
Adds a new signal for general notifications and sends it every time the notify task is triggered

### DIFF
--- a/src/pretix/base/signals.py
+++ b/src/pretix/base/signals.py
@@ -392,6 +392,16 @@ as the first argument.
 As with all event-plugin signals, the ``sender`` keyword argument will contain the event.
 """
 
+refund_requested = EventPluginSignal()
+"""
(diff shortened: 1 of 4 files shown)


src/pretix/base/signals.py
```diff
@@ -279,6 +279,13 @@ def connect(self, receiver, sender=None, weak=True, dispatch_uid=None):
 notification settings!
 """
 
+notification = EventPluginSignal()
+"""
+This signal is sent out when a notification is send.
```
raphaelm, line 284: The documentation is missing information on the parameters that will be available

luelista:
```suggestion
This signal is sent out when a notification is sent.
```

raphaelm (approved):

PR #3732, opened 2023-11-20
LogEntry: Add a direct relationship to organizer

raphaelm:
We currently are not able to tie log entries directly to organizers. This is problematic when we want to delete organizers with all their data, since we can't easily identify all logentries that need to be deleted. This PR adds

The migration is obviously not the one we could run on pretix production to backfill millions of rows, we'd need to start a very long-running backfill process for that. Therefore, this PR introduces this field but does not yet make use of it.


src/pretix/base/migrations/0251_logentry_organizer.py
```diff
@@ -0,0 +1,56 @@
...
+        )
+    )
+    for ct in ContentType.objects.all():
+        try:
+            model = apps.get_model(ct.app_label, ct.model)
+        except LookupError:
+            continue
+        if "organizer" in model._meta.fields:
+            LogEntry.objects.filter(
+                organizer_link__isnull=True, event__isnull=False, content_type=ct,
```
wiffbi, line 26: Do we need the `event__isnull=False` here? We already have the organizer in the model so we probably can leave that out? Or should we even do ``event__isnull=True`` as we already updated all LogEntries that have an event attached in line 14-16.

Also, having ``event__isnull=True`` would update LogEntries, that are related to a model with an organizer, but not an event. Not sure, we have logs for this though …

src/pretix/base/migrations/0251_logentry_organizer.py
```diff
@@ -0,0 +1,56 @@
...
+            ).update(
+                organizer_link_id=Subquery(model.objects.filter(pk=OuterRef('object_id')).values('organizer_id'))
+            )
+        elif "event" in model._meta.fields:
+            LogEntry.objects.filter(
+                organizer_link__isnull=True, event__isnull=False, content_type=ct,
+            ).update(
+                organizer_link_id=Subquery(model.objects.filter(pk=OuterRef('object_id')).values('event__organizer_id'))
+            )
```
wiffbi, line 35: Do we need this at all? LInes 13-17 updated all LogEntries with events attached, so as far as I understand it, this should not add anything – it just takes the route through the model, that is linked to the event and then takes the organizer.

src/pretix/base/migrations/0251_logentry_organizer.py
```diff
@@ -0,0 +1,56 @@
...
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("pretixbase", "0250_eventmetaproperty_filter_public"),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name="logentry",
+            name="organizer_link",
```
wiffbi, line 46: What is the reason you called this `organizer_link` instead of just `organizer`?

src/pretix/base/models/log.py
```diff
@@ -126,7 +127,9 @@ def notification_type(self):
     def organizer(self):
         from .organizer import Organizer
 
-        if self.event:
+        if self.organizer_link:
+            return self.event.organizer_link
```
wiffbi, line 131: I guess this is a typo?

```suggestion
            return self.organizer_link
```

wiffbi (commented):
Looks good, just a question regarding naming and one regarding updating log-entries without an event, but an organizer attached (through the log-entry’s related model).

src/pretix/base/migrations/0251_logentry_organizer.py
raphaelm: I think this should have been event__isnull=True, yes

raphaelm: > Also, having event__isnull=True would update LogEntries, that are related to a model with an organizer, but not an event. Not sure, we have logs for this though …

Yes, like a Customer account?

src/pretix/base/migrations/0251_logentry_organizer.py
raphaelm: Yes, this theoretically should not match anything if all data is correct

src/pretix/base/migrations/0251_logentry_organizer.py
raphaelm: Because we already have a `def organizer` and it wouldn't be backwards-compatible otherwise. We can rename it once we migrated all data on production, if we want to

src/pretix/base/migrations/0251_logentry_organizer.py
wiffbi: Ah, if correct. That is good, more robust code. :)

src/pretix/base/migrations/0251_logentry_organizer.py
wiffbi: Ah, thanks for explaining this!

PR #3736, opened 2023-11-22
Update translations

pretix-translations:


c0de-bender:
Hey, we made a mistake yesterday in these commits when uploading the file on translate.pretix.eu by using these settings:

2023-11-22_1-pretix-polish-upload-mistake-settings

We started to also translate fuzzy strings, but of course not all of them, and the file yesterday didn't actually contain any changes to fuzzy strings yet. This upload mistake had the following effect:

- it removed most of the "fuzzy" / "needs editing" markers (of strings for which the English source changed and the Polish translation is outdated),
  - thereby flagging the outdated Polish translations as approved.
- it removed all the old original English strings for these.

When comparing our version of the file (left) to the upload result file (right – we downloaded it from the translation website, but it's the same as in the first commit above), this is how the mistake looked in a file comparison:

diff-after-upload-mistake

_We tried to fix this_ by uploading the same file again with these settings:
2023-11-22_2-pretix-polish-upload-fix-settings

This kind of worked – it marked all the strings which were previously flagged "fuzzy" as fuzzy again. However, when comparing the result file to our version, the committed result file was still missing the old English source strings:
diff-after-upload-partial-fix

Interestingly, the translation website [still knows about the source change and the old English text](https://translate.pretix.eu/translate/pretix/pretix/pl/?checksum=90bf6ee2bb81b842):

translation-website-still-knows-about-source-change


### Questions

a) Shall we revert our mistake and also restore the old English source texts in the file? I could easily make a Pull Request for that.

b) How can the website still know the old English text when the information is gone from the translation file?
b.1) Is it somehow possible to copy this information from the website's "database" back into the file?

c) I saw changes [in your "merge commit"](https://github.com/pretix/pretix/pull/3736/commits/a0089ae58eed0b62963c05669a488d857bde0602) of this Pull Request, mostly deleting leftover outdated Polish translations, old English strings and their fuzzy markers, flagging those strings as "Untranslated" instead of "Needs editing". Would you actually prefer that for all fuzzy strings?

d) More an observation than a question: I saw that actually our upload mistake didn't remove all fuzzy markers and old English strings (it kept 93 out of previously 913) – I don't see any pattern there and don't know why it only removed most of them, but not all.

PR #3737, opened 2023-11-23
Stripe: Add Support for Affirm Pay Later

pc-coholic:
This PR does not only add support for Affirm, but also slightly changes the way SCA and `StripePaymentIntentMethod` behave.

- `StripePaymentIntentMethod`s do not absolutely require a `payment_method_id` anymore. If there is none, the Idempotency-key is generated based on the payment.full_id instead of include the `payment_method_id`. 
- The `SCAView` and `SCAReturnView` now check a new property to determine how the redirect should be handled. As - right now - only Credit Cards are using the SCA view, they will keep being displayed in the iFrame. Other payment methods may opt to use a proper redirect and not be iFramed. The Javascript has been outfitted with some extra code to take care of this.

Regarding Affirm, we are using a very barebones approach:
- create payment intent (without providing any additional data since it is only recommended but not required), and by not passing a Stripe-JS-created payment_method_id
- redirect the user to Affirm
- Have them come back

Webhooks, etc. have not been tested, but according to Stripe's documentation, a successfull Affirm-payment will behave exactly the same as any PI/PM-payment.


src/pretix/plugins/stripe/payment.py
```diff
@@ -421,6 +421,16 @@ def settings_form_fields(self):
                      help_text=_('Needs to be enabled in your Stripe account first.'),
                      required=False,
                  )),
+                ('method_affirm',
+                 forms.BooleanField(
+                     label=_('Affirm'),
+                     disabled=self.event.currency not in ['USD', 'CAD'],
+                     help_text=' '.join([
+                         _('Needs to be enabled in your Stripe account first.'),
+                         _('Only available for payments between $50 and $30,000.')
```
raphaelm, line 430:
```suggestion
                         str(_('Needs to be enabled in your Stripe account first.'),)
                         str(_('Only available for payments between $50 and $30,000.'))
```

PR #3739, opened 2023-11-23
Improve UI to configure unavailable items handling (Z#23131828)

luelista:
This PR adds the possibility to have products, that are not available currently due to the "available_from" or "available_until" dates, still visible in the frontend, with an according notice for the user.

grafik

In the control panel, this is implemented using a new ButtonGroupRadioSelect element next to the relevant date fields. For consistency, the require_voucher field gets the same ButtonGroupRadioSelect element to toggle the hide_without_voucher option. 
grafik

available_from and available_until get new available_from_mode and available_until_mode database fields with the options "hide" and "info", which correspond directly to the options of the ButtonGroupRadioSelect. 

require_voucher and hide_without_voucher are mapped as follows:

| require_voucher | hide_without_voucher | "Require voucher" checkbox |  ButtonGroupRadioSelect  |
|-|-|-|-|
| False | False | unchecked | disabled |
| True | False | checked |  info |
| False | True |  (invalid) | (invalid) |
| True | True |  checked | hide |

We check in the Item.save(), ItemUpdateForm.clean() and ItemUpdateGeneral.get_context_data() to prevent the "invalid" combination.

TODO:

- [x] add available_from_mode and available_until_mode to API and docs
- [x] use dynamic translation texts for unavailability_reason in widget


src/pretix/base/models/items.py
```diff
@@ -485,11 +492,21 @@ class Item(LoggedModel):
...
     available_until = models.DateTimeField(
         verbose_name=_("Available until"),
         null=True, blank=True,
         help_text=_('This product will not be sold after the given date.')
     )
+    available_until_mode = models.CharField(
+        choices=UNAVAIL_MODES,
+        default=UNAVAIL_MODE_HIDDEN,
+        max_length=16,
+    )
```
raphaelm, line 509: These should also be exposed (and documented) in the API :)

src/pretix/control/templates/pretixcontrol/button_group_radio_option.html
```diff
@@ -0,0 +1,2 @@
+<label{% if widget.attrs.id %} for="{{ widget.attrs.id }}"{% endif %} class="btn btn-primary-if-active form-field-boundary {% if widget.attrs.checked %} active{% endif %}" title="{{ widget.label }}" data-toggle="tooltip">
+{% include "django/forms/widgets/input.html" %} <span class="fa fa-{{ widget.attrs.icon }}"></span></label>
```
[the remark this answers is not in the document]
luelista: added the CSS rule

src/pretix/control/templates/pretixcontrol/item/index.html
```diff
@@ -152,27 +152,33 @@
...
-                        {% bootstrap_field form.hide_without_voucher layout="control" %}
-                        {% bootstrap_field form.require_bundling layout="control" %}
+
+                        <div class="form-group">
+                            <div class="col-md-10"></div>
+                            <div class="col-md-2 text-right">{% trans "Visibility" %} <span class="fa fa-question-circle-o" data-toggle="tooltip" title="{% trans "xxxxx" %}"></span>
+                            </div>
+                        </div>
+
+                        {% bootstrap_field form.sales_channels layout="control" horizontal_field_class="col-md-7" %}
```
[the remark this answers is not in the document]
wiffbi: As I said, not a hill I will die on. I like it either way :) , just thought maybe we could save some space and also visually make these toggles part of the current setting and not a kind of extra column, that feels like a separated visibility toggle.

src/pretix/base/models/items.py
luelista: First time adding something to the API, could you have a look if it's correct? https://github.com/pretix/pretix/pull/3739/commits/c27c7292695ac862970557b3477c12b4307aee3e

raphaelm (changes requested):
Unfortunately, I noticed a larger thing to talk about: I think we forgot to consider variations and subevents so far.

In addition to `Item.available_from/until` we also have two more ways to make an item unavailable by time:

- `ItemVariation.available_from/until`
- `SubEventItem.available_from/until` / `SubEventItemVariation.available_from/until`

Both are relevant here as they are basically the same thing on another level. `SubEventItem(Variation)` is basically the only useful way to use `SubEventItem`  in an event series where you want products to hide or show up e.g. ~24h before the event.

So we have two options now:

- The `ItemVariation`, `SubEventItem`, and `SubEventItemVariation` models get their own `available_from_mode` and `available_until_mode` which makes at least the subevent form way more complicated
- Everything is controlled by the `Item` flags, which makes the UI this PR adds less intuitive because suddenly it *does* make a different what mode is set even when the constraint itself is kept empty

Both sound acceptable to me, but the current implementation just ignores it entirely and that won't hold up for long.

luelista:
I'm not sure where to put the indicator icon in the subevent date view. I like Variant B a little better, but I was only able to implement it in a somewhat hacky way (see CSS code below, which moves the icon around using `position: relative`). Any opinions?

### Variant A
image

### Variant B
image
```
.unavail-mode-indicator {
  position: relative;
  top: 32px;
  right: 10px;
}
```

doc/api/resources/items.rst
```diff
@@ -50,8 +50,16 @@ sales_channels                          list of strings            Sales channel
                                                                    ``"web"`` or ``"resellers"``. Defaults to ``["web"]``.
 available_from                          datetime                   The first date time at which this item can be bought
                                                                    (or ``null``).
+available_from_mode                     string                     If ``hide`` (the default), this item is hidden in the shop
+                                                                   if unavailable due to the available_from setting.
```
raphaelm, line 54:
```suggestion
                                                                   if unavailable due to the ``available_from`` setting.
```

doc/api/resources/items.rst
```diff
@@ -50,8 +50,16 @@ sales_channels                          list of strings            Sales channel
...
+                                                                   If ``info``, the item is visible, but can't be purchased,
+                                                                   and a note explaining the unavailability is displayed.
 available_until                         datetime                   The last date time at which this item can be bought
                                                                    (or ``null``).
+available_until_mode                    string                     If ``hide`` (the default), this item is hidden in the shop
+                                                                   if unavailable due to the available_until setting.
```
raphaelm, line 60:
```suggestion
                                                                   if unavailable due to the ``available_until`` setting.
```

doc/api/resources/items.rst
```diff
@@ -156,8 +164,16 @@ variations                              list of objects            A list with o
                                                                    available.
 ├ available_from                        datetime                   The first date time at which this variation can be bought
                                                                    (or ``null``).
+├ available_from_mode                   string                     If ``hide`` (the default), this variation is hidden in the shop
+                                                                   if unavailable due to the available_from setting.
```
raphaelm, line 168:
```suggestion
                                                                   if unavailable due to the ``available_from`` setting.
```

doc/api/resources/items.rst
```diff
@@ -156,8 +164,16 @@ variations                              list of objects            A list with o
...
+                                                                   If ``info``, the variation is visible, but can't be purchased,
+                                                                   and a note explaining the unavailability is displayed.
 ├ available_until                       datetime                   The last date time at which this variation can be bought
                                                                    (or ``null``).
+├ available_until_mode                  string                     If ``hide`` (the default), this variation is hidden in the shop
+                                                                   if unavailable due to the available_until setting.
```
raphaelm, line 174:
```suggestion
                                                                   if unavailable due to the ``available_until`` setting.
```

src/pretix/control/templates/pretixcontrol/subevents/fragment_unavail_mode_indicator.html
```diff
@@ -0,0 +1,6 @@
+{% load i18n %}
+{% if mode == "hide" %}
+<span class="pull-right text-muted unavail-mode-indicator" data-toggle="tooltip" title="{% trans "Hide product if unavailable" %}. {% trans "To change this option, use the product settings of the base event." %}"><span class="fa fa-eye-slash"></span></span>
```
raphaelm, line 3:
```suggestion
<span class="pull-right text-muted unavail-mode-indicator" data-toggle="tooltip" title="{% trans "Hide product if unavailable" %}. {% if f.variation %}{% trans "You can change this option in the variation settings." %}{% else %}{% trans "You can change this option in the product settings." %}{% endif %}"><span class="fa fa-eye-slash"></span></span>
```

src/pretix/control/templates/pretixcontrol/subevents/fragment_unavail_mode_indicator.html
```diff
@@ -0,0 +1,6 @@
...
+{% else %}
+<span class="pull-right text-muted unavail-mode-indicator" data-toggle="tooltip" title="{% trans "Show info text if unavailable" %}. {% trans "To change this option, use the product settings of the base event." %}"><span class="fa fa-info-circle"></span></span>
```
raphaelm, line 5:
```suggestion
<span class="pull-right text-muted unavail-mode-indicator" data-toggle="tooltip" title="{% trans "Show info text if unavailable" %}. {% if f.variation %}{% trans "You can change this option in the variation settings." %}{% else %}{% trans "You can change this option in the product settings." %}{% endif %}"><span class="fa fa-info-circle"></span></span>
```

src/pretix/control/templates/pretixcontrol/subevents/detail.html
```diff
@@ -145,11 +145,13 @@ <h4 class="panel-title">
...
-                                <label for="{{ f.available_from.id_for_label }}" class="text-muted">{% trans "Available from" %}</label><br>
-                                {% bootstrap_field f.available_from form_group_class="" layout="inline" %}
+                                <label for="{{ f.available_from.id_for_label }}" class="text-muted">{% trans "Available from" %}</label>
+                                {% include "pretixcontrol/subevents/fragment_unavail_mode_indicator.html" with mode=f.available_from_mode %}<br>
+                                {% bootstrap_field f.available_from form_group_class="foo" layout="inline" %}
                             </div>
                             <div class="col-md-4">
-                                <label for="{{ f.available_until.id_for_label }}" class="text-muted">{% trans "Available until" %}</label><br>
+                                <label for="{{ f.available_until.id_for_label }}" class="text-muted">{% trans "Available until" %}</label>
+                                {% include "pretixcontrol/subevents/fragment_unavail_mode_indicator.html" with mode=f.available_until_mode %}<br>
```
raphaelm, line 154: This should probably be added to `bulk_edit.html` and `bulk_add.html` as well

raphaelm (commented):
Very good work, just a tiny few final remarks before this can ship :)

> I'm not sure where to put the indicator icon in the subevent date view. I like Variant B a little better, but I was only able to implement it in a somewhat hacky way (see CSS code below, which moves the icon around using position: relative). Any opinions?

I actually like variant A better. Variant B is pretty, but makes it less clear that it is something you can interact with.

src/pretix/control/templates/pretixcontrol/subevents/detail.html
luelista: Done ✅

PR #3741, opened 2023-11-24
Check-in: New error reason for unapproved orders

raphaelm:


doc/api/resources/checkinlists.rst
```diff
@@ -632,7 +632,8 @@ Order position endpoints
                                        set this to ``false``. In that case, questions will just be ignored. Defaults
                                        to ``true``.
    :<json boolean canceled_supported: When this parameter is set to ``true``, the response code ``canceled`` may be
-                                      returned. Otherwise, canceled orders will return ``unpaid``.
+                                      returned. Otherwise, canceled orders will return ``unpaid``. (**Deprecated**, will
+                                      be removed and always ``true`` in the future.)
```
pc-coholic, line 636: Unclear to me, what exactly will be always `true` in the future. Perhaps rephrase this more along the lines of "In the future, the reason `unpaid` may always be given, regardless if `canceled_supported` is set or not").

pc-coholic (changes requested):

PR #3762, opened 2023-12-06
Rename LogEntry.organizer_link to LogEntry.organizer

raphaelm:
Now that the migration has completed on pretix.eu, we can rename the field on django-level (I'm keeping the DB name to save us yet another hard migration)


wiffbi (approved):
Looks good! Had to look up, what/why the AlterField and RenameField do what they do, but basically no-op on DB, just change the name.

PR #3764, opened 2023-12-07
PDF-Editor: fix controls when resizing multiple objects

wiffbi:


src/pretix/static/pretixcontrol/js/ui/editor.js
```diff
@@ -466,7 +466,15 @@ var editor = {
         $("#toolbox-position-x").val(editor._px2mm(o.left).toFixed(2));
         $("#toolbox-position-y").val(editor._px2mm(bottom).toFixed(2));
 
-        if (o.type === "barcodearea") {
+        if (o.type === 'activeSelection') {
+            // limit controls to the ones which are common to every element in selection
+            o.forEachControl(function(control, k) {
+                o.setControlVisible(k, o.getObjects().every(function(obj) {
+                    // special case „text“ in group does not support any controls
+                    return obj.type !== "text" && obj.type !== "textarea" && obj.isControlVisible(k)
```
luelista, line 474:
```suggestion
            o.forEachControl(function(control, key) {
                o.setControlVisible(key, o.getObjects().every(function(obj) {
                    // special case „text“ in group does not support any controls
                    return obj.type !== "text" && obj.type !== "textarea" && obj.isControlVisible(key)
```

luelista (approved):
works for me :)

PR #3768, opened 2023-12-10
API: Prefetch customer in order list view

stapelberg:
Before this change, each order row needed an additional database query just to resolve the customer id (integer) to customer identifier (string).

With this change, django does a database JOIN.

This is a performance optimization. In my setup, querying a few hundred orders now takes ≈1.5s instead of 2 to 2.5s


raphaelm (approved):

stapelberg:
Not sure why the “Build / Packaging (3.11) (pull_request)” step is failing, but it does not seem to be related to my change. Can you take a look please?

raphaelm:
> Not sure why the “Build / Packaging (3.11) (pull_request)” step is failing, but it does not seem to be related to my change. Can you take a look please?

Yup, broken on master, not your fault!

PR #3769, opened 2023-12-11
Bank transfer: Improve refund handling

raphaelm:
The bank import has weird handling of negative amounts. This is understandable, since we don't have sufficient information on what the user actually wants. This PR makes the handling less weird (although not entirely non-weird).


wiffbi (approved):
Looks good!

PR #3787, opened 2023-12-28
Fix redirect loop caused when both 2fa and password change are forced

kiancross:
When both two factor authentication and a password change are being forced, a redirect loop occurs. This commit resolves this by allowing the password change to occur between 2fa setup is forced. More specifically, 2fa is no longer forced whilst a forced password change is in operation.


raphaelm:
Thank you!

PR #3801, opened 2024-01-08
Name formats: Allow empty salutation

raphaelm:
Currently, if a name format with salutation is selected, we show the following choices:

* `---` (pre-selected, but choosing this is not allowed)
* Mr
* Ms
* Mx

The idea behind this design choice was:

- We do not allow the empty option since we don't want people skipping the field out of pure laziness
- We always provide an option for people who are neither Mr not Ms. If someone chooses Mx, that will later not be used e.g. in emails, instead, we'll be addressing the person simple as `{given_name} {family_name}`

Now, we've had a complaint that not only some people might not want to use Mx either, but furthermore some people might not want to *disclose* which salutation is appropriate for them, therefore an empty option should be selectable. I still think it is a good idea to disallow the pre-selected option to be used to increase data quality by forcing people to choose something, but this PR introduces an additional option that is really empty.

* `---` (pre-selected, but choosing this is not allowed)
* (empty option, may be selected)
* Mr
* Ms
* Mx

For technical reasons, the new option can't be the empty string on HTML level, otherwise the `required` HTML attribute would no longer work, so there's some encoding and decoding in the form layer required.

### DIFF
--- a/src/pretix/base/forms/questions.py
+++ b/src/pretix/base/forms/questions.py
@@ -125,7 +125,7 @@ class NamePartsWidget(forms.MultiWidget):
             if fname == 'title' and self.titles:
                 widgets.append(Select(attrs=a, choices=[('', '')] + [(d, d) for d in self.titles[1]]))
             elif fname == 'salutation':
-                widgets.append(Select(attrs=a, choices=[('', '---')] + PERSON_NAME_SALUTATIONS))
+                widgets.append(Select(attrs=a, choices=[('', '---'), ('empty', '')] + PERSON_NAME_SALUTATIONS))
             else:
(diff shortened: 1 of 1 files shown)


luelista (approved):

src/pretix/base/forms/questions.py
```diff
@@ -136,7 +136,11 @@ def decompress(self, value):
         data = []
         for i, field in enumerate(self.scheme['fields']):
             fname, label, size = field
-            data.append(value.get(fname, ""))
+            fval = value.get(fname, "")
+            print(field, repr(fval))
```
felixrindt, line 140:
```suggestion
```

PR #3806, opened 2024-01-11
Add upper limit on positions in an order

raphaelm:
Sometimes, people create orders with *thousands* of tickets in them. And then stuff breaks. Backend pages break, forms break, apps break, because pretix is not designed for that.

This PR tries to reduce the danger by setting an upper limit of 500 tickets per order.

The main reason I did not do this before was worry it could be read as "up to 500 everything will always be fine", which is also not true. But at least this will prevent the extreme cases.


wiffbi (approved):
Looks good to me. Should we add a max-attribute (form_kwargs) to the settings input? Might be problematic with existing settings that are above `PRETIX_MAX_ORDER_SIZE` though when saving settings.

PR #3822, opened 2024-01-19
Subevent bulk creation: Handle available_from of items the same way as dates

raphaelm:
The "bulk copy" of subevents had inconsistent handling of `presale_start`/`presale_end` vs `available_from`/`available_until` of specific products. This PR unifies the behaviour.


wiffbi (approved):
Works like a charm. Only thing is the time-part on relative date input, which I originally left empty and gets translated to the current time of the event to copy. Not a big deal and I do not see a way around it, just something I noticed.

PR #3828, opened 2024-01-23
Allow to set amount and date when manually confirming a payment

raphaelm:


wiffbi (approved):
The only thing I stumbled upon is that when setting today’s date, it uses now() with current time, when setting a date manually, I cannot add a time. Not that I want to, just something I noticed, which is a bit inconsistent with user-input, but on the other totally okay! So, looks good to me. :)

PR #3829, opened 2024-01-23
Fix auto-selected date in calendar with long-running events (Z#23141457)

raphaelm:


wiffbi (changes requested):
I think, this is not working right yet. I tested it locally with a long running event and at first it seemed to work right as it set „now“ as the starting time. But from my understanding of the code, that did not make any sense, as it should have selected the event that is currently running and set the starting time of the widget to event.date_from. So I dug deeper and it actually did not select the currently running event, but did not select any event. When I added an event in the future, it actually selected the event in the future – looking at the SQL generated it showed that using Python’s `now()` instead of SQL’s/djangos `django.db.models.functions.Now()` creates Microseconds differences inside the same query (might be my local dev-setup, but it does), which effectively made the `filter(effective_date__gte=now())` filter out any currently running events. So when I replaced `now()` with `Now()` it acutally selected the currently running event, but of course set the startdate to this event’s date_from, which still results in starting in the past.

So I think the approach with the annotation is right (when using SQLs `Now()`), but you need to actually use the `effective_date` to set `self.year` and `self.month` instead of date_from – probably not practically accessible so you need to extend the if-check before setting datetime_from to the same condition as the SQL-effective_date-Case-assignment.

wiffbi (approved):
Works now! You missed to replace one instance of now() – it actually has practically no effect, but just to make sure I added a comment.

PR #3830, opened 2024-01-23
Add "Partitioned" flag to our cookies

raphaelm:
Chrome is deprecating third-party cookies that may be used for tracking. The 1% trial is already running:

https://developers.google.com/privacy-sandbox/blog/cookie-countdown-2023oct

You can opt-in for the trial at chrome://flags/#test-third-party-cookie-phaseout

This breaks the widget entirely. The only suitable fix is marking our cookies as "Partitioned":
https://developers.google.com/privacy-sandbox/3pcd/chips

Unfortunately, Firefox and Safari not yet support that. If they do some day, it will also fix our cross-site issues there.

This PR sets Partitioned on all cookies that we set SameSite=None on, in all browsers that are known not to break with unknown cookie flags.

### DIFF
--- a/src/pretix/helpers/cookies.py
+++ b/src/pretix/helpers/cookies.py
@@ -44,6 +44,8 @@ def set_cookie_without_samesite(request, response, key, *args, **kwargs):
         # This will only work on secure cookies as well
         # https://www.chromestatus.com/feature/5633521622188032
         response.cookies[key]['secure'] = is_secure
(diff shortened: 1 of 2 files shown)


luelista (approved):
LGTM, tested in Firefox (Partitioned is sent by Pretix but ignored by Firefox) and in Chrome (Partitioned seems to work correctly)

PR #3836, opened 2024-01-25
Let plugins allow/prevent the download of individual tickets in an order

luelista:
The pretix-shipping plugin needs this to prevent the download only of the tickets which should be shipped physically to the customer, but not the tickets which should stay available for Print-at-home.

This modifies the `allow_ticket_download` signal so that it can not only return True or False to allow or block ticket download for an order as a whole, but alternatively return a list of OrderPositions, making only those downloadable.

https://github.com/pretix/pretix/pull/3836/files#diff-efaade2fa7bebe18526b3532280a3d750061a038953066c3a43b38be7d87c8d1R644-R652

### DIFF
--- a/src/pretix/base/models/orders.py
+++ b/src/pretix/base/models/orders.py
@@ -1144,7 +1144,6 @@ class Order(LockModel, LoggedModel):
     @property
     def positions_with_tickets(self):
         signal_response = allow_ticket_download.send(self.event, order=self)
-        print("signal_response", signal_response)
         if all([r is True for rr, r in signal_response]):
             return self.positions_with_tickets_ignoring_plugins
         elif any([r is False for rr, r in signal_response]):
--- a/src/pretix/base/models/orders.py
+++ b/src/pretix/base/models/orders.py
@@ -44,7 +44,7 @@ from datetime import datetime, time, timedelta
 from decimal import Decimal
 from functools import reduce
 from time import sleep
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, List, Union, Iterable
 from zoneinfo import ZoneInfo
 
 import dateutil
@@ -79,7 +79,7 @@ from pretix.base.i18n import language
 from pretix.base.models import Customer, User
 from pretix.base.reldate import RelativeDateWrapper
 from pretix.base.settings import PERSON_NAME_SCHEMES
-from pretix.base.signals import order_gracefully_delete
+from pretix.base.signals import order_gracefully_delete, allow_ticket_download
 
 from ...helpers import OF_SELF
 from ...helpers.countries import CachedCountries, FastCountryField
@@ -1199,6 +1199,16 @@ class Order(LockModel, LoggedModel):
         self._transaction_key_reset()
         _transactions_mark_order_clean(self.pk)
         return create
+
+    @property
+    def plugins_allow_ticket_download(self):
+        signal_response = allow_ticket_download.send(self.event, order=self)
+        if all([r == True for rr, r in signal_response]):
+            return True
+        elif any([r == False for rr, r in signal_response]):
+            return False
+        else:
+            return set.intersection(*[set(r) for rr, r in signal_response if isinstance(r, Iterable)])
 
 
 def answerfile_name(instance, filename: str) -> str:
--- a/src/pretix/base/models/orders.py
+++ b/src/pretix/base/models/orders.py
@@ -44,7 +44,7 @@ from datetime import datetime, time, timedelta
 from decimal import Decimal
 from functools import reduce
 from time import sleep
-from typing import Any, Dict, List, Union, Iterable
+from typing import Any, Dict, Iterable, List, Union
 from zoneinfo import ZoneInfo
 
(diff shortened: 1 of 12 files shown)


src/pretix/presale/views/order.py
```diff
@@ -175,23 +173,44 @@ class TicketPageMixin:
...
+            )
+        ctx['cart'] = self.get_cart(
+            answers=True, downloads=ctx['can_download'],
+            queryset=qs,
+            order=self.order
+        )
+
+        ctx['tickets_with_download'] = [p for p in ctx['cart']['positions'] if p in positions_with_tickets]
+        for allowed_op in ctx['tickets_with_download']:
+            allowed_op.ticket_download_allowed = True
```
wiffbi, line 212: I am not sure about code style that this sets an otherwise undefined and undocumented property on the OrderPosition. Should we convert this to a propery on the OrderPosition that does the lookup through self.order.positions_with_tickets? I don’t have a opinion on this though.

src/pretix/base/models/orders.py
```diff
@@ -1137,12 +1137,19 @@ def resend_link(self, user=None, auth=None):
...
+    def positions_with_tickets_ignoring_plugins(self):
+        return (op for op in self.positions.select_related('item') if op.generate_ticket)
+
     @property
     def positions_with_tickets(self):
-        for op in self.positions.select_related('item'):
-            if not op.generate_ticket:
-                continue
-            yield op
+        signal_response = allow_ticket_download.send(self.event, order=self)
```
wiffbi, line 1146: I know this is nit-picking, but semantically IMHO positions_with_tickets and positions, whose tickets can be downloaded are different. So I am not sure we should reuse the property-name, but instead add another property e.g. `positions_with_downloadable_tickets`?

wiffbi (commented):
Looks good, although I have not tested extensively. I have some comments regarding naming and maybe performance/yield?

src/pretix/base/models/orders.py
```diff
@@ -1137,12 +1137,19 @@ def resend_link(self, user=None, auth=None):
                 attach_tickets=True,
             )
 
+    @property
+    def positions_with_tickets_ignoring_plugins(self):
+        return (op for op in self.positions.select_related('item') if op.generate_ticket)
```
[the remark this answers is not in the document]
raphaelm: We don't really care too much about orders with thousands of order positions (in fact, #3806 makes sure you can't create them), so this should not be too much of an issue. Moreover, this here is a generator expression, not a list comprehension, so it should even be equivalent to the yield statement with regards to lazy evaluation.

src/pretix/base/models/orders.py
luelista: In theory, yes. I had it named differently in the meantime (something along the lines of "positions where download is allowed by plugins"). However, as this property is [documented](https://docs.pretix.eu/en/latest/development/api/ticketoutput.html#pretix.base.ticketoutput.pretix.base.ticketoutput.BaseTicketOutput.BaseTicketOutput.generate_order) as the property to use when generating tickets for download, I think we should keep the name.

src/pretix/presale/views/order.py
luelista: I found an even better way, I can just replace `{% if line.ticket_download_allowed %}` with `{% if line in tickets_with_download %}` in the template and get rid of the property completely :)

src/pretix/presale/views/order.py
```diff
@@ -175,23 +173,42 @@ class TicketPageMixin:
...
+                    ).order_by().values('position').annotate(c=Count('*')).values('c')
+                )
+            )
+        ctx['cart'] = self.get_cart(
+            answers=True, downloads=ctx['can_download'],
+            queryset=qs,
+            order=self.order
+        )
+
+        ctx['tickets_with_download'] = [p for p in ctx['cart']['positions'] if p in positions_with_tickets]
```
wiffbi, line 210: Couldn’t this just be `positions_with_tickets`? Or can `positions_with_tickets` contain order-positions, that are not in cart? This loop basically just brings the `positions_with_tickets` in the correct order needed for the cart, but currently `tickets_with_download` is AFAICT only used in if-checks and the ordering there is not necessary. So:

```suggestion
        ctx['tickets_with_download'] = positions_with_tickets
```

wiffbi (approved):
Just a tiny comment regarding `tickets_with_download` needing the cart-order – maybe you can double check. Other than that looks good to go!

src/pretix/presale/views/order.py
luelista: `positions_with_tickets` contains all positions (with tickets) from the order, on the regular `OrderDetails` page this should be a subset of `ctx['cart']['positions']`, but on the `OrderPositionDetails` page, this contains only the current position (and its addons, if any). For the `fragment_downloads.html` fragment to display the buttons properly it's necessary that only the relevant tickets are stored in `tickets_with_download`.

wiffbi: Thanks for clarifying!

PR #3839, opened 2024-01-26
Quota form: Change item selection field per context

ser8phin:
Change the quota forms item checkbox selection to a multiselect widget if it is displayed within a date (subevent) form so it loads better in some browers

### DIFF
--- a/src/pretix/control/forms/item.py
+++ b/src/pretix/control/forms/item.py
@@ -67,7 +67,7 @@ from pretix.control.forms import (
     ButtonGroupRadioSelect, ItemMultipleChoiceField, SizeValidationMixin,
     SplitDateTimeField, SplitDateTimePickerWidget,
 )
-from pretix.control.forms.widgets import Select2
+from pretix.control.forms.widgets import Select2, Select2Multiple
 from pretix.helpers.models import modelcopy
 from pretix.helpers.money import change_decimal_field
 
@@ -207,6 +207,11 @@ class QuestionOptionForm(I18nModelForm):
 
 
 class QuotaForm(I18nModelForm):
+    itemvars = forms.ChoiceField(
+        label=_("Products"),
+        required=True
+    )
+
     def __init__(self, **kwargs):
         self.instance = kwargs.get('instance', None)
         self.event = kwargs.get('event')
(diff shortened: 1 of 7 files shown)


src/pretix/control/forms/widgets.py
```diff
@@ -77,3 +77,18 @@ def options(self, name, value, attrs=None):
 
 class Select2ItemVarQuota(Select2ItemVarQuotaMixin, forms.Select):
     pass
+
+
+class Select2ItemVarQuotaMulti(Select2Mixin, forms.SelectMultiple):
+    def options(self, name, value, attrs=None):
+        for i, v in enumerate(value):
```
raphaelm, line 84: After reading this function, I have still no idea what it does and why we need it. Could you add a short comment explaining its necessity?

luelista:
If I select one of the "Any variations" items from the select2:
grafik

I get a KeyError in this dictionary access: https://github.com/pretix/pretix/pull/3839/files#diff-9f8c7504cf91d3d6a95a3ac0f2a640f1001dcad5f57c3f4e6b55b710546bfe20R88

The dictionary seems to contain only the variation-less products (v = "1234"), and the variations (v = "1235-1"), but not the products with variations (v = "1235"), which causes the KeyError.

src/pretix/control/forms/widgets.py
ser8phin: I added an inline comment as well now, but to explain: we need to have this to get multi-selection while having no queryset to use as we combine items and variations into the choice list (if we had, we could use the Select2 without change). We tried adjusting the Select2ItemVarQuotaMixin but it broke so much, so this is the cleanest way I myself can see.

src/pretix/control/forms/item.py
```diff
@@ -231,12 +237,26 @@ def __init__(self, **kwargs):
...
+                },
+                choices=choices,
+            )
+        else:
+            self.fields['itemvars'] = forms.MultipleChoiceField(
+                label=_('Products'),
+                required=True,
+                choices=choices,
+                widget=forms.CheckboxSelectMultiple
+            )
```
raphaelm, line 259: This feels a little redundant now. Would

```suggestion
            self.fields['itemvars'].choices = choices
            self.fields['itemvars'].widget = forms.CheckboxSelectMultiple
```
Work the same? The `choices` part could even move out of the `if/else`.

src/pretix/control/forms/item.py
```diff
@@ -207,14 +207,20 @@ class Meta:
...
+    itemvars = forms.MultipleChoiceField(
+        label=_("Products"),
+        required=True,
+    )
+
     def __init__(self, **kwargs):
         self.instance = kwargs.get('instance', None)
         self.event = kwargs.get('event')
         items = kwargs.pop('items', None) or self.event.items.prefetch_related('variations')
+        multiselect = kwargs.pop('multiselect', None)
```
raphaelm, line 219: I don't feel "multiselect" really explains what this does. We always allow multiple selections, don't we? But I have trouble coming up with a better name. Maybe `searchable_selection` since that's the primary positive thing select2 gives us?

src/pretix/control/forms/widgets.py
```diff
@@ -77,3 +77,19 @@ def options(self, name, value, attrs=None):
 
 class Select2ItemVarQuota(Select2ItemVarQuotaMixin, forms.Select):
     pass
+
+
+class Select2ItemVarQuotaMulti(Select2Mixin, forms.SelectMultiple):
```
raphaelm, line 82:
```suggestion
class Select2ItemVarMulti(Select2Mixin, forms.SelectMultiple):
```
We have a item/var selection, not an item/var/quota selection, so that seems more appropriate?

raphaelm (commented):
I have no functional remarks, but a few minor code style remarks

src/pretix/control/forms/item.py
ser8phin: Naming things is always difficult, isn't it? I also cannot think of anything better, so `searchable-selection` it is

PR #3846, opened 2024-01-30
Add simplified invoice renderer

raphaelm:
Adds an invoice renderer variant that

- has more logo space
- does not have the sender in the "address window"
- has no event name included
- has the event date as part of the body of the invoice

This also renames the modern invoice renderer because the name no longer makes sense


luelista (approved):
grafik

PR #3847, opened 2024-01-30
Fix #65 -- Disallow None value for product default prices

### LINKED ISSUE #65: A item default price of None does not make sense

raphaelm:


pc-coholic:
Played around with it and didn't see any issues.

pc-coholic (approved):

PR #3848, opened 2024-01-30
Cart: More useful error message if some selected products are sold

raphaelm:


src/pretix/base/services/cart.py
```diff
@@ -113,6 +113,15 @@ def __init__(self, *args):
         'Some of the products you selected are no longer available in '
         'the quantity you selected. Please see below for details.'
     ),
+    'unavailable_listed': gettext_lazy(
+        'Some of the products you selected are no longer available.'
```
luelista, line 117:
```suggestion
        'Some of the products you selected are no longer available. '
```

src/pretix/base/services/cart.py
```diff
@@ -1132,9 +1143,15 @@ def _perform_operations(self):
                     voucher_available_count = min(voucher_available_count, vouchers_ok[op.voucher])
 
                 if quota_available_count < 1:
-                    err = err or error_messages['unavailable']
+                    err = err or error_messages['unavailable_listed']
```
luelista, line 1146: If I understand correctly, if we have both errors in the same operation (unavailable_listed or in_part_listed), we display whatever error message occurs first, but with all affected products? This might be confusing... OTOH I guess it's a pretty narrow edge case so it's probably okay to ignore.

luelista (approved):

src/pretix/base/services/cart.py
raphaelm: Well... yes. A list of error messages would probably be cleaner.

PR #3858, opened 2024-02-02
Let plugins prevent the download of individual tickets in an order

luelista:
second attempt at https://github.com/pretix/pretix/pull/3836

⚠️ merge at the same time as the matching change to pretix-shipping

### DIFF
--- a/src/pretix/base/models/orders.py
+++ b/src/pretix/base/models/orders.py
@@ -44,7 +44,7 @@ from datetime import datetime, time, timedelta
 from decimal import Decimal
 from functools import reduce
 from time import sleep
-from typing import Any, Dict, List, Union
+from typing import Any, Dict, Iterable, List, Union
 from zoneinfo import ZoneInfo
 
 import dateutil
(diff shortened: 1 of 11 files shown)


luelista:
Original PR had a bug where if no plugin intervention happened (either no plugins respond to `allow_ticket_download` or all respond with `True`, meaning allow all downloads), only the first ticket in an order (instead of all tickets) could be downloaded from the OrderDetails page.

(Left with bug, right fixed)
grafik

Reason was that (case A) if no plugin intervenes, `positions_with_tickets` just passes through the iterator from `positions_with_tickets_ignoring_plugins`, whereas otherwise (case B), it returns the set intersection of `positions_with_tickets_ignoring_plugins` and the responses of all plugins, so that only tickets allowlisted by all plugins can be downloaded, or an empty list if any plugin blocks all downloads. 
Response of case A was only iterable one time, whereas case B was iterable multiple times. For each ticket on the OrderDetails page, the response is iterated over by a `if line in tickets_with_download` check, which in case A only succeeds the first time, so for the first ticket.

raphaelm (approved):

PR #3860, opened 2024-02-05
Stripe: Add link to Stripe App

raphaelm:


pc-coholic:
Also, since we're now officially providing an app: Perhaps we should mention also somewhere close to the link/button that webhooks are not auto-installed and still need to be handled manually.

raphaelm:
> LGTM except the app-icon that we have on file with Stripe

As discussed, not reproducible

> Also, installing the app yielded an error for me, but it app was actually installed and the keys visibel from within the "installed apps"-section. Not sure if we should add an additional link to [dashboard.stripe.com/settings/apps](https://dashboard.stripe.com/settings/apps) or if this will make things more confusing...

Reproducible, but we'll just report it to Stripe as a bug…

> Also, since we're now officially providing an app: Perhaps we should mention also somewhere close to the link/button that webhooks are not auto-installed and still need to be handled manually.

Good point…  But something to add in the Stripe plugin settings page, I suppose?

raphaelm:
See https://github.com/pretix/stripe-app-pretix/pull/1 for remaining feedback

PR #3861, opened 2024-02-06
Docs/Widget: Add Payment Method Domain Registration Instructions (Z#23143782)

pc-coholic:


doc/user/events/widget.rst
```diff
@@ -429,4 +429,26 @@ Hosted or pretix Enterprise are active, you can pass the following fields:
             });
         </script>  
 
+
+Offering wallet payments (Apple Pay, Google Pay) within the widget
+------------------------------------------------------------------
+
+Some payment providers (such as Stripe) also offer Apple or Google Pay. But in order to use them, the domain of the
+payment needs to be approved first. Unless otherwise noted within the specific payment provider plugin, pretix will
+take care of the domain verification process for you automatically, when taking your shop live. However, pretix can
```
raphaelm, line 438: I don't like the combination of phrasing this generic with "such as Stripe" and "pretix will take care automatically" when in fact Stripe is the ONLY provider where we can take care automatically.

pc-coholic: I rephrased it to make it clear, that we can only do Stripe right now.

PR #3869, opened 2024-02-08
CheckinlistPDF: Export attendee company based on op; fallback to IA (Z#23144201)

pc-coholic:


src/pretix/plugins/checkinlists/exporters.py
```diff
@@ -382,8 +382,9 @@ def get_story(self, doc, form_data):
                 iac = ""
 
             name = op.attendee_name or (op.addon_to.attendee_name if op.addon_to else '') or ian
-            if iac:
-                name += "<br/>" + iac
+            company = op.company or (op.addon_to.company if op.addon_to else '') or iac
+            if company:
+                name += "<br/>" + company
```
wiffbi, line 387: It was like this before, but maybe it is worth a look: this generates an empty line if no name is given. It looks a bit strange in the PDF. Not sure how often that happens, that e.g. a company orders tickets, but does not provide individual names.
```suggestion
                if name:
                    name += "<br>"
                name += company
```

wiffbi (approved):
Looks and works like a charm. Just a little note regarding empty lines when no name, but a company is given.

PR #3873, opened 2024-02-11
Widget: fix availability of variations whose base item is unavailable

luelista:
in the following example, the product "T-Shirt":
- has an unavailable_from date which is not reached
- and unavailable_from_mode = "info":

current behaviour of widget: (t-shirt variations seem to be available, but adding to cart fails)
grafik

fixed behaviour of widget: (all t-shirt variations are set to "Not yet available")
grafik

behaviour of regular presale is unchanged:
grafik

### DIFF
--- a/src/pretix/static/pretixpresale/js/widget/widget.js
+++ b/src/pretix/static/pretixpresale/js/widget/widget.js
@@ -256,7 +256,7 @@ Vue.component('availbox', {
             }
         },
         unavailability_reason_message: function () {
-            var reason = this.item.has_variations ? this.variation.current_unavailability_reason : this.item.current_unavailability_reason;
+            var reason = this.item.current_unavailability_reason || this.variation?.current_unavailability_reason;
             if (reason) {
                 return strings["unavailable_" + reason] || reason;
             }
--- a/src/pretix/static/pretixpresale/js/widget/widget.js
+++ b/src/pretix/static/pretixpresale/js/widget/widget.js
@@ -211,7 +211,7 @@ Vue.component('availbox', {
         + strings.sold_out
         + '</div>'
         + '<div class="pretix-widget-waiting-list-link"'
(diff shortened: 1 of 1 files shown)


src/pretix/static/pretixpresale/js/widget/widget.js
```diff
@@ -256,7 +256,7 @@ Vue.component('availbox', {
             }
         },
         unavailability_reason_message: function () {
-            var reason = this.item.has_variations ? this.variation.current_unavailability_reason : this.item.current_unavailability_reason;
+            var reason = this.item.current_unavailability_reason || this.variation?.current_unavailability_reason;
```
wiffbi, line 259: I am not sure about our browser support for the widget, but I think we try to keep it as backwards compatible as possible. So IMHO we should avoid `?.` notation here. I am not sure, why that change is needed here – or can unavailability_reason be set on the containing item even if it has variations? I assumed `unavailability_reason` is always set on each variation, if an item has variations. I might be wrong though.

wiffbi: Sorry, I have not read you description correctly, that the item is not available, but variations are. So your code is right, but I still think we should avoid `?.` for backwards-compatibility.

```suggestion
            var reason = this.item.current_unavailability_reason || (this.variation ? this.variation.current_unavailability_reason : '');
```

luelista: Yes, this is a recent feature ([2020](https://caniuse.com/?search=%3F.%20)), but, we already use this anyway (I had checked it with ctrl-F for `?.` on this file, which has has three other matches), so no new dependency introduced.

wiffbi: Oh, you are right. Sorry, if nobody complained before, then we can leave it like that.

PR #3893, opened 2024-02-13
Event-specific fonts and Web-Embedded Fonts (Z#23130701)

pc-coholic:
This PR is - to a certain extend - a revival of #1666.

- Add `register_event_fonts` signal
This will allow event-organizers to provide fonts that are restricted to their events only. Of course/unfortunately, this will also mean, that these fonts won't be available on the organizer level. If we could find a way around this - especially since customers do have the option to inherit their shop-design from the organizer, I'd be very open to that.

- Add the capability to embed webfonts
Probably wrong nomenclature here, but for this PR this is meant as "fonts stored locally on the pretix server" vs. "fonts embedded from a 3rd party website". This is about the latter ones.
Syntax for `register_fonts` and `register_event_fonts` is the same: Instead of providing a local/static path, a `https`-URL is provided.

Implementation notes:
- `get_fonts()` has been extended with two parameters, `event` and `pdf_only`. Instead of having multiple locations within the pretix-source call the function and then manually filter out the fonts that are marked as `pdf_only`, we do that filtering now in a central location. This has also the added benefit of being able to filter out any "webfonts", since they have no business in the PDF-generation.
- `get_font_stylesheet()` has been simplified a little bit, since with special treatment for local- and webfonts, a lot of code got duplicated...
- When a webfont is present, the middleware is also automatically including it in the CSP. We could probably limit this to fonts, that are selected as `primary_font`, if we want to keep the CSP smaller.
- I am not super happy about adding `self.fields['primary_font'].choices` back to `forms/event.py` - but I don't think we have another option since the `DEFAULT`-dict in `base/settings.py` has no concept whatsoever of an possible active event.
- No attempt has been made to include the event-fonts for the invoice renderers. It would probably also only take the same statement as for the issue above; but has exactly the same issues in every other regard, too.


src/pretix/base/pdf.py
```diff
@@ -939,6 +939,12 @@ def _draw_textarea(self, canvas: Canvas, op: OrderPosition, order: Order, o: dic
         if o['italic']:
             font += ' I'
 
+        # Since pdfmetrics.registerFont is global, we want to make sure that no one tries to sneak in a font, they
+        # should not have access to.
+        if font not in list(get_fonts(order.event, pdf_only=True).keys()) + ['Open Sans']:
```
raphaelm, line 944: Running the signal for every text box we render does not sound terribly efficient

src/pretix/presale/style.py
```diff
@@ -225,17 +226,67 @@ def regenerate_organizer_css(organizer_id: int, regenerate_events=True):
...
-        f.update(value)
+        received_fonts.update(value)
+
+    if event:
+        for recv, value in register_event_fonts.send(event):
+            received_fonts.update(value)
+
+    for font, payload in received_fonts.items():
+        if pdf_only:
+            if any(v.startswith('https') for v in list(nested_dict_values(payload))):
```
raphaelm, line 276: see above, I'd either go with ``"//" in`` or ``.startswith('https://')``. Your approach is just calling for a font named "https font" ;)

raphaelm (commented):
> This will allow event-organizers to provide fonts that are restricted to their events only. Of course/unfortunately, this will also mean, that these fonts won't be available on the organizer level. If we could find a way around this - especially since customers do have the option to inherit their shop-design from the organizer, I'd be very open to that.

I don't see a way that is simpler than rewriting how our plugins work…

>    I am not super happy about adding self.fields['primary_font'].choices back to forms/event.py - but I don't think we have another option since the DEFAULT-dict in base/settings.py has no concept whatsoever of an possible active event.

That seems fine to me.

src/pretix/presale/style.py
```diff
@@ -225,17 +226,67 @@ def regenerate_organizer_css(organizer_id: int, regenerate_events=True):
...
+        "bolditalic": {
+            ...
+        },
+        "pdf_only": False,   # if True, font is not usable on the web,
+    }
+}
+"""
+
+
+def get_fonts(event: Event = None, pdf_only=False):
```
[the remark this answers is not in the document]
pc-coholic: Fair point. Will change the signature.

src/pretix/base/pdf.py
pc-coholic: Adding a 30 second cache; open to discuss a longer runtime.

raphaelm: I was thinking about a refactor that does this once per render, not once per text field on that render. With the cache, we'll still need to hit the cache via network on every text field?

pc-coholic: Moved to __init__

src/pretix/control/templates/pretixcontrol/pdf/webfonts.css
```diff
@@ -14,7 +15,7 @@
 
 {% for family, styles in fonts.items %}
 {% for style, formats in styles.items %}
-{% if "sample" not in style %}
+{% if "sample" not in style and "pdf_only" not in style %}
```
raphaelm, line 18: Well, the plugin API does allow to have `style["pdf_only"] = False` in which case this condition would be wrong, wouldn't it? :thinking:

pc-coholic: I did check it for a font that doesn't have `pdf_only`, that has it set to `True` and one that has it set to `False` - in either case it worked for me...

But perhaps we're looking at two different things here?

pc-coholic: As per direct discussion: Works as intended.

pc-coholic:
Ready to merge upon my return

PR #3930, opened 2024-02-19
added working  install command for pretix cliques

manuliner:


pc-coholic:
This mostly looks like purely stylistic changes to the indentation and adding the cliques plugin.

But since this is the main pretix repo, this doesn't make really sense - we're not going to ship out the cliques plugin by default for all pretix docker users, sorry.

If you need plugins installed, please build your own Dockerfile as explained here: https://docs.pretix.eu/en/latest/admin/installation/docker_smallscale.html

manuliner:
wasn't intended. wrong button pressed. sorry

PR #3941, opened 2024-02-29
Allow multiple returnurl prefixes (Z#23145768)

luelista:


src/pretix/plugins/returnurl/views.py
```diff
@@ -31,10 +33,12 @@
 
 
 class ReturnSettingsForm(SettingsForm):
-    returnurl_prefix = forms.URLField(
-        label=_("Base redirection URL"),
-        help_text=_("Redirection will only be allowed to URLs that start with this prefix."),
+    returnurl_prefix = forms.RegexField(
+        label=_("Base redirection URLs"),
+        help_text=_("Redirection will only be allowed to URLs that start with one of these prefixes. Enter one or more allowed URL prefix per line. "),
```
raphaelm, line 38: Can we add that it must have a ``/`` after the hostname in here? It might be very confusing why the validation fails otherwise.

luelista: Done

PR #3948, opened 2024-03-01
Add migration to force slash after hostname in returnurl prefixes

luelista:


src/pretix/plugins/returnurl/migrations/0002_auto_20240301_1355.py
```diff
@@ -0,0 +1,38 @@
...
+        new_value = "\n".join(new_prefixes)
+        setting.value = new_value
+        cache.delete('hierarkey_{}_{}'.format('event', setting.object_id))
+        setting.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('returnurl', '0001_initial'),
```
raphaelm, line 33: This needs to depend on a recent migration in pretixbase, otherwise on a fresh installation django might try to run this migration before EventSettingsStore is created, which will fail loudly

PR #3956, opened 2024-03-04
Prevent transferring files from priv/ to pub/ on event clone

luelista:


raphaelm (commented):
Hmm, on pretix.eu they also take names like `2/priv/cachedfiles/` i.e. where the CDN node key is before the prefix :thinking:

luelista:
This should work now, I don't really like that regex solution but don't have any better idea that's readily implementable. Ideally, we could somehow detect which setting is supposed to contain a `File`, which is possible for those which are configured via `add_default`, but some plugins don't seem to do that (e.g. `ticketoutput_pdf_background`)?

raphaelm:
Hmm. Yeah, the regex is probably still insufficient since I believe we have legacy/priv/ for files older than the CDN? :see_no_evil: 

> Ideally, we could somehow detect which setting is supposed to contain a File

That sounds a lot safer…

>  which is possible for those which are configured via add_default, but some plugins don't seem to do that (e.g. ticketoutput_pdf_background)?

Yes, but that's something we could fix, we don't seem to have "dynamic" keys. pretix Hosted production only knows these keys:

```
In [3]: for k in Event_SettingsStore.objects.filter(value__startswith="file://").values("key").order_by("key").distinct():
   ...:     print(k)
   ...: 
{'key': 'exhibitors_logo_image'}
{'key': 'invoice_logo_image'}
{'key': 'logo_image'}
{'key': 'mail_attachment_new_order'}
{'key': 'og_image'}
{'key': 'payment_epaybl_client_cert'}
{'key': 'ticketoutput_passbook_background'}
{'key': 'ticketoutput_passbook_background2x'}
{'key': 'ticketoutput_passbook_background3x'}
{'key': 'ticketoutput_passbook_icon'}
{'key': 'ticketoutput_passbook_icon2x'}
{'key': 'ticketoutput_passbook_icon3x'}
{'key': 'ticketoutput_passbook_logo'}
{'key': 'ticketoutput_passbook_logo2x'}
{'key': 'ticketoutput_passbook_logo3x'}
```

so that's pretty realistic to do.

raphaelm:
ticketoutput_pdf_background seems like a dead link to the past of pretix where there was only one layout per event, and apparently there is not a single value left

luelista:
checking whether all keys are properly declared as file using `add_default`...
- [ ] `{'key': 'exhibitors_logo_image'}`
- [ ] `{'key': 'invoice_logo_image'}`
- [ ] `{'key': 'logo_image'}`
- [ ] `{'key': 'mail_attachment_new_order'}`
- [ ] `{'key': 'og_image'}`
- [ ] `{'key': 'payment_epaybl_client_cert'}`
- [ ] `{'key': 'ticketoutput_passbook_background'}`
- [ ] `{'key': 'ticketoutput_passbook_background2x'}`
- [ ] `{'key': 'ticketoutput_passbook_background3x'}`
- [ ] `{'key': 'ticketoutput_passbook_icon'}`
- [ ] `{'key': 'ticketoutput_passbook_icon2x'}`
- [ ] `{'key': 'ticketoutput_passbook_icon3x'}`
- [ ] `{'key': 'ticketoutput_passbook_logo'}`
- [ ] `{'key': 'ticketoutput_passbook_logo2x'}`
- [ ] `{'key': 'ticketoutput_passbook_logo3x'}`

luelista:
Ready once the dependencies are merged:

- https://github.com/raphaelm/django-hierarkey/pull/18
- https://github.com/pretix/pretix-passbook/pull/64
- https://code.rami.io/pretix/pretix-epaybl/-/merge_requests/4
- https://code.rami.io/pretix/pretix-exhibitors/-/merge_requests/13

raphaelm:
I have released https://pypi.org/project/django-hierarkey/1.2.0/

raphaelm (approved):

PR #3961, opened 2024-03-07
Time machine mode [Z#23129725]

luelista:
### screenshots

image

image

### assumptions

- time machine is only allowed on shops in test mode
- we only use the fake datetime for availability checks, and for ticket validity periods, not for order and payment dates (order date, expiry date, payment date, etc)

### Implementation notes

In a [previous attempt](https://github.com/pretix/pretix/pull/3960) to implement this, we tried to pass the `time_machine_now` (the faked date time) through to every check, which was really messy. Therefore we now implemented a middleware which puts the datetime into thread local storage, from where it can be accessed via the `time_machine_now()` method. However we still need to manually transfer it when calling into async tasks (cart manipulation and order confirmation).


src/pretix/base/models/orders.py
```diff
@@ -540,7 +541,7 @@ def touch(self):
         self.save(update_fields=['last_modified'])
 
     def set_expires(self, now_dt=None, subevents=None):
-        now_dt = now_dt or now()
+        now_dt = now_dt or time_machine_now()
```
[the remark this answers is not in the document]
luelista: I had a situation where including it was useful for testing stuff. But I think in general it's probably better to use real-now here.

src/pretix/presale/templates/pretixpresale/event/base.html
```diff
@@ -33,6 +33,21 @@
...
+            <div class="container">
+                <span class="fa fa-clock-o" aria-hidden="true"></span>
+                {% trans "You are currently using the time machine. The ticket shop is rendered as if it were" %} {{ request.now_dt }}
+                <a href="{{ site_url }}{% url "control:event.live" event=event.slug organizer=event.organizer.slug %}" class="btn-link">
+                    {% trans "Change" %}
+                </a>
+                <form action="{{ site_url }}{% url "control:event.live" event=event.slug organizer=event.organizer.slug %}" method="post" class="download-btn-form">
+                    {% csrf_token %}
+                    <button type="submit" class="btn-link" value="true" name="timemachine_disable">{% trans "Go back to current time" %}</button>
+                </form>
```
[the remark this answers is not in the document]
luelista: This should work now that the form is in the frontend, too

src/pretix/base/timemachine.py
```diff
@@ -0,0 +1,80 @@
...
+
+class TimeMachineMiddleware:
+    tls = threading.local()
+
+    def __init__(self, get_response=None):
+        self.get_response = get_response
+        super().__init__()
+
+    def __call__(self, request):
+        if hasattr(request, 'event') and hasattr(request, '_namespace') and request._namespace == 'presale' and \
```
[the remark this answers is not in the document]
luelista: Okay, I moved it into a contextmanager which I call from `event_view_wrapper` as well as `EventMiddleware`

src/pretix/control/templates/pretixcontrol/event/live.html
```diff
@@ -124,6 +124,35 @@ <h1>{% trans "Shop status" %}</h1>
...
+            {% trans "Time machine" %}
+        </div>
+        <div class="panel-body">
+            <form action="" method="post" class="">
+                {% csrf_token %}
+                {% bootstrap_form_errors timemachine_form %}
+
+                <p>{% trans "Test your shop as if it were a different date and time." %}</p>
+
+                {% bootstrap_field timemachine_form.now_dt layout="control" show_label=False field_class="col-md-6" %}
```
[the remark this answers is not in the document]
luelista: Seems to work with layout="inline"? The paddings were due to the col-md-6 without a .row around it, fixed that as well

doc/development/setup.rst
```diff
@@ -211,5 +211,14 @@ with the documentation a lot, you might find it useful to use sphinx-autobuild::
 Then, go to http://localhost:8081 for a version of the documentation that automatically re-builds
 whenever you change a source file.
 
+Working with frontend assets
+----------------------------
+
+To update the frontent styles, run the following commands inside your virtual environment.::
```
raphaelm, line 217:
```suggestion
To update the frontend styles of shops with a custom styling, run the following commands inside
your virtual environment.::
```

src/pretix/static/pretixpresale/scss/_event.scss
```diff
@@ -187,18 +187,19 @@ article.item-with-variations .product-row:last-child:after {
 div.front-page {
     margin-top: 30px;
 }
-.offline-banner {
-    background: $brand-danger;
-    color: white;
+.offline-banner, .timemachine-banner {
```
raphaelm, line 190: timemachine-banner no longer exists, does it?

src/pretix/control/views/event.py
```diff
@@ -954,6 +954,9 @@ class EventLive(EventPermissionRequiredMixin, TemplateView):
     permission = 'can_change_event_settings'
     template_name = 'pretixcontrol/event/live.html'
 
+    def setup(self, request, *args, **kwargs):
+        super().setup(request, *args, **kwargs)
+
```
raphaelm, line 959: Is this redundant now?

wiffbi (changes requested):
Looks good! Just some minor adjustments around CSS maybe.

I know payments are mostly plugins, but would some minimal support be viable? E.g. [GiftCardPayment::checkout_prepare](https://github.com/pretix/pretix/blob/master/src/pretix/base/payment.py#L1444) (not sure we need the `payment_prepare` and `execute_payment` adapted though) and `GiftCard::expired`?

Also maybe `Membership::is_valid` might be a nice addition?

As well as Discounts with [apply_discounts](https://github.com/pretix/pretix/blob/master/src/pretix/base/services/pricing.py#L170-L171)?

I think waitinglist-auto-disable was merged after this PR was branched off, so probably do a rebase and check [autodisable](https://github.com/pretix/pretix/blob/master/src/pretix/base/models/event.py#L237), so waitinglists respect autodisable on timemachine.

src/pretix/presale/templates/pretixpresale/event/timemachine.html
```diff
@@ -0,0 +1,44 @@
...
+                {% bootstrap_form_errors timemachine_form %}
+
+                <p>{% trans "Test your shop as if it were a different date and time." %}</p>
+
+                <div class="row">
+                    <div class="col-md-6">
+                        {% bootstrap_field timemachine_form.now_dt layout="inline" show_label=False  %}
+                    </div>
+                    <div class="col-md-6 text-right">
+                        <button type="submit" class="btn btn-default btn-lg btn-save">
```
wiffbi, line 29:
```suggestion
                        <button type="submit" class="btn btn-primary btn-lg btn-save">
```

src/pretix/presale/views/event.py
```diff
@@ -913,8 +920,49 @@ def post(self, request, *args, **kwargs):
...
-            if 'event_access' not in parentdata:
+            if 'child_session_{}'.format(request.event.pk) not in parentdata:
                 raise PermissionDenied(_('Please go back and try again.'))
 
         request.session['pretix_event_access_{}'.format(request.event.pk)] = parent
         return redirect_to_url(eventreverse(request.event, 'presale:event.index'))
+
+
+class TimemachineForm(forms.Form):
+    now_dt = forms.SplitDateTimeField(
```
wiffbi, line 931: Can we make that timestamp not required and if left empty, just deactivate time machine? I tried to deactivate time machine by emptying the inputs, then realized there is a deactivate-button and it did not submit. Alternatively instead of allowing empty values (which I think I prefer) would be to have the deactivate-button in its own `<form>`.

src/pretix/static/pretixpresale/scss/main.scss
```diff
@@ -39,6 +39,7 @@ $body-bg: #f5f5f5 !default;
 h1 a, .btn {
     text-decoration: none;
 }
+a.stealth, a.stealth:hover { color: inherit; text-decoration: none; }
```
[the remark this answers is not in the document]
raphaelm: Is it only because of the space or also because of the icon itself otherwise being underlined?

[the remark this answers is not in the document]
wiffbi: In my tests the icon was never underlined. But might be browser-dependent (Firefox and Chrome) …

src/pretix/presale/templates/pretixpresale/event/base.html
```diff
@@ -111,9 +111,37 @@ <h1>
...
+                    <p>
+                        {% eventurl event "presale:event.timemachine" as time_machine_link %}
+                        {% blocktrans trimmed with time_machine_link=time_machine_link %}
+                            To view your shop at different points in time, you can enable
+                            <a href="{{ time_machine_link }}" class="stealth"><span class="fa fa-clock-o" aria-hidden="true"></span> <u>time machine</u></a>.
+                        {% endblocktrans %}
+                    </p>
+                {% elif request.event_domain or request.organizer_domain %}
+                    <p>
+                        {% url "control:event.transfer_session" event=event.slug organizer=event.organizer.slug as time_machine_link %}
```
raphaelm, line 138: This does not seem to work, ``{% url %}`` usually generates relative URLs, so this will link to /control, but we would need to link to the main domain (from settings.SITE_URL)

raphaelm (changes requested):

src/pretix/static/pretixpresale/scss/_event.scss
luelista:
```suggestion
.offline-banner {
```

src/pretix/presale/views/event.py
luelista: > allowing empty values

This causes some weird behavior if only one of the fields is empty, where I think an error message should be displayed instead of just disabling (and displaying a "success" message). I'll go with the two forms approach.

src/pretix/presale/views/event.py
```diff
@@ -913,8 +920,49 @@ def post(self, request, *args, **kwargs):
...
+            )
+        )
+
+    def get_context_data(self, **kwargs):
+        ctx = super().get_context_data(**kwargs)
+        ctx['timemachine_form'] = self.timemachine_form
+        return ctx
+
+    def post(self, request, *args, **kwargs):
+        if request.POST.get("timemachine_disable"):
```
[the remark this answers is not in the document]
luelista: Should be fixed by the check added in `setup()`

src/pretix/presale/templates/pretixpresale/event/base.html
luelista: It works for me, in a somewhat ugly way though: It links to `"/control/event/.../transfer_session/"`, so the browser requests `https://customdomain/control/event/.../transfer_session/`, which gets in turn redirected to `https://maindomain/control/event/.../transfer_session/`. Is this different on your machine?

This seems to be the same behaviour as the "View in backend" button on the order detail page.

I'm not quite sure what the best way to fix this would be: Currently, settings (and therefore SITE_URL) is not available in the default context of presale. 
- I could change that and just use `time_machine_link=settings.SITE_URL|add:time_machine_link`. 
- Or I could try and fix the `url` template tag (or, more realistically, create a new derived one) which always calls build_absolute_url (or do we have something like that already?).

Recommendations?

luelista:
> Looks good! Just some minor adjustments around CSS maybe.
> 
> I know payments are mostly plugins, but would some minimal support be viable? E.g. [GiftCardPayment::checkout_prepare](https://github.com/pretix/pretix/blob/master/src/pretix/base/payment.py#L1444) (not sure we need the `payment_prepare` and `execute_payment` adapted though) and `GiftCard::expired`?
> 
> Also maybe `Membership::is_valid` might be a nice addition?
> 
> As well as Discounts with [apply_discounts](https://github.com/pretix/pretix/blob/master/src/pretix/base/services/pricing.py#L170-L171)?

added: https://github.com/pretix/pretix/pull/3961/commits/3e234c004c2709154923c4aa721369e42c6faa72

src/pretix/base/models/event.py
```diff
@@ -235,7 +235,7 @@ def waiting_list_active(self):
         if not self.settings.waiting_list_enabled:
             return False
         if self.settings.waiting_list_auto_disable:
-            return self.settings.waiting_list_auto_disable.datetime(self) > now()
+            return self.settings.waiting_list_auto_disable.datetime(self) > time_machine_now()
```
raphaelm, line 238: I just wanted do note it explicitly: waiting_list_auto_disable also influences quota availability (see base/services/quotas.py), where do *not* want the time machine, as that could lead to wrong contents in the quota cache or overbookings. So it's all good as it is, just don't want anyone to add it there as well because it might look like a good idea on first sight :)

src/pretix/presale/templates/pretixpresale/event/base.html
luelista: Added a new absmainurl template tag based on the other multidomain url template tags

src/pretix/multidomain/templatetags/eventurl.py
```diff
@@ -45,11 +45,13 @@ def render(self, context):
             for k, v in self.kwargs.items()
         }
         view_name = self.view_name.resolve(context)
-        event = self.event.resolve(context)
+        event = self.event.resolve(context) if self.event is not False else False
         url = ''
         try:
             if self.absolute:
```
raphaelm, line 51: It took me a few minutes too understand all the different cases this can be called with now… if you have time… would this be worth some tests?

raphaelm (approved):
LGTM, let's merge this after your vacation  :)

src/pretix/multidomain/templatetags/eventurl.py
luelista: Added tests and improved the docstring

PR #3962, opened 2024-03-07
Widget: pass utm-params from embedding page to presale (Z#23146288)

wiffbi:


raphaelm (approved):

raphaelm (changes requested):
Oh no, I noticed a problem right after approach. When the widget is set to open in a new tab (e.g. no SSL), the form where you select products opens in a new tab as a POST request that then gets redirected. So the utm parameters never get passed to a page where JS runs. Do we need to pull them through the redirect on the server side?

wiffbi:
Oh, yes, that is a good catch! I will look into this.

raphaelm (approved):

PR #3965, opened 2024-03-08
Presale: add data-attribute price to article

wiffbi:
To make parsing and calculating prices in presale-frontend easier (e.g. for  e-commerce tracking), this PR adds an attribute `data-price` to each `<article>` in presale.

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
+++ b/src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
@@ -98,7 +98,7 @@
                     </div>
                     <div class="variations {% if not event.settings.show_variations_expanded %}variations-collapsed{% endif %}">
                         {% for var in item.available_variations %}
-                            <article aria-labelledby="item-{{ item.pk }}-{{ var.pk }}-legend"{% if var.description %} aria-describedby="item-{{ item.pk }}-{{ var.pk }}-description"{% endif %} class="row product-row variation" id="item-{{ item.pk }}-{{ var.pk }}"{% if not var.free_price %} data-price="{{ var.price }}"{% endif %}>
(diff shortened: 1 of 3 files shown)


src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
```diff
@@ -228,7 +228,7 @@ <h5 id="item-{{ item.pk }}-{{ var.pk }}-legend">{{ var }}</h5>
                     </div>
                 </article>
             {% else %}
-                <article aria-labelledby="item-{{ item.pk }}-legend"{% if item.description %} aria-describedby="item-{{ item.pk }}-description"{% endif %} class="row product-row simple" id="item-{{ item.pk }}">
+                <article aria-labelledby="item-{{ item.pk }}-legend"{% if item.description %} aria-describedby="item-{{ item.pk }}-description"{% endif %} class="row product-row simple" id="item-{{ item.pk }}"{% if not item.free_price %} data-price="{{ item.price }}"{% endif %}>
```
luelista, line 231: `item` hat keinen `price` sondern nur einen `default_price`

```suggestion
                <article aria-labelledby="item-{{ item.pk }}-legend"{% if item.description %} aria-describedby="item-{{ item.pk }}-description"{% endif %} class="row product-row simple" id="item-{{ item.pk }}"{% if not item.free_price %} data-price="{{ item.default_price }}"{% endif %}>
```

luelista (commented):
sollte man das evtl an den anderen produktlisten auch einfügen (voucher.html, fragment_addon_choice.html)?

src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
wiffbi: Doh, did not realize this as my only test-product, that was not either free_price or a variation-product, was free … Thanks!

wiffbi:
You are right about adding it to voucher and add-ons forms too. I did not implement that in tracking yet, but that will come sooner than later, so I will add it. Thanks!

wiffbi:
I changed to the universal property `display_price`, which is available on items as well as variations. I also unlocalized these values.

PR #3988, opened 2024-03-15
Enforce uniqueness of order codes and ticket secrets

raphaelm:
pretix works with the assumptions that order codes and ticket secrets are unique within an organizer, and if this assumption is violated, things break in bad ways. So far, we validated this constraint at the respective boundaries, but not in the database, leading to a possible race condition when creating orders through the API.

This PR adds additional foreign keys, uniqueness constraints and tests for strengthen this constraint. The migration intentionally does not backfill the foreign key for the entire table, as that migration would run for a looooong time on large systems. We can think about that in a second step.


luelista (commented):
Das macht auf jeden Fall ein paar Tests kaputt, die aktuell darauf dependen, dass sie mehrere Orders mit gleichem Code anlegen können. Soweit ich es sehe, ist das aber nicht wirklich beabsichtigt/entscheidend in den Tests. Hab entsprechend die Codes einfach geändert.

luelista (approved):

PR #3990, opened 2024-03-15
Treat partially paid expired orders as overpaid orders (Z#23147757)

raphaelm:


luelista:
I think we also need to change it here:
https://github.com/pretix/pretix/blob/master/src/pretix/control/forms/filter.py#L313

PR #3994, opened 2024-03-15
Fix #3974 -- Webhooks: Add type of checkin

### LINKED ISSUE #3974: Webhooks - pretix.event.checkin, triggered on both Entry and Exit scans
### Problem and impact

The web hook pretix.event.checkin triggers on both a Pretix Scan 'Entry' scan and a Pretix Scan 'Exit' scan.

I haven't been able to find a description of intended behaviour in the docs or a previous issue entry addressing the issue. 
Given that their is a pretix.event.checkin.reverted trigger and that the received webhook data, always presents the action as 'pretix.event.checkin' with no other entry to differentiate as an Entry or Exit scan, I am assuming that this is not the intended behaviour. 

### Expected behaviour

Trigger webhook pretix.event.checkin when an Entry scan is added
Trigger webhook pretix.event.checkin.reverted when an Exit scan is added.

### Steps to reproduce

Self-hosted Pretix, version: 2024.1.0
Setup a webhook with only pretix.event.checkin ticked.
Scan a ticket as entering event, webhook should be triggered.
Scan the ticket as exiting the event, and the webhook will again be triggered.

### Screenshots

_No response_

### Link

_No response_

_No response_

_No response_

### Version

2024.1.0

raphaelm:


pc-coholic (approved):
LGTM, but I did think about adding a quick line to api/webhooks.rst to mention the new field...

pc-coholic:
That being said: What exactly is first_checkin doing here? Is that a remanent of something old? I was not able to find any reference to that anywhere else...

PR #3997, opened 2024-03-16
Allow customization of cache and log directory

mweinelt:
On systems that follow the FHS it may be desirable to separate logs and cache files into dedicated base directories (e.g. /var/log/pretix or /var/cache/pretix).


raphaelm:
@cla-bot[bot] check

raphaelm (approved):

PR #4002, opened 2024-03-19
Generalize import process from orders to more models

raphaelm:


doc/development/api/import.rst
```diff
@@ -3,11 +3,12 @@
...
 
-Extending the order import process
-==================================
+Extending the import process
+============================
 
-It's possible through the backend to import orders into pretix, for example from a legacy ticketing system. If your
-plugins defines additional data structures around orders, it might be useful to make it possible to import them as well.
+It's possible through the backend to import objects into pretix, for example orders from a legacy ticketing system. If
+your plugins defines additional data structures around those objects, it might be useful to make it possible to import
```
wiffbi, line 10:
```suggestion
your plugin defines additional data structures around those objects, it might be useful to make it possible to import
```

src/pretix/base/modelimport_vouchers.py
```diff
@@ -0,0 +1,378 @@
...
+    def static_choices(self):
+        return [
+            (str(p.pk), '{} – {}'.format(p.item, p.value)) for p in self.items
+        ]
+
+    def clean(self, value, previous_values):
+        if value:
+            matches = [
+                p for p in self.items
+                if str(p.pk) == value or any((v and v == value) for v in i18n_flat(p.value)) and p.item_id == previous_values['item'].pk
```
wiffbi, line 215: Should we make sure, that we only allow variations of the item.pk even when a variant.pk is given?
```suggestion
                if (str(p.pk) == value or any((v and v == value) for v in i18n_flat(p.value))) and p.item_id == previous_values['item'].pk
```

wiffbi (commented):
When I tested it locally, pretix-c-mb threw an error due to DataImportError not being available – as it moved modelimport. Did you check all plugins that might use this? Or should we temporarily add the services.orderimport back in?

src/pretix/control/views/modelimport.py
```diff
@@ -64,28 +68,16 @@
...
-
+class BaseImportView(TemplateView):
     def post(self, request, *args, **kwargs):
         if 'file' not in request.FILES:
-            return redirect_to_url(reverse('control:event.orders.import', kwargs={
-                'event': request.event.slug,
-                'organizer': request.organizer.slug,
-            }))
+            return redirect_to_url(request.path)
         if not request.FILES['file'].name.lower().endswith('.csv'):
```
wiffbi, line 75: Should we allow .txt files as well?

src/pretix/base/modelimport.py
```diff
@@ -0,0 +1,287 @@
...
+            charset = chardet.detect(data)['encoding']
+        except ImportError:
+            charset = file.charset
+    data = data.decode(charset or "utf-8", mode)
+    # If the file was modified on a Mac, it only contains \r as line breaks
+    if '\r' in data and '\n' not in data:
+        data = data.replace('\r', '\n')
+
+    try:
+        dialect = csv.Sniffer().sniff(data.split("\n")[0], delimiters=";,.#:")
```
wiffbi, line 65: This throws an error CSV-file with one-column – e.g. when I want to import a list of voucher-codes the csv-file could have only one column. I know I can use the „Create multiple vouchers“-feature, but maybe we can support that special case here?

raphaelm: Oof. Yes, I get it, but I don't see how we could to that while keeping autodetection of the delimiter?

src/pretix/control/views/modelimport.py
raphaelm: I think that only leads to additional confusion?

wiffbi: You are probably right!

src/pretix/base/modelimport.py
wiffbi: Wouldn’t – instead of returning None – setting `dialect = 'excel'` (the default anyways) work? AFAICT the worst that could happen, is that we have a CSV-file with an unknown delimiter parsed as one value per line. IMHO even slightly better UI-wise than the current, generic error message: the user might realize it’s due to the delimiter when trying to assign columns and only one long column is available.

```
    try:
        dialect = csv.Sniffer().sniff(data.split("\n")[0], delimiters=";,.#:")
    except csv.Error as e:
        dialect = 'excel'
```

raphaelm: That could work, but could also be quite a footgun when you can upload almost any file then…

wiffbi: You are probably right.

wiffbi (approved):

PR #4004, opened 2024-03-20
Waiting list: Allow to set auto-disable date (Z#23141338)

raphaelm:


wiffbi (approved):
Looks good!

Would we makse sense to add a hint in [control](https://github.com/pretix/pretix/blob/master/src/pretix/control/templates/pretixcontrol/waitinglist/index.html#L9) that the waiting-list is deactivated? Other than that good to go IMHO.

raphaelm:
Done, but only for non-series events

PR #4016, opened 2024-03-27
Add database.disable_server_side_cursors option

### LINKED ISSUE #4014: Allow setting DISABLE_SERVER_SIDE_CURSORS for PostgreSQL configurations
### Problem and impact

We use Pretix with a PostgreSQL server and a PGBouncer in-between. We observe for example exports (in this case an Excel export) breaking on db cursor errors.

We are currently testing if passing [`DISABLE_SERVER_SIDE_CURSORS`](https://docs.djangoproject.com/en/3.0/ref/settings/#disable-server-side-cursors) solves the problem, it is however currently not possible to pass this to Pretix in the `pretix.cfg` file or (as far as we were able to determine) via an environment variable or equivalent.

Would adding an option for this be possible if the problem is solved with this db setting?

### Expected behaviour

All exports work even when using PGBouncer.

### Steps to reproduce

_No response_

### Screenshots

_No response_

### Link

_No response_

_No response_

_No response_

### Version

_No response_

thegcat:
Fixes #4014.

### DIFF
--- a/doc/admin/config.rst
+++ b/doc/admin/config.rst
@@ -157,6 +157,7 @@ Example::
     host=localhost
     port=3306
     advisory_lock_index=1
+    disable_server_side_cursors=0
     sslmode=require
     sslrootcert=/etc/pretix/postgresql-ca.crt
     sslcert=/etc/pretix/postgresql-client-crt.crt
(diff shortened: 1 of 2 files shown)


thegcat:
I was unsure if this should be documented, as the other more advanced DB options are not in the docs either? Happy to add docs if you’d like some.

raphaelm:
Yes, I think adding it to the docs would be good.

PR #4035, opened 2024-04-02
Allow admins to generate emergency 2FA tokens

raphaelm:
Admins can already disable 2FA for a user entirely, so we can improve things by giving them a little-bit-less dangerous option for helping an user out of an impossible situation.


src/pretix/control/views/users.py
```diff
@@ -150,6 +152,32 @@ def get_success_url(self):
...
+        self.object = get_object_or_404(User, pk=self.kwargs.get("id"))
+
+        d, __ = StaticDevice.objects.get_or_create(user=self.object, name='emergency')
+        token = d.token_set.create(token=get_random_string(length=12, allowed_chars='1234567890'))
+        self.object.log_action('pretix.user.settings.2fa.emergency', user=self.request.user)
+
+        messages.success(request, _(
+            'The emergency token for this user is "{token}". It can only be used once. The previously known emergency '
+            'token for this user remain active. Please make sure to transmit this code only over an authenticated '
+            'channel (other than email, if possible).'
```
wiffbi, line 170: Hm, „previously known“ sounds to me like it is not known anymore. But the sentence ends by stating that the token is in fact still known. A bit confusing IMHO. Also, does an emergency token always exist? So maybe this:

```suggestion
            'The emergency token for this user is "{token}". It can only be used once. Please make sure to transmit '
            'this code only over an authenticated channel (other than email, if possible). Any previous emergency '
            'tokens for this user remain active.'
```

wiffbi (approved):
Looks good, just a minor text change maybe.

PR #4056, opened 2024-04-09
Do not allow currency codes that to not represent money

raphaelm:


pc-coholic (approved):

src/pretix/_base_settings.py
```diff
@@ -235,7 +235,12 @@
     )
 }
 
-CURRENCIES = list(currencies)
+CURRENCIES = [
+    c for c in currencies
+    if c.alpha_3 not in {
+        'XAF', 'XAG', 'XAU', 'XBA', 'XBB', 'XBC', 'XBD', 'XDR', 'XOF', 'XPD', 'XPF', 'XPT', 'XSU', 'XTS', 'XUA',
```
wiffbi, line 241: Why are `XAF`, `XOF`, `XPF` excluded? There are even CURRENCY_PLACES defined for them (see lines 258-260.

src/pretix/base/migrations/0263_auto_20240409_0732.py
```diff
@@ -0,0 +1,26 @@
+# Generated by Django 4.2.10 on 2024-04-09 07:32
+
+from django.db import migrations
+
+
+def change_currencies(apps, schema_editor):
+    Event = apps.get_model("pretixbase", "Event")
+    Event.objects.filter(
+        currency__in={
+            'XAF', 'XAG', 'XAU', 'XBA', 'XBB', 'XBC', 'XBD', 'XDR', 'XOF', 'XPD', 'XPF', 'XPT', 'XSU', 'XTS', 'XUA',
```
wiffbi, line 10: See comment above regarding `XAF`, `XOF` and `XPF`.

wiffbi (approved):
Looks good. Why do we exclude `XAF`, `XOF`, `XPF`? We might then also remove their definitions in `CURRENCY_PLACES` …

src/pretix/_base_settings.py
raphaelm: Good catch, missed those!

PR #4061, opened 2024-04-11
Checkin: Allow to use presence state in rules

raphaelm:


robbi5:
Is this rule feature something where we should really warn that offline usage across multiple devices is not possible?

raphaelm:
Isn't that the case for all the rule features that work on previous entries?

robbi5:
Well... yeah 😅

robbi5 (approved):

PR #4067, opened 2024-04-12
Introduce locking to prevent duplicate invoices (Z#23150548)

raphaelm:
This is not a perfect solution as it does not handle all code paths to create invoices, but it handles all that seem likely to be triggered concurrently

### DIFF
--- a/src/pretix/api/views/order.py
+++ b/src/pretix/api/views/order.py
@@ -1911,12 +1911,12 @@ class InvoiceViewSet(viewsets.ReadOnlyModelViewSet):
     @transaction.atomic()
     def reissue(self, request, **kwargs):
         inv = self.get_object()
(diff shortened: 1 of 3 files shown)


src/pretix/api/views/order.py
```diff
@@ -1905,16 +1908,18 @@ def regenerate(self, request, **kwargs):
             return Response(status=204)
 
     @action(detail=True, methods=['POST'])
+    @transaction.atomic()
     def reissue(self, request, **kwargs):
         inv = self.get_object()
+        order = Order.objects.select_for_update(of=OF_SELF).get(pk=inv.order_id)
```
wiffbi, line 1914: Do we need to already lock the order here? Why not after validation right before we `generate_invoice()` ? Might be micro-optimization though …

wiffbi (commented):
Looks good. Not sure though, if we need to add locking the the invoice as well so it cannot be cancelled twice?

PR #4090, opened 2024-04-18
API: Do not write log entry for events when no changes are made

raphaelm:


wiffbi (approved):
Looks good. Still amazed by python dict-comparison …

PR #4093, opened 2024-04-19
Seating plans: Do not allow GUIDs starting or ending with spaces

raphaelm:


wiffbi (approved):
Looks good. The pattern forces the seat_guid to be at least two chars long – is that a problem? I could not come up with a regex that only requires one char though.

raphaelm:
I thought the same and I think it should be acceptable

PR #4104, opened 2024-04-22
Fix N+1 query in event calendar found by sentry

raphaelm:
This fixes a performance issue introduced by the new settings lookup in #4036

### DIFF
--- a/src/pretix/presale/views/organizer.py
+++ b/src/pretix/presale/views/organizer.py
@@ -561,6 +561,13 @@ def add_subevents_for_days(qs, before, after, ebd, timezones, event=None, cart_n
     for se in qs:
         if se.presale_is_running:
             quotas_to_compute += se.active_quotas
+            for q in se.active_quotas:
+                # save database lookups later
+                q.subevent = se
+                if event is not None:
+                    q.event = event
+                else:
+                    q.event = se.event
 
     qcache = {}
     if quotas_to_compute:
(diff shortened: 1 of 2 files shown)


src/pretix/presale/views/organizer.py
```diff
@@ -561,6 +561,13 @@ def add_subevents_for_days(qs, before, after, ebd, timezones, event=None, cart_n
     for se in qs:
         if se.presale_is_running:
             quotas_to_compute += se.active_quotas
+            for q in se.active_quotas:
+                # save database lookups later
+                q.subevent = se
+                if event is not None:
+                    q.event = event
+                else:
+                    q.event = se.event
```
luelista, line 570: why is this necessary? shouldn't this already be filled in by the `.select_related('event', 'subevent')` in `EventMixin.annotated` (i see that it isn't, but i don't understand why)?

luelista (approved):

src/pretix/presale/views/organizer.py
raphaelm: Even if it *would*, ``select_related`` would give it a **new** instance of the ``Event()`` class. We want, if possible, to use **one** instance of the ``Event()`` class so we only need to fetch the hierarkey data from cache once :)

raphaelm: (That's one reason why we sometimes use prefetch_related where select_related would be sufficient, but also why we do some hacks like this here…)

PR #4121, opened 2024-04-30
Change default password hash to argon2id

raphaelm:


wiffbi (approved):
That PR sent me from django docs to Argon2 to the Password Hashing Competition … Thanks fot that wormhole! :)

PR #4133, opened 2024-05-06
Vouchers: Fix validation of quota when copying a blocking voucher (Z#23152799)

raphaelm:


wiffbi (approved):
Looks good. Would it makes sense to change `__init__` to only set `self.initial_instance_data` if instance has a pk? I am not sure whether that would change semantics though, so probably leave as you suggest.

src/pretix/control/forms/vouchers.py
```diff
@@ -234,8 +234,8 @@ def clean(self):
         )
         if check_quota:
             Voucher.clean_quota_check(
-                data, cnt, self.initial_instance_data, self.instance.event,
-                self.instance.quota, self.instance.item, self.instance.variation
+                data, cnt, self.initial_instance_data,
+                self.instance.event, self.instance.quota, self.instance.item, self.instance.variation
```
wiffbi, line 238: Basically undo the linebreak introduced by the `if` that was originally in the PR
```suggestion
                data, cnt, self.initial_instance_data, self.instance.event,
                self.instance.quota, self.instance.item, self.instance.variation
```

wiffbi (approved):
Looks good, just a minor undo for less noise in the PR.

PR #4134, opened 2024-05-06
PDF editor: Add variables for purchase date (Z#23152887)

raphaelm:


wiffbi (commented):
Looks good. I would have placed them after `now_time` as that belongs to the group of `now_date` and `now_datetime`. Also any reason why no `purchase_time` placeholder? There are for e.g. now_* as well as valid_from_*

wiffbi (approved):

PR #4136, opened 2024-05-07
Prevent parallel refunds for the same order (Z#23152965)

raphaelm:


src/pretix/control/views/orders.py
```diff
@@ -1376,9 +1136,264 @@ def choose_form(self):
...
-            'start_form': self.start_form
+            'start_form': self.start_form,
+            'last_known_refund_id': self.order.refunds.aggregate(m=Max("id"))["m"] or 0,
         })
 
+    @transaction.atomic()
+    def perform_refund(self, comment, full_refund, payments):
+        order = Order.objects.select_for_update(of=OF_SELF).get(pk=self.order.pk)
+
+        if self.request.POST.get("last_known_refund_id") != str(self.order.refunds.aggregate(m=Max("id"))["m"] or 0):
```
wiffbi, line 1147: This is probably a theoretical problem, but should we allow `last_known_refund_id` to be `None`– might be that during deployment a POST-request from old code gets to new code? Not that it is a problem though if one has to re-initiate the refund for these probably non-existent edge-cases.

wiffbi (commented):
Looks good. Just a small comment. Do you think, that this double-submit-blocker (or better parallel-refund-blocker) shoud have a test?

PR #4139, opened 2024-05-07
Improve order secret handling

luelista:
- use hmac.compare_digest for all secret comparisons
- use salted_hmac with sha256 instead of plain sha1 for hashed secrets
- move secret handling into helper functions

after this is merged, our payment provider plugins can be refactored to use tagged_secret / get_with_secret_check instead of hashlib.sha1

### DIFF
--- a/src/pretix/base/models/orders.py
+++ b/src/pretix/base/models/orders.py
@@ -35,6 +35,7 @@
 
 import copy
 import hashlib
+import hmac
 import json
 import logging
 import operator
@@ -59,7 +60,7 @@ from django.db.models.functions import Coalesce, Greatest
 from django.db.models.signals import post_delete
 from django.dispatch import receiver
 from django.urls import reverse
-from django.utils.crypto import get_random_string
+from django.utils.crypto import get_random_string, salted_hmac
 from django.utils.encoding import escape_uri_path
 from django.utils.formats import date_format
 from django.utils.functional import cached_property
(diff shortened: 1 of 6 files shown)


src/pretix/base/models/orders.py
```diff
@@ -1219,6 +1220,38 @@ def create_transactions(self, is_new=False, positions=None, fees=None, dt_now=No
...
+        if not qs:
+            qs = Order.objects
+        dummy = "0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef"[:secret_length]
+        try:
+            order = qs.get(code=code)
+            if not hmac.compare_digest(
+                order.tagged_secret(tag, secret_length) if tag else order.secret,
+                received_secret[:secret_length].lower() if tag else received_secret.lower()
+            ) and not (
+                    # TODO: remove this clause after a while (compatibility with old secrets currently in flight)
```
raphaelm, line 1237: but not remove the `not tag` part, right? otherwise OrderDetailMixin would break

src/pretix/base/models/orders.py
```diff
@@ -1219,6 +1220,38 @@ def create_transactions(self, is_new=False, positions=None, fees=None, dt_now=No
         _transactions_mark_order_clean(self.pk)
         return create
 
+    def tagged_secret(self, tag, secret_length=64):
+        return salted_hmac(key_salt=tag, value=self.secret, algorithm="sha256").hexdigest()[:secret_length]
+
+    @staticmethod
+    def get_with_secret_check(code, received_secret, tag, secret_length=64, qs=None):
```
raphaelm, line 1227: Django style question: Would this better be a method on a custom QuerySet class so I can chain it with .filter() or .annotate() etc?

src/pretix/base/models/orders.py
luelista: i though about removing the whole part in parentheses, like this:
```diff
-              ) and not (
-                   # TODO: remove this clause after a while (compatibility with old secrets currently in flight)
-                    tag and hmac.compare_digest(
-                        hashlib.sha1(order.secret.lower().encode()).hexdigest(),
-                        received_secret.lower()
-                    )
```


why would this break OrderDetailMixin? from OrderDetailMixin, tag=None is passed in, so i assume the following condition would be checked, regardless of the second clause being removed or not (the `tag and` short-circuits the second clause to False anyway if no tag is present):

```
            if not hmac.compare_digest(
                 order.secret,
                 received_secret.lower()
            )
```

src/pretix/base/models/orders.py
luelista: Chaining as in `event.orders.get_with_secret_check(...).annotate(...)`? I think it wouldn't be possible to implement it that way, as we need the Order object in here for the secret check? 

The other way round, would work I think, which would just change the API slightly from `Order.get_with_secret_check(..., qs=event.orders.annotate(...))` to `event.orders.annotate(...).get_with_secret_check(...)` which seems nicer though, yes. I'll have a look into implementing that.

src/pretix/base/models/orders.py
raphaelm: oh, yes, my mistake

src/pretix/base/models/orders.py
luelista: could you confirm that this is in fact an okay way to specify the Manager for orders:
https://github.com/pretix/pretix/pull/4139/files#diff-a88712c33526e582606fb2e095cea07e0972a201338a955b91cf2dfc675fa3bbR318

in other places it's done in a more verbose way with a separate manager (e.g. [ItemQuerySetManager](https://github.com/pretix/pretix/blob/master/src/pretix/base/models/items.py#L289)), I think my way is equivalent with less duplicated code but I'm not sure if there is some subtlety I'm missing?

raphaelm: as_manager() is certainly fine, and if it works with scopes, fine with me

raphaelm (approved):

PR #4162, opened 2024-05-22
Widget: Allow passing more invoice data fields

raphaelm:


wiffbi (approved):
Looks good and works as expected. Not sure the naming is ideal as IMHO the new fields semantically do not feel like being part of an address, so instead name them e.g. `data-invoice-internal-reference`, `data-invoice-vat-it`, `data-invoice-custom-field`? On the other hand `data-invoice-vat-it` could be mistaken for the VAT-ID of the seller, so maybe just leave it as is.

PR #4177, opened 2024-05-24
Harden timing when getting order with secret check

wiffbi:
When getting an order through `OrderQuerySet::get_with_secret_check`, if the order was found, but the secret was wrong, two hash-compares were executed which could potentially lead to timing attacks – basically if the response takes twice as long, it is a correct order-code, but a wrong hash.


luelista (approved):
Thanks, good catch :)

PR #4182, opened 2024-05-29
Fix email preview curly brackets mismatch

wiffbi:
This PR adds a modification to the email preview handling so it does not break, when trying to preview text with not matching } or {. It also respects escaped {{. To have the reg-ex respect escaped {{ and }}, it masks {{ and }} in a parallel string, that is tested for correctness and matches in the test-string are then transferred through the original string into a cleaned-string.

I am not sure though, that this is the best way possible. It does render the mismatching first { or last } as plain { or }. An alternative would be to instead render the last { or first } plain, so one can see the error in formatting easier as the matched placeholder would not work.


wiffbi:
#4184 is a better solution

PR #4184, opened 2024-05-29
Improve validation of email templates

luelista:


wiffbi (commented):
I like your solution better than mine! Maybe change the error message to something like `There is an error with your placeholder syntax. Please check your opening and closing curly brackets „{ }“ on your placeholders. Please note: to use literal { or }, you need to double them as {{ and }}.`

I also like your idea to surround this message with a `div.alert.alert-danger`.

PR #4185, opened 2024-05-29
Draft: Cross selling

luelista:


src/pretix/base/models/items.py
```diff
@@ -110,18 +113,138 @@ class ItemCategory(LoggedModel):
...
+        verbose_name=_("Cross-selling condition products"),
+        related_name="matched_by_cross_selling_categories",
+    )
 
     class Meta:
         verbose_name = _("Product category")
         verbose_name_plural = _("Product categories")
         ordering = ('position', 'id')
 
+    def cross_sell_visible(self, cart, sales_channel):
```
raphaelm, line 149: A cart here is a list of cartpositions, not the weird "cart dictionary" we have in some other places?

src/pretix/base/models/discount.py
```diff
@@ -37,6 +37,12 @@
 from pretix.base.models import fields
 from pretix.base.models.base import LoggedModel
 
+ITEM_ID = 0
+SUBEVENT_ID = 1
+LINE_PRICE_GROSS = 2
+IS_ADDON_TO = 3
+VOUCHER_DISCOUNT = 4
```
raphaelm, line 44: Not the point of this PR, but if we do this, maybe we should jump straight to a namedtuple? Still an improvement this way.

src/pretix/base/models/discount.py
```diff
@@ -268,30 +278,61 @@ def _apply_min_count(self, positions, condition_idx_group, benefit_idx_group, re
...
+
+            # how many discount applications are possible according to benefitting products in cart
+            possible_applications_benefit = ceil(len(benefit_idx_group) / self.benefit_only_apply_to_cheapest_n_matches)
+
+            n_groups = min(possible_applications_cond, possible_applications_benefit)
             consume_idx = condition_idx_group[:n_groups * self.condition_min_count]
             benefit_idx = benefit_idx_group[:n_groups * self.benefit_only_apply_to_cheapest_n_matches]
+
+            if collect_potential_discounts is not None:
+                if n_groups * self.benefit_only_apply_to_cheapest_n_matches > len(benefit_idx_group):
```
raphaelm, line 299: Should this be a tighter condition than just `>`? If I have a group discount of "if you buy 10 tickets, you get one free", I probably want to advertise that to someone buying 9 tickets, but not someone buying 1 ticket.

src/pretix/presale/checkoutflow.py
```diff
@@ -605,10 +630,96 @@ def forms(self):
...
+        ]
+
+    @cached_property
+    def cross_selling_is_applicable(self):
+        return any(len(items) > 0 for (category, items) in self.cross_selling_data)
+
+    @cached_property
+    def cross_selling_data(self):
+        class DummyCategory:
+            def __init__(self, rule, subevent=None):
```
raphaelm, line 652: Why "rule"? Took me a while to understand what is passed in there (type annotation would also help understanding here)

src/pretix/presale/checkoutflow.py
```diff
@@ -486,11 +488,34 @@ class AddOnsStep(CartMixin, AsyncAction, TemplateFlowStep):
...
+        cur_step_identifier = request.resolver_match.kwargs.get('step')
+        is_past_this_step = any(step.identifier == cur_step_identifier for step in request._checkout_flow[request._checkout_flow.index(self) + 1:])
+        if is_past_this_step:
+            applicable = self.cart_session.get('_checkoutflow_addons_applicable', None)
+            if applicable is not None:
+                return applicable
+
+        # check whether cross-selling is applicable
+        applicable = self.cross_selling_is_applicable
+        self.cart_session['_checkoutflow_addons_applicable'] = applicable
```
raphaelm, line 508: We were talking about the need of caching here. To think about this more, do you have first estimates on how many database queries this part takes?

raphaelm (commented):
I did a first read. I can't say I fully understood the complex logic, but I also maybe did not put in the time required to do so (I absolutely will before we merge this, no worries, but at that point hopefully a great test suite will make it easier to play with it and understand the edge cases :wink:). I think the PR is on the right track (although the added complexity scares me a little).

src/pretix/presale/checkoutflow.py
```diff
@@ -605,10 +630,96 @@ def forms(self):
...
+                for subevent in subevents
+            )
+        else:
+            result = (
+                (category, self._items_for_cross_selling(None, items_qs, discount_info))
+                for (category, items_qs, discount_info) in categories
+            )
+        return [(category, items) for (category, items) in result if len(items) > 0]
+
+    def _items_for_cross_selling(self, subevent, items_qs, discount_info):
```
raphaelm, line 672: Would it make sense to move this into a separate module outside of the checkout step with clear boundaries (like a get_grouped_items, but one layer higher)? That would at least make it more testable and maybe even more understandable if the boundary is clearer

src/pretix/presale/checkoutflow.py
luelista: that was a leftover from the pretix_cross_selling plugin where I used cross-selling-rules instead of categories

src/pretix/base/models/items.py
```diff
@@ -110,16 +113,116 @@ class ItemCategory(LoggedModel):
...
+                    [
+                        (cp.item_id, cp.subevent_id, cp.line_price_gross, bool(cp.addon_to), cp.is_bundled,
+                         cp.listed_price - cp.price_after_voucher)
+                        for cp in cart
+                    ],
+                    collect_potential_discounts=potential_discounts_by_cartpos
+                )
+
+                # technically, this is a dict, but we use it as an OrderedSet here
+                potential_discount_set = dict.fromkeys(info for lst in potential_discounts_by_cartpos.values() for info in lst)
```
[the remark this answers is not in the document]
luelista: It's just used to flatten the dict of lists into an ordered set. I don't think this is possible without nesting?
I added a comment to clarify: https://github.com/pretix/pretix/pull/4185/files#diff-5479c6a8cd3d750ee44dfa7f5eb7ea287f2475e072ad596c1ebe9f08a03e94a7R180-R181

src/pretix/presale/checkoutflow.py
luelista: yep, lives in `pretix/base/services/cross_selling.py` now

src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
```diff
@@ -136,11 +136,11 @@ <h5 id="item-{{ item.pk }}-{{ var.pk }}-legend">{{ var }}</h5>
                                         <div class="input-group input-group-price">
                                             <span class="input-group-addon">{{ event.currency }}</span>
                                             <input type="number" class="form-control input-item-price"
-                                                   id="price-variation-{{ item.pk }}-{{ var.pk }}"
+                                                   id="{{ id_prefix }}price-variation-{{ item.pk }}-{{ var.pk }}"
                                                    {% if not ev.presale_is_running %}disabled{% endif %}
                                                    placeholder="0"
                                                    min="{% if event.settings.display_net_prices %}{{ var.display_price.net|money_numberfield:event.currency }}{% else %}{{ var.display_price.gross|money_numberfield:event.currency }}{% endif %}"
-                                                   name="price_{{ item.id }}_{{ var.id }}"
+                                                   name="{{ id_prefix }}price_{{ item.id }}_{{ var.id }}"
```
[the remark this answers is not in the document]
luelista: but it's a prefix to these item.id / variation.id strings...? but maybe `subevent_prefix` is better because it will contain the subevent id?

src/pretix/presale/checkoutflow.py
luelista: I managed to get a constant number of queries with regards to the number of items per category, categories in general, and number of cart positions, however for every cross-selling category that's actually rendered, a separate bunch of queries is performed by get_grouped_items. It might be possible to merge this somehow into a single call to get_grouped_items and split it up again afterwards, but I'm not sure whether that's worth the additional complexity... After all, I think in a usual shop there will be only one or a hand full of cross-selling categories visible in each order.

src/pretix/base/models/discount.py
luelista: done

src/pretix/base/models/items.py
```diff
@@ -110,16 +113,116 @@ class ItemCategory(LoggedModel):
...
         verbose_name_plural = _("Product categories")
         ordering = ('position', 'id')
 
+    def cross_sell_visible(self, cart, sales_channel):
+        """
+        If this category should be visible in the cross-selling step for a given cart and sales_channel, this method
+        returns a queryset of the items that should be displayed, as well as a dict giving additional information on them.
+
+        :returns: (QuerySet<Item>, dict<item_pk: (max_count, discount_rule)>)
+            max_count is `inf` if the item should not be limited
```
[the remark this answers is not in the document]
luelista: > I'd assume inf means I can buy it infinite times
yes, that's what it means here, inf = not limited = can be bought infinite times

this also works nicely with using the `max` function to operate on these values

src/pretix/base/models/items.py
```diff
@@ -110,16 +113,116 @@ class ItemCategory(LoggedModel):
                     'only be bought in combination with a product that has this category configured as a possible '
                     'source for add-ons.')
     )
+    CROSS_SELLING_MODES = (
```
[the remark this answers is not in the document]
luelista: I'm a bit hesitant to change that, as it'd need changes in many places, we'd need to keep API compatibility anyway, and I don't see it as a much better representation (it would match the UI, but on the other hand, these are two quite different properties of a category - is_addon is something that's really enforced in the cart logic, whereas the cross-selling mode is just presentational).

I don't have a strong opinion on it, so if you do think it's a lot better, I can change it, but it's quite some work and I'm not sure whether it's worth the effort.

src/pretix/base/models/items.py
```diff
@@ -111,6 +110,33 @@ class ItemCategory(LoggedModel):
...
+        choices=CROSS_SELLING_CONDITION,
+        null=True,
+        max_length=10,
+    )
+    cross_selling_match_products = models.ManyToManyField(
+        'pretixbase.Item',
+        blank=True,
+        verbose_name=_("Cross-selling condition products"),
+        related_name="matched_by_cross_selling_categories",
+    )
```
raphaelm, line 140: I think the PR is lacking code to handle events being cloned and then re-assigning the cloned items to the cloned categories correctly for this relationship.

src/pretix/base/models/items.py
```diff
@@ -111,6 +110,33 @@ class ItemCategory(LoggedModel):
...
+    cross_selling_mode = models.CharField(
+        choices=CROSS_SELLING_MODES,
+        null=True,
+        max_length=5
+    )
+    CROSS_SELLING_CONDITION = (
+        ('always', _('Always show in cross-selling step')),
+        ('products', _('Only if the cart contains one of these products')),
+        ('discounts', _('Only show products affected by discount rules')),
```
raphaelm, line 126: I'm not sure if "affected" is totally clear here to the user, but I have not yet come up with a better idea.

src/pretix/base/services/cross_selling.py
```diff
@@ -0,0 +1,201 @@
...
+        # - max_count for product: sum up max_counts
+        # - discount_rule for product: take first discount_rule
+
+        def discount_info(item, infos_for_item):
+            infos_for_item = list(infos_for_item)
+            return (
+                item,
+                sum(max_count for (item, discount_rule, max_count, i) in infos_for_item),
+                next(discount_rule for (item, discount_rule, max_count, i) in infos_for_item)
+            )
```
raphaelm, line 140: Wouldn't we need subevent here as well for proper handling in case discount.subevent_mode is set to SUBEVENT_MODE_SAME or am I missing a piece?

src/pretix/base/services/cross_selling.py
```diff
@@ -0,0 +1,201 @@
...
+        ]
+
+    def _prepare_items(self, subevent, items_qs, discount_info):
+        items, _btn = get_grouped_items(
+            self.event,
+            subevent=subevent,
+            voucher=None,
+            channel=self.sales_channel,
+            base_qs=items_qs,
+            allow_addons=True,
```
raphaelm, line 163: Really?

src/pretix/base/services/cross_selling.py
```diff
@@ -0,0 +1,201 @@
...
+                    testmode=self.event.testmode
+                )
+                if self.customer else None
+            ),
+        )
+        new_items = list()
+        for item in items:
+            max_count = inf
+            if item.pk in discount_info:
+                (max_count, discount_rule) = discount_info[item.pk]
```
raphaelm, line 177: This will probably also gain a subevent dependency to make SUBEVENT_MODE_SAME work

src/tests/base/test_cross_selling.py
```diff
@@ -0,0 +1,634 @@
...
+        Regular Ticket    42.00          42.00
+        Regular Ticket    42.00          42.00
+
+        Reduced Ticket    23.00          11.50
+        ''',
+        recommendations='''             Price     Discounted Price    Max Count
+        Tickets     Reduced Ticket      23.00                11.50            1
+        '''
+    )
+    print("The interesting part:")
```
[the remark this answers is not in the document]
luelista: no :D 
```suggestion
```

src/tests/base/test_cross_selling.py
```diff
@@ -0,0 +1,634 @@
...
+        ''',
+        recommendations='''             Price     Discounted Price    Max Count
+        Tickets     Reduced Ticket      23.00                11.50            2
+        '''
+    )
+
+
+@scopes_disabled()
+@pytest.mark.django_db
+@pytest.mark.skip("currently unsupported (cannot give discount to specific product on minimum cart value)")
```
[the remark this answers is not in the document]
luelista: I think I left this in so far because this was one of our example use cases (the champagne example), and it's not supported now...? So we should at least think about whether we want to support this in one way or another, and remove it otherwise.

src/pretix/base/services/cross_selling.py
luelista: Probably not... (shouldn't make a difference as it's mutually exclusive to appearing in cross-selling queryset, but should still be False)
```suggestion
            allow_addons=False,
```

src/pretix/base/models/items.py
luelista: In line with your "celebration" wording from Slack maybe "Only show products that qualify for a discount according to discount rules"?

raphaelm: :+1:

src/pretix/base/services/cross_selling.py
```diff
@@ -0,0 +1,201 @@
...
+                    previous_price = item.display_price
+                    new_price = (
+                        previous_price * (
+                            (Decimal('100.00') - discount_rule.benefit_discount_matching_percent) / Decimal('100.00'))
+                    )
+                    item.display_price = new_price
+
+            # reduce order_max by number of items already in cart (prevent recommending a product the user can't add anyway)
+            item.order_max = min(
+                item.order_max - sum(1 for pos in self.cartpositions if pos.item_id == item.pk),
```
wiffbi, line 195: This fails, when I have an item with variants. I am not sure, how we should handle variants here though. Remove variants, which should not be recommended?

src/pretix/api/serializers/item.py
```diff
@@ -441,7 +441,22 @@ class ItemCategorySerializer(I18nAwareModelSerializer):
 
     class Meta:
         model = ItemCategory
-        fields = ('id', 'name', 'internal_name', 'description', 'position', 'is_addon')
+        fields = (
+            'id', 'name', 'internal_name', 'description', 'position',
+            'is_addon', 'cross_selling_mode',
+            'cross_selling_condition', 'cross_selling_match_products'
```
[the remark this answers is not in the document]
luelista: added to resources/categories.rst

src/pretix/static/pretixpresale/scss/_checkout.scss
```diff
@@ -102,6 +102,9 @@
     }
   }
 }
+.cross-selling .panel-body > section:first-child > h3 {
+  margin-top: 10px; font-size: 21px;
+}
```
wiffbi, line 107: Why does only the first section get a smaller category-headline? Looks a bit strange if I have multiple cross-selling categories. Also – this ist very detailed nit-picking – the `margin-top` adds to much space compared to other addons headline inside a `.panel-body`. Also the default 45px margin-top for h3 is way bigger than the space when having multiple addon-categories. To match the spacing of addons, it needs to be:

```suggestion
.cross-selling .panel-body h3 {
  font-size: 21px;
  line-height: inherit;
  margin-top: 20px;
}
.cross-selling .panel-body > *:first-child > h3:first-child {
  margin-top: 0;
}
```

Note the `line-height: inherit` which increases the line-height to the same amount as the `<legend>` elements for the addons have and thus creates exactly the same space at the top of the `.panel-body`.

Arguably the `margin-top: 20px` might be a bit small, but if we were to increase it to e.g. 30px, we would need to change `.addons fieldset` as well.

src/pretix/static/pretixcontrol/scss/_forms.scss
```diff
@@ -588,6 +588,9 @@ table td > .checkbox input[type="checkbox"] {
     border-bottom: 1px solid #ccc;
   }
 }
+.form-horizontal .big-radio-wrapper .radio label > span {
+  display: block;
+}
```
wiffbi, line 593: To match the existing big-radio in product settings, we need to make the radio-labels bold and add a little space at the top of the help-text. Furthermore I added more spacing to the top of the control-label to match the baseline of the first radio-label (this is an improvement to the big-radio in product-settings as well).

```suggestion
.form-horizontal .control-label:has(+.big-radio-wrapper),
.form-horizontal .control-label:has(+div > .big-radio) {
  padding-top: 16px;
}
.form-horizontal .big-radio-wrapper .radio label {
  font-weight: bold;
}
.form-horizontal .big-radio-wrapper .radio label > span {
  display: block;
  font-weight: normal;
}
```
Bildschirmfoto 2024-10-11 um 10 05 50
Bildschirmfoto 2024-10-11 um 10 06 16

src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -411,8 +414,10 @@ var form_handlers = function (el) {
                 }
             };
         update();
-        dependency.closest('.form-group').find('[name=' + dependency.attr("name") + ']').on("change", update);
-        dependency.closest('.form-group').find('[name=' + dependency.attr("name") + ']').on("dp.change", update);
+        dependency.each(function() {
+            $(this).closest('.form-group').find('[name=' + $(this).attr("name") + ']').on("change", update);
+            $(this).closest('.form-group').find('[name=' + $(this).attr("name") + ']').on("dp.change", update);
```
wiffbi, line 419: As it is the same element, it can be compressed to:
```suggestion
            $(this).closest('.form-group').find('[name=' + $(this).attr("name") + ']').on("change dp.change", update);
```

src/pretix/base/services/cart.py
```diff
@@ -1611,13 +1610,14 @@ def clear_cart(self, event: Event, cart_id: str=None, locale='en', sales_channel
...
 @app.task(base=ProfiledEventTask, bind=True, max_retries=5, default_retry_delay=1, throws=(CartError,))
-def set_cart_addons(self, event: Event, addons: List[dict], cart_id: str=None, locale='en',
+def set_cart_addons(self, event: Event, addons: List[dict], add_to_cart_items: List[dict], cart_id: str=None, locale='en',
                     invoice_address: int=None, sales_channel='web', override_now_dt: datetime=None) -> None:
     """
-    Removes a list of items from a user's cart.
+    Assigns addons to eligible products in a user's cart, adding and removing the addon products as necessary to
+    ensure the requested addon state.
     :param event: The event ID in question
     :param addons: A list of dicts with the keys addon_to, item, variation
```
wiffbi, line 1619: Should we document the `add_to_cart_items` parameter here as well? Or is it kind of self-explanatory like locale, invoice_adress, etc. Addons is documented, so I guess we should add it.

wiffbi (changes requested):
Looks good! I especially like the changes with `namedtuple` in discount.py and the `with`-statement in fragment_product_list.html – makes both of them so much more readable! I just have some small comments regarding labels and a little CSS to improve the UI/adapt even more to the exisiting big-radio.

src/pretix/base/services/cart.py
luelista:
```suggestion
    :param addons: A list of dicts with the keys addon_to, item, variation
    :param add_to_cart_items: A list of dicts with the keys item, variation, count, custom_price, voucher, seat ID
```

src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
```diff
@@ -5,20 +5,28 @@
...
-    {% if tup.0 %}
-        <section aria-labelledby="category-{{ tup.0.id }}"{% if tup.0.description %} aria-describedby="category-info-{{ tup.0.id }}"{% endif %}>
-            <h3 id="category-{{ tup.0.id }}">{{ tup.0.name }}</h3>
-            {% if tup.0.description %}
-                <div id="category-info-{{ tup.0.id }}">{{ tup.0.description|localize|rich_text }}</div>
+{% for tup in items_by_category %}{% with category=tup.0 items=tup.1 form_prefix=tup.2 %}
+    {% if category %}
+        <section aria-labelledby="{{ form_prefix }}category-{{ category.id }}"{% if category.description %} aria-describedby="{{ form_prefix }}category-info-{{ category.id }}"{% endif %}>
+            <h3 id="{{ form_prefix }}category-{{ category.id }}">{{ category.name }}
+                {% if category.category_has_discount %}
```
wiffbi, line 12: If we add the subevent-property to the DummyCategory, then this would be the HTML.
```suggestion
                {% if category.subevent %}
                    &nbsp; <small><span class='fa fa-calendar'></span> {{ category.subevent }}</small>
                {% endif %}
                {% if category.category_has_discount %}
                    &nbsp;
```
Bildschirmfoto 2024-10-11 um 13 10 40

src/pretix/base/services/cross_selling.py
```diff
@@ -0,0 +1,231 @@
...
+
+class DummyCategory:
+    """
+    Used to create fake category objects for displaying the same cross-selling category multiple times,
+    once for each subevent
+    """
+
+    def __init__(self, category: ItemCategory, subevent):
+        self.id = category.id
+        self.name = f"{category.name} ({subevent})"
```
wiffbi, line 43: I’d love to display the subevent as a formatted HTML-string like so:

Bildschirmfoto 2024-10-11 um 12 52 04

Do you have an idea how to best add it? Give the ItemCategory a subevent property and handle it in the HTML-template? In quick tests, I did not get `__html__` to work, but I would not want `__str__` to return HTML.

```suggestion
        self.name = f"{category.name}"
        self.subevent = subevent
```

wiffbi (approved):
Another tiny UI idea regarding subevent-date-display. Other than that, looks and works. IMHO ready for staging. 👍

PR #4191, opened 2024-06-02
Replace SCSS compilation with CSS variables

raphaelm:
Currently, theming of shops (e.g. colors etc) is implemented through SCSS variables. This is mainly historically because that's what Bootstrap3 natively supports and what works with all browsers. However, it has some major drawbacks. Mostly, our production deployment is getting close to 100k events from 15k organizers. Re-compiling the SCSS for every one of them after every change of the source is slowly getting problematic operationally and can also cause weird intermittent issues since it takes half an our after CSS changes arrive in all shops after a change was deployed.

Now that Internet Explorer is dead for a while, we can finally use CSS variables instead.

Bootstrap 3 does not support that, but Bootstrap 3 is unmaintained anyways and since an in-place upgrade to a new Bootstrap version would be impossible anyways, there is nothing holding us back from changing ("forking") our embedded version of Bootstrap 3.4.1. So this PR changes Bootstrap in a way that we can use CSS variables as theme inputs!

By consequence, after this step pretix would absolutely no longer work in Internet Explorer. I think that is fine at this point – our access logs still find lots of entries with MSIE and Trident, but it looks like that is almost all bots.

Also, this removes the `sass_preamble` and `sass_postamble` signals. Plugins can just inject HTML with their custom CSS into the header/body for all known use cases. Plugins that need changes:

- [x] pretix-c-mb
- [x] pretix-c-wzb
- [x] pretix-c-quadriga
- [x] pretix-stay22
- [x] pretix-closer2event
- [x] pretix-seating


raphaelm:
Apparently, we used #8E44B3 as the default primary color for a while but #7F4A91 in the widget? I don't quite get it, but this PR also unifies it to 8E44B3

src/pretix/presale/style.py
```diff
@@ -312,8 +167,53 @@ def get_font_stylesheet(font_name, event: Event = None):
...
+            '$in-font-family-sans-serif: "{}", "Open Sans", "OpenSans", "Helvetica Neue", Helvetica, Arial, sans-serif '
+            '!default'.format(
+                font
+            )
+        )
+
+    if widget:
+        sassrules.append("$widget: true;")
+
+    with open(finders.find("pretixbase/scss/_variables.scss"), "r") as f:
```
luelista, line 210: vllt die datei dann auch `_theme_vars.scss` oder so nennen? irgendwie die unterscheidung zwischen der `_bootstrap_vars` und `_variables` klarer machen

src/pretix/presale/style.py
```diff
@@ -312,8 +167,53 @@ def get_font_stylesheet(font_name, event: Event = None):
...
+        source_scss = f.read()
+        sassrules.append(source_scss)
+
+    sassdir = os.path.join(settings.STATIC_ROOT, "pretixbase/scss")
+    css = sass.compile(
+        string="\n".join(sassrules),
+        include_paths=[sassdir]
+    )
```
luelista, line 218: ist sass.compile performant genug, dass wir das direkt beim rendern der response so machen sollten? (ich vermute die antwort ist ja, kurzes testen hat ergeben dass es auf meinem laptop ca 10ms unterschied ausmacht wenn ich diesen block durch ein `css = "\n".join(sassrules)` ersetze)

ich frage mich, brauchen wir an dieser stelle überhaupt noch sass? wir machen ja quasi nur noch template-ersetzung mit einer handvoll farb-berechnungen. das könnte man ja ggf auch mit einem jinja2-template lösen. aber solange es keine probleme verursacht können wir auch einfach bei sass bleiben.

luelista (approved):

luelista:
this also needs changes:
- [ ] pretixeu

src/pretix/static/pretixbase/scss/_variables.scss
```diff
@@ -7,69 +8,127 @@
...
-$gray-lightest: lighten(#000, 97.25%);
+$widget: false !default;
 
-$font-family-sans-serif:  "Open Sans", "OpenSans", "Helvetica Neue", Helvetica, Arial, sans-serif !default;
-$text-color: #222222 !default;
-$text-muted: #767676 !default;
-$input-color-placeholder: lighten(#000, 70%) !default;
+// Input variables that may be overridden by event themes.
+$in-font-family-sans-serif: "Open Sans", "OpenSans", "Helvetica Neue", Helvetica, Arial, sans-serif !default;
+$in-brand-primary: #7f5a91 !default;
```
wiffbi, line 18: Shouldn’t this be 

```suggestion
$in-brand-primary: #8e44b3 !default;
```

src/pretix/presale/context.py
```diff
@@ -80,10 +81,22 @@ def _default_context(request):
 
     if hasattr(request, 'event') and request.event:
```
wiffbi, line 82: This might be overkill, but would it make sense to check, if any customization on event-level was done? I guess it is safe to assume that at least on organizer-level there are customizations, but on event-level? That way we would not need to compile (and re-compile) for every event of an organizer, but most of the time just on organizer-level and just include the organizer-theme.css on events as well. Problem is, how do we check settings are on organizer-level and not event-level when event-settings inherit from organizer automagically … Maybe not worth the effort?

wiffbi (approved):
Looks good. Just a minor comment regarding a wrong color – that is always overwritten with the correct one in the theme.css, but nonetheless. And a question, whether it is worth it to optimize if event-level customization is not set to return organizer-level customization.

src/pretix/presale/style.py
raphaelm: Yeah I had the same question and the same experimental result. We could probably do it in a Django template but I did not quickly find a library that does all of the color computations efficiently. libsass is native code, so a python implementation of the color computations is even likely to be slower. We would save on the parsing and import time though

wiffbi: Maybe some time in the future, we could swap to [CSS color-mix](https://developer.mozilla.org/en-US/blog/color-palettes-css-color-mix/)? Then no pre/post-processor would be needed. Although to have this somewhat backwards-compatible would mean the same effort to have this PR with `var()` colors compatible with IE – ie. doubling all color-related properties with the unmixed and mixed color.

raphaelm: Oh I did not notice color-mix being well supported now, I looked for stuff like the lighten and the contrasts and the functions I found were not even in recent Firefox yet

PR #4193, opened 2024-06-03
Web checkin: Render special cases of pending state in search (Z#23154934)

raphaelm:


wiffbi (approved):
Looks good. In pretix-control the state pending_valid is blue/$brand-info – probably not useful in the context of check-in, which an either yay or nay is best. So good to go IMHO.

PR #4194, opened 2024-06-03
Bank transfer: Ignore checksum for blocklist (Z#23154934)

raphaelm:


src/pretix/plugins/banktransfer/payment.py
```diff
@@ -591,7 +593,12 @@ def payment_refund_supported(self, payment: OrderPayment) -> bool:
         except ValidationError:
             return False
         else:
-            return not any(iban.startswith(b) for b in (self.settings.refund_iban_blocklist or '').splitlines() if b)
+            def _compare(iban, prefix):  # Compare IBAN with pretix ignoring the check digits
+                iban = iban[:2] + "XX" + iban[4:]
+                prefix = prefix[:2] + "XX" + prefix[4:]
+                return iban.startswith(prefix)
```
wiffbi, line 599: As we remove both check-digits, wouldn’t it be enough to just do

```
iban = iban[:2] + iban[4:]
prefix = prefix[:2] + prefix[4:]
```

Or even shorten it to:

```suggestion
                return iban[:2].startswith(prefix[:2]) and iban[4:].startswith(prefix[4:])
```

wiffbi (approved):
Looks good! I just added a micro-optimization, which you are more than free to ignore.

src/pretix/plugins/banktransfer/payment.py
raphaelm: Sure. Is it more or less readable though=

PR #4215, opened 2024-06-11
Improve List Sorting UI

luelista:
- product list:
  - allow move between categories 
  - replace redundant column "category" (which is already visible as group headers) with "base price" (very open to suggestions for better columns to show here)
- unclutter UI a bit by hiding up/down arrows if drag-drop is available

new:

image

before:

image


src/pretix/static/pretixcontrol/js/ui/dragndroplist.js
```diff
@@ -1,48 +1,46 @@
...
         }
-
+        let didSort = false, lastClick = 0;
+        function handleMouseUp() {
+            if (Date.now() - lastClick < 6000) {
+                container.find(".sortable-up, .sortable-down").removeClass("sr-only");
+            }
+            lastClick = Date.now();
+        }
+        container.find(".dnd-sort-handle").on("mouseup", handleMouseUp);
```
wiffbi, line 21: This behaves a bit strange IMHO. When I click a drag-handle, but do not drag far enough (ie. outside of the button), it is registered as a click/mouseup. When I then click any other button inside that group, decide not to drag, I get the sortable buttons. I think the 6s delay is to long, maybe 2s might be enough? And it should either apply to all drag-buttons together or (IMHO better) to each button individually. Also, when a successfull drag was done, I think the whole handleMouseUp-Listener could be removed (although that would also mean to remove the help-text/title regarding double-click on that drag-element – so better just leave it connected)?

Also maybe call the method as what it does instead of when its called, so something like `checkIfSortButtonsNeeded`?

src/pretix/static/pretixcontrol/js/ui/dragndroplist.js
```diff
@@ -1,48 +1,46 @@
...
         Sortable.create(container.get(0), {
             filter: ".sortable-disabled",
             handle: ".dnd-sort-handle",
+            group: container.data("dnd-group"),
             onMove: function (evt) {
                 return evt.related.className.indexOf('sortable-disabled') === -1;
             },
             onStart: function (evt) {
                 container.addClass("sortable-dragarea");
```
wiffbi, line 30: I have not looked to much into it, but when I start dragging, all groups are „dimmed“ except the `container`. Not sure how we easily fix this, but not break it for other draggable sortable lists. Maybe we should change the `data-dnd-url` to a global one on the whole table here? Not sure how this then behaves with the category headlines. So maybe live with it or have some other solution to not dim the other groups.

src/pretix/static/pretixcontrol/scss/main.scss
```diff
@@ -841,9 +841,20 @@ h1 .label {
...
+}
+.table-items tbody[data-dnd-url]:after {
+    content:'';
+    display: table-row;
+    height: 0.5em;
+}
 .sortable-sorting tbody:not(.sortable-dragarea) {
     opacity: .4;
 }
+th span.not-bold {
```
wiffbi, line 855: Maybe a more generic `.font-normal` with a font-weight and -style: normal would be more practical? But feel free to leave it as-is.

```suggestion
.font-normal {
    font-style: normal;
```

src/pretix/base/models/items.py
```diff
@@ -122,6 +122,16 @@ def __str__(self):
...
+    def get_category_type_display(self):
+        if self.is_addon:
+            return _('Add-On products')
+        else:
+            return None
+
+    @property
+    def category_type(self):
+        return 'addon' if self.is_addon else 'normal'
+
```
wiffbi, line 134: I do not see it used elsewhere (yet?), so is it worth the added „complexity“ instead of doing a `{% if c.is_addon %} <span class="font-normal">({% trans "Add-On products" %})</span>{% endif %}` in the (currently only one) template?

src/pretix/static/pretixcontrol/js/ui/dragndroplist.js
```diff
@@ -1,48 +1,46 @@
 /*global $, Sortable*/
 $(function () {
     $("[data-dnd-url]").each(function(){
-        var container = $(this),
+        const container = $(this),
             url = container.data("dnd-url"),
             handle = $('<span class="btn btn-default btn-sm dnd-sort-handle"><i class="fa fa-arrows"></i></span>');
```
wiffbi, line 6: I know, not part of this PR, but maybe just improve it to a real button?
```suggestion
            handle = $('<button type="button" class="btn btn-default btn-sm dnd-sort-handle"><i class="fa fa-arrows"></i></button>');
```

src/pretix/static/pretixcontrol/js/ui/dragndroplist.js
```diff
@@ -1,48 +1,46 @@
...
-                }
+                if (!didSort) handleMouseUp();
             },
             onSort: function (evt){
-                var container = $(evt.to),
-                    ids = container.find("[data-dnd-id]").toArray().map(function (e) { return e.dataset.dndId; });
+                if (evt.target !== evt.to) return;
+                didSort = true;
 
+                const ids = container.find("[data-dnd-id]").toArray().map(function (e) { return e.dataset.dndId; });
```
wiffbi, line 43: Weee, TIL about `Element.dataset` – thanks!

wiffbi (commented):
Looks good. Just a minor thing regarding dimming other categories although they are a valid drag-target. And maybe we need some optimization around that double-click-behaviour to not limit it to a group. Or in practical terms it might maybe be enough to just shorten the click-interval to 1 or 2 seconds (and simply keep it on a per group basis).

src/pretix/base/models/items.py
luelista: I did this way, because this PR is somewhat of a "backport" from [the cross-selling PR](https://github.com/pretix/pretix/pull/4185), where I first implemented this and then split it out into this separate PR. Once the cross-selling feature lands, we need a more complex logic which I think justifies the additional code here:
```python

    def __str__(self):
        name = self.internal_name or self.name
        category_type = self.get_category_type_display()
        if category_type:
            return _('{category} ({category_type})').format(category=str(name), category_type=category_type)
        return str(name)

    def get_category_type_display(self):
        if self.is_addon:
            return _('Add-On products')
        elif self.cross_selling_mode:
            return self.get_cross_selling_mode_display()
        else:
            return None

    @property
    def category_type(self):
        return 'addon' if self.is_addon else self.cross_selling_mode or 'normal'

```

wiffbi: Makes total sense, thanks for clearing that up!

src/pretix/static/pretixcontrol/js/ui/dragndroplist.js
luelista: lieber nicht, das verändert irgendwie das click-handling was die browser machen, jedenfalls wird mit einem `<button>` das drag-and-drop *manchmal* nicht ausgelöst, während es mit einem `<span>` zuverlässig ist

src/pretix/static/pretixcontrol/js/ui/dragndroplist.js
luelista: habs das verhalten nochmal ein bisschen getuned :) 

ich habs auf "alle buttons merken sich die zeit gemeinsam" geändert weil das einfach zu implementieren ist und ich finde es macht nicht so einen großen unterschied, aber die zeit reduziert.

wiffbi: Super! Mir würde fast noch weniger Zeit (z.B. 800ms) reichen, aber jetzt triggered es nicht mehr so schnell (und dadurch gefühlt fast zufällig).

src/pretix/static/pretixcontrol/js/ui/dragndroplist.js
```diff
@@ -1,48 +1,57 @@
...
-
-                var disabledDown = container.find(".sortable-down:disabled"),
-                    lastDown = container.find(">tr[data-dnd-id] .sortable-down").last();
-                if (disabledDown.length && disabledDown.get(0) !== lastDown.get(0)) {
-                    disabledDown.prop("disabled", false);
-                    lastDown.prop("disabled", true);
+                if (!didSort) {
+                    maybeShowSortButtons();
+                } else {
+                    $("[data-dnd-url] .sortable-up, [data-dnd-url] .sortable-down").addClass("sr-only");
```
wiffbi, line 46: I am not sure, we want to hide the sort buttons again once one has sorted by drag-n-drop. On the other hand, there is no way to hide the sort-buttons again thanks to sessionStorage. So I guess we leave it like that?

src/pretix/static/pretixcontrol/js/ui/dragndroplist.js
```diff
@@ -1,48 +1,57 @@
...
-                if (disabledUp.length && disabledUp.get(0) !== firstUp.get(0)) {
-                    disabledUp.prop("disabled", false);
-                    firstUp.prop("disabled", true);
-                }
-
-                var disabledDown = container.find(".sortable-down:disabled"),
-                    lastDown = container.find(">tr[data-dnd-id] .sortable-down").last();
-                if (disabledDown.length && disabledDown.get(0) !== lastDown.get(0)) {
-                    disabledDown.prop("disabled", false);
-                    lastDown.prop("disabled", true);
```
wiffbi, line 39: Why did you remove the disabled property changing? I know it is an edge case, but when I dragged once, then failed and accidentally double clicked so that the up/down-buttons are shown, it looks a bit weird that a sorting button in the middle is disabled.

wiffbi (commented):
Looks good. Just two questions regarding showing/hiding the sort buttons and disabling the buttons.

src/pretix/static/pretixcontrol/js/ui/dragndroplist.js
luelista: naja die idee wäre wenn du doch drag-und-drop bedienen kannst, brauchst du wahrscheinlich diese buttons doch nicht... und das mit dem sessionStorage finde ich schon sinnvoll weil du sie sonst nach jeder einzelnen sortieraktion wieder neu einblenden musst

wiffbi (approved):

PR #4216, opened 2024-06-11
Stripe: Support for TWINT

raphaelm:


src/pretix/plugins/stripe/payment.py
```diff
@@ -1938,3 +1947,25 @@ def _payment_intent_kwargs(self, request, payment):
...
+    public_name = 'TWINT'
+    method = 'twint'
+    confirmation_method = 'automatic'
+    explanation = _(
+        'This payment method is available to users of the Swiss app TWINT. Please have your app '
+        'ready.'
+    )
+
+    def is_allowed(self, request: HttpRequest, total: Decimal=None) -> bool:
+        return super().is_allowed(request, total) and request.event.currency == "CHF" and total < Decimal("5000.00")
```
wiffbi, line 1964: According to stripe docs maximum amount is 5.000 – so I read it as included.
```suggestion
        return super().is_allowed(request, total) and request.event.currency == "CHF" and total <= Decimal("5000.00")
```

wiffbi (approved):
Looks good, just a tiny thing – I think the 5000 CHF is up to and including.

PR #4235, opened 2024-06-17
LogEntryType registry

luelista:
Move display of LogEntry details from the `logentry_display` and `logentry_object_link` signals to a class hierarchy based approach. For each action_type, an instance of a subclass of `LogEntryType` is registered in the `log_entry_types` registry. Analogous to EventPluginSignal, this registry is an `EventPluginRegistry`, so it keeps track of the plugin the LogEntryType is defined in.

This PR is a first step towards making GDPR-compliant data shredding in log entries more reliable, followed by #4714 and further changes.


luelista:
image

src/pretix/base/models/log.py
```diff
@@ -261,3 +202,134 @@ def bulk_postprocess(cls, objects):
         to_wh = [o.id for o in objects if o.webhook_type]
         if to_wh:
             notify_webhooks.apply_async(args=(to_wh,))
+
+
+class LogEntryTypeRegistry(EventPluginRegistry):
```
raphaelm, line 207: This should, like email placeholders etc, probably get its own module instead of living here with the models, since it's… a lot

src/pretix/base/models/log.py
```diff
@@ -261,3 +202,134 @@ def bulk_postprocess(cls, objects):
...
+                self.register(clz(action_type=action_type, plain=plain))
+        return reg
+
+
+log_entry_types = LogEntryTypeRegistry({'action_type': lambda o: getattr(o, 'action_type')})
+
+
+class LogEntryType:
+    def __init__(self, action_type=None, plain=None):
+        assert self.__module__ != LogEntryType.__module__  # must not instantiate base classes, only derived ones
```
raphaelm, line 220: wouldn't `type(self) != LogEntryType` be more correct here?

src/pretix/control/logdisplay.py
```diff
@@ -620,95 +357,479 @@ def pretixcontrol_logentry_display(sender: Event, logentry: LogEntry, **kwargs):
...
+    elif block_name.startswith('api:'):
+        return _('Blocked because of an API integration')
 
-    if logentry.action_type == 'pretix.team.member.added':
-        return _('{user} has been added to the team.').format(user=data.get('email'))
 
-    if logentry.action_type == 'pretix.team.member.removed':
-        return _('{user} has been removed from the team.').format(user=data.get('email'))
+@log_entry_types.new_from_dict({
+    'pretix.event.order.modified': _('The order details have been changed.'),
```
raphaelm, line 386: Doesn't this ignore that many of these will need custom shredders?

src/pretix/plugins/sendmail/signals.py
```diff
@@ -116,21 +119,28 @@ def control_nav_import(sender, request=None, **kwargs):
...
-        'pretix.plugins.sendmail.rule.order.email.sent': _('A scheduled email was sent to the order'),
-        'pretix.plugins.sendmail.rule.order.position.email.sent': _('A scheduled email was sent to a ticket holder'),
-        'pretix.plugins.sendmail.rule.deleted': _('An email rule was deleted'),
-    }
-    if logentry.action_type in plains:
-        return plains[logentry.action_type]
+@log_entry_types.new('pretix.plugins.sendmail.sent', _('Mass email was sent to customers or attendees.'))
+@log_entry_types.new('pretix.plugins.sendmail.sent.waitinglist', _('Mass email was sent to waiting list entries.'))
+class SendmailPluginLogEntryType(EventLogEntryType):
+    pass
```
raphaelm, line 125: I wonder if we even should use something like

```suggestion
log_entry_types.new('pretix.plugins.sendmail.sent', _('Mass email was sent to customers or attendees.'))(EventLogEntryType)
log_entry_types.new('pretix.plugins.sendmail.sent.waitinglist', _('Mass email was sent to waiting list entries.'))(EventLogEntryType)
```

explicitly here instead of creating many empty classes, what do you think?

src/pretix/base/signals.py
```diff
@@ -202,6 +219,37 @@ def connect(self, receiver, sender=None, weak=True, dispatch_uid=None):
         super().connect(receiver, sender=None, weak=True, dispatch_uid=None)
 
 
+class Registry:
```
raphaelm, line 222: This can certainly use some code documentation with a simple example in a docstring or soemthing like that to quicker understand the terminology used here around keys, accessors, items, …

src/pretix/base/models/log.py
luelista: no, because I want all the "base types" (which are all types declared in this file) to be "abstract" / not directly instantiable. the main reason for that is that the __module__ is later used for checking in which plugin a LogEntryType is defined, so we can [mark the disabled ones separately](https://github.com/pretix/pretix/pull/4235/files/4e2799ff6570e2bb77920a1a54161293d642791b#diff-2bd781e18aa91b7ca1ab4364813c7938c72728986aec122af0ed13bc9f1a70feR62). I'll add that as a comment, because I had to think a bit to remember the reason myself :sweat_smile:

src/pretix/plugins/sendmail/signals.py
luelista: No, we need the empty classes here because this way they are declared in the correct package, so the registry can detect which plugin the LogEntryTypes come from. If you have a better solution for that though, I'd be very open, because I'm not a huge fan of the empty classes myself, but I couldn't think of a better way so far

src/pretix/base/signals.py
```diff
@@ -202,6 +219,115 @@ def connect(self, receiver, sender=None, weak=True, dispatch_uid=None):
...
+
+    def get(self, **kwargs):
+        (key, value), = kwargs.items()
+        return self.by_key.get(key).get(value, (None, None))
+
+    def filter(self, **kwargs):
+        return ((entry, meta)
+                for entry, meta in self.registered_entries
+                if all(value == meta[key] for key, value in kwargs.items())
+                )
```
raphaelm, line 312:
```suggestion
        return (
            (entry, meta)
            for entry, meta in self.registered_entries
            if all(value == meta[key] for key, value in kwargs.items())
        )
```

src/pretix/base/signals.py
```diff
@@ -202,6 +219,115 @@ def connect(self, receiver, sender=None, weak=True, dispatch_uid=None):
...
+        return reg
+
+    def get(self, **kwargs):
+        (key, value), = kwargs.items()
+        return self.by_key.get(key).get(value, (None, None))
+
+    def filter(self, **kwargs):
+        return ((entry, meta)
+                for entry, meta in self.registered_entries
+                if all(value == meta[key] for key, value in kwargs.items())
```
raphaelm, line 311: This will crash if key is not contained in meta for any registered item, is that what we want?

luelista: meta is supposed to contain the same keys for all items, as it's generated using the accessor functions provided to the constructor, so this should not cause any exceptions. I could change it got meta.get(key) but it shouldn't make any difference.

doc/development/api/plugins.rst
```diff
@@ -121,6 +121,7 @@ This will automatically make pretix discover this plugin as soon as it is instal
 through ``pip``. During development, you can just run ``python setup.py develop`` inside
 your plugin source directory to make it discoverable.
 
+.. _`Signals`:
```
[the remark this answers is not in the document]
luelista: :+1:

```suggestion
.. _`signals`:
```

src/pretix/base/logentrytypes.py
```diff
@@ -0,0 +1,238 @@
...
+                # ...
+
+        :param data: action types and descriptions
+                     ``{"some_action_type": "Plain text description", ...}``
+        """
+        def reg(clz):
+            for action_type, plain in data.items():
+                self.register(clz(action_type=action_type, plain=plain))
+            return clz
+        return reg
```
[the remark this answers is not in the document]
luelista: Should not be necessary as we return `clz` completely unchanged, and don't wrap anything (we just add create an instance of it).

src/pretix/base/logentrytypes.py
```diff
@@ -0,0 +1,238 @@
...
+log_entry_types = LogEntryTypeRegistry({'action_type': lambda o: getattr(o, 'action_type')})
+
+
+class LogEntryType:
+    """
+    Base class for a type of LogEntry, identified by its action_type.
+    """
+
+    def __init__(self, action_type=None, plain=None):
+        assert self.__module__ != LogEntryType.__module__  # must not instantiate base classes, only derived ones
```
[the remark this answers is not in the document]
luelista: Good to know! I'd say it's safe enough to assume that if this assert fails, it will already be triggered during debugging or test execution. As LogEntryTypes are typically instanciated only once at application startup, it's unlikely this only fails in a specific situation on production. 

But it also wouldn't hurt to change it I guess?
```suggestion
        if self.__module__ == LogEntryType.__module__
            raise Exception("LogEntryType: must not instantiate base classes, only derived ones")
```

src/pretix/base/logentrytypes.py
```diff
@@ -0,0 +1,238 @@
...
+        """
+        To be used for shredding personally identified information contained in the data field of a LogEntry of this
+        type.
+        """
+        raise NotImplementedError
+
+
+class EventLogEntryType(LogEntryType):
+    """
+    Base class for any LogEntry type whose content_object is either an `Event` itself or belongs to a specific `Event`.
```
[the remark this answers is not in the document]
luelista:
```suggestion
    Base class for any `LogEntry` type whose `content_object` is either an `Event` itself or belongs to a specific `Event`.
```

src/pretix/base/signals.py
```diff
@@ -202,6 +219,115 @@ def connect(self, receiver, sender=None, weak=True, dispatch_uid=None):
...
+            @some_registry.register
+            def my_new_entry(foo):
+              # ...
+        """
+        for obj in objs:
+            meta = {k: accessor(obj) for k, accessor in self.keys.items()}
+            tup = (obj, meta)
+            for key, value in meta.items():
+                self.by_key[key][value] = tup
+            self.registered_entries.append(tup)
```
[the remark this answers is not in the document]
luelista: Yes, it should
```suggestion
            self.registered_entries.append(tup)
        if len(objs) == 1:
            return objs[0]
```

src/pretix/base/signals.py
```diff
@@ -202,6 +219,115 @@ def connect(self, receiver, sender=None, weak=True, dispatch_uid=None):
...
+            @animal_sound_registry.new("woof")
+            class AnimalSound:
+              def __init__(self, sound):
+                # ...
+        """
+        def reg(clz):
+            obj = clz(*args, **kwargs)
+            self.register(obj)
+            return clz
+        return reg
```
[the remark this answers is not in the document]
luelista: I'd say we don't

doc/development/api/plugins.rst
```diff
@@ -153,6 +154,28 @@ in the ``installed`` method:
 Note that ``installed`` will *not* be called if the plugin is indirectly activated for an event
 because the event is created with settings copied from another event.
 
+.. _`Registries`:
```
luelista, line 157:
```suggestion
.. _`registries`:
```

doc/development/implementation/logging.rst
```diff
@@ -72,24 +73,87 @@ following ready-to-include template::
...
+                'PAYMENT.SALE.DENIED': _('Payment denied.'),
+                # ...
+            }.get(event_type, f"({event_type})")
+            return _('PayPal reported an event: {}').format(text)
+
+.. automethod:: pretix.base.logentrytypes.LogEntryType.display
+
+If your new model object does not belong to an `Event`, you need to implement
+
+meow
```
luelista, line 150:
```suggestion
If your new model object does not belong to an `Event`, you need to inherit directly from ``LogEntryType`` instead
of ``EventLogEntryType``, providing your own implementation of ``get_object_link_info`` if object links should be
displayed.

.. autoclass:: pretix.base.logentrytypes.LogEntryType
   :members: get_object_link_info

```

doc/development/implementation/logging.rst
```diff
@@ -72,24 +73,87 @@ following ready-to-include template::
...
+
+.. autoclass:: pretix.base.logentrytypes.Registry
+   :members: new
+
+.. autoclass:: pretix.base.logentrytypes.LogEntryTypeRegistry
+   :members: new, new_from_dict
```
luelista, line 156:
```suggestion
```

doc/development/api/plugins.rst
```diff
@@ -153,6 +154,28 @@ in the ``installed`` method:
...
+In a ``Registry``, classes are collected at application startup, along with a unique key (in case
+of LogEntryType, the ``action_type``) as well as which plugin registered them.
+
+To register a class, you can use one of several decorator provided by the Registry object:
+
+.. code-block:: python
+
+   @log_entry_types.new('my_pretix_plugin.some.action', _('Some action in My Pretix Plugin occured.'))
+   class MyPretixPluginLogEntryType(EventLogEntryType):
+       pass
```
luelista, line 174:
```suggestion
To register a class, you can use one of several decorators provided by the Registry object:

.. autoclass:: pretix.base.logentrytypes.LogEntryTypeRegistry
   :members: register, new, new_from_dict
```

doc/development/api/plugins.rst
```diff
@@ -153,6 +154,28 @@ in the ``installed`` method:
...
+In a ``Registry``, classes are collected at application startup, along with a unique key (in case
+of LogEntryType, the action_type) as well as which plugin registered them.
+
+To register a class, you can use one of several decorator provided by the Registry object:
+
+.. code-block:: python
+
+   @log_entry_types.new('my_pretix_plugin.some.action', _('Some action in My Pretix Plugin occured.'))
+   class MyPretixPluginLogEntryType(EventLogEntryType):
+       pass
```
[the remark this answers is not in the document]
luelista: I agree... I changed this to just include the code documentation for the decorators, instead of providing example code directly, because I couldn't think of a simpler example

src/pretix/base/signals.py
```diff
@@ -202,6 +219,115 @@ def connect(self, receiver, sender=None, weak=True, dispatch_uid=None):
...
+        """
+        :param keys: dictionary {key: accessor_function}
+                     When a new entry is registered, all accessor functions are called with the new entry as parameter.
+                     Their return value is stored as the metadata value for that key.
+        """
+        self.registered_entries = list()
+        self.keys = keys
+        self.by_key = {key: {} for key in self.keys.keys()}
+
+    def register(self, *objs):
```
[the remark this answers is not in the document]
luelista: I'm pretty sure it's not thread-safe, but I don't think it needs to be - as far as I understand, imports are at least locked per imported module, so any given module will only be imported once?
If it were imported twice, thread-safety wouldn't really help, as two distinct class objects would be constructed and registered, which we couldn't really detect here.

However, currently this is not even safe against register being called twice with the same object. I'll switch self.registered_entries to a `set`, to protect against that. Not sure yet if I want to ignore or throw in that case, though.

src/pretix/base/logentrytypes.py
```diff
@@ -0,0 +1,252 @@
...
+        """
+        Returns the message to be displayed for a given logentry of this type.
+
+        :return: `str` or `LazyI18nString`
+        """
+        if hasattr(self, 'plain'):
+            plain = str(self.plain)
+            if '{' in plain:
+                data = defaultdict(lambda: '?', logentry.parsed_data)
+                return plain.format_map(data)
```
wiffbi, line 126: Should we catch any formatting errors? `plain` is coming from translation, so might contain a typo or somehow else contain a misconfigured or mismatching {}. Otherwise it might fail hard. As a fallback just return plain?

src/pretix/base/logentrytypes.py
```diff
@@ -0,0 +1,252 @@
...
+        return str(content_object)
+
+
+class OrderLogEntryType(EventLogEntryType):
+    object_link_wrapper = _('Order {val}')
+    object_link_viewname = 'control:event.order'
+    object_link_argname = 'code'
+
+    def object_link_argvalue(self, order):
+        return order.code
```
wiffbi, line 186: What if we need more than one additional argument for the URL? Probably currently not the case anywhere, but wouldn't it be more flexible to have a `object_link_args`, which returns a dict that gets merged with the other URL-args?

```suggestion

    def object_link_args(self, order):
        return {'code': order.code}
```

Maybe for the default case, where the `logentry.content_object` is the only additional URL-param, we could keep the `object_link_argname` and change the original `object_link_argvalue` to

```
    def object_link_args(self, content_object):
        """Return the identifier used in a link to content_object."""
        if self.object_link_argname:
            return {self.object_link_argname: content_object.id}
        return {}
```

src/pretix/base/logentrytypes.py
```diff
@@ -0,0 +1,252 @@
...
+        raise NotImplementedError
+
+
+class EventLogEntryType(LogEntryType):
+    """
+    Base class for any `LogEntry` type whose `content_object` is either an `Event` itself or belongs to a specific `Event`.
+    """
+
+    def get_object_link_info(self, logentry) -> dict:
+        if hasattr(self, 'object_link_viewname') and hasattr(self, 'object_link_argname') and logentry.content_object:
```
wiffbi, line 161: See below for using multiple args to the URL:

```suggestion
        if hasattr(self, 'object_link_viewname') and logentry.content_object:
```

src/pretix/base/logentrytypes.py
```diff
@@ -0,0 +1,252 @@
...
+            return {
+                'href': reverse(self.object_link_viewname, kwargs={
+                    'event': logentry.event.slug,
+                    'organizer': logentry.event.organizer.slug,
+                    self.object_link_argname: self.object_link_argvalue(logentry.content_object),
```
wiffbi, line 166: See below regarding multiple URL-params …
```suggestion
                    **self.object_link_args(logentry.content_object),
```

wiffbi (approved):
Looks good! Probably not needed, but would it make sense to support multiple additional URL-params for the `self.object_link_viewname`? Probably makes it more complex, so feel free to ignore.

Otherwise just an idea to maybe handle an exception, if the plain-message format-string is malformed (as it is coming from translation).

PR #4111, opened 2024-04-24
New implementation of sales channels

raphaelm:
core:
- [x] new model
- [x] data migration
  - [x] test migration of discounts/items again when they are not on all channgels
  - [x] in-place migration plan for pretix hosted
- [x] auto generate for organizers
- [x] ordering of sales channels
- [x] explicit dependency of sales channels on event-level plugins
- [x] organizer ui to create/manage channels
  - [x] tests
- [x] api
- [x] api docs
- [x] api backwards compatibility
- [x] api to create sales channels
  - [x] tests
- [x] payment providers
- [x] invoice generate setting
- [x] mail setting
- [x] exporters
- [x] auto-checkin
- [x] allow_blocked_seats_for_channel
- [x] cart services
- [x] order services
- [x] apply discounts
- [x] cloning events across organizers
  - [x] ticketoutputpdf
- [x] control forms
- [x] control templates
- [x] control views
- [x] statistics
- [x] stripe
- [x] ticketoutputpdf
- [x] ticketoutputpdf API
- [x] presale
- [x] tests

Open questions:

- [ ] does it make sense that `label` is an i18nstring or should it me monolingual since I don't see it being used externally?

Plugins that need changes:

- [x] pretix-c-accounting-tools
- [x] pretix-c-bre…
- [x] pretix-c-hei…
- [x] pretix-c-mb
- [x] pretix-doistep
- [x] pretix-downstream-events
- [x] pretix-l…merchant
- [x] pretix-mobilepdf
- [x] pretix-offlinesales
- [x] pretix-posbackend
- [x] pretix-presale-saml
- [x] pretix-printable
- [x] pretix-reports
- [x] pretix-resellers
- [x] pretix-seating
- [x] pretix-shipping
- [x] pretix-superbundles
- [x] pretix-ticketswap
- [x] pretix-vacc-utils
- [x] pretix-vr-ticket
- [x] billing
- [x] privacy
- [x] pretixeu organizer setup

PR #4236, opened 2024-06-17
CartManager: Allow to specify an explicit expiry time

pc-coholic:
This PR should only be merged after #4111, since it's already leveraging the new signature of `CartManager` introduced in said PR.


raphaelm (approved):

pc-coholic:
Cherrypicked to main

PR #4237, opened 2024-06-17
item_forms: Allow signals to return None

pc-coholic:


src/pretix/control/views/item.py
```diff
@@ -1326,7 +1326,8 @@ def plugin_forms(self):
             if isinstance(resp, (list, tuple)):
```
wiffbi, line 1326: Wouldn’t it be easier to just
```suggestion
            if not resp:
                continue
            if isinstance(resp, (list, tuple)):
```

src/pretix/control/views/item.py
```diff
@@ -1326,7 +1326,8 @@ def plugin_forms(self):
...
                 forms.extend(resp)
             else:
-                forms.append(resp)
+                if resp:
+                    forms.append(resp)
```
wiffbi, line 1330: See above. Exit early, then decide what to do?
```suggestion
                forms.append(resp)
```

wiffbi (approved):
Just a suggestion with a different coding style regarind exiting/returning/continueing early.

PR #4255, opened 2024-06-24
Refactor FieldRenderers

luelista:


src/pretix/control/forms/renderers.py
```diff
@@ -21,90 +21,16 @@
...
-        # Empty label, e.g. checkbox
-        attrs.setdefault('class', '')
-        attrs['class'] += ' label-empty'
-
-    builder = '<{tag}{attrs}>{content}{opt}</{tag}>'
-    return format_html(
-        builder,
-        tag='label',
-        attrs=mark_safe(flatatt(attrs)) if attrs else '',
-        opt=mark_safe('<br><span class="optional">{}</span>'.format(pgettext('form', 'Optional'))) if optional else '',
```
wiffbi, line 56: We need to find a way to overwrite how fields are marked as optional in presale vs. control. You removed the text „Optional“ from below a label in control.

wiffbi (changes requested):
I have not teste thoroughly, but this PR removes the additional label „Optional“ in pretix-control formfield-labels. We probably need a way to overwrite parts of render_label in the subclasses?

PR #4269, opened 2024-06-27
Fix missing CSS in seatingframe

wiffbi:
The change from precompiled CSS to Custom Properties broke seatingframe as no precompiled CSS is now loaded there. This PR adds the results from presale.signals.html_head to its HTML-head. This adds the needed CSS-files for seating to be displayed correctly.

CAVEAT: I am not sure though, what the implications are when using additional plugins as presale.signals.html_head does totally different things than just adding CSS – I looked through my plugins locally and they either filter for specific pages when adding stuff or „just“ add tracking javascript.


wiffbi:
Further testing revealed that jQuery is not loaded on the seatingframe page which causes at least pretix_tracking to trigger an error. Hm, should we load jQuery? Or how do we add the CSS while keeping that page stripped down?

PR #4273, opened 2024-06-29
Easier PCI DSS compliance for payment pages

raphaelm:


src/tests/presale/test_checkout.py
```diff
@@ -151,6 +151,28 @@ def _enable_country_specific_taxing(self):
...
+            'company': 'Foo',
+            'name': 'Bar',
+            'street': 'Baz',
+            'zipcode': '1234',
+            'city': 'Here',
+            'country': 'AT',
+            'email': 'admin@localhost'
+        }, follow=True)
+        assert b'TRACKING SCRIPT' not in payment_r.content
+
```
wiffbi, line 175: Should we test all possible payment pages/URLs listed under „Zahlungsseiten“ on our PCI-DSS-page to make sure that all of them correctly set the `request.pci_dss_payment_page = true`?

wiffbi (approved):
Looks good! I am still thinking about how to not depend on the plugin to do the right thing when embedding scripts through html_head … but I can not come up with anything better.

src/tests/presale/test_checkout.py
raphaelm: PCI rule 11.6.1 will require us to build a "monitoring" for the scripts on all payment pages on pretix.eu anyways, so I think we can skip that.

raphaelm:
I've added a note to the plugin signal

PR #4277, opened 2024-07-01
Add test for no pending migrations

raphaelm:
Idea and reasoning: https://adamj.eu/tech/2024/06/23/django-test-pending-migrations/

### DIFF
--- a/src/tests/test_no_missing_migrations.py
+++ b/src/tests/test_no_missing_migrations.py
@@ -1,0 +1,18 @@
+from io import StringIO
+
+import pytest
+from django.core.management import call_command
+
+
+@pytest.mark.django_db
+def test_no_pending_migrations():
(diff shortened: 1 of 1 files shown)


wiffbi (commented):
I am unsure, how this is supposed to work. When I change e.g. the order-model’s email-field to `null=False` or even add a new field to the order model, I would expect this test to fail. But at least in my local setup, it does not fail. :( Runnind the underlying command, I get this as output:

```
Migrations for 'pretixbase':
  pretix/base/migrations/0269_order_email2_alter_order_email.py
    - Add field email2 to order
    - Alter field email on order
```

So I think, it does not return an error-code and thus does not raise the SystemExit?

raphaelm:
Dann lassen wir das erstmal, dann ist es die energie nicht wert.

PR #4278, opened 2024-07-01
Drop nullability on Order.organizer and OrderPosition.organizer

raphaelm:
Followup to make #3988 effective. This is already in place (through manual SQL) on pretix hosted, this PR brings the same to pretix Community.

Downside: The migration would fail for anyone who currently has non-unique order codes or position secrets in their database. How can we give good guidance for those to proceed with the update?


src/pretix/base/migrations/0269_alter_order_organizer_alter_orderposition_organizer.py
```diff
@@ -0,0 +1,54 @@
...
+            "pretixbase",
+            "0268_remove_subevent_items_remove_subevent_variations_and_more",
+        ),
+    ]
+
+    operations = [
+        migrations.RunSQL(
+            "UPDATE pretixbase_order "
+            "SET organizer_id = (SELECT e.organizer_id FROM pretixbase_event e WHERE e.id = pretixbase_order.event_id) "
+            "WHERE pretixbase_order.id IN (SELECT id FROM pretixbase_order o2 WHERE o2.organizer_id IS NULL);"
```
wiffbi, line 20: I am not too firm in databases – what is the reasoning behind this subquery instead of just `WHERE organizer_id IS NULL`?

src/pretix/base/migrations/0269_alter_order_organizer_alter_orderposition_organizer.py
```diff
@@ -0,0 +1,54 @@
...
+        ),
+        migrations.AlterField(
+            model_name="order",
+            name="organizer",
+            field=models.ForeignKey(
+                on_delete=django.db.models.deletion.CASCADE,
+                related_name="orders",
+                to="pretixbase.organizer",
+            ),
+        ),
```
wiffbi, line 44: Isn’t this just the same as the AlterField above?
```suggestion
```

wiffbi (commented):
Looks good.

Regarding your question of guidance: when would this migration fail? If understand the migration/AddConstraint in #3988 correctly, that migration would have already failed if an order-code/secret existed twice (over the whole database) as both of them would have gotten an organizer=null – did we give any guidance on this? This migration here would only fail, if an organizer previously had one secret or order-code where the organizer is null and now after #3988 added the same order-code/secret but with the organizer-column not being null. Now when back-filling organizer_id it would fail. That to me seems less likely than #3988 failing. Did we give any guidance there? Maybe we can get away with just a hint at installation? Or can we output something to stdout when running `RunPython` to check whether such error-conditions apply? The question though is still, what do we suggest, that the admin does? What if one of those offending orders/orderpositions is cancelled? Could we automatically remove/regenerate its code/secret? Hm, not really ideal either. Tricky, I don’t know what best to suggest to do here.

src/pretix/base/migrations/0269_alter_order_organizer_alter_orderposition_organizer.py
raphaelm: Nothing, bad takeover from our internal migration (where the subquery had a LIMIT) on my part, will fix

src/pretix/base/migrations/0302_resolve_duplicate_codes_and_secrets.py
```diff
@@ -0,0 +1,58 @@
...
+    # sysadmins manually.
+    OrderPosition = apps.get_model("pretixbase", "OrderPosition")
+    Order = apps.get_model("pretixbase", "Order")
+
+    qs = OrderPosition.all.values("secret", "order__event__organizer_id").order_by().annotate(c=Count("*")).filter(c__gt=1)
+    for row in qs:
+        affected = OrderPosition.all.filter(
+            **{k: v for k, v in row.items() if k != "c"}
+        ).order_by("pk")
+        logger.error(f"Found {row['c']} tickets with with the same secret \"{row["secret"]}\" in organizer {row['order__event__organizer_id']}, all except one will be changed")
```
pc-coholic:
```suggestion
        logger.error(f"Found {row['c']} tickets with with the same secret \"{row['secret']}\" in organizer {row['order__event__organizer_id']}, all except one will be changed")
```

src/pretix/base/migrations/0302_resolve_duplicate_codes_and_secrets.py
```diff
@@ -0,0 +1,58 @@
...
+                    f"Ticket {a.pk} has new secret {a.secret}"
+                )
+            a.save(update_fields=["organizer_id", "secret"])
+
+    qs = Order.objects.values("code", "event__organizer_id").order_by().annotate(c=Count("*")).filter(c__gt=1)
+    for row in qs:
+        affected = Order.objects.filter(
+            **{k: v for k, v in row.items() if k != "c"}
+        ).order_by("pk")
+        logger.error(f"Found {row['c']} orders with with the same code \"{row["code"]}\" in organizer {row['event__organizer_id']}, all except one will be changed")
```
pc-coholic:
```suggestion
        logger.error(f"Found {row['c']} orders with with the same code \"{row['code']}\" in organizer {row['event__organizer_id']}, all except one will be changed")
```

pc-coholic (changes requested):
Needs a small change for the f-string, but apart from that: :shipit:

raphaelm:
Check if you are affected:

```
select e.organizer_id, o.code, count(*)
from pretixbase_order o
left join pretixbase_event e on e.id = o.event_id
group by e.organizer_id, o.code
having count(*) > 1;

select e.organizer_id, op.secret, count(*)
from pretixbase_orderposition op
left join pretixbase_order o on o.id = op.order_id
left join pretixbase_event e on e.id = o.event_id
group by e.organizer_id, op.secret
having count(*) > 1;
```

PR #4282, opened 2024-07-02
API: Add details of seats

raphaelm:


wiffbi (approved):
Looks good.

Should we at some point update the VoucherSerializer to have a InlineSeatSerializer as well instead of just the seat_guid? Would be a breaking change though.

raphaelm:
Yeah, breaking changes are hard :(

PR #4285, opened 2024-07-03
API: Add various filtering options (Z#23158339)

raphaelm:


doc/api/resources/item_variations.rst
```diff
@@ -164,6 +164,7 @@ Endpoints
       }
 
    :query integer page: The page number in case of a multi-page result set, default is 1
+   :query boolean search: Filter the list by the value of the variation (substring search).
```
wiffbi, line 167:
```suggestion
   :query string search: Filter the list by the value of the variation (substring search).
```

doc/api/resources/items.rst
```diff
@@ -392,6 +392,7 @@ Endpoints
       }
 
    :query integer page: The page number in case of a multi-page result set, default is 1
+   :query boolean search: Filter the list by internal name or name of the item (substring search).
```
wiffbi, line 395:
```suggestion
   :query string search: Filter the list by internal name or name of the item (substring search).
```

src/pretix/api/views/order.py
```diff
@@ -138,6 +141,16 @@ def subevent_after_qs(self, qs, name, value):
...
 
+        def provider_qs(self, qs, name, value):
+            qs = qs.annotate(
+                has_payment_with_provider=Exists(
+                    OrderPayment.objects.filter(
+                        Q(order=OuterRef('pk')) & Q(provider=value)
+                    )
+                )
+            ).filter(has_payment_with_provider=True)
+            return qs
```
wiffbi, line 152: That looks a bit complicated to me as we anyway filter for true, ie. only existing payments. So for me it is easier to read:

```suggestion
            return qs.filter(Exists(
                OrderPayment.objects.filter(order=OuterRef('pk'), provider=value)
            ))
```

wiffbi (changes requested):
Looks good, I will commit and merge my suggested changes myself (two typos) and one simplification in qs-filtering.

PR #4300, opened 2024-07-11
New column and sort options for check-in-list export  (Z#23158159)

luelista:
- [x] Allow sorting by order datetime
- [x] Add product and order check-in text to export

image

### DIFF
--- a/src/pretix/plugins/checkinlists/exporters.py
+++ b/src/pretix/plugins/checkinlists/exporters.py
@@ -119,7 +119,7 @@ class CheckInListMixin(BaseExporter):
                      choices=[
                          ('name', _('Attendee name')),
                          ('code', _('Order code')),
-                         ('datetime', _('Order date')),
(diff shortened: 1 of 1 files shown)


wiffbi (approved):
Looks good! Maybe change the naming of the sorting-option, but probably not needed.

src/pretix/plugins/checkinlists/exporters.py
```diff
@@ -229,6 +230,8 @@ def _get_queryset(self, cl, form_data, prefetch=True):
             )
         elif sort == 'code':
             qs = qs.order_by(*o, 'order__code')
+        elif sort == 'order_datetime':
+            qs = qs.order_by(*o, 'order__datetime')
```
wiffbi, line 234: Just read the client’s request, that they want the last orders at the top. So 

```suggestion
            qs = qs.order_by(*o, '-order__datetime')
```

PR #4303, opened 2024-07-15
Auto-fill questions from invoice address (Z#23143497)

wiffbi:
This PR adds auto-filling questions from invoice address fields by matching id-suffixes, placeholder-attributes and (visible) labels to placeholder-attributes.

It removes the code to alyways update all questions when one of the answers changes to just updating the question matched to the one answer, that was changed.

### DIFF
--- a/src/pretix/static/pretixpresale/js/ui/main.js
+++ b/src/pretix/static/pretixpresale/js/ui/main.js
@@ -380,21 +380,14 @@ $(function () {
         " #id_city, #id_country, #id_state").change(function () {
         if (copy_to_first_ticket) {
             var $first_ticket_form = $(".questions-form").first().find("[data-addonidx=0]");
-            $first_ticket_form.find("input[id*=attendee_email]").val($("#id_email").val());
-            $first_ticket_form.find("input[id$=company]").val($("#id_company").val());
-            $first_ticket_form.find("textarea[id$=street]").val($("#id_street").val());
-            $first_ticket_form.find("input[id$=zipcode]").val($("#id_zipcode").val());
-            $first_ticket_form.find("input[id$=city]").val($("#id_city").val());
-
-            $first_ticket_form.find("select[id$=state]").val($("#id_state").val());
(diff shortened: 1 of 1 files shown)


src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -380,21 +380,16 @@ $(function () {
...
-            $first_ticket_form.find("select[id$=state]").val($("#id_state").val());
-            if ($first_ticket_form.find("select[id$=country]").val() !== $("#id_country").val()) {
-                $first_ticket_form.find("select[id$=country]").val($("#id_country").val()).trigger('change');
+            $first_ticket_form.find("[id$=" + this.id.substring(3) + "]").val(this.value);
+            if (this.placeholder) {
+                $first_ticket_form.find("[placeholder='" + this.placeholder + "']").val(this.value);
+            }
+            var label = $("label[for=" + this.id +"]").first().contents().filter(function () {
+                return this.nodeType != Node.ELEMENT_NODE || !this.classList.contains("sr-only");
+            }).text();
```
raphaelm, line 389: Have you tested the PR with a country question (which I belive is the most important for our customer requesting this)?

This did not work for me for multiple reasons:

- The `var label = …` needs a `.trim()`, otherwise the label is deemed to be `" Land"` in my browser
- The lookup down below does not work due to `<select>` not having a placeholder

I think we actually need to match label to label here instead of label to placeholder?

raphaelm (changes requested):

src/pretix/static/pretixpresale/js/ui/main.js
wiffbi: Sorry, I tested with input and textarea for address, not a country-dropdown. Sorry! I will look into this!

wiffbi:
I added a `trigger("change")` so initial values of the invoice form (especially country) are prefilled.

raphaelm (approved):

PR #4309, opened 2024-07-16
Do not take "optional" as part of error message label

raphaelm:
Before:
image

After:

image


wiffbi (approved):
I probably never encountered an error for an optional field when developing this, so sorry for not doing this properly in the first place. Also, I probably should take your approach to [my auto-fill PR](https://github.com/pretix/pretix/pull/4303/files#diff-e8598d047e821416e9d543b198cfcebe40f9a376a0deccfde0a3e2bc65bb9415R387) to make it more robust – I basically took the easy way out and just took the firstChild.textcontent, which with the current HTML is enough.

PR #4310, opened 2024-07-16
Add webhook for pretix.event.order.deleted (Z#23159520)

raphaelm:


wiffbi (approved):
For a second I thought, why is this not a `ParametrizedOrderWebhookEvent`? But of course, once the order is deleted, there is no order.

PR #4312, opened 2024-07-17
API: add api_meta to order

wiffbi:
This PR adds api_meta_info to the order-object. This meta-info is only available through the API. A new signal for plugins to write to this meta-object on order-creation is added as well.


doc/api/resources/orders.rst
```diff
@@ -742,6 +744,8 @@ Updating order fields
 
    * ``comment``
 
+   * ``api_meta_info`` (you always need to supply the full object, or ``null`` to delete the current value)
```
raphaelm, line 747:
```suggestion
   * ``api_meta`` (you always need to supply the full object, or ``{}`` to delete the current value)
```

src/pretix/base/models/orders.py
```diff
@@ -299,6 +299,11 @@ class Order(LockModel, LoggedModel):
         verbose_name=_("Meta information"),
         null=True, blank=True
     )
+    api_meta_info = models.JSONField(
+        verbose_name=_("API-Meta information"),
```
raphaelm, line 303:
```suggestion
        verbose_name=_("API meta information"),
```

src/pretix/base/models/orders.py
```diff
@@ -299,6 +299,11 @@ class Order(LockModel, LoggedModel):
...
+        null=True, blank=True,
```
raphaelm, line 304: Why should this be nullable? Do we need the semantic difference between NULL and {}?

```suggestion
        null=False, blank=True,
```

doc/api/resources/orders.rst
```diff
@@ -42,6 +42,7 @@ payment_date                          date                       **DEPRECATED AN
 payment_provider                      string                     **DEPRECATED AND INACCURATE** Payment provider used for this order
 total                                 money (string)             Total value of this order
 comment                               string                     Internal comment on this order
+api_meta_info                         object                     Meta-data for that order. Only available through API.
```
raphaelm, line 45: I'm not happy about the naming, see slack discussion

src/pretix/base/migrations/0269_order_api_meta_info.py
```diff
@@ -0,0 +1,18 @@
...
+
+    dependencies = [
+        ('pretixbase', '0268_remove_subevent_items_remove_subevent_variations_and_more'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='order',
+            name='api_meta_info',
+            field=models.JSONField(default=dict, null=True),
```
raphaelm, line 16:
```suggestion
            field=models.JSONField(default=dict, null=False),
```

doc/api/resources/orders.rst
```diff
@@ -742,6 +744,8 @@ Updating order fields
 
    * ``comment``
 
+   * ``api_meta`` (you always need to supply the full object, or ``null`` to delete the current value)
```
raphaelm, line 747:
```suggestion
   * ``api_meta``
```

doc/api/resources/orders.rst
```diff
@@ -42,6 +42,7 @@ payment_date                          date                       **DEPRECATED AN
 payment_provider                      string                     **DEPRECATED AND INACCURATE** Payment provider used for this order
 total                                 money (string)             Total value of this order
 comment                               string                     Internal comment on this order
+api_meta                              object                     Meta-data for that order. Only available through API.
```
raphaelm, line 45:
```suggestion
api_meta                              object                     Meta data for that order. Only available through API, no guarantees
                                                                 on the content structure. You can use this to save references to your system.
```

raphaelm (approved):

src/pretix/base/migrations/0269_order_api_meta_info.py
wiffbi: Fixed

PR #4335, opened 2024-07-25
adds an `order_info_top` signal for the control component

chrisdeely:
I'm working on a plugin to capture additional customer details and display them on the order summary page of the `control` component. The only currently suitable signal is `control.signals.order_info` which allows injection considerable far down the page.

`order_details_top` will permit plugins to insert relevant content to the top-level "Order details" section as needed.

Ideally, I think the Order Details section could be rewritten to move the logic outside of the template and instead have the signal collect a list of dictionaries with property titles and values to display. 

I'm very new to this project, django & python in general, so I don't know if that is advisable or not. I'm willing to take a look at it if that would be worthwhile.

### DIFF
--- a/src/pretix/control/signals.py
+++ b/src/pretix/control/signals.py
@@ -255,6 +255,15 @@ As with all plugin signals, the ``sender`` keyword argument will contain the eve
 Additionally, the argument ``order`` and ``request`` are available.
 """
 
+order_info_top = EventPluginSignal()
+"""
+Arguments: ``order``, ``request``
+
+This signal is sent out to display additional information on top of the order detail page ("Order details" section)
+
+As with all plugin signals, the ``sender`` keyword argument will contain the event.
+"""
+
 order_position_buttons = EventPluginSignal()
 """
 Arguments: ``order``, ``position``, ``request``
--- a/src/pretix/control/templates/pretixcontrol/order/index.html
+++ b/src/pretix/control/templates/pretixcontrol/order/index.html
@@ -177,6 +177,7 @@
                 </div>
                 <div class="panel-body">
                     <dl class="dl-horizontal">
+                        {% eventsignal event "pretix.control.signals.order_info_top" order=order request=request %}
                         <dt>{% trans "Order code" %}</dt>
                         <dd>{{ order.code }}</dd>
                         <dt>{% trans "Order date" %}</dt>


raphaelm:
Thanks for the proposal!

Is the information your plugin is adding really more important than the order code and date? For a plugin that collects some additional data, I'd rather put it between the phone number and invoices. I also think a name along the lines of `order_details_…` like your PR title is better than `order_info_…` like your code. Maybe even `order_contact_display` to have some relation to `contact_form_fields` on the other end to make it more symmetric.

chrisdeely:
> Is the information your plugin is adding really more important than the order code and date? For a plugin that collects some additional data, I'd rather put it between the phone number and invoices.

I suppose that is debatable, and highly dependent on context. In my personal case, yes, the data being added to the Order is far more interesting to the end user than an Order identifier. I can relocate the signal elsewhere if desired, but what are your thoughts on changing the template altogether to simply render a list of `details` and allow plugins to add/remove/reorder those details as needed?

> I also think a name along the lines of `order_details_…` like your PR title is better than `order_info_…` like your code. Maybe even `order_contact_display` to have some relation to `contact_form_fields` on the other end to make it more symmetric.

Apologies for the typo, I have synchronized the PR name with the actual signal name in the code for clarity. The signal name was copied directly from an analogous signal surfaced within the `presale` component, so I kept that consistent. I am open to any name you would prefer.

raphaelm:
> I suppose that is debatable, and highly dependent on context. In my personal case, yes, the data being added to the Order is far more interesting to the end user than an Order identifier. I can relocate the signal elsewhere if desired, but what are your thoughts on changing the template altogether to simply render a list of details and allow plugins to add/remove/reorder those details as needed?

Generally, I'd be open to that, if it does not add too much complexity. For example, I'm worried about transforming the current elements of the `<dl>` list since some of them contain logic with buttons, labels, etc.

I'd happily accept a signal that allows plugins to contribute to a dictionary(-like) structure that gets displayed here, but still, I thini it belongs below the data from the core system.

> The signal name was copied directly from an analogous signal surfaced within the presale component, so I kept that consistent.

I don't think it is analogous/consistant, though, since the signal within the presale component allows adding *more boxes of content* at the top, while your PR adds content *within a box*, which is kinda different?

PR #4336, opened 2024-07-25
API: Allow to filter enabled webhooks (Z#23160605)

raphaelm:


src/pretix/api/views/webhooks.py
```diff
@@ -19,18 +19,26 @@
...
+import django_filters
+from django_filters.rest_framework import FilterSet, DjangoFilterBackend
 from rest_framework import viewsets
 
 from pretix.api.models import WebHook
 from pretix.api.serializers.webhooks import WebHookSerializer
 from pretix.helpers.dicts import merge_dicts
 
 
+class EventFilter(FilterSet):
```
robbi5, line 31: Shouldn't it be more named like the corresponding ViewSet? :)

```suggestion
class WebHookFilter(FilterSet):
```

luelista (approved):

PR #4349, opened 2024-07-29
Bulk voucher creation: Use event default language for default text

raphaelm:


src/pretix/control/forms/vouchers.py
```diff
@@ -332,6 +335,12 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self._set_field_placeholders('send_subject', ['event', 'name'])
         self._set_field_placeholders('send_message', ['event', 'voucher_list', 'name'])
+
+        with language(self.instance.event.settings.locale, self.instance.event.settings.region):
+            for f in ("send_subject", "send_message"):
+                print(repr(self.fields[f].initial), self.instance.event.settings.locale)
```
wiffbi, line 341:
```suggestion
```

src/pretix/control/forms/vouchers.py
```diff
@@ -332,6 +335,12 @@ def __init__(self, *args, **kwargs):
...
+                self.fields[f].initial = str(_(self.fields[f].initial))
```
wiffbi, line 342: See above, we do not need `gettext_noop`, that actual solution here is the conversion to `str()` which fixes the translation into a string with the event’s locale in the with-statement.

```suggestion
                self.fields[f].initial = str(self.fields[f].initial)
```

src/pretix/control/forms/vouchers.py
```diff
@@ -269,16 +272,16 @@ class VoucherBulkForm(VoucherForm):
         label=_("Subject"),
         widget=forms.TextInput(attrs={'data-display-dependency': '#id_send'}),
         required=False,
-        initial=_('Your voucher for {event}')
+        initial=gettext_noop('Your voucher for {event}')
```
wiffbi, line 275: I do not think we need noop here as we do not want to use that text as literal text lateron. What we actually do is force-convert the lazy-translation string to a fixed translated string in `__init__`.

```suggestion
        initial=_('Your voucher for {event}')
```

src/pretix/control/forms/vouchers.py
```diff
@@ -269,16 +272,16 @@ class VoucherBulkForm(VoucherForm):
...
     )
     send_message = forms.CharField(
         label=_("Message"),
         widget=MarkdownTextarea(attrs={'data-display-dependency': '#id_send'}),
         required=False,
-        initial=_('Hello,\n\n'
-                  'with this email, we\'re sending you one or more vouchers for {event}:\n\n{voucher_list}\n\n'
-                  'You can redeem them here in our ticket shop:\n\n{url}\n\nBest regards,  \n'
-                  'Your {event} team')
+        initial=gettext_noop('Hello,\n\n'
```
wiffbi, line 281: See above

```suggestion
        initial=_('Hello,\n\n'
```

wiffbi (changes requested):
I do not think we need `gettext_noop` here. The trick is in the __init__-function converting the lazy translation to a fixed string within the event’s default locale.

PR #4372, opened 2024-08-02
Add codification of tax rates

raphaelm:
- [x] Data model
- [x] UI
- [x] Tests
- [x] API support
- [x] Add support in pretix-zugferd https://github.com/pretix/pretix-zugferd/pull/3
- [x] Plugins with fee creation
  - [x] shipping 
  - [x] servicefees https://github.com/pretix/pretix-servicefees/pull/36
  - [x] resellers

Post merge:
- [ ] Add support in pretix-posbackend
- [ ] Add support in pretixPOS (receipts, dsfinvk-export)


pc-coholic:
I played around with it and I only found that changing existing orders seems to be not working as expected:
- Enter existing event
- Supplement existing tax rule with a tax code
- Open existing order with a product using said tax rule
- At this moment: The order position does not have a tax code attributed to it (since it was created before the introduction of the tax code)
- Edit the order, requesting a recalculation of the taxes and re-issue of the invoice
- Expected result: tax code is added to order position, but it is not
- Second attempt: change the taxrule in the dropdown to the same tax rule that is currently applied and recalculate taxes
- Expected result: tax code is added to order position, but it is not

The only way around this that I found, was to change the tax rule of the order position to a wrong one and then back to the correct one.

src/pretix/base/models/tax.py
```diff
@@ -121,6 +129,152 @@ def __eq__(self, other):
...
+    (
+        _("Special cases"),
+        (
+            ("L", pgettext_lazy("tax_code", "Canary Islands general indirect tax")),
+            ("M", pgettext_lazy("tax_code", "Tax for production, services and importation in Ceuta and Melilla")),
+            ("B", pgettext_lazy("tax_code", "Transferred (VAT), only in Italy")),
+        )
+    ),
+    (
+        _("Except with specific reason"),
```
luelista, line 192:
```suggestion
        _("Exempt with specific reason"),
```

raphaelm:
>    Enter existing event
>    Supplement existing tax rule with a tax code
>    Open existing order with a product using said tax rule
>    At this moment: The order position does not have a tax code attributed to it (since it was created before the introduction of the tax code)
>    Edit the order, requesting a recalculation of the taxes and re-issue of the invoice
>    Expected result: tax code is added to order position, but it is not

The implemented version is that a change to the same tax rule is a no-op that is not even passed to OrderChangedManager. I'm not saying you wrote that line, but it's consistent with the rest of the order change manger, so I'd keep it that way.

>    Second attempt: change the taxrule in the dropdown to the same tax rule that is currently applied and recalculate taxes
>    Expected result: tax code is added to order position, but it is not

Agreed, that is wrong. Will fix.

PR #4383, opened 2024-08-08
Show weekday in subevent dates [Z#23161884]

luelista:
Only displayed for single-day events, to avoid complex logic:
image


raphaelm (approved):
Code looks good, could you look after the two remaining failing tests?

PR #4385, opened 2024-08-11
settings: also enable DEBUG for runserver_plus, not just runserver

stapelberg:
When using runserver_plus from the django-extensions package (for serving a development instance with a TLS certificate), I noticed the DEBUG setting was not set correctly, which resulted in static files not being served correctly.


raphaelm:
Change sounds good, I don't really see why the CI does not like it, let's re-run

PR #4415, opened 2024-08-23
Event list: Do not include old or inactive subevents in event list (Z#23163541)

raphaelm:


luelista (approved):
wrong Z# ticket number in title?

PR #4416, opened 2024-08-26
Control: fix initial sidebar display for mobile

wiffbi:
Originally window.load was bound to inside a jquery-ready call, which sometimes does not work as the window.load event was already fired.

The diff looks a bit weird, but I basically just moved the binding to window.load/resize outside the jquery-ready call.


luelista (approved):
👍

PR #4422, opened 2024-08-27
Social Dancing - Pretix SSO

dkelsen:


dkelsen:
Apologies, I didn't mean to open a PR here. You can safely ignore it.

raphaelm:
No worries! Note the special provisions of our license, though. Changing the logo is fine, fully changing the "powered by …" to no longer include us is not, see:

https://github.com/pretix/pretix/blob/7f7c95aedb7df780b1f2a8ecec700404ad9d9b57/LICENSE#L24-L27
https://docs.pretix.eu/en/latest/license/faq.html#what-do-i-need-to-do-if-i-modify-pretix

Also do not forget to link to the changed source code :)

dkelsen:
Yes, that I am aware of. While we're at it, perhaps you could clarify a question of mine that I could not find an answer to in the license: does this (and how would it) apply to emails and generated tickets?

Thank you for the thorough documentation and your hard work on the project!

raphaelm:
> does this (and how would it) apply to emails and generated tickets?

I don't think so. Generated tickets 100% not, emails 99% not, but we're not going to try to enforce it there ;)

PR #4452, opened 2024-09-09
Render progress info on non-javascript fallback page for celery tasks

raphaelm:
Looks like this:
image

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/waiting.html
+++ b/src/pretix/presale/templates/pretixpresale/waiting.html
@@ -27,12 +27,14 @@
                 </div>
             </div>
             {% if steps %}
-                <p class="steps">
+                <ol class="steps">
                     {% for step in steps %}
-                        <span class="fa fa-fw {% if step.done %}fa-check text-success{% else %}fa-cog fa-spin text-muted{% endif %}"></span>
-                        {{ step.label }}<br>
+                        <li>
+                            <span class="fa fa-fw {% if step.done %}fa-check text-success{% else %}fa-cog fa-spin text-muted{% endif %}"></span>
+                            {{ step.label }}
+                        </li>
                     {% endfor %}
-                </p>
+                </ol>
             {% endif %}
         {% endif %}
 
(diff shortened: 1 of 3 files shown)


src/pretix/presale/templates/pretixpresale/waiting.html
```diff
@@ -21,6 +21,21 @@
...
+                <div class="progress-bar progress-bar-success progress-bar-{{ percentage|floatformat:0 }}">
+                </div>
+            </div>
+            {% if steps %}
+                <p class="steps">
+                    {% for step in steps %}
+                        <span class="fa fa-fw {% if step.done %}fa-check text-success{% else %}fa-cog fa-spin text-muted{% endif %}"></span>
+                        {{ step.label }}<br>
+                    {% endfor %}
+                </p>
```
wiffbi, line 35: Regarding a11y, I would make this an `<ol>`, something like this:

```suggestion
                <ol class="steps">
                    {% for step in steps %}
                        <li><span class="fa fa-fw {% if step.done %}fa-check text-success{% else %}fa-cog fa-spin text-muted{% endif %}"></span>
                        {{ step.label }}<span class="sr-only">, {% if step.done %}{% trans "done" %}{% else %}{% trans "in progress" %}{% endif %}"></span></li>
                    {% endfor %}
                </ol>
```

src/pretix/static/pretixpresale/scss/waiting.scss
```diff
@@ -18,3 +18,16 @@ body {
...
+.progress {
+    max-width: 450px;
+    margin: 0 auto 20px;
+}
+.steps {
+    max-width: 450px;
+    text-align: left;
+    margin: 0 auto 20px;
+}
+@for $i from 0 through 100 {
```
wiffbi, line 31: Not sure we need that fine granularity – how fine will the task report the percentage anyway? Maybe just every 5 or even 10 percent steps? On the other hand, this will probably be well compressed with gzip, so it should not matter much adding 100 lines of CSS.

wiffbi (approved):
Looks good, except for a minor change regarding a11y. Not sure the progressbar needs a sr-only text as well, IMHO it does not.

src/pretix/static/pretixpresale/scss/waiting.scss
raphaelm: I'd rather have more granularity here than less, for some really slow tasks even sub-percent progress is visible.

PR #4469, opened 2024-09-17
Order import: Allow to set check-in attention and text (Z#23165806)

raphaelm:


pc-coholic (approved):
I just wanted to leave a comment saying how importing multiple lines into a single order will only use the last position to be imported as the decision maker - but it's the same with the email-address when importing that way... So works as intended, I guess :-)

PR #4474, opened 2024-09-19
Sendmail: Fix incorrect filtering of subevents when mailing to attendees (Z#23166274)

raphaelm:


src/pretix/plugins/sendmail/tasks.py
```diff
@@ -76,7 +79,7 @@ def send_mails_to_orders(event: Event, user: int, subject: dict, message: dict,
                         list_id__in=checkin_lists or []
                     )
                 ),
-            ).prefetch_related('addons'):
+            ).prefetch_related('addons', 'subevent'):
```
wiffbi, line 82: Probably not worth the optimization, but should we prefetch `subevent` only if at least one of the new subevent-parameters is provided?

wiffbi (approved):
Looks good, just minor comments very probably not worth changing.

src/pretix/plugins/sendmail/tasks.py
```diff
@@ -99,6 +102,15 @@ def send_mails_to_orders(event: Event, user: int, subject: dict, message: dict,
                 if p.attendee_email == o.email and send_to_order:
                     continue
 
+                if subevent and p.subevent_id != subevent:
+                    continue
+
+                if subevents_from and p.subevent.date_from < subevents_from:
```
wiffbi, line 108: Should we include subevents that started before `subevents_from` but end after `subevents_from`? IMHO not, but just to make sure.

src/pretix/plugins/sendmail/tasks.py
raphaelm: I think that is in fact not worth the complexity

src/pretix/plugins/sendmail/tasks.py
raphaelm: That would be contrary to how the filter parameters are labeled

PR #4482, opened 2024-09-23
Allow to set fallback secret keys

raphaelm:


wiffbi (approved):
I thought of using its own option-group in cfg and use e.g. `config.items('django_secret_fallbacks')`, but that is not really more elegant than those indexed keys. So looks good!

PR #4489, opened 2024-09-27
Avoid showing empty add-on step (Z#23167007)

raphaelm:
Do not show add-on step if none of the add-on items is available. This is not a perfect solution, there are still cases where an empty add-on step will show up, e.g.:

- Products disabled only for a specific subevent
- Sold-out add-ons combined with the "hide sold-out products" flag

Still, it's a good performance-functionality compromise this way and an improvement on the status quo.

### DIFF
--- a/src/pretix/base/models/items.py
+++ b/src/pretix/base/models/items.py
@@ -303,6 +303,48 @@ def filter_available(qs, channel='web', voucher=None, allow_addons=False):
     return qs.filter(q)
 
 
+def filter_variations_available(qs, channel='web', voucher=None, allow_addons=False):
+    # Channel can currently be a SalesChannel or a str, since we need that compatibility, but a SalesChannel
+    # makes the query SIGNIFICANTLY faster
+    from .organizer import SalesChannel
+
+    assert isinstance(channel, (SalesChannel, str))
+    q = (
+        Q(active=True)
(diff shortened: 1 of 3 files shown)


src/pretix/base/models/items.py
```diff
@@ -303,6 +303,48 @@ def filter_available(qs, channel='web', voucher=None, allow_addons=False):
...
+
+    if isinstance(channel, str):
+        q &= Q(Q(all_sales_channels=True) | Q(limit_sales_channels__identifier=channel))
+        q &= Q(Q(item__all_sales_channels=True) | Q(item__limit_sales_channels__identifier=channel))
+    else:
+        q &= Q(Q(all_sales_channels=True) | Q(limit_sales_channels=channel))
+        q &= Q(Q(item__all_sales_channels=True) | Q(item__limit_sales_channels=channel))
+
+    if not allow_addons:
+        q &= Q(Q(item__category__isnull=True) | Q(item__category__is_addon=False))
```
wiffbi, line 331: These two lines seem new, but what is the difference to line 319 above, which is always added to the queryset? Not sure, this here has any effect/makes a difference.

wiffbi (commented):
I am unsure about one addition around `not allow_addons`. Other than that, the `_is_applicable` looks reasonable.

PR #4512, opened 2024-10-08
Fix reauth loops with redirect style authentication plugins

### LINKED ISSUE #4498: Reauth with Redirection Auth Backend does not work
### Problem and impact

Using OIDC Authentication using a pluggable authentication backend users will get thrown into a loop when they need to reauth, i.e. reauth page > OIDC authenticator > oidc callback > reauth page.

The logs suggest that the login was successful:

```test
INFO 2024-10-01 17:28:02,653 pretix.control.views.auth auth Backend login successful for user 7.
```

But the session seems to lack some information to get out of the reauth.

### Expected behaviour

The successful login after the reauth should lead to a working sessions.

### Steps to reproduce

1. Login using https://github.com/thegcat/pretix-oidc
2. Wait 3 hours and a bit
3. Reload a Pretix page
4. Click on the "Continue" button on the reauth page
5. Some OIDC dance
6. Be back on the reauth page.

I have followed the different URLs/redirection in my browser. The order is:
1. GET https://pretix/control/reauth/
2. GET https://oidc-provider/authorize
3. GET https://pretix/oidc/callback/ (from I can gather the "Backend login successful" happens here)
4. GET https://pretix/control/
5. GET https://pretix/control/reauth/

### Screenshots

_No response_

### Link

_No response_

_No response_

_No response_

### Version

2024.9.0

thegcat:
Fixes #4498

### DIFF
--- a/src/tests/control/test_auth.py
+++ b/src/tests/control/test_auth.py
@@ -41,7 +41,7 @@ from django.contrib.auth.tokens import (
     PasswordResetTokenGenerator, default_token_generator,
 )
 from django.core import mail as djmail
-from django.test import TestCase, override_settings
+from django.test import RequestFactory, TestCase, override_settings
 from django.utils.timezone import now
 from django_otp.oath import TOTP
 from django_otp.plugins.otp_totp.models import TOTPDevice
@@ -50,6 +50,7 @@ from webauthn.authentication.verify_authentication_response import (
 )
 
 from pretix.base.models import Organizer, Team, U2FDevice, User
+from pretix.control.views.auth import process_login
 from pretix.helpers import security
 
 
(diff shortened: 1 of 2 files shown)


src/tests/control/test_auth.py
```diff
@@ -892,6 +893,19 @@ def test_log_out_after_relative_timeout_really_enforced(self):
...
+        session['pretix_auth_long_session'] = True
+        session['pretix_auth_login_time'] = int(time.time()) - 3600 * 5
+        session['pretix_auth_last_used'] = int(time.time()) - 3600 * 3 - 60
+        session.save()
+
+        request = RequestFactory().get("/")
+        request.session = self.client.session
+        process_login(request, self.user, keep_logged_in=True)
+
+        assert request.session['pretix_auth_last_used'] >= int(time.time())
```
raphaelm, line 907: This only works "by accident" and is not stable. After all, this line is running *after* process_login, so we could already be in the next second if our computer is slow, or if process_login ran at a time with .999999 and we have bad luck. Let's allow some slack:

```suggestion
        assert request.session['pretix_auth_last_used'] >= int(time.time()) - 60
```

It still tests what the test is trying to do.

raphaelm (commented):
LGTM with one caveat in the test

raphaelm:
(Ignore the test failure on postgres, we're trying to find that bug but it is not related)

src/tests/control/test_auth.py
thegcat: Works for me, I added it to the test commit.

raphaelm (approved):

PR #4533, opened 2024-10-17
Add action buttons to "missing quota" warnings

luelista:
image

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/item/base.html
+++ b/src/pretix/control/templates/pretixcontrol/item/base.html
@@ -12,10 +12,22 @@
     {% endif %}
     {% if object.id and not object.quotas.exists %}
         <div class="alert alert-warning">
-            {% blocktrans trimmed %}
-                Please note that your product will <strong>not</strong> be available for sale until you have added your
-                item to an existing or newly created quota.
-            {% endblocktrans %}
+            <div class="row">
+                <div class="col-lg-8">
+                    {% blocktrans trimmed %}
+                        Please note that your product will <strong>not</strong> be available for sale until you have added your
+                        item to an existing or newly created quota.
+                    {% endblocktrans %}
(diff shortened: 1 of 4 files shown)


src/pretix/control/templates/pretixcontrol/item/base.html
```diff
@@ -12,10 +12,22 @@ <h1>{% trans "Create product" %}</h1>
...
+                    {% blocktrans trimmed %}
+                        Please note that your product will <strong>not</strong> be available for sale until you have added your
+                        item to an existing or newly created quota.
+                    {% endblocktrans %}
+                </div>
+                <div class="col-lg-4 text-right">
+                    <a class="btn btn-default btn-sm" href="{% url "control:event.items.quotas" organizer=request.event.organizer.slug event=request.event.slug %}">
+                        <i class="fa fa-wrench"></i> {% trans "Manage quotas" %}
+                    </a>
+                    <a class="btn btn-default btn-sm" href="{% url "control:event.items.quotas.add" organizer=request.event.organizer.slug event=request.event.slug %}?{% if object.has_variations %}{% for var in object.variations.all %}product={{ object.pk }}-{{ var.pk }}&{% endfor %}{% else %}product={{ object.pk }}{% endif %}">
```
raphaelm, line 26: Could we even pass the product name as a default for the quota name in this one?

luelista: You mean, so it doesn't list all variation names in case of a product with variations, but only the product name? If so, I'd rather change that in the name generation logic in `quota.js`. However, I don't think we should do that, because it could cause confusion if more variations are added later on.

raphaelm: I forgot that quota.js will already handle this, nevermind

PR #4546, opened 2024-10-22
Fix undefined currency in paypal2

wiffbi:
This fixes a bug, where `currency` and `value` can be undefined even if they should not.

Should we add an logger.error as well? Not sure the else-case should ever be expected valid in the first place?


wiffbi:
Well, at least it does not throw a 500 but a `PaymentException` with at least a somewhat helpful error message for the user in `execute_payment` and returns an empty JSON-response in `XHRView`. Less noise (although it happens rarely anyway) in sentry as well.

wiffbi:
I will close this as discussed.

PR #4558, opened 2024-10-25
CI: Install dependencies with uv for speedup

raphaelm:


raphaelm:
Improvement ~20s per job

PR #4561, opened 2024-10-25
CI: Try to run with 4 threads

raphaelm:
GitHub has upgraded their standard workers from 3 to 4 CPUs since we last checked. Let's see if CI jobs run faster if we utilize them.


raphaelm:
Interestingly, this almost makes it slower. 3 seems like a good sweet spot, then.

PR #4577, opened 2024-10-28
Cross-selling: Use different label if there are no add-on products

raphaelm:
Before:

image

After:

Screenshot 2024-10-28 at 12-13-22 Step 1 of 5 Add-on products – Checkout Cross-selling test shop

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/checkout_addons.html
+++ b/src/pretix/presale/templates/pretixpresale/event/checkout_addons.html
@@ -16,7 +16,7 @@
             {% trans "A product in your cart is only sold in combination with add-on products that are no longer available. Please contact the event organizer." %}
         </div>
     {% endif %}
-    <form class="form-horizontal" method="post" data-asynctask
+    <div class="form-horizontal" method="post" data-asynctask
             data-asynctask-headline="{% trans "We're now trying to book these add-ons for you!" %}">
         {% csrf_token %}
         <div class="panel-group addons" id="questions_group">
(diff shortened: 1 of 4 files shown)


src/pretix/presale/templates/pretixpresale/event/checkout_addons.html
```diff
@@ -16,7 +16,7 @@
             {% trans "A product in your cart is only sold in combination with add-on products that are no longer available. Please contact the event organizer." %}
         </div>
     {% endif %}
-    <form class="form-horizontal" method="post" data-asynctask
+    <div class="form-horizontal" method="post" data-asynctask
```
wiffbi, line 19: Why did you remove the form? It does not work without it in my tests.

```suggestion
    <form class="form-horizontal" method="post" data-asynctask
```

wiffbi (changes requested):
I guess it was accidental, that you remoced the form-element? Otherwise the addon-step does not work in my tests.

raphaelm:
argh, when the IDE tries to be clever but isn't...

src/pretix/presale/templates/pretixpresale/event/checkout_addons.html
```diff
@@ -57,16 +57,28 @@ <h3 class="panel-title">
         </div>
 
         {% if cross_selling_data %}
-            <details class="panel panel-default cross-selling" open>
+            {% if forms %}
+                <details class="panel panel-default cross-selling" open>
+            {% else %}
+                <div class="panel panel-default cross-selling">
```
wiffbi, line 63: Hm, just stumbled upon this. It creates invalid HTML in combination with the summary element. Not that it’s super bad, but not ideal for a11y. So could you move the whole summary-block with headline „more recomandations“ into the if-block and in the else-block replace the summary-tag with a div.

wiffbi (changes requested):
Looks good, but just stumbled upon a potential a11y problem regarding details/summary.

wiffbi (approved):

PR #4588, opened 2024-10-29
Badge export: Allow to filter by order date (Z#23168742)

raphaelm:


luelista (commented):
maybe we should also allow to sort results by order datetime to make it easier to spot potential overlaps (e.g. if some orders came in later the same day)?
```
        elif form_data.get('order_by') == 'order_date':
            qs = qs.order_by('order__datetime')
```

luelista (approved):

PR #4597, opened 2024-10-31
Make API security profiles pluggable

raphaelm:


luelista (changes requested):
can we now remove [PretixPosSecurityProfile](https://github.com/pretix/pretix/pull/4597/files#diff-5694aab37b22e536165e9ad7db7ebb6e43f4f5d4d344e27f60b59e12dbd00022R187) in here?

PR #4599, opened 2024-10-31
Hide input dependencies recursively if input will be disabled

wiffbi:


src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -377,7 +377,7 @@ var form_handlers = function (el) {
         var dependent = $(this),
             dependency = findDependency($(this).attr("data-display-dependency"), this),
             update = function (ev) {
-                var enabled = dependency.toArray().some(function(d) {
+                var enabled = dependency.attr('disabled') ? false : dependency.toArray().some(function(d) {
```
luelista, line 380:
```suggestion
                var enabled = dependency.toArray().some(function(d) {
                    if (dependency.attr('disabled')) return false;
```

PR #4606, opened 2024-11-04
API: Expose cancellation_date on order endpoint (Z#23170733)

pc-coholic:


doc/api/resources/orders.rst
```diff
@@ -104,6 +104,7 @@ url                                   string                     The full URL to
 payments                              list of objects            List of payment processes (see below)
 refunds                               list of objects            List of refund processes (see below)
 last_modified                         datetime                   Last modification of this object
+cancellation_date                     datetime                   Time of order creation (or ``null``)
```
robbi5, line 107:
```suggestion
cancellation_date                     datetime                   Time of order cancellation (or ``null``)
```

doc/api/resources/orders.rst
```diff
@@ -104,6 +104,7 @@ url                                   string                     The full URL to
 payments                              list of objects            List of payment processes (see below)
 refunds                               list of objects            List of refund processes (see below)
 last_modified                         datetime                   Last modification of this object
+cancellation_date                     datetime                   Time of order cancellation (or ``null``)
```
raphaelm, line 107: Maybe add a note that this will not be set for partial cancellations and is not reliable for orders canceled, reactivated and canceled again?

PR #4616, opened 2024-11-07
Stripe: Disable SOFORT

raphaelm:


src/pretix/plugins/stripe/payment.py
```diff
@@ -1600,6 +1600,17 @@ class StripeSofort(StripeRedirectMethod):
     method = 'sofort'
     redirect_in_widget_allowed = False
 
+    def is_allowed(self, request: HttpRequest, total: Decimal=None) -> bool:
+        # Stripe<>giropay is shut down November 29th
```
pc-coholic, line 1604: Sofort

pc-coholic (changes requested):
We should probably also schedule a second PR to merge with the next release to remove the `method_sofort`-checkbox from the settings

src/pretix/plugins/stripe/payment.py
raphaelm, line 1604:
```suggestion
        # Stripe<>Sofort is shut down November 29th
```

pc-coholic (approved):

PR #4620, opened 2024-11-08
Fix creating to many ScheduledMails on subevent creation

wiffbi:
When havin a event series with a ScheduledMail-rule, that is bound to a single subevent, when creating a new subevent, the subevent filter for that rule was ignored and a ScheduledMail was created for the newly created subevent.

This PR basically adds a `.filter(subevent=None)` when looping over `event.sendmail_rules`. It furthermore adds some minor optimizations such as checking for subevent in the for-loop (subevent will always be set as the signal will only be run on subevent-creation) and filtering for existing_rules on db-level, not in the for-loop.

### DIFF
--- a/src/pretix/plugins/sendmail/signals.py
+++ b/src/pretix/plugins/sendmail/signals.py
@@ -65,11 +65,10 @@ def scheduled_mail_create(sender, **kwargs):
     with scope(organizer=event.organizer):
         existing_rules = ScheduledMail.objects.filter(subevent=subevent).values_list('rule_id', flat=True)
         to_create = []
-        for rule in event.sendmail_rules.all():
-            if rule.pk not in existing_rules and subevent:
-                sm = ScheduledMail(rule=rule, event=event, subevent=subevent)
-                sm.recompute()
-                to_create.append(sm)
+        for rule in event.sendmail_rules.filter(subevent=None).exclude(id__in=existing_rules):
+            sm = ScheduledMail(rule=rule, event=event, subevent=subevent)
+            sm.recompute()
+            to_create.append(sm)
         ScheduledMail.objects.bulk_create(to_create)
 
 


src/pretix/plugins/sendmail/signals.py
```diff
@@ -65,11 +65,10 @@ def scheduled_mail_create(sender, **kwargs):
     with scope(organizer=event.organizer):
         existing_rules = ScheduledMail.objects.filter(subevent=subevent).values_list('rule_id', flat=True)
         to_create = []
-        for rule in event.sendmail_rules.all():
-            if rule.pk not in existing_rules and subevent:
-                sm = ScheduledMail(rule=rule, event=event, subevent=subevent)
-                sm.recompute()
-                to_create.append(sm)
+        for rule in event.sendmail_rules.filter(subevent=None).exclude(id__in=existing_rules):
```
raphaelm, line 68: what about the subevent=subevent case?

wiffbi: The subevent=subevent-case would not need a new `ScheduledMail`. If it is a newly created subevent, there cannot be any sendmail_rule containing it, so we only need to add ScheduledMails to rules that apply to all (and all future) subevents (therefore subevent=None). If the subevent already existed and is updated and sendmail_rules exist for it, then there also should already exist ScheduledMails for these sendmail_rules, as they got created when the sendmail_rule was created.

Recomputation of existing ScheduledMails due to changes in subevents regarding date and time is done in the periodic task `sendmail_run_rules`. We could of course do it here as well, but it wasn’t before, so I thought I leave it that way.

raphaelm: Okay, makes sense

raphaelm (approved):

PR #4623, opened 2024-11-08
Event creation: Prevent accidentally creating events without tax rate

raphaelm:


src/pretix/control/forms/event.py
```diff
@@ -136,6 +136,10 @@ class EventWizardBasicsForm(I18nModelForm):
         choices=settings.LANGUAGES,
         label=_("Default language"),
     )
+    no_taxes = forms.BooleanField(
+        label=_("Do not use taxes"),
```
wiffbi, line 140: As one could later define more complex taxes, this label is kind of misleading. Should we either label it differently („I don’t want to specify taxes now“) and/or should we add a `help_text` such as „You can always configure tax rates later“).

wiffbi: Sorry, forgot to add suggestion.

```suggestion
        label=_("I don’t want to specify taxes now"),
        help_text=_("You can always configure tax rates later."),
```

PR #4627, opened 2024-11-12
Allow to use custom domains for some but not all events (Z#23153875)

raphaelm:


src/pretix/multidomain/urlreverse.py
```diff
@@ -43,36 +43,41 @@
...
+                domain = None, None
         event.cache.set('domain' + suffix, domain or 'none')
         setattr(event, '_cached_domain' + suffix, domain or 'none')
     elif domain == 'none':
         setattr(event, '_cached_domain' + suffix, 'none')
         domain = None, None
     else:
         setattr(event, '_cached_domain' + suffix, domain)
-    return domain if return_info or not isinstance(domain, tuple) else domain[0]
+    return domain if return_mode or not isinstance(domain, tuple) else domain[0]
```
luelista, line 80: the `or not isinstance(domain, tuple)` condition feels like it either never triggers or if it does, can be a source for a very weird bug, where get_event_domain might return a non-tuple value when a tuple is expected?

raphaelm: Yup, that case can probably go, I don't know why it is there

raphaelm:
```suggestion
    return domain if return_mode else domain[0]
```

src/pretix/control/forms/organizer.py
```diff
@@ -133,63 +133,108 @@ def clean_slug(self):
...
+            organizer=self.organizer,
+        )
+        self.add_fields(form, None)
+        return form
+
+    def clean(self):
+        super().clean()
+        data = [f.cleaned_data for f in self.forms]
+
+        if len([d for d in data if d["mode"] == KnownDomain.MODE_ORG_DOMAIN and not d.get("DELETE")]) > 1:
```
luelista, line 226: to avoid KeyError if the formset is sent with empty forms
```suggestion
        if len([d for d in data if d.get("mode") == KnownDomain.MODE_ORG_DOMAIN and not d.get("DELETE")]) > 1:
```

luelista (changes requested):

luelista (approved):

PR #4631, opened 2024-11-14
Allow "open in new tab" for event typeahead

raphaelm:


wiffbi:
It kind of works, but only for right-click and „open in new …“. When testing I generally stumbled upon e.g. keyboard-navigation (which does not work at all in the Dashboard’s typeahead due to `$query.onblur`) as well as command-click not working as expected on Mac (should open a new window as well). I played around a bit and changed the JS to use `focusin/focusout` instead of `blur` on the `$query` and also added support for the Escape-key. Should I push my changes to this PR or create new one?

raphaelm:
Feel free to push :)

PR #4654, opened 2024-11-22
[A11y] Improve customer account pages

wiffbi:
This PR gives the customer account pages an overhaul – improving not only accessibility but adding better support for mobile displays. Instead of using tabs for lists of orders, memberships, etc. this PR introduces a subnavigation to the customer account pages and adds pages for each list of orders, memberships, etc.

This PR also adds two helper template-tags:

- `{% icon "check" %)` which makes it quicker to write and easier to read/spot icon-usage in the html-source – we might at some point use it to replace all icon usages in pretix?
- `{% textbubble "warning" icon="exclamation-triangle" %}Text goes here{% endtextbubble %}` allows for easier markup of status text bubbles and replaces bootstrap’s `.label` with a colored label, that uses accessible color-combinations. The `icon` parameter is optional.


src/pretix/presale/templates/pretixpresale/organizers/customer_memberships.html
```diff
@@ -0,0 +1,110 @@
...
+                                        <span class="text-muted" data-toggle="tooltip" title="{% trans "Membership is transferable" %}">
+                                            {% icon "users" %}
+                                        </span>
+                                    {% endif %}
+                                </dt>
+                                <dd class="text-muted">
+                                    {% if not m.membership_type.transferable %}
+                                        {% icon "id-badge" %}
+                                    {% endif %}
+                                    {{ m.attendee_name|default_if_none:"" }}
```
luelista, line 38: Either add some text for the default_if_none case, or remove the whole line, otherwise it looks a bit broken /unclear what the icon means without attendee name:
image

luelista (approved):

src/pretix/presale/templates/pretixpresale/organizers/customer_memberships.html
```diff
@@ -0,0 +1,110 @@
...
+                                    {% if m.membership_type.transferable %}
+                                        <span class="text-muted" data-toggle="tooltip" title="{% trans "Membership is transferable" %}">
+                                            {% icon "users" %}
+                                        </span>
+                                    {% endif %}
+                                </dt>
+                                <dd class="text-muted">
+                                    {% if not m.membership_type.transferable %}
+                                        {% icon "id-badge" %}
+                                    {% endif %}
```
luelista, line 37: This icon should be displayed if an attendee_name is displayed and hidden otherwise (and maybe we should add a title/sr-only text saying "Attendee name:").
```suggestion
                                    {% if m.attendee_name %}
                                        {% icon "id-badge" %}
                                    {% endif %}
```

src/pretix/presale/templates/pretixpresale/organizers/customer_memberships.html
```diff
@@ -0,0 +1,110 @@
...
+                                    {{ m.attendee_name|default_if_none:"" }}
+                                </dd>
```
luelista, line 39:
```suggestion
                              {% if m.attendee_name %}
                                  <dd class="text-muted">
                                        {% icon "id-badge" %}
                                        <span class="sr-only">{% trans "Attendee name:" %}</span>
                                        {{ m.attendee_name|default_if_none:"" }}
                                  </dd>
                              {% endif %}
  ```

src/pretix/presale/templates/pretixpresale/organizers/customer_memberships.html
wiffbi: Good catch – in my tests, when it is not transferable, I always had a name when membership was not transferable. Not sure how you managed to get in that state? Ah, maybe changing the membership later! I will fix as you suggest. :)

luelista (approved):

PR #4659, opened 2024-11-25
Update address field logic (Z#23163120)

luelista:
This PR consolidates the front-end (javascript) code to show/hide and mark as required/optional certain address components depending on the selected country. Information on the required fields is loaded on demand. It's loaded in the same request as the list of (federal) states of a country, so no additional request is required.

Example response for Germany:
image

this resolves the following problems:
- the state select box was always marked as optional, even if it was required because the whole address was required and a country with "state in address" was selected - this is now handled correctly
- even for countries where we don't know about the correct address fields, we always required street, zipcode and city, if the whole address form was obligatory (but only street address if the whole address form was optional) - now we always only require a street address, which is a multi line field, allowing customers to enter their address in free form.

### DIFF
--- a/src/pretix/static/pretixbase/js/addressform.js
+++ b/src/pretix/static/pretixbase/js/addressform.js
@@ -24,6 +24,7 @@ $(function () {
                         return;  // Lost race
                     }
                     var selected_value = dependents.state.prop("data-selected-value");
+                    if (selected_value) dependents.state.prop("data-selected-value", "");
                     dependents.state.find("option").filter(function (t) {return !!$(this).attr("value")}).remove();
                     if (data.data.length > 0) {
                         $.each(data.data, function (k, s) {
@@ -35,7 +36,7 @@ $(function () {
                         });
                     }
                     for(var k in dependents) {
-                        const options = data[k], dependent = dependents[k]; console.log(options, dependent)
+                        const options = data[k], dependent = dependents[k];
                         if ('visible' in options) {
                             if (options.visible) {
                                 dependent.closest(".form-group").show().toggleClass('required', isRequired);
@@ -54,6 +55,7 @@ $(function () {
                     dependency.closest(".form-group").find("label .fa-spin").remove();
                 });
             };
+        dependents.state.prop("data-selected-value", dependents.state.val());
         update();
         dependency.on("change", update);
     });
--- a/src/pretix/static/pretixbase/js/addressform.js
+++ b/src/pretix/static/pretixbase/js/addressform.js
@@ -1,0 +1,44 @@
+$(function () {
+    "use strict";
+
+    $("select[name$=state]:not([data-static])").each(function () {
+        var dependent = $(this),
+            counter = 0,
+            dependency = $(this).closest(".panel-body, form").find('select[name$=country]'),
+            update = function (ev) {
+                counter++;
+                var curCounter = counter;
+                dependent.prop("disabled", true);
+                dependency.closest(".form-group").find("label").prepend("<span class='fa fa-cog fa-spin'></span> ");
+                $.getJSON('/js_helpers/states/?country=' + dependency.val(), function (data) {
+                    if (counter > curCounter) {
+                        return;  // Lost race
+                    }
+                    var selected_value = dependent.prop("data-selected-value");
+                    dependent.find("option").filter(function (t) {return !!$(this).attr("value")}).remove();
+                    if (data.data.length > 0) {
+                        $.each(data.data, function (k, s) {
+                            var o = $("<option>").attr("value", s.code).text(s.name);
+                            if (s.code == selected_value || (selected_value && selected_value.indexOf && selected_value.indexOf(s.code) > -1)) {
+                                o.prop("selected", true);
+                            }
+                            dependent.append(o);
+                        });
+                        dependent.closest(".form-group").show();
+                        dependent.prop('required', dependency.prop("required"));
+                    } else {
+                        dependent.closest(".form-group").hide();
+                        dependent.prop("required", false);
+                    }
+                    dependent.prop("disabled", false);
+                    dependency.closest(".form-group").find("label .fa-spin").remove();
+                });
+            };
+        if (dependent.find("option").length === 1) {
+            dependent.closest(".form-group").hide();
+        } else {
+            dependent.prop('required', dependency.prop("required"));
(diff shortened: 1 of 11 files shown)


src/pretix/static/pretixbase/js/addressform.js
```diff
@@ -0,0 +1,63 @@
...
+                                dependent.prop('required', isRequired);
+                            } else {
+                                dependent.closest(".form-group").hide();
+                                dependent.prop("required", false);
+                            }
+                        }
+                        if ('required' in options) {
+                            dependent.closest(".form-group").toggleClass('required', options.required && isRequired);
+                            dependent.prop('required', options.required && isRequired);
+                        }
```
wiffbi, line 52: I’d remove the `if ('visible' in options)` as if one country happens to hide the street, then it will not show up again as the js_helper does not always provide the key `visible` in its reponse. To make it more robust, I would add a default-visibility of true like so – also, I made the code more compact:

```suggestion
                        const options = data[k],
                            dependent = dependents[k],
                            visible = 'visible' in options ? options.visible : true,
                            required = 'required' in options && options.required && isRequired && visible;

                        dependent.closest(".form-group").toggle(visible).toggleClass('required', required);
                        dependent.prop("required", required);
```

src/pretix/static/pretixbase/js/addressform.js
```diff
@@ -0,0 +1,63 @@
+$(function () {
+    "use strict";
+
+    $("select[data-country-information-url]").each(function () {
+        let counter = 0;
+        const dependency = $(this),
+            url = this.getAttribute('data-country-information-url'),
```
wiffbi, line 7: I’d add the loading-indicator once to the DOM and just hide and show it. Maybe even fade?
```suggestion
            loader = $("<span class='fa fa-cog fa-spin'></span>").hide().prependTo(dependency.closest(".form-group").find("label")),
            url = this.getAttribute('data-country-information-url'),
```

src/pretix/static/pretixbase/js/addressform.js
```diff
@@ -0,0 +1,63 @@
...
+                'zipcode': form.find("input[name$=zipcode]"),
+                'street': form.find("textarea[name$=street]"),
+                'state': form.find("select[name$=state]"),
+                'vat_id': form.find("input[name$=vat_id]"),
+            },
+            update = function (ev) {
+                counter++;
+                const curCounter = counter;
+                for (var k in dependents) dependents[k].prop("disabled", true);
+                dependency.closest(".form-group").find("label").prepend("<span class='fa fa-cog fa-spin'></span> ");
```
wiffbi, line 21:
```suggestion
                loader.fadeIn();
```

src/pretix/static/pretixbase/js/addressform.js
```diff
@@ -0,0 +1,63 @@
...
+                                dependent.prop("required", false);
+                            }
+                        }
+                        if ('required' in options) {
+                            dependent.closest(".form-group").toggleClass('required', options.required && isRequired);
+                            dependent.prop('required', options.required && isRequired);
+                        }
+                    }
+                    for (var k in dependents) dependents[k].prop("disabled", false);
+                    dependency.closest(".form-group").find("label .fa-spin").remove();
```
wiffbi, line 55: Do not remove the loading-spinner, but just hide it. I would move to the `.always` callback of the xhr.

```suggestion
                }).always(function() {
                    loader.fadeOut();
```

wiffbi (changes requested):
Looks good! I have a few opinionated code changes for the javascript, e.g. adding the loader once to the DOM and just toggle it. You can ignore that if you want, but I would definitely add a `.fail` handler to the jquery-xhr and not use a counter for race-conditions, but abort the previous xhr if any.

src/pretix/static/pretixbase/js/addressform.js
```diff
@@ -0,0 +1,63 @@
+$(function () {
+    "use strict";
+
+    $("select[data-country-information-url]").each(function () {
+        let counter = 0;
+        const dependency = $(this),
+            url = this.getAttribute('data-country-information-url'),
+            form = $(this).closest(".panel-body, form, .profile-scope"),
```
wiffbi, line 8:
```suggestion
            form = dependency.closest(".panel-body, form, .profile-scope"),
```

src/pretix/static/pretixbase/js/addressform.js
luelista: > as the js_helper does not always provide the key visible in its reponse

How so? The js_helper always returns the exact same set of keys, just different values. So all fields that can get hidden also will be made visible again, if required. 

But I like your more compact logic better, I'll just have to double check whether it really behaves the same (I think so, but need to try it - or did you already test it?)

src/pretix/static/pretixbase/js/addressform.js
```diff
@@ -0,0 +1,63 @@
...
+                        if ('required' in options) {
+                            dependent.closest(".form-group").toggleClass('required', options.required && isRequired);
+                            dependent.prop('required', options.required && isRequired);
+                        }
+                    }
+                    for (var k in dependents) dependents[k].prop("disabled", false);
+                    dependency.closest(".form-group").find("label .fa-spin").remove();
+                });
+            };
+        dependents.state.prop("data-selected-value", dependents.state.val());
```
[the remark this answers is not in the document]
wiffbi: For me it adds unnecessary complexity to the code. When I tested, I came across a somewhat edge-case, but nonetheless: I selected US, then selected another country, where the XHR failed (I made changes to the py-code) and then selected US again and the selected state was wiped. This just shows e.g. on a slow connection, this could happen as well. So that is why I would change it to a general kind of if-the-state-was-selected-select-it-again type of thing.

src/pretix/static/pretixbase/js/addressform.js
wiffbi: You are right, currently there cannot be any setup where e.g. street would get a key `visible` only in the case it is hidden. But I guess to make it fail-safe, it might be a good idea to default to something, so if something else hid the element, it gets unhidden.

I did test it locally, but it adds one more thing: an input, that is hidden, never can be required – that is why the `&& visible` is at the end of the reuqired-assignment. I think that is what we want, but wanted to point that out.

PR #4662, opened 2024-11-27
Presale: Frontend enforcement of max item per order

stranck:
The maximum number of item per order, even if set, is not "visually" enforced on the user: The user is free to select how many items he wants, and the check is done (serverside) only after he clicks "add to cart". This can be confusing for the end user, specially if the `max_items_per_order` setting is set to 1. This commit creates a "live" check done client side, really similar to the one used for addons, which stops the user from surpassing the previous settings.

The diff of the `fragment_product_list.html` file looks broken to me. What I really did was simply wrapping the big, outern, for inside a section I will later use inside `main.js`

### DIFF
--- a/src/pretix/base/services/tax.py
+++ b/src/pretix/base/services/tax.py
@@ -62,6 +62,9 @@ class VATIDTemporaryError(VATIDError):
 
 def _validate_vat_id_NO(vat_id, country_code):
     # Inspired by vat_moss library
+    if not vat_id.startswith("NO"):
+        # prefix is not usually used in Norway, but expected by vat_moss library
+        vat_id = "NO" + vat_id
     try:
         vat_id = vat_moss.id.normalize(vat_id)
     except ValueError:
--- a/src/tests/base/test_vat_id_validation.py
+++ b/src/tests/base/test_vat_id_validation.py
@@ -195,4 +195,27 @@ def test_NO_id_valid():
 
     assert validate_vat_id('NO974760673 MVA', 'NO') == 'NO974760673MVA'
 
+
+@responses.activate
+def test_NO_id_valid_without_prefix():
+    responses.add(
+        responses.GET,
+        'https://data.brreg.no/enhetsregisteret/api/enheter/974760673',
+        body='{"organisasjonsnummer":"974760673","navn":"REGISTERENHETEN I BRØNNØYSUND","organisasjonsform":{"kode":'
+             '"ORGL","beskrivelse":"Organisasjonsledd","_links":{"self":{"href":"https://data.brreg.no/enhetsregisteret/api/'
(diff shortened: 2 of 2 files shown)


raphaelm:
The problem is that this only solves the problem in some cases. On the shop frontpage there could be products in the cart already which this logic does not know about. Therefore, so far I believed it is not worth the complexity. @wiffbi what do you think??

stranck:
I can try seeing if I have a way to also have them included in the total! :D
A question regarding this: in which cases do we already have a pre-filled cart before the manual insertion by the user? With widgets and [cart share](https://github.com/pretix/pretix-cartshare), or is there any other use case?

stranck:
In the meanwhile, I added checks against the pre-existent positions in the cart:
- The number of items (not add-ons, just items) in the cart is now added to the selected total
- If the number of items in the cart already fulls the max item per order count, the buttons/inputs to add new items to the cart are disabled

Sorry for the late hour!

wiffbi:
Sorry for my late reply, I wanted to properly evaluate this. It is a nice addition, but I stumbled upon a few things - not necessarily show-stoppers, but just to make sure, we are aware:

- It does not give feedback to users as to why it disables inputs. I know, when reaching the max-attributes value while inputting quantities, the browser does not give feedback either. But disabling the inputs when the cart already is at max-capacity feels a bit surprising especially as there is no explanation given. Maybe we could make it behave similar to what we do in the add-on step with `data-addon-max-count`? The inputs are enabled, but when you exceed the limit, a little popup is shown telling you that you reached the maximum possible. I am still investigating this regarding to accessibility though, as it currently does not read out that message on e.g. VoiceOver either. Maybe we could unify that behaviour with one set of common data-attributes and one script.
- It does not work with seating as seating dynamically adds „product inputs“ to the form on seat selection. Not sure that is a big problem at all, seating just does not get the benefit of this PR. Seating actually has its own min/max-mechanism which shows a message close to the submit-button and disables its own submit-button.

Generally I like the improvement, although I think we need to work on the UX regarding communicating the blocking of input. It might even be a good idea to solve it similar to what seating does – it keeps the inputs active, but shows an error message close to the action-button (add to cart), so the user understands the problem. Regarding accessibility this even has the added benefit to be a live-region, so the error can be read out aloud. Not sure this would work for add-ons (currently it blocks input and shows a title-attribute/popup), we probably need a way to show the error closer to the affected fieldset. And additionally some error close to the continue-button? Hm …

The way pretix handles this currently is not the most elegant way, but it provides proper error messages to the user to understand the problem – especially for hidden settings like max-items-per-order. I do think we can and should improve the UX here with what you suggest, but IMHO we need to find a way to let the user know what is happending and why, so they can adjust accordingly.

I will look into this UI/UX as I am working on at the accessibility implications of the add-on-step max-limitation anyways.

stranck:
Oops, accidentally closed this PR. I still have the intention on working on this, due to life's issues I still don't have time yet sadly :((

PR #4669, opened 2024-11-30
Make BaseInvoiceForm more reusable

raphaelm:


src/pretix/static/pretixbase/js/addressform.js
```diff
@@ -8,7 +8,8 @@ $(function () {
             url = this.getAttribute('data-country-information-url'),
             form = dependency.closest(".panel-body, form, .profile-scope"),
             isRequired = dependency.closest(".form-group").is(".required"),
-            dependents =  {
+            htmlRequired = dependency.closest(".no-required-attributes").length == 0,
```
wiffbi, line 11: For my brain, it is easier to process it as an overwrite-option that forces anything to be optional. Might just be my brain …

```suggestion
            forceOptional = dependency.closest(".no-required-attributes").length > 0,
```

Should we add a comment, that plugins can add this CSS-class to make everything optional? Because in pretix itself, there is no use of this CSS-class elsewhere.

src/pretix/static/pretixbase/js/addressform.js
```diff
@@ -39,7 +40,7 @@ $(function () {
                             required = 'required' in options && options.required && isRequired && visible;
 
                         dependent.closest(".form-group").toggle(visible).toggleClass('required', required);
-                        dependent.prop("required", required);
+                        dependent.prop("required", required && htmlRequired);
```
wiffbi, line 43: Especially used in this context IMHO naming it `forceOptional` is easier to read/understand for my brain.

```suggestion
                        dependent.prop("required", required && !forceOptional);
```

wiffbi (approved):
Looks good, just an opinion on naming and maybe adding a comment regarding usage of the sepcial CSS-class.

PR #4675, opened 2024-12-03
[A11y] Improve number inputs for screen-readers

wiffbi:
This PR groups the number inputs and its rocker-buttons into a fieldset with a legend, which improves the screenreader experience by reading that legend as well when focusing the +/- rocker buttons.

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/fragment_addon_choice.html
+++ b/src/pretix/presale/templates/pretixpresale/event/fragment_addon_choice.html
@@ -6,13 +6,14 @@
 {% load eventsignal %}
 {% load rich_text %}
 {% for c in form.categories %}
+    {% with category_idx=forloop.counter %}
     <fieldset data-addon-max-count="{{ c.max_count }}"{% if c.multi_allowed %} data-addon-multi-allowed{% endif %}>
         <legend>{{ c.category.name }}</legend>
         {% if c.category.description %}
             {{ c.category.description|rich_text }}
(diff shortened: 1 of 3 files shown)


src/pretix/presale/templates/pretixpresale/event/fragment_addon_choice.html
```diff
@@ -6,13 +6,14 @@
...
 {% for c in form.categories %}
+    {% with category_idx=forloop.counter %}
     <fieldset data-addon-max-count="{{ c.max_count }}"{% if c.multi_allowed %} data-addon-multi-allowed{% endif %}>
         <legend>{{ c.category.name }}</legend>
         {% if c.category.description %}
             {{ c.category.description|rich_text }}
         {% endif %}
         {% if c.min_count == c.max_count %}
-            <p class="addon-count-desc">
+            <p class="addon-count-desc" id="addon-count-desc-{{ category_idx }}">
```
raphaelm, line 16: This is a fragment template that will be included once per position, so these IDs won't really be unique, will they?

raphaelm (changes requested):

src/pretix/presale/templates/pretixpresale/event/fragment_addon_choice.html
wiffbi: 🤦  You are right! I will add the `{{ form.pos.pk }}`. Thanks!

PR #4679, opened 2024-12-04
Fix shipping form  (Z#23175243)

luelista:


luelista:
resolved by https://github.com/pretix/pretix/pull/4681

PR #4685, opened 2024-12-04
Move cursor fix for disabled fieldsets from bootstrap to presale

wiffbi:
This PR undoes commit a747ab1, which changed a bootstrap css – although we probably never receive an update to bootstrap 3, I guess we should not change it if possible. Furthermore it adds slideUp/Down-animation through Javascript to the panels as using transform with plain CSS does not properly due to auto-height and padding issues.

### DIFF
--- a/src/pretix/static/pretixpresale/js/ui/main.js
+++ b/src/pretix/static/pretixpresale/js/ui/main.js
@@ -329,10 +329,6 @@ $(function () {
         var $input = $("input", this);
         if (!$input.prop("checked")) $input.prop('checked', true).trigger("change");
     });
-    $(".accordion-radio input").on("change", function() {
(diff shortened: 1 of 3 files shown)


src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -329,6 +329,10 @@ $(function () {
         var $input = $("input", this);
         if (!$input.prop("checked")) $input.prop('checked', true).trigger("change");
     });
+    $(".accordion-radio input").on("change", function() {
+        $(this).closest("fieldset").find(".panel-body").slideDown();
+        $(this).closest(".panel-group").find(".accordion-panel:has(.accordion-radio input:not(:checked)) .panel-body").slideUp();
+    });
```
wiffbi, line 335:
```suggestion
```

PR #4686, opened 2024-12-04
Fix fieldset accordion css

wiffbi:
The new fieldset-accordion markup was worng by using p and div inside legend-elements. This PR fixes the CSS when changing them to correctly label and span.

### DIFF
--- a/src/pretix/static/pretixpresale/scss/_forms.scss
+++ b/src/pretix/static/pretixpresale/scss/_forms.scss
@@ -135,9 +135,13 @@ a.btn, button.btn {
   margin: 0;
 }
 .panel-default>.accordion-radio>.panel-heading, fieldset.accordion-panel>legend>.panel-heading {
+  display: block;
   color: #333;
   background-color: #f5f5f5;
   padding: 8px 15px;
+  margin: 0;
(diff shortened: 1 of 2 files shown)


raphaelm (approved):

src/pretix/static/pretixpresale/scss/_forms.scss
```diff
@@ -135,9 +135,13 @@ a.btn, button.btn {
   margin: 0;
 }
 .panel-default>.accordion-radio>.panel-heading, fieldset.accordion-panel>legend>.panel-heading {
+  display: block;
   color: #333;
   background-color: #f5f5f5;
   padding: 8px 15px;
+  margin: 0;
+  line-height: 1;
```
luelista, line 143:
```suggestion
```
this makes the headers smaller, is that intended?

wiffbi: 🤦  You are right, I just tested it for centering and somehow did not recognize it made them smaller. Thank you! I took that super magic number 1.428571429 from other css-parts. :)

```suggestion
  line-height: 1.428571429;
```

PR #4687, opened 2024-12-05
Event dashboard: Use intcomma in numbers (Z#23175343)

raphaelm:


wiffbi:
Should we only do this in the dashboard or generally, so numbers are always displayed the same? `USE_THOUSAND_SEPARATOR = True` feels like a big change, but it makes all numbers being formatted the same.

raphaelm:
Yeah, I think fixing this in money_filter is already global enough, and for non-monetary numbers doing it as-needed is okay

luelista (approved):

PR #4710, opened 2024-12-16
[A11y] Improve focus handling for widget overlay

wiffbi:
To be able to move focus inside the widgets overlay iframe, this PR moves the widget’s close-button before the iframe in the HTML source. It also adds a few aria-attributes and improves access to the iframe – instead of setting the src directly it provides a prop, that can be set.


src/pretix/static/pretixpresale/js/widget/widget.js
```diff
@@ -2003,9 +1996,41 @@ var create_overlay = function (app) {
...
         methods: {
+        },
+        watch: {
+            frame_src: function (newValue) {
+                this.$el.querySelector("iframe").src = newValue;
+            },
+            frame_loading: function (newValue) {
+                if (newValue) {
+                    this.$el.querySelector("iframe").src = this.frame_src;
+                }
```
raphaelm, line 2014: Can you explain the idea behind this, maybe even in a comment?

Doesn't that possibly cause
```
        this.$root.overlay.frame_loading = true;
        this.$root.overlay.frame_src = redirect_url;
```
to trigger two page loads, one to the old and one to the new URL?

src/pretix/static/pretixpresale/js/widget/widget.js
```diff
@@ -2003,9 +1996,41 @@ var create_overlay = function (app) {
...
+                    var btn = this.$el?.querySelector(".pretix-widget-frame-close a");
+                    this.$nextTick(function () {
+                        btn?.focus();
+                    });
+                } else {
+                    this.prevActiveElement?.focus();
+                }
+            },
+            error_message: function (newValue) {
+                if (newValue) {
```
raphaelm, line 2028: Hmm, wouldn't it make sense to focus the "Continue" button of the error message here? In my test, I wasn't really well able to reach it.

src/pretix/static/pretixpresale/js/widget/widget.js
wiffbi: The problem I faced, was that frame_src was not changed when I re-opened an existing cart. This always made the loading-spinner stick. Originally the iframe.src was always set and that happened to trigger a load-event. To mimick this behaviour, I actively set the iframe.src on frame_loading. Probably not the most elegant and definitely not the proper solution. I will look into this more. Sorry!

wiffbi: I totally reworked that logic. I do not set frame_loading manually, but decide on the value (and previously value) of frame_src. When manually setting the frame.src even with the same value, it relaoded. For the new mechanism with the frame_src-watcher, one needs to unset the URL of the iframe.src when closed. Does not make a difference technically to the iframe (other than that the hidden iframe contains no content). IMHO that makes the iframe-loading a bit cleaner in the code.

src/pretix/static/pretixpresale/js/widget/widget.js
wiffbi: I added focusing the button – Safari did that automatically probably due to aria-live, but Firefox did not. The focus-management happens in the `@after-enter="focusButton"` call in the `transition`.

raphaelm (approved):

PR #4726, opened 2024-12-29
Allow ticket QR code colour to be configured

kiancross:
This PR introduces a feature enabling users to customise the QR code colour in the ticket editor.


wiffbi:
Thank you for your PR. We will look into this. It might take some time though as this affects several PDF-creation implementations, as not only the server but also our Android-apps can create PDFs. We also need to verify how RGB-to-CMYK conversion is handled when e.g. using a black-and-white-printer – something as important as a ticket-qr-code ideally should not be printed rastered IMHO.

kiancross:
@wiffbi Let me know if there is anything I can do to speed up the review process!

raphaelm (approved):
Finally got around to create a PR for the Java implementation, pending internal review now: https://github.com/pretix/libpretixprint/pull/6

PR #4727, opened 2024-12-31
Name order and Examples Defined in PERSON_NAME_SCHEMES, Adaption to Japanese addressing system

7m4gmh:
The reasons for the corrections are as follows:

### "family_nospace_given" is incorrectly defined as having the order of “given name, family name,” with the example of “毛泽东” (Mao Zedong) written in simplified Chinese characters.

This results in names being displayed as “泽东毛” (phonetically "ZedongMao"), which is inconsistent with the example and opposite to the name order in cultural contexts where Chinese characters are used.

Furthermore, using the simplified Chinese version of “毛泽东” may be difficult to understand in regions that do not use simplified characters. (Simplified Chinese is used primarily in Mainland China and Singapore.)  Additionally, the name of a leader of the Chinese Communist Party may be considered inappropriate in some regions.

I propose replacing this example with “孫文” (Sun Wen, the official name of “Sun Yat-sen”).

The characters in this name are identical in both simplified and traditional Chinese, and they also match the character style used in Hong Kong, Macao, Japan (Kanji) and Korea (Hanja). Moreover, Sun Yat-sen is regarded as the “Father of the Nation” in both mainland China and Taiwan and is generally recognized as a politically neutral figure.

### "full_transcription" currently lists “庄司” (Shoji) as an example of a full name, but this is clearly incorrect.

“庄司” is the kanji representation of a Japanese surname "Shoji" and is not a full name. According to Japanese law, a full name must include both a family name and a given name. The only exception to this rule applies to members of the Imperial Family.

Therefore, I suggest revising the example to “明仁” (Akihito, Emperor Emeritus of Japan).

It should also be noted that in some regions of Asia, legal names consist only of given names without surnames. If pretix plans to support languages from such regions in the future, it would be more appropriate to use examples of full names from those languages in this section. (e.g., Burmese, traditional Mongolian, Javanese, etc.)

スクリーンショット 2025-01-01 3 16 45
スクリーンショット 2025-01-01 3 17 50

### DIFF
--- a/src/pretix/base/settings.py
+++ b/src/pretix/base/settings.py
@@ -3609,8 +3609,8 @@ PERSON_NAME_SCHEMES = OrderedDict([
         'concatenation': lambda d: str(d.get('full_name', '')),
         'concatenation_all_components': lambda d: str(d.get('full_name', '')) + " (" + d.get('latin_transcription', '') + ")",
         'sample': {
-            'full_name': '明仁',
-            'latin_transcription': 'Akihito',
+            'full_name': 'Aye',
+            'latin_transcription': 'Aye',
             '_scheme': 'full_transcription',
         },
     }),
--- a/src/pretix/base/settings.py
+++ b/src/pretix/base/settings.py
@@ -3557,8 +3557,8 @@ PERSON_NAME_SCHEMES = OrderedDict([
             str(p) for p in [d.get('family_name', ''), d.get('given_name', '')] if p
         ),
         'sample': {
-            'given_name': '泽东',
-            'family_name': '毛',
+            'family_name': '孫',
+            'given_name': '文',
             '_scheme': 'family_nospace_given',
         },
     }),
@@ -3609,8 +3609,8 @@ PERSON_NAME_SCHEMES = OrderedDict([
         'concatenation': lambda d: str(d.get('full_name', '')),
         'concatenation_all_components': lambda d: str(d.get('full_name', '')) + " (" + d.get('latin_transcription', '') + ")",
         'sample': {
-            'full_name': '庄司',
-            'latin_transcription': 'Shōji',
+            'full_name': '明仁',
+            'latin_transcription': 'Akihito',
             '_scheme': 'full_transcription',
         },
     }),
--- a/src/pretix/base/settings.py
+++ b/src/pretix/base/settings.py
@@ -3609,8 +3609,8 @@ PERSON_NAME_SCHEMES = OrderedDict([
         'concatenation': lambda d: str(d.get('full_name', '')),
         'concatenation_all_components': lambda d: str(d.get('full_name', '')) + " (" + d.get('latin_transcription', '') + ")",
         'sample': {
-            'full_name': 'Aye',
-            'latin_transcription': 'Aye',
+            'full_name': '山田花子',
+            'latin_transcription': 'Yamada Hanako',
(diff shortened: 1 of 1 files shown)


raphaelm:
Thank you very much! We'll get back to this next week after winter vacation.

7m4gmh:
Thank you as always. Your concerns are very well-founded.
I've revised the file.

The example of a name without a family name has been changed to ‘Aye,’ a very common name in Myanmar.  This name is gender-neutral and is comparable to ‘John’ or ‘Mary’ in English as a typical first name.  It should also be noted that Burmese people do not have family names.    (In Burmese transcript, this name is written as 'အေး'.  It looks just like '632:'!)

However, I am unsure if this change will be acceptable to everyone.  This is because, living in Asia, I can easily understand the existence of such names, but for those in Europe, it might seem like an unusual name.

raphaelm:
> Thank you as always. Your concerns are very well-founded.
> I've revised the file.

Ah, sorry, I think we still have a misunderstanding. I think this makes it worse? Now both fields of the example contain the same name. That makes it hard to use the example to help understanding how the name works.

I hope it's okay if I use your name as an example at hand, I only use it because it's on your website so I assume your full name is not private information (happy to delete the comment otherwise).

The intention of this name format was:

- full_name: 梅本聖
- latin_stranscription: Hijiri Umemoto

To paraphrase: The idea was to allow users to enter both their *native* spelling of the name (in whatever language or name format they like!) as well as a latin-character-set friendly version of the name (in whatever order they like). One purpose of this at e.g. an international scientific conference would be to print both on the name tag of a person, so they have their name in its original form on their name tag (which is a respectful thing to allow them to have) as well as a version that allows people from other countries to read or search that name.

Now, obviously, neither the "Aye" example nor the "Akihito" example is *wrong*, but I think they fall short at illustrating the concept. "Aye", because no difference between the fields is necessary, "Akihito" because it uses an edge-case of the definition of "full name" which might confuse people who are aware of the differences in Japan.

The fact that we have these misunderstandings might mean that the idea behind this name format is flawed or not as useful as we thought it was – in which case we'd be happy to add more formats, but at least for backwards compatibility I'd like to keep it and improve on the example / description.

7m4gmh:
Thank you for your detailed explanation regarding the intention and purpose behind the name format. I now have a much clearer understanding that this system is designed to support both “native script” and “Latin script” name representations, enhancing usability in international contexts.

As for using my name as an example, I have no objections since it is already publicly available on the website. In fact, I’d be glad if it helps clarify the concept in a meaningful way!

raphaelm:
Ah, I was using your name as an example for the purposes of this comment here, I would also be fine with it making it into the codebase but if you have a suggestion for a good generic example, that would also be fine. Your name was just the one I was most confident making a "correct" example here, seeing that I speak no Japanese at all. (I had two years of Mandarin in school, which means I can tell the difference between Kanji from Kana but that is absolutely where my knowledge ends :D)

Are there some name combinations in Japan that is so common that it's a good example, like there are tens of thousands of "James Smith" in the US?

7m4gmh:
In Japan, we have names like ‘Yamada Taro’ (male) and ‘Yamada Hanako’ (female), which function similarly to ‘John Doe’ in English. While there aren’t many real people with those names, everyone understands them as sample names.

raphaelm:
Perfect!

PR #4735, opened 2025-01-07
Add option to restrict anonymous access to order URLs

kiancross:
By default, users who place orders while logged in can still access their order URLs without authentication. This raises potential security risks, particularly if order confirmation emails are forwarded.

This commit introduces an organiser-level setting to disable anonymous access for such orders. When enabled, unauthenticated attempts to access URLs starting with `/order/`, which are intended for the customer, are redirected to the login page. Upon successful authentication, the user is redirected back to the original order URL.

It is important to note that this change does not impact routes intended for attendees (e.g., `/ticket/*`), which remain accessible without authentication.

Implements feature discussed in https://github.com/pretix/pretix/discussions/4725.


kiancross:
Are we aiming for full testing coverage on the PR?

raphaelm:
> Are we aiming for full testing coverage on the PR?

No, it's okay to apply some human thinking to where testing makes more sense or less sense. No need to pass 100%!

kiancross:
In that case, I will add a test later this week for admin access to the order URL. Then, I think all reasonable test cases are covered (but welcome feedback if you think I have missed any).

kiancross:
@raphaelm I've added the test for admin access, so I think this is ready for you to review.

kiancross:
@raphaelm Merge conflict fixed.

kiancross:
Thanks for the review. I'll take a look at the minor comments in the coming days.
 
> If the new option is set, what do we even expect for this scenario? Do we need to force them to activate their account and log in before they can proceed with the order (meh, bad for conversion and raises some weird questions related to isolated widget sessions) or do we need to give them temporary access in some other way? How do we make the customer experience of this case acceptable?

I think, in this case, we treat the order in the same way as when it is not associated with a customer's account: allow access. It sounds like the rationale for not logging the customer in automatically is that the account's state can't be considered fully 'created' until the email is verified. And so long as that is the case, I think it is acceptable to allow access. Once the account is verified and we know the account is in a 'good' state, the login is forced.

How does that sound?

raphaelm:
> I think, in this case, we treat the order in the same way as when it is not associated with a customer's account: allow access. It sounds like the rationale for not logging the customer in automatically is that the account's state can't be considered fully 'created' until the email is verified. And so long as that is the case, I think it is acceptable to allow access. Once the account is verified and we know the account is in a 'good' state, the login is forced.

Ah, so your idea is "the protection only kicks in if `customer.is_verified`"? That should work :thinking:  I was thinking more complicated with some kind of temporary token in the session after the order, but that's much more complicated and error-prone.

kiancross:
> > I think, in this case, we treat the order in the same way as when it is not associated with a customer's account: allow access. It sounds like the rationale for not logging the customer in automatically is that the account's state can't be considered fully 'created' until the email is verified. And so long as that is the case, I think it is acceptable to allow access. Once the account is verified and we know the account is in a 'good' state, the login is forced.
> 
> Ah, so your idea is "the protection only kicks in if `customer.is_verified`"? That should work 🤔 I was thinking more complicated with some kind of temporary token in the session after the order, but that's much more complicated and error-prone.

Yes this is what I was thinking. If you're happy with the approach, I'll go ahead, and also address your other comments!

raphaelm:
Yeah, sounds good! Let's try if people break it in the real world :D

src/pretix/presale/views/order.py
```diff
@@ -236,8 +261,13 @@ class OrderDetails(EventViewMixin, OrderDetailMixin, CartMixin, TicketPageMixin,
...
     def get(self, request, *args, **kwargs):
         self.kwargs = kwargs
-        if not self.order:
+        if self.order is None:
             raise Http404(_('Unknown order code or not authorized to access this order.'))
+        elif self.order is False:
+            return redirect_to_url(
+                eventreverse(self.request.organizer, 'presale:organizer.customer.login', kwargs={}) +
+                '?next=' + quote(self.request.path_info + '?' + self.request.GET.urlencode())
+            )
```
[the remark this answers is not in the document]
kiancross: I've addressed this - most of it could be refactored into the `OrderDetailsMixin`.

kiancross:
Hi @raphaelm. Happy New Year! I've finally had time to address your comments and so this is ready for you to review again.

src/pretix/presale/views/order.py
```diff
@@ -154,8 +174,13 @@ def get_position_url(self):
...
     def get(self, request, *args, **kwargs):
-        if not self.order:
+        if self.order is None:
             raise Http404(_('Unknown order code or not authorized to access this order.'))
+        elif self.order is False:
+            return redirect_to_url(
+                eventreverse(self.request.organizer, 'presale:organizer.customer.login', kwargs={}) +
+                '?next=' + quote(self.request.path_info + '?' + self.request.GET.urlencode())
+            )
         if self.order.check_email_confirm_secret(kwargs.get('hash')) and not self.order.email_known_to_work:
```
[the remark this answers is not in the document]
kiancross: I'm not 100% sure what you meant by this comment, however, now that this has been refactored, it might be inadvertantly resolved. Let me know if not!

src/pretix/base/settings.py
```diff
@@ -171,6 +171,17 @@ def unserialize(cls, s):
...
         )
     },
+    'customer_accounts_require_login_for_order_access': {
+        'default': 'False',
+        'type': bool,
+        'form_class': forms.BooleanField,
+        'serializer_class': serializers.BooleanField,
+        'form_kwargs': dict(
+            label=_("Require login to access order confirmation pages"),
+            help_text=_("If enabled, users who were logged in at the time of purchase must also log in to access their order information."),
```
raphaelm:
```suggestion
            help_text=_("If enabled, users who were logged in at the time of purchase must also log in to access their order information. "
                        "If a customer account is created while placing an order, the restriction only becomes active after the customer account was activated."),
```

src/pretix/presale/views/order.py
```diff
@@ -100,13 +101,44 @@
...
             )
+
+            if has_event_access_permission(self.request, 'can_view_orders'):
+                return order
+
+            if order.customer is None or not order.customer.is_verified or self._allow_anonymous_access():
+                return order
+
+            if not self.request.customer:
+                return False
```
raphaelm, line 146: Why does this function sometimes return False and sometimes None now?

src/pretix/presale/views/order.py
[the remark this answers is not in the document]
raphaelm: I don't understand it any more either.

src/tests/presale/test_orders.py
```diff
@@ -1605,3 +1605,286 @@ def test_answer_download_token(self):
...
+
+        self.order.customer = customer
+        self.order.save()
+
+        self.orga.settings.customer_accounts_require_login_for_order_access = False
+
+        response = self.client.get(
+            '/%s/%s/ticket/%s/%s/%s/' % (self.orga.slug, self.event.slug, self.order.code,
+                                         self.ticket_pos.positionid, self.ticket_pos.web_secret)
+        )
```
raphaelm: For brevity and maintainability, could we merge these tests into one that does just both HTTP requests? Same for the ones below where we test the same scenario for order and position

src/tests/presale/test_orders.py
```diff
@@ -1605,3 +1605,286 @@ def test_answer_download_token(self):
...
+            '/%s/%s/order/%s/%s/' % (self.orga.slug, self.event.slug, self.order.code, self.order.secret)
+        )
+        assert response.status_code == 200
+        doc = BeautifulSoup(response.content.decode(), "lxml")
+        assert len(doc.select(".cart-row")) > 0
+        assert "pending" in doc.select(".order-details")[0].text.lower()
+        assert "Peter" in response.content.decode()
+        assert "Lukas" not in response.content.decode()
+
+    def test_require_login_for_order_access_disabled_position_auth(self):
```
raphaelm: This case is not really interesting, honestly, but why not

src/pretix/presale/views/order.py
```diff
@@ -100,13 +101,44 @@
...
+    def verify_order_access(self):
+        o = self.order
+
+        if o is None:
+            raise Http404(_('Unknown order code or not authorized to access this order.'))
+
+        if o is False:
+            login_url = eventreverse(self.request.organizer, 'presale:organizer.customer.login', kwargs={})
+            next_url = quote(self.request.get_full_path())
+            return redirect_to_url(f'{login_url}?next={next_url}')
```
raphaelm: This breaks if the event page and the organizer page are on different domains, because the login will only be valid for the organizer domain and the event page is never accessible. We have a flow to support this, with this change, it will work:

```suggestion
            if self.request.event_domain:
                next_url = quote(self.request.scheme + "://" + self.request.get_host() + self.request.get_full_path())
                return redirect_to_url(f'{login_url}?next={next_url}&request_cross_domain_customer_auth=true')
            else:
                next_url = quote(self.request.get_full_path())
                return redirect_to_url(f'{login_url}?next={next_url}')
```

src/pretix/presale/views/order.py
kiancross: The function returns `order` if the order can be accessed, `False` if the order requires authentication to access, and `None` if the order cannot be accessed at all.

This is required for the checks in `verify_order_access` (and has always been the case in this PR), although I did refactor the code in bcf30a67e66731a66f221e550ef6b8bdf79214a4 to remove nesting.

raphaelm: Worth a docstring then probably ;)

raphaelm: Worth a docstring then probably ;)

src/pretix/presale/views/order.py
kiancross: I slightly changed the fix due an exception. Would be worth that you're happy with the amended version!

src/pretix/presale/views/order.py
kiancross: Done!

src/pretix/presale/views/order.py
```diff
@@ -99,17 +99,57 @@
...
+
+        return None
+
+    """
+    Returns the order object when access is permitted, returns `False` when the
+    order exists but requires authentication, and returns `None` when the order
+    does not exist or access is denied entirely.
+    """
     @cached_property
     def order(self):
```
raphaelm:
```suggestion
@cached_property
def order(self):
        """
        Returns the order object when access is permitted, returns `False` when the
        order exists but requires authentication, and returns `None` when the order
        does not exist or access is denied entirely.
        """
```

PR #4740, opened 2025-01-13
New invoice generation setting: User request but paid only

raphaelm:


luelista (changes requested):
`pretix/src/pretix/control/views/orders.py:1587` should also check for the `user_paid` option

raphaelm:
Oops, how did I miss this, thanks!

PR #4744, opened 2025-01-13
Invoice renderer: Group invoice lines even with addons (Z#23173618)

raphaelm:
Before:
image

After:
image

### DIFF
--- a/src/pretix/base/invoice.py
+++ b/src/pretix/base/invoice.py
@@ -99,22 +99,17 @@ def addon_aware_groupby(iterable, key, is_addon):
 
     for i in iterable:
         if is_addon(i):
-            packed_groups[-1][1].append(i)
+            packed_groups[-1].append(i)
         else:
-            packed_groups.append((i, []))
-    # Each packed_groups element contains (parent product, list of addon products)
+            packed_groups.append([i])
+    # Each packed_groups element contains a list with the parent product as first element, and any addon products following
 
     def _reorder(packed_groups):
         # Emit the products as individual products again, reordered by "all parent products, then all addon products"
         # within each group.
-        for group_grouper, grouped in groupby(packed_groups, key=lambda g: key(g[0]) + tuple(key(a) for a in g[1])):
-            grouped = list(grouped)
-            for grp in grouped:
-                yield grp[0]
-            for i in range(max(len(grp[1]) for grp in grouped)):
-                for grp in grouped:
-                    if len(grp[1]) > i:
-                        yield grp[1][i]
+        for _, repeated_groups in groupby(packed_groups, key=lambda g: tuple(key(a) for a in g)):
+            for repeated_items in zip(*repeated_groups):
+                yield from repeated_items
 
     return groupby(_reorder(packed_groups), key)
 
--- a/src/pretix/base/invoice.py
+++ b/src/pretix/base/invoice.py
@@ -62,6 +62,63 @@ from pretix.presale.style import get_fonts
 logger = logging.getLogger(__name__)
 
 
+def addon_aware_groupby(iterable, key, is_addon):
+    """
+    We use groupby() to visually group identical lines on an invoice. For example, instead of
+
+    Product 1       5.00 EUR
+    Product 1       5.00 EUR
+    Product 1       5.00 EUR
+    Product 2       7.00 EUR
+
+    We want to print
+
+    3x Product 1    5.00 EUR = 15.00 EUR
+    Product 2       7.00 EUR
+
+    However, this fails for setups with addon-products since groupby() only groups consecutive
+    lines with the same identity. So in
+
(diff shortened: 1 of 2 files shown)


src/pretix/base/invoice.py
```diff
@@ -62,6 +62,63 @@
...
+    # Each packed_groups element contains (parent product, list of addon products)
+
+    def _reorder(packed_groups):
+        # Emit the products as individual products again, reordered by "all parent products, then all addon products"
+        # within each group.
+        for group_grouper, grouped in groupby(packed_groups, key=lambda g: key(g[0]) + tuple(key(a) for a in g[1])):
+            grouped = list(grouped)
+            for grp in grouped:
+                yield grp[0]
+            for i in range(max(len(grp[1]) for grp in grouped)):
```
felixrindt, line 114: Yikes. This line was hard to understand. Just as an exercise I thought about how I would do it: using `zip_longest` and checking for None instead of the index :3

luelista (approved):

src/pretix/base/invoice.py
```diff
@@ -62,6 +62,63 @@
 logger = logging.getLogger(__name__)
 
 
+def addon_aware_groupby(iterable, key, is_addon):
```
luelista, line 65:
```suggestion
parent_and_addons = namedtuple('parent_and_addons', ('parent', 'addons'))


def addon_aware_groupby(iterable, key, is_addon):
```

src/pretix/base/invoice.py
```diff
@@ -62,6 +62,63 @@
...
+        # Emit the products as individual products again, reordered by "all parent products, then all addon products"
+        # within each group.
+        for group_grouper, grouped in groupby(packed_groups, key=lambda g: key(g[0]) + tuple(key(a) for a in g[1])):
+            grouped = list(grouped)
+            for grp in grouped:
+                yield grp[0]
+            for i in range(max(len(grp[1]) for grp in grouped)):
+                for grp in grouped:
+                    if len(grp[1]) > i:
+                        yield grp[1][i]
```
luelista, line 117:
```suggestion
    for i in iterable:
        if is_addon(i):
            packed_groups[-1].addons.append(i)
        else:
            packed_groups.append(parent_and_addons(parent=i, addons=[]))
    # Each packed_groups element contains (parent product, list of addon products)

    def _reorder(packed_groups):
        # Emit the products as individual products again, reordered by "all parent products, then all addon products"
        # within each group.
        for group_grouper, grouped in groupby(packed_groups, key=lambda g: key(g.parent) + tuple(key(a) for a in g.addons)):
            grouped = list(grouped)
            for grp in grouped:
                yield grp.parent
            for i in range(max(len(grp.addons) for grp in grouped)):
                for grp in grouped:
                    if len(grp.addons) > i:
                        yield grp.addons[i]
```

luelista (changes requested):
It might be more readable to use a namedtuple here

src/pretix/base/invoice.py
luelista: The addon lists (contents of `grouped[n][1]`) are always the same length, by definition (because we use them as grouping key above). Therefore we can just zip them and don't need to filter at all:
```
            parents, addon_groups = zip(*grouped)
            yield from parents
            for addons in zip(*addon_groups):
                yield from addons
```

src/pretix/base/invoice.py
luelista:
```suggestion
    for i in iterable:
        if is_addon(i):
            packed_groups[-1].addons.append(i)
        else:
            packed_groups.append(parent_and_addons(parent=i, addons=[]))
    # Each packed_groups element contains (parent product, list of addon products)

    def _reorder(packed_groups):
        # Emit the products as individual products again, reordered by "all parent products, then all addon products"
        # within each group.
        for group_grouper, grouped in groupby(packed_groups, key=lambda g: key(g.parent) + tuple(key(a) for a in g.addons)):
            parents, addon_groups = zip(*grouped)
            yield from parents
            for addons in zip(*addon_groups):
                yield from addons
```

luelista, line 117: Even simpler way to do it:

```suggestion
    for i in iterable:
        if is_addon(i):
            packed_groups[-1].append(i)
        else:
            packed_groups.append([i])
    # Each packed_groups element contains a list with the parent product as first element, and any addon products following

    def _reorder(packed_groups):
        # Emit the products as individual products again, reordered by "all parent products, then all addon products"
        # within each group.
        for _, repeated_groups in groupby(packed_groups, key=lambda g: tuple(key(a) for a in g)):
            for repeated_items in zip(*repeated_groups):
                yield from repeated_items
```

luelista (changes requested):
Alternatively, as we don't really need to special case the parent product, there is an even simpler way to do it:

PR #4748, opened 2025-01-14
Fix typo in documentation

n-thumann:
This PR fixes a typo in the documentation, just noticed that during reading :)


raphaelm:
It's fine, we don't require one for docs, the bot just isn't clever enough

PR #4756, opened 2025-01-16
Allow 0% tax rate on event creation

luelista:
(still warn if tax rate is not filled at all)


raphaelm:
Should pretix/control/views/main.py:316 then also be a none check so a 0% rule is created?

luelista:
oops, yes :face_with_peeking_eye:

raphaelm (approved):

PR #4762, opened 2025-01-17
Invoice rendering: Remove transparency from logos (Z#23179391)

raphaelm:
Valid ZUGFeRD invoices must be compliant with PDF/A-3. pretix-zugferd ensures this by passing them through ghost script. Unfortunately, if the logo contains transparency, this will still fail with:

```
      <validationReport jobEndStatus="normal" profileName="PDF/A-3B validation profile" statement="PDF file is not compliant with Validation Profile requirements." isCompliant="false">
        <details passedRules="144" failedRules="1" passedChecks="4239" failedChecks="1">
          <rule specification="ISO 19005-3:2012" clause="6.2.10" testNumber="2" status="failed" failedChecks="1">
            <description>If the document does not contain a PDF/A OutputIntent, then all Page objects that contain transparency shall include the Group key, and the attribute dictionary that forms the value of that Group key shall include a CS entry whose value shall be used as the default blending colour space</description>
            <object>PDPage</object>
            <test>gOutputCS != null || containsGroupCS == true || containsTransparency == false</test>
            <check status="failed">
              <context>root/document[0]/pages[0](4 0 obj PDPage)</context>
              <errorMessage>The page contains transparent objects with no blending colour space defined</errorMessage>
            </check>
          </rule>
        </details>
      </validationReport>
```

I was unable to figure out a way to fix this in GhostScript, so the easy fix is to remove the transparency, as our invoices always have a white background anyways.

### DIFF
--- a/src/pretix/base/invoice.py
+++ b/src/pretix/base/invoice.py
@@ -389,6 +389,10 @@ class ClassicInvoiceRenderer(BaseReportlabInvoiceRenderer):
                 logger.exception("Can not resize image")
                 pass
             try:
+                # Valid ZUGFeRD invoices must be compliant with PDF/A-3. pretix-zugferd ensures this by passing them
+                # through ghost script. Unfortunately, if the logo contains transparency, this will still fail.
+                # I was unable to figure out a way to fix this in GhostScript, so the easy fix is to remove the
+                # transparency, as our invoices always have a white background anyways.
                 ir.remove_transparency()
             except:
                 logger.exception("Can not remove transparency from logo")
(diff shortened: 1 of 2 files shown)


wiffbi (approved):
I looked into this whether there is an alternative by including an OutputIntent, but did not manage to get it working. So I guess this is a good, pragmatic approach. Just maybe add a comment on why we remove the transparency?

As an alternative I tried to convert the invoice-PDF in pretix-zufgerd to `-dCompatibilityLevel=1.3` in the hopes this will remove transparency before running the PDF-A ghostscript, but that somehow turned everything into an image. So, I did not look further.

PR #4763, opened 2025-01-17
Generate invoice earlier in payment method change process (Z#23179304)

raphaelm:


wiffbi (approved):
Looks good. I would change the naming of the boolean var new_invoice to is_new_invoice to make a better distinction to the other object-vars new_fee, etc.

PR #4768, opened 2025-01-20
Define LogEntryTypes for all actions in pretix core, improve content_object handling

luelista:
This creates LogEntryType definitions for all missing `action_type`s (order changes, check-in events, settings changes of PaymentProviders and TicketOutputs).

This also checks whether the stored `content_object` is of the expected model type, preventing incorrect links. 

Refactoring:
- Moving the base LogEntryType definitions for our models to their own file was required to allow referencing model classes without import loops.
- HTML escaping was moved into `make_link` to make it less likely to oversee in the LogEntryType definitions
- pretix.event.order.deleted is now logged with the deleted Order model as content_object, matching the other `*.deleted` action_types

### DIFF
--- a/src/pretix/control/logdisplay.py
+++ b/src/pretix/control/logdisplay.py
@@ -96,8 +96,7 @@ class OrderItemChanged(OrderChangeLogEntryType):
         new_item = str(event.items.get(pk=data['new_item']))
         if data['new_variation']:
             new_item += ' - ' + str(ItemVariation.objects.get(item__event=event, pk=data['new_variation']))
-        return _('Position #{posid}: {old_item} ({old_price}) changed '
-                              'to {new_item} ({new_price}).').format(
+        return _('Position #{posid}: {old_item} ({old_price}) changed to {new_item} ({new_price}).').format(
             posid=data.get('positionid', '?'),
             old_item=old_item, new_item=new_item,
             old_price=money_filter(Decimal(data['old_price']), event.currency),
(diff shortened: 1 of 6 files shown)


src/pretix/control/logdisplay.py
```diff
@@ -68,128 +68,212 @@
...
+
+    def display_prefixed(self, event: Event, logentry: LogEntry, data):
+        return _('A fee of {old_price} was removed.').format(
             old_price=money_filter(Decimal(data['old_price']), event.currency),
         )
-    elif logentry.action_type == 'pretix.event.order.changed.cancel':
+
+
+@log_entry_types.new()
+class OrderCancelled(OrderChangeLogEntryType):
```
raphaelm, line 194:
```suggestion
class OrderCanceled(OrderChangeLogEntryType):
```
both spellings are correct, but we use the one l everywhere

src/pretix/control/logdisplay.py
```diff
@@ -198,195 +282,145 @@ def _display_order_changed(event: Event, logentry: LogEntry):
...
+    def display(self, logentry: LogEntry, data):
         return _('The user confirmed the following message: "{}"').format(
-            bleach.clean(logentry.parsed_data.get('msg'), tags=set(), strip=True)
+            bleach.clean(data.get('msg'), tags=set(), strip=True)
         )
 
-    if logentry.action_type == 'pretix.event.order.canceled':
-        comment = logentry.parsed_data.get('comment')
+
+class OrderCancelledLogEntryType(OrderLogEntryType):
```
raphaelm, line 389:
```suggestion
class OrderCancelledLogEntryType(OrderLogEntryType):
```
both spellings are correct, but we use the one l everywhere

src/pretix/control/logdisplay.py
```diff
@@ -486,17 +521,16 @@ class VoucherRedeemedLogEntryType(VoucherLogEntryType):
...
 
-    def display(self, logentry):
-        data = json.loads(logentry.data)
-        data = defaultdict(lambda: '?', data)
+    def display(self, logentry, data):
         url = reverse('control:event.order', kwargs={
             'event': logentry.event.slug,
             'organizer': logentry.event.organizer.slug,
-            'code': data['order_code']
+            'code': data('order_code', '?')
```
raphaelm, line 528:
```suggestion
            'code': data.get('order_code', '?')
```

probably?

raphaelm (approved):

src/pretix/control/logdisplay.py
luelista:
```suggestion
class OrderCanceledLogEntryType(OrderLogEntryType):
```

PR #4776, opened 2025-01-23
Implement hidden_if_item_available_mode option  (Z#23177008)

luelista:
image


luelista:
Currently, it displays the same message as for items that are only available at a later date:
> Not available yet

Not sure if there is a better wording?

src/pretix/api/serializers/item.py
```diff
@@ -272,7 +272,7 @@ class Meta:
                   'require_voucher', 'hide_without_voucher', 'allow_cancel', 'require_bundling',
                   'min_per_order', 'max_per_order', 'checkin_attention', 'checkin_text', 'has_variations', 'variations',
                   'addons', 'bundles', 'original_price', 'require_approval', 'generate_tickets',
-                  'show_quota_left', 'hidden_if_available', 'hidden_if_item_available', 'allow_waitinglist',
+                  'show_quota_left', 'hidden_if_available', 'hidden_if_item_available', 'hidden_if_item_available_mode', 'allow_waitinglist',
```
raphaelm, line 275: If the API changes, docs and tests need to change, too

raphaelm (approved):
> Not sure if there is a better wording?

I couldn't come up with a better one that fits most use cases.

PR #4791, opened 2025-01-29
OIDC: Allow to add query parameters to Authorization URL

pc-coholic:
Z#23180758

I guess, we could also just blindly append the `query_parameters` to the URL and assume that everything has already been properly urlencoded... But this seems safer...

### DIFF
--- a/src/pretix/base/customersso/oidc.py
+++ b/src/pretix/base/customersso/oidc.py
@@ -139,6 +139,17 @@ def oidc_validate_and_complete_config(config):
                 )
             )
 
+    if "query_parameters" in config and config["query_parameters"]:
+        for qp in config["query_parameters"].split("&"):
+            # Very rudimentary check to avoid the most common footguns:
+            # - No ? in the query parameters
+            # - Max of one = (to split key and value)
+            # - One key, one value. Not just keys with no value.
+            if not (qp.count('=') == 1 and qp.count('?') == 0 and len(list(filter(None, qp.split('=')))) == 2):
+                raise ValidationError(
+                    _(f'Query parameter {qp} is invalid.')
+                )
+
     config['provider_config'] = provider_config
     return config
 
@@ -154,6 +165,11 @@ def oidc_authorize_url(provider, state, redirect_uri):
         'state': state,
         'redirect_uri': redirect_uri,
     }
+
+    if "query_parameters" in provider.configuration and provider.configuration["query_parameters"]:
+        for qp in provider.configuration["query_parameters"].split("&"):
+            params[qp.split("=")[0]] = qp.split("=")[1]
+
     return endpoint + '?' + urlencode(params)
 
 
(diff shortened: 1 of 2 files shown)


src/pretix/base/customersso/oidc.py
```diff
@@ -139,6 +139,17 @@ def oidc_validate_and_complete_config(config):
...
+        for qp in config["query_parameters"].split("&"):
+            # Very rudimentary check to avoid the most common footguns:
+            # - No ? in the query parameters
+            # - Max of one = (to split key and value)
+            # - One key, one value. Not just keys with no value.
+            if not (qp.count('=') == 1 and qp.count('?') == 0 and len(list(filter(None, qp.split('=')))) == 2):
+                raise ValidationError(
+                    _(f'Query parameter {qp} is invalid.')
+                )
+
```
luelista, line 152: I'd just parse and re-encode it using the library functions -- so the user sees what will be transmitted in the end:
```suggestion
    if "query_parameters" in config and config["query_parameters"]:
         config["query_parameters"] = urllib.parse.urlencode(
             urllib.parse.parse_qsl(config["query_parameters"])
         )
```

src/pretix/base/customersso/oidc.py
```diff
@@ -154,6 +165,11 @@ def oidc_authorize_url(provider, state, redirect_uri):
         'state': state,
         'redirect_uri': redirect_uri,
     }
+
+    if "query_parameters" in provider.configuration and provider.configuration["query_parameters"]:
+        for qp in provider.configuration["query_parameters"].split("&"):
+            params[qp.split("=")[0]] = qp.split("=")[1]
+
```
luelista, line 172: Again, I'd recommend using parse_qsl. This also allows the user to actually use correctly escaped values. Otherwise, they would be double-escaped here.
```suggestion

    if "query_parameters" in provider.configuration and provider.configuration["query_parameters"]:
        params.update(urllib.parse.parse_qsl(provider.configuration["query_parameters"]))

```

src/pretix/base/customersso/oidc.py
pc-coholic: I was thinking yesterday that there was such a function in urllib, but apparently I was too cooked to actually find it... Thank you for the proposal; that's a lot better!

src/pretix/base/customersso/oidc.py
```diff
@@ -139,6 +139,10 @@ def oidc_validate_and_complete_config(config):
                 )
             )
 
+    if "query_parameters" in config and config["query_parameters"]:
+        config["query_parameters"] = urlencode(
+            parse_qsl(config["query_parameters"])
+        )
     config['provider_config'] = provider_config
```
luelista, line 146:
```suggestion

    config['provider_config'] = provider_config
```

luelista (approved):

PR #4818, opened 2025-02-07
Fix log display for checkins and order splits (Z#23181229)

luelista:
- Fixes link in pretix.event.order.changed.split
- Adds link to existing order in pretix.event.order.changed.split_from
- Fixes display of checkin entries without datetime in data
- Adds additional info for admins (action type, linked content object):
  image


wiffbi (approved):
Cool, I like the „syntax-highlighting“ for those JSON-strings.

PR #4822, opened 2025-02-07
Allow plugins to add data to the order API (Z#23179688)

raphaelm:


src/pretix/api/serializers/media.py
```diff
@@ -47,6 +47,9 @@ class Meta:
 class NestedOrderPositionSerializer(OrderPositionSerializer):
     order = NestedOrderMiniSerializer()
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
```
wiffbi, line 52: I guess this can be omitted?
```suggestion
```

wiffbi (approved):
Looks good, just a minor comment ragarding a superflous `__init__`

PR #4824, opened 2025-02-07
OIDC: Implement PKCE in OP and RP

raphaelm:
~This is implemented against the spec, but not tested against third-party servers yet~


raphaelm:
Ok, validated the client implementation against Okta server.

I don't have a reliable client implementation to validate our server against at hand, so I'd trust the tests there.

pc-coholic (changes requested):
This breaks for example with LinkedIn.

I think we should not blindly send PKCE if the OIDC-endpoint does not advertise its support for it.

https://www.linkedin.com/oauth/.well-known/openid-configuration vs. https://accounts.google.com/.well-known/openid-configuration

The latter does contain a section
```json
"code_challenge_methods_supported": [
    "plain",
    "S256"
],
```

raphaelm:
> I think we should not blindly send PKCE if the OIDC-endpoint does not advertise its support for it. 

That's what RFC 7636 wants :roll_eyes:

```
   As the OAuth 2.0 [[RFC6749](https://www.rfc-editor.org/rfc/rfc6749)] server responses are unchanged by this
   specification, client implementations of this specification do not
   need to know if the server has implemented this specification or not
   and SHOULD send the additional parameters as defined in [Section 4](https://www.rfc-editor.org/rfc/rfc7636#section-4) to
   all servers.
```

I'll change it ;)

pc-coholic:
Standards. So much more fun if people actually followed them :grin:

pc-coholic (approved):
Works for me now :)

PR #4832, opened 2025-02-11
Improve efficiency of bulk operations

raphaelm:


src/pretix/base/models/log.py
```diff
@@ -165,6 +165,16 @@ def parsed_data(self):
...
 
+    @classmethod
+    def bulk_create_and_postprocess(cls, objects):
+        if connections['default'].features.can_return_rows_from_bulk_insert:
+            cls.objects.bulk_create(objects)
+            cls.bulk_postprocess(objects)
+        else:
+            for le in objects:
+                le.save()
+            cls.bulk_postprocess(objects)
```
wiffbi, line 176: One could extract the `cls.bulk_postprocess(objects)` from the if-else to make it a tiny bit more obvious that only the create is different in bulk.

```suggestion
        else:
            for le in objects:
                le.save()
        cls.bulk_postprocess(objects)
```

src/pretix/control/views/subevents.py
```diff
@@ -657,47 +657,59 @@ class SubEventBulkAction(SubEventQueryMixin, EventPermissionRequiredMixin, View)
...
                         'active': True
-                    }
-                )
+                    }, save=False
+                ))
                 obj.active = True
                 obj.save(update_fields=['active'])
+
+            LogEntry.bulk_create_and_postprocess(log_entries)
             messages.success(request, pgettext_lazy('subevent', 'The selected dates have been enabled.'))
```
wiffbi, line 684: Wouldn’t it make sense to combine these two (enable/disable). Code-wise it is just one boolean and a swap in success-messages:

```
        if request.POST.get('action') == 'disable' or request.POST.get('action') == 'enable':
            active = request.POST.get('action') == 'enable'
            log_entries = []
            for obj in self.get_queryset():
                log_entries.append(obj.log_action(
                    'pretix.subevent.changed', user=self.request.user, data={
                        'active': active
                    }, save=False
                ))
                obj.active = active
                obj.save(update_fields=['active'])

            LogEntry.bulk_create_and_postprocess(log_entries)
            if active:
                msg = pgettext_lazy('subevent', 'The selected dates have been enabled.')
            else:
                msg = pgettext_lazy('subevent', 'The selected dates have been disabled.')
            messages.success(request, msg)
```
Wouldn’t it even be possible to bulk-change the active state?

src/pretix/control/views/subevents.py
```diff
@@ -657,47 +657,59 @@ class SubEventBulkAction(SubEventQueryMixin, EventPermissionRequiredMixin, View)
...
                             'active': False
-                        }
-                    )
+                        }, save=False,
+                    ))
                     obj.active = False
                     obj.save(update_fields=['active'])
+
+            if to_delete:
+                CartPosition.objects.filter(addon_to__subevent_id__in=to_delete).delete()
```
wiffbi, line 709: This was in the original code already, but don’t addons have the subevent_id as well? I probably don’t have enough data in my local database, especially not added through the API to verify this, so I guess we better keep this?

src/pretix/control/views/vouchers.py
```diff
@@ -619,19 +619,22 @@ def post(self, request, *args, **kwargs):
...
             })
         elif request.POST.get('action') == 'delete_confirm':
+            log_entries = []
             for obj in self.objects:
                 if obj.allow_delete():
-                    obj.log_action('pretix.voucher.deleted', user=self.request.user)
+                    log_entries.append(obj.log_action('pretix.voucher.deleted', user=self.request.user, save=False))
                     CartPosition.objects.filter(addon_to__voucher=obj).delete()
                     obj.cartposition_set.all().delete()
                     obj.delete()
```
wiffbi, line 628: Should we add a `to_delete` bulk-delete similar to the one in subevents?

src/pretix/control/views/waitinglist.py
```diff
@@ -160,10 +160,15 @@ def post(self, request, *args, **kwargs):
...
-                    obj.delete()
+            with transaction.atomic():
+                log_entries = []
+                to_delete = []
+                for obj in self.get_queryset(force_filtered=True):
+                    if not obj.voucher_id:
+                        log_entries.append(obj.log_action('pretix.event.orders.waitinglist.deleted', user=self.request.user, save=False))
+                        to_delete.append(obj.pk)
+                LogEntry.bulk_create_and_postprocess(log_entries)
+                WaitingListEntry.objects.filter(id__in=to_delete).delete()
```
wiffbi, line 171: Probably does not make difference, but in the above code-parts we first do the change and then write the log. Maybe a micro-optimization regarding the transaction, but I guess if something fails it is in the delete-call and that way, the log-entries would not have to rollback?

wiffbi (approved):
Looks good, just a few questions around the optimizations. Other than that, it looks fine and works – I only tested subevent create/delete though.

src/pretix/control/views/subevents.py
raphaelm: Honestly, I'm torn if it's easier to read/follow.

src/pretix/control/views/subevents.py
raphaelm: Weirdly, we seem to need to enforce deletion order in some cases, i.e. delete the addons first, then the regular ones, otherwise we get a foreign key issue. I think this shouldn't be necessary theoretically, but it comes up in sentry often enough, I don't think I added these for fun :\

PR #4837, opened 2025-02-12
API: Fix crash on invalid input (PRETIXEU-B7Q)

raphaelm:


wiffbi (approved):
I probably would have moved the `isinstance`-check to the start of the if to circumvent unnecessary string-search, but I guess this happens so rarely and doesn’t cost much resources it is not even worth the effort to think about it. :)

PR #4844, opened 2025-02-13
Address form: Add provinces for Italy

raphaelm:
In Italy, the province is part of the mailing address:
http://www.bitboost.com/ref/international-address-formats/italy/

This is used in the same way that e.g. a US state is part of an address. So far, our field has always been named "State", even though it reflected e.g. the "Provinces" of Canada. Nobody complained about this, because structurally, the thirteen provinces by Canada are on a similar level of government as e.g. a US state. I'm sure they Canadians don't like them being named states, but they are probably used to it.

In contrast, in Italy the "province" is not structurally on the same level as a "state". The equivalent to a "state" in Italy would be the 20 "regions". However, these are not used in the addresses – the ~100 "provinces" are. So this PR also allows country-dependent naming of the field.

### DIFF
--- a/src/pretix/base/forms/questions.py
+++ b/src/pretix/base/forms/questions.py
@@ -83,8 +83,8 @@ from pretix.base.services.tax import (
     VATIDFinalError, VATIDTemporaryError, validate_vat_id,
 )
 from pretix.base.settings import (
-    COUNTRIES_WITH_STATE_IN_ADDRESS, PERSON_NAME_SALUTATIONS,
-    PERSON_NAME_SCHEMES, PERSON_NAME_TITLE_GROUPS,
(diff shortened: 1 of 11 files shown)


wiffbi (commented):
Looks good and works as expected. On the checkout-confirm page, it still shows „state“ as the template uses a static translated value. Not sure what the easiest way would be to change here – maybe even remove the `<dt>`-labels and just output the adress? the wrong label is shown in checkout-confirm, order-success and in order-overview in pretix-control (search for `{% trans "State" context "address" %}`)

Also the shipping-plugin does not yet use the new labels in confirm, etc., but only on the input form.

wiffbi (approved):

PR #4845, opened 2025-02-13
Address form: Fix "copy from above" for the "State" field

raphaelm:


wiffbi (approved):
Just a tiny change to safe some space and make the one line say „change value“.

I first thought, this should actually be somehow handled by addressform.js, but then one cannot set a value on the select, when it has that value not available yet. So best to do it here.

PR #4861, opened 2025-02-19
Fix selector for copying order/invoice answers to first ticket

### LINKED ISSUE #4860: Issues validating multi-select if a field of different type but with the same name exists.
### Problem and impact

When trying to order a ticket, if there are two custom fields:
- A "Choose multiple from a list"
- Another field with **type text** that has the **same name** as one of the selectable options of the "Choose multiple from a list"

Then, if I tick the box for the option with the conflicted name, Pretix returns an error about the string I put into the text field not being of the same type as would be required for the option ("Bitte eine gültige Auswahl treffen. `<value of text field>` ist keine gültige Auswahl."). I don't get why, since those two fields shouldn't be related at all.

### Expected behaviour

The "Choose multiple from a list" field should be validated correctly and independently of any other fields. Using Internal identifier of fields instead of the user-facing names.

### Steps to reproduce

1. Create a new event with a product
2. Go to fields and add two fields to the product:
    - "Choose multiple from a list", where one of the options is called "E-Mail"
    - "Text (one line)", with name "E-Mail"
3. Open the shop
4. Try to buy the product, filling in the "E-Mail" text field and ticking the "E-Mail" Checkbox
5. Observe validation error on the "Choose multiple from a list" field

### Screenshots

Image
Image

### Link

_No response_

_No response_

_No response_

### Version

2025.1.0

wiffbi:
This fixes #4860 

It furthermore fixes copying the selected country by triggering a change-event which refreshes the state-list as well.

Note: this ignores any checkbox/radio inputs, so if (or when) we add these type of inputs to order questions/fields, one needs to adapt this. As copying checkboxes checked states is anyway totally different to selected options/input-values, this would need to be adapted anyways then.

### DIFF
--- a/src/pretix/static/pretixpresale/js/ui/main.js
+++ b/src/pretix/static/pretixpresale/js/ui/main.js
@@ -386,22 +386,29 @@ $(function () {
         if (copy_to_first_ticket) {
             var $first_ticket_form = $(".questions-form").first().find("[data-addonidx=0]");
             $first_ticket_form.find("[id$=" + this.id.substring(3) + "]").val(this.value);
+
+            // when matching by placeholder or label, make sure to always match the tagName of the element
+            // filter any checkbox/radio inputs as copied input is either select or type=email|text, but could be
+            // matched when custom answer options have the same label, see https://github.com/pretix/pretix/issues/4860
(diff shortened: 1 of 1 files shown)


wiffbi:
See #4865 for a better approach

PR #4865, opened 2025-02-20
Refactor copy_to_first_ticket, fixes issue #4860

luelista:


src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -383,29 +389,21 @@ $(function () {
...
+        var source = this;
+        var source_label = source.placeholder || get_label_text_for_id(source);
+
+        var $first_ticket_form = $(".questions-form").first().find("[data-addonidx=0]");
+        var $candidates = $first_ticket_form.find(source.tagName + ":not([type='checkbox'], [type='radio'])");
+        var $match = $candidates.filter(function() {
+            return (
+                this.id.endsWith(source.id.substring(3))
+                || this.placeholder === source_label
+                || get_label_text_for_id(this.id) === source_label
```
luelista, line 403:
```suggestion
                || (this.placeholder || get_label_text_for_id(this.id)) === source_label
```

wiffbi: AFAICT this basically disables any label-matching as all of our inputs have a placeholder-attribute. So I would not do this. I think it’s fine how you solved it before – taking my suggestion regarding matching source.placeholder as well as source.label.

src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -383,29 +389,21 @@ $(function () {
...
-                    if (text == label) {
-                        $("#" + this.getAttribute("for")).val(v);
-                    }
-                });
-            }
-        }
+        if (!copy_to_first_ticket) return;
+
+        var source = this;
+        var source_label = source.placeholder || get_label_text_for_id(source);
```
wiffbi, line 395: This removes an important? matching bit: with this change we either match to the placeholder of the source or to the label whereas originally we matched source.placeholder==input.placeholder (line 390) as well as source.label== input.placeholder (line 397). In most cases this does not make a difference as the placeholder is either the same as the label (most fields) or the placeholder is anyways the one we want to match (name_parts). There is one exception: „Address“ has a placeholder of „Street and Number“ and we want to match that input not only with our own system question (which also has „Address“ as label and „Street and Number“ as placeholder) but with existing custom questions which should also be matched when they are being labelled „Address“. With this change, they would need to be labelled „Street and Number“. I think this should be:

```suggestion
        var source_label = get_label_text_for_id(source);
```

Also see suggestion further down.

src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -383,29 +389,21 @@ $(function () {
...
+
+        var $first_ticket_form = $(".questions-form").first().find("[data-addonidx=0]");
+        var $candidates = $first_ticket_form.find(source.tagName + ":not([type='checkbox'], [type='radio'])");
+        var $match = $candidates.filter(function() {
+            return (
+                this.id.endsWith(source.id.substring(3))
+                || this.placeholder === source_label
```
wiffbi, line 402: See above regarding comparing to source.placeholder and a source’s label.
```suggestion
                || this.placeholder === source.placeholder
                || this.placeholder === source_label
```

wiffbi (changes requested):
I have small change regarding matching to source.placeholder as well as a source’s label. Other than that, looks good. Let me know when you committed the suggestion, I want to test once more, just to make sure. Thanks!

src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -383,29 +389,22 @@ $(function () {
...
-                    if (text == label) {
-                        $("#" + this.getAttribute("for")).val(v);
-                    }
-                });
-            }
-        }
+        if (!copy_to_first_ticket) return;
+
+        var source = this;
+        var source_label = get_label_text_for_id(source);
```
wiffbi, line 395:
```suggestion
        var source_label = get_label_text_for_id(source.id);
```

src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -383,29 +389,22 @@ $(function () {
...
+        var $candidates = $first_ticket_form.find(source.tagName + ":not([type='checkbox'], [type='radio'])");
+        var $match = $candidates.filter(function() {
+            return (
+                this.id.endsWith(source.id.substring(3))
+                || this.placeholder === source.placeholder
+                || this.placeholder === source_label
+                || get_label_text_for_id(this.id) === source_label
+            );
+        }).first();
+        $match.val(this.value).trigger("change");
```
wiffbi, line 407: `trigger("change")` is needed to make the country input work, but basically disables all copying of input after the first one. See line 410. As a quick fix, we could change line 410 from 
`attendee_address_fields.change(function () {`
to
`attendee_address_fields.on("input", function () {` which is not triggered by the change-event here.

wiffbi (changes requested):
I do have two bugs commented, but while we are at refactoring this, I’d rather not even have the listeners on the inputs in the first place, if `copy_to_first_ticket == false`. And ideally, instead returning early in the event listener, we remove the eventlistener with jQuery’s `off()`. What do you think? We could even get rid of the `copy_to_first_ticket` variable and just use `if (attendee_address_fields.some(function(el){return el.value})) {…}`

Or we just leave it like that and just change my two small comments. IMHO that is good enough.

src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -383,29 +389,22 @@ $(function () {
...
     }).trigger("change");
     attendee_address_fields.change(function () {
```
luelista, line 409:
```suggestion
    attendee_address_fields.on("input", function () {
```

wiffbi (approved):

PR #4875, opened 2025-02-26
New mechanism to transfer cookie consent from the widget (Z#23181715)

raphaelm:
The current approach of the "data-consent" field of the widget has two problems:

- There is unclear precedence if there is already a value stored in the widget. We inverted the precedence in #4862, but it is still weird. The problem is, that we have different usage scenaries, such as a well-built main website that has proper cookie consent management and a user that only ever interacts with pretix through the widget on that page. In that case, we want later updates to the cookie consent choices on the main website to also apply to the widget. The other case is a user who is also interacting with the pretix directly and has already chosen their consent through the pretix UI and then stumbles over the widget and there is likely no intention of revoking prior consent. Therefore, I propose the precedence implemented in this PR:
  - If the consent is passed through the widget and this is the first interaction of the user with pretix, use that consent and save it in permanent storage.
  - If the consent is passed through the widget and there is already a different consent stored for the user with pretix, use the consent passed through the widget *but only for the running tab*, without changing the permanently stored consent.
- It's part of widget_data, which is later saved to cart_session. This means that we need to handle updates to the widget_data (#4870) in a way that might not be suitable for other usages of widget_data. The bigger problem, though, is that cart_session and hence widget_data is not available on pages that are not related to a cart, such as the order confirmation page at the end of the checkout or e.g. the waiting list. If you currently open the widget for a sold-out event and click the "waiting list" button in the widget, the iframe opens but the data-consent option is basically ignored because there is no cart to store it in. This PR moves the consent out of widget_data and treats it as a transient parameter that is passed in through a GET parameter and is then passed to the JavaScript component *once*.

### DIFF
--- a/src/pretix/presale/context.py
+++ b/src/pretix/presale/context.py
@@ -52,7 +52,6 @@ from .signals import (
     footer_link, global_footer_link, global_html_footer, global_html_head,
     global_html_page_header, html_footer, html_head, html_page_header,
 )
-from .views.cart import cart_session, get_or_create_cart_id
 from .views.theme import _get_source_cache_key
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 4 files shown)


wiffbi (approved):
Works as intended. It changes the visibility of the footer-link for managing pretix’s cookie consent. Not sure we want this.

PR #4891, opened 2025-03-06
Pass widget_data to new tab (Z#23176995)

luelista:
This allows pre-filled questions and other widget configuration to be used even if Safari third-party tracking prevention is enabled


wiffbi:
See #4894

luelista:
Doesn't work like this

PR #4898, opened 2025-03-10
New consistent representation of date ranges with times (Z#23184657)

raphaelm:


src/pretix/base/models/event.py
```diff
@@ -180,35 +181,47 @@ def get_date_range_display(self, tz=None, force_show_end=False, as_html=False) -
...
         if (not self.settings.show_date_to and not force_show_end) or not self.date_to:
             df, dt = self.date_from, self.date_from
+            show_times = try_to_show_times
         else:
             df, dt = self.date_from, self.date_to
-        return daterange(df.astimezone(tz), dt.astimezone(tz), as_html)
+            show_times = try_to_show_times and self.settings.show_times and (
+                # Show date to if start and end are on the same day ("08:00-10:00")
+                dt.astimezone(tz).date() == df.astimezone(tz).date() or
+                # Show date to if start and end are on consecutive days and less than 24h ("23:00-03:00")
```
luelista, line 190:
```suggestion
                # Show times if start and end are on the same day ("08:00-10:00")
                dt.astimezone(tz).date() == df.astimezone(tz).date() or
                # Show times if start and end are on consecutive days and less than 24h ("23:00-03:00")
```

src/pretix/base/models/event.py
```diff
@@ -180,35 +181,47 @@ def get_date_range_display(self, tz=None, force_show_end=False, as_html=False) -
...
-    def get_time_range_display(self, tz=None, force_show_end=False) -> str:
-        """
-        Returns a formatted string containing the start time and sometimes the end time
-        of the event with respect to the current locale and to the ``show_date_to``
-        setting. Dates are not shown. This is usually used in combination with get_date_range_display
-        """
-        tz = tz or self.timezone
+            if as_html:
+                d = format_html(
+                    d + ' <span data-time="{}" data-timezone="{}" data-time-short>{}</span>',
```
luelista, line 207: What do you think about using a <time datetime> tag here as well?

```suggestion
                    d + ' <time datetime="{}" data-timezone="{}" data-time-short>{}</time>',
```

With a small update to the [corresponding javascript](https://github.com/pretix/pretix/blob/master/src/pretix/static/pretixpresale/js/ui/main.js#L536-L537)
```
    $("span[data-timezone], small[data-timezone], time[data-timezone]").each(function() {
        var t = moment.tz($(this).attr("datetime") || $(this).attr("data-time"), $(this).attr("data-timezone"))
```

wiffbi: I stumbled upon this in my a11y tests as well, so yes, please use `<time>` here as well. 👍

luelista (approved):

PR #4907, opened 2025-03-14
Fix image size in help-text for logo-images

wiffbi:
pretix only offers a max-width of 1120px for logo-images – instead of changing the layout, update the help-text.

### DIFF
--- a/src/pretix/base/settings.py
+++ b/src/pretix/base/settings.py
@@ -2883,7 +2883,7 @@ Your {organizer} team"""))  # noqa: W291
             ext_whitelist=settings.FILE_UPLOAD_EXTENSIONS_IMAGE,
             max_size=settings.FILE_UPLOAD_MAX_SIZE_IMAGE,
             help_text=_('If you provide a logo image, we will by default not show your event name and date '
-                        'in the page header. By default, we show your logo with a size of up to 1140x120 pixels. You '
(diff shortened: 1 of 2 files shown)


raphaelm (changes requested):
This just changes the text to a different kind of wrong. The actual sizes used are:

- 1120 pixels if there is a page background color
- 1140 pixels if there is no page background color 
- 1170 pixels if event_logo_image_large is set (regardless of background color)

The 1170 pixels are indicated in the event_logo_image_large settings but I'm not sure how to make the 1140/1120 difference clear. We could try to get rid of the difference by reducing the page width with a white background color also to 1120, but … this is not an area I touch lightly.

wiffbi:
Good catch! As I do not have a white background set in my test-events, I did not realize this behaviour. That is a tough one … What about:

`If you use a white background, we show your logo with a size of up to 1140x120 pixels. Otherwise the maximum size is 1120x120 pixels.`

raphaelm:
That's probably the worst best option we have :D

wiffbi:
Sorry, I let this sit for a while. In my a11y-PR regarding font-size and meta-userzoom, I thought about changing the width to 1120px as you described, as we are anyway changing how the shop looks with the font-size change. But changing the width does not seem necessary and for users who optimized their header-images it is probably a bigger hassle to adapt again, so I would not touch it for now.

raphaelm (approved):

PR #4917, opened 2025-03-16
Date picker: Guess default time 23:59 for end dates (Z#23182900)

raphaelm:
When a filter form asks users to select a datetime range, it is a common scenario that users want to filter just by date. However, this will easily lead to off-by-one-errors, as "Feb 1st, 00:00 to Feb 28th, 00:00" is missing the last day of February. This PR tries to detect these fields and use a different initial value for the time field.

### DIFF
--- a/src/pretix/static/pretixcontrol/js/ui/main.js
+++ b/src/pretix/static/pretixcontrol/js/ui/main.js
@@ -205,7 +205,7 @@ var form_handlers = function (el) {
                     return;
                 }
                 if ($timepicker.val() === "") {
-                    if ($(this).attr("name").includes("_until") || $(this).attr("name").includes("_end")) {
+                    if (/_(until|end|to)(_|$)/.test($(this).attr("name"))) {
                         date.set({'hour': 23, 'minute': 59, 'second': 59});
                     } else {
                         date.set({'hour': 0, 'minute': 0, 'second': 0});
--- a/src/pretix/static/pretixcontrol/js/ui/main.js
+++ b/src/pretix/static/pretixcontrol/js/ui/main.js
@@ -205,7 +205,11 @@ var form_handlers = function (el) {
                     return;
                 }
                 if ($timepicker.val() === "") {
-                    date.set({'hour': 0, 'minute': 0, 'second': 0});
+                    if ($(this).attr("name").includes("_until") || $(this).attr("name").includes("_end")) {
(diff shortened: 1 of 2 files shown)


luelista (commented):
I like the idea a lot, some notes:

- this is not limited in any way to filter forms, e.g. "End of presale" or Product "Available Until" are affected as well. Not to say they shouldn't be included, but it's a broader change than just filters.
- we should add `_to` (e.g. Advanced Order Search "Order placed before" (`created_to`), or "Event End Time" (`date_to`)
- I feel somewhat uncomfortable with just searching in-string, esp. for `_to`, I'd like some kind of `/_to(_|$)/` check better (I guess you mainly want to match strings ending in `_whatever` or `_whatever_1` etc?)

src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -205,7 +205,11 @@ var form_handlers = function (el) {
                     return;
                 }
                 if ($timepicker.val() === "") {
-                    date.set({'hour': 0, 'minute': 0, 'second': 0});
+                    if ($(this).attr("name").includes("_until") || $(this).attr("name").includes("_end")) {
```
luelista, line 208:
```suggestion
                    if (/_(until|end|to)(_|$)/.test($(this).attr("name"))) {
```

src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -91,7 +91,11 @@ var form_handlers = function (el) {
                     return;
                 }
                 if ($timepicker.val() === "") {
-                    date.set({'hour': 0, 'minute': 0, 'second': 0});
+                    if ($(this).attr("name").includes("_until") || $(this).attr("name").includes("_end")) {
```
luelista, line 94:
```suggestion
                    if (/_(until|end|to)(_|$)/.test($(this).attr("name"))) {
```

raphaelm:
> this is not limited in any way to filter forms, e.g. "End of presale" or Product "Available Until" are affected as well. Not to say they shouldn't be included, but it's a broader change than just filters. (A more accurate way to match appropriate datetime-fields would be adding a data-attribute - but of course that would require lots of changes in field definitons)

For "end of presale" and "available until" these still make sense and I still think it's a positive change. The one place I'm not so sure is (Sub)Event.date_to, that's why I omitted "to", but if you think it makes sense there as well, happy to change it.

PR #4922, opened 2025-03-17
SMTP config: Do not accept non-ASCII passwords

raphaelm:
Python's SMTP implementation can't seem to handle non-ascii passwords, so let's not let people enter them.

### DIFF
--- a/src/pretix/control/forms/mailsetup.py
+++ b/src/pretix/control/forms/mailsetup.py
@@ -25,6 +25,8 @@ import socket
 from django import forms
 from django.conf import settings
 from django.core.exceptions import ValidationError
+from django.core.validators import RegexValidator
+from django.utils.text import format_lazy
 from django.utils.translation import gettext_lazy as _
 
 from pretix.base.forms import SecretKeySettingsField, SettingsForm
@@ -54,6 +56,15 @@ class SMTPMailForm(SettingsForm):
     smtp_password = SecretKeySettingsField(
         label=_("Password"),
         required=False,
+        validators=[RegexValidator(
+            r"^[A-Za-z0-9!\"#$%&'()*+,-./:;<=>?@\^_`{}|~]+$",
+            message=format_lazy(
+                _("The password contains characters not supported by our email system. Please only use characters "
+                  "A-Z, a-z, 0-9, and common special characters ({characters})."),
+
+                characters=r'!"#$%%&\'()*+,-./:;<=>?@\^_`{}|~'
+            )
+        )]
     )
     smtp_use_tls = forms.BooleanField(
         label=_("Use STARTTLS"),


src/pretix/control/forms/mailsetup.py
```diff
@@ -54,6 +56,15 @@ class SMTPMailForm(SettingsForm):
     smtp_password = SecretKeySettingsField(
         label=_("Password"),
         required=False,
+        validators=[RegexValidator(
+            r"^[A-Za-z0-9!\"#$%&'()*+,-./:;<=>?@\^_`{}|~]+$",
+            message=format_lazy(
+                _("The password contains characters not supported by our email system. Please only use characters "
+                  "A-Z, a-z, 0-9, and common special characters ({characters})."),
+
+                characters=r'!"#$%%&\'()*+,-./:;<=>?@\^_`{}|~'
```
wiffbi, line 65: Did check all these special chars are supported? When reading the [issue on cpython](https://github.com/python/cpython/issues/73936) I found [this linked issue](https://github.com/janeczku/calibre-web/issues/2977) that stated backticks cause an error. I also found that hash (#) seem to cause errors in some sendmail implementations (can’t find the link currently, sorry!).

luelista: `-` should always be last in a character class, otherwise it specifies a range. However this case, the result is accidentally (or trickily) correct, because
```
>>> ord(",")
44
>>> ord("-")
45
>>> ord(".")
46
```

src/pretix/control/forms/mailsetup.py
```diff
@@ -54,6 +56,15 @@ class SMTPMailForm(SettingsForm):
     smtp_password = SecretKeySettingsField(
         label=_("Password"),
         required=False,
+        validators=[RegexValidator(
+            r"^[A-Za-z0-9!\"#$%&'()*+,-./:;<=>?@\^_`{}|~]+$",
```
luelista, line 60:
```suggestion
            r"^[A-Za-z0-9!\"#$%&'()*+,./:;<=>?@\^_`{}|~-]+$",
```
`-` should always be last in a character class, otherwise it specifies a range. However this case, the result is accidentally (or trickily) correct, because
```
>>> ord(",")
44
>>> ord("-")
45
>>> ord(".")
46
```

raphaelm: I'd love to say this was intentional, which it kinda was since I typed the characters down from an ASCII table, but let's better not leave it as a work of art :D

src/pretix/control/forms/mailsetup.py
raphaelm: The linked issue that talks about backticks is inconsistent, it gives ` (ASCII character 96) as an example but the error message shows ˋ (unicode character U+02CB, "Modifier Letter Grave Accent") to be the problem. I'm not claiming anyone will understand this, but ASCII backtick should be fine.

I did not test all of the characters, I just went by the "can't handle non-ascii characters" error message and think that would be okay for now. We can always reduce the charset further.

PR #4927, opened 2025-03-18
Discounts: Add applicability date ranges (Z#23151897)

ser8phin:
Adds optional start and end times for discounts in between which an event series taking place may be discounted. Possible use case: I want to give a discount for on one specific weekend during my event series, while not applying the discount to any other events in the series.

TODO:

- [x] add date fields to model
- [x] add date checks/compare function
- [x] fix condition checking / queryset(s)
- [x] add tests
- [x] make UI adjustments
- [x] make API adjustments
- [x] add migration after final pull before undrafting the PR (before this, all automated tests will fail!)


src/pretix/base/models/discount.py
```diff
@@ -171,6 +172,17 @@ class Discount(LoggedModel):
                     "access to sold-out quota will still receive the discount."),
     )
 
+    event_date_from = models.DateTimeField(
+        verbose_name=_("Available for events from"),
+        null=True,
+        blank=True,
+    )
+    event_date_until = models.DateTimeField(
+        verbose_name=_("Available for events until"),
```
ser8phin, line 181:
```suggestion
        verbose_name=_("Available for dates until"),
```

src/pretix/base/models/discount.py
```diff
@@ -171,6 +172,17 @@ class Discount(LoggedModel):
                     "access to sold-out quota will still receive the discount."),
     )
 
+    event_date_from = models.DateTimeField(
+        verbose_name=_("Available for events from"),
```
ser8phin, line 176:
```suggestion
        verbose_name=_("Available for dates from"),
```

src/pretix/base/models/discount.py
```diff
@@ -352,6 +364,20 @@ def apply(self, positions: Dict[int, PositionInfo],
...
             limit_products = {p.pk for p in self.condition_limit_products.all()}
 
+        subevents_in_range = None
+        if self.event_date_from or self.event_date_until:
+            subevent_ids = {
+                subevent_id
+                for (item_id, subevent_id, line_price_gross, is_addon_to, voucher_discount)
+                in positions.values()
+            }
+            subevents_qs = SubEvent.objects.all().filter(pk__in=subevent_ids)
```
raphaelm, line 374: Instead of doing this, would it be worth modifying the signature of the `apply()` and the calling `apply_discounts()` function such that the subevent object or the subevent date is passed down? Otherwise, this creates pretty useless SQL queries. 

For example, if we start looking from CartManager, it starts with

```
        discount_results = apply_discounts(
            self.event,
            self._sales_channel.identifier,
            [
                (cp.item_id, cp.subevent_id, cp.line_price_gross, bool(cp.addon_to), cp.is_bundled, cp.listed_price - cp.price_after_voucher)
                for cp in positions
            ]
        )
```

So we *could* just pass additional information about the subevent and would not require any additional SQL queries at all.

I see two options:

## Option 1

We pass down entire subevent objects:

```
diff --git a/src/pretix/base/models/discount.py b/src/pretix/base/models/discount.py
index be2b3d0ac..5777e5377 100644
--- a/src/pretix/base/models/discount.py
+++ b/src/pretix/base/models/discount.py
@@ -36,7 +36,7 @@ from django_scopes import ScopedManager
 from pretix.base.decimal import round_decimal
 from pretix.base.models.base import LoggedModel
 
-PositionInfo = namedtuple('PositionInfo', ['item_id', 'subevent_id', 'line_price_gross', 'is_addon_to', 'voucher_discount'])
+PositionInfo = namedtuple('PositionInfo', ['item_id', 'subevent', 'line_price_gross', 'is_addon_to', 'voucher_discount'])
```

This would be the simplest option, but at the same time keeping PositionInfo fully "scalar" is very nice for testability. So I think I would prefer Option 2:

## Option 2

```
diff --git a/src/pretix/base/models/discount.py b/src/pretix/base/models/discount.py
index be2b3d0ac..b38d6ebbc 100644
--- a/src/pretix/base/models/discount.py
+++ b/src/pretix/base/models/discount.py
@@ -36,7 +36,7 @@ from django_scopes import ScopedManager
 from pretix.base.decimal import round_decimal
 from pretix.base.models.base import LoggedModel
 
-PositionInfo = namedtuple('PositionInfo', ['item_id', 'subevent_id', 'line_price_gross', 'is_addon_to', 'voucher_discount'])
+PositionInfo = namedtuple('PositionInfo', ['item_id', 'subevent_id', 'subevent_date_from', 'line_price_gross', 'is_addon_to', 'voucher_discount'])
```

raphaelm (changes requested):

doc/api/resources/discounts.rst
```diff
@@ -35,6 +35,10 @@ subevent_mode                            strings                    Determines h
                                                                     ``"same"`` (discount is only applied for groups within
                                                                     the same date), or ``"distinct"`` (discount is only applied
                                                                     for groups with no two same dates).
+event_date_from                          datetime                   The first date time of a subevent to which this discount can be applied
+                                                                    (or ``null``).
+event_date_until                         datetime                   The last date time of a subevent to which this discount can be applied
+                                                                    (or ``null``).
```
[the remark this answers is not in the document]
wiffbi: Hm, is it a useful option to apply it to Event.date_from/until? IMHO not really and thinking about it, I’d then rename both parameters to `subevent_date_from` and `subevent_date_to` (instead of just event_date_XXX) – which already in the name makes it clear, what it is for.

src/pretix/base/models/discount.py
wiffbi: That is a good idea – I don’t know why we did not consider changing the signature as this could potentially lead to quite a few SQL-queries if we have lots of discounts with daterange-limits. Although this is probably rather an edge-case, we still already have the info available, so why not use it. Also the only bigger change seems to be in api/serializers/order.py as the OrderPosition-object there AFAICT does not yet have the subevent pre-populated.

I think option 2 is the more practical one as well.

ser8phin: > as the OrderPosition-object there AFAICT does not yet have the subevent pre-populated

Yep, just needs an ittle bittle change with some consequences... let's see how this goes :+1:

doc/api/resources/discounts.rst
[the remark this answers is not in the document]
ser8phin: > We could potentially also decide this differently and apply it to Event.date_from
or
> rename both parameters to subevent_date_from and subevent_date_to
renaming makes sense, unless we do want to use it for all events as well, i think. What do you think @raphaelm ?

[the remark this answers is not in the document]
raphaelm: The one scenario where I can see it useful for events would be with preparing the discount on a template event and then cloning the event many times… But it's a big edge-case. I'm torn, but I'm leaning towards renaming it to subevent_date_from/until

src/pretix/base/models/discount.py
```diff
@@ -355,11 +368,15 @@ def apply(self, positions: Dict[int, PositionInfo],
...
-            for idx, (item_id, subevent_id, line_price_gross, is_addon_to, voucher_discount) in positions.items()
+            for idx, (item_id, subevent_id, subevent_date_from, line_price_gross, is_addon_to, voucher_discount) in
+            positions.items()
             if (
                 (self.condition_all_products or item_id in limit_products) and
                 (self.condition_apply_to_addons or not is_addon_to) and
                 (not self.condition_ignore_voucher_discounted or voucher_discount is None or voucher_discount == Decimal('0.00'))
+                and (not subevent_id or (
+                    self.event_date_from is None or self.event_date_from and subevent_date_from >= self.event_date_from)) and (
+                        self.event_date_until is None or self.event_date_until and subevent_date_from < self.event_date_until)
```
raphaelm, line 379: I think this can be reduced to 

```suggestion
                    self.event_date_from is None or subevent_date_from >= self.event_date_from)) and (
                        self.event_date_until is None or subevent_date_from < self.event_date_until)
```

or what am I missing?

raphaelm (commented):
LGTM with
- one simplification
- the open suggestion to rename the new fields to subevent_date_from/until

src/pretix/base/models/discount.py
ser8phin: Not missing anything, as far as I can tell, just a leftover snip from sometime earlier that is now removed

src/pretix/base/models/discount.py
```diff
@@ -171,6 +173,17 @@ class Discount(LoggedModel):
                     "access to sold-out quota will still receive the discount."),
     )
 
+    subevent_date_from = models.DateTimeField(
+        verbose_name=_("Available for dates from"),
```
raphaelm, line 177:
```suggestion
        verbose_name=pgettext_lazy("subevent", "Available for dates starting from"),
```
The word "date" in English is quite ambiguous, so we need to ensure e.g. German translaters will translate with "Termin" not "Datum". Also, we can make it clearer that we only look at the start of the event

src/pretix/control/templates/pretixcontrol/items/discount.html
```diff
@@ -26,6 +26,8 @@ <h1>{% trans "Automatic discount" %}</h1>
                     {% bootstrap_field form.condition_ignore_voucher_discounted layout="control" %}
                     {% if form.subevent_mode %}
                         {% bootstrap_field form.subevent_mode layout="control" %}
+                        {% bootstrap_field form.event_date_from layout="control" %}
+                        {% bootstrap_field form.event_date_until layout="control" %}
```
ser8phin, line 30:
```suggestion
                        {% bootstrap_field form.subevent_date_from layout="control" %}
                        {% bootstrap_field form.subevent_date_until layout="control" %}
```

src/pretix/base/models/discount.py
```diff
@@ -171,6 +173,17 @@ class Discount(LoggedModel):
...
+        null=True,
+        blank=True,
+    )
+    subevent_date_until = models.DateTimeField(
+        verbose_name=_("Available for dates until"),
```
raphaelm, line 182:
```suggestion
        verbose_name=pgettext_lazy("subevent", "Available for dates starting until"),
```

raphaelm (approved):

src/pretix/base/models/discount.py
ser8phin:
```suggestion
        verbose_name=pgettext_lazy("subevent", "Available for dates starting before"),
```
sounds better then, right?

PR #4934, opened 2025-03-18
Fix Exception when displaying mail_history (PRETIXEU-BB0)

luelista:
Log entries of type pretix.plugins.banktransfer.order.email.invoice had type(invoices) == int instead of list

### DIFF
--- a/src/pretix/control/views/orders.py
+++ b/src/pretix/control/views/orders.py
@@ -2491,10 +2491,12 @@ class OrderEmailHistory(EventPermissionRequiredMixin, OrderViewMixin, ListView):
     def get_context_data(self, **kwargs):
         ctx = super().get_context_data(**kwargs)
         for l in ctx["logs"]:
-            if l.parsed_data.get("invoices"):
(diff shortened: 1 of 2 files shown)


src/pretix/control/views/orders.py
```diff
@@ -2491,10 +2491,12 @@ def get_queryset(self):
     def get_context_data(self, **kwargs):
         ctx = super().get_context_data(**kwargs)
         for l in ctx["logs"]:
-            if l.parsed_data.get("invoices"):
+            invoice_ids = l.parsed_data.get("invoices")
+            if invoice_ids:
+                if type(invoice_ids) is int: invoice_ids = [invoice_ids]
```
wiffbi, line 2496:
```suggestion
                if type(invoice_ids) is int:
                    invoice_ids = [invoice_ids]
```

wiffbi (approved):
Just a code-style/flake8-issue … let’s hope uv fixes its bug and we get CI running again. :)

PR #4936, opened 2025-03-19
PDF editor: Allow Alt+Arrow for small adjustments (Z#23186382)

raphaelm:


src/pretix/static/pretixcontrol/js/ui/editor.js
```diff
@@ -1115,7 +1115,7 @@ var editor = {
     },
 
     _on_keydown: function (e) {
-        var step = e.shiftKey ? editor._mm2px(10) : editor._mm2px(1);
+        var step = e.shiftKey ? editor._mm2px(10) : (e.altKey ? editor._mm2px(0.1) : editor._mm2px(1));
```
wiffbi, line 1118: What about? Is a bit easier to read for me, but that is IMHO.

```suggestion
        var step = editor._mm2px(e.shiftKey ? 10 : (e.altKey ? 0.1 : 1));
```

wiffbi (approved):
Just a tiny comment, which seems easier to read to me. But that is just IMHO.

PR #4949, opened 2025-03-25
API: fix event creation sales-channel-migration

wiffbi:
We stumbled over an issue where copying an event through the API had a `data["sales_channels"]` being `None`, which caused the serialization to fail. This PR only migrates legacy sales_channels if there are any.

### DIFF
--- a/src/pretix/api/serializers/__init__.py
+++ b/src/pretix/api/serializers/__init__.py
@@ -80,7 +80,14 @@ class SalesChannelMigrationMixin:
             raise ValueError("organizer not in context")
 
     def to_internal_value(self, data):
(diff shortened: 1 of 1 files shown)


raphaelm:
I would argue that `"sales_channels": None` is invalid input to the API and should be rejected with error 400 instead of being silently accepted to mean the default, or did we ever document this as valid?

wiffbi:
I’d agree, but the POST-request, that I read regarding the issue, did not include any information to sales_channels in its request-body, only `"all_sales_channels": true,` or not event that.

raphaelm (approved):

PR #4960, opened 2025-03-28
Fix bulk edit of revoked devices (Z#23187227)

luelista:


src/pretix/control/views/organizer.py
```diff
@@ -944,13 +944,16 @@ def get_queryset(self):
...
 
         if 'device' in self.request_data and '__ALL' not in self.request_data:
             qs = qs.filter(
                 id__in=self.request_data.getlist('device')
             )
+        elif self.request.method == 'GET' or '__ALL' in self.request_data:
+            if self.filter_form.is_valid():
+                qs = self.filter_form.filter_qs(qs)
+        else:
+            raise PermissionDenied()
```
raphaelm, line 956: Is that the best thing we can return here, not some kind of bad request?

raphaelm (approved):

src/pretix/control/views/organizer.py
luelista: BadRequest is better :)

I stole the PermissionDenied from the other places with similar checks, e.g. https://github.com/pretix/pretix/blob/303ab9fd4a6142d7ff86ff4d47a866bf12b29c30/src/pretix/control/views/orders.py#L220

PR #4961, opened 2025-03-28
Show link to stripe payment receipt in backend (Z#23187178)

luelista:


raphaelm (changes requested):
I don't think this belongs into the log entry, I think we should sow this as part of rendering the OrderPayment:

image

luelista:
Like that? 

image

raphaelm (approved):

PR #4977, opened 2025-04-02
PDF-Editor: use panel-head as topbar for common commands/tools and preview/save

wiffbi:
This PR adds a new topbar to the PDF-editor

Bildschirmfoto 2025-04-02 um 08 22 25

- copy-to-clipboard does not create savepoint anymore
- add duplicate-button (does not overwrite clipboard), remove cut/copy/paste-buttons
- change from deprecated event.keyCode to event.key


src/pretix/control/templates/pretixcontrol/pdf/index.html
```diff
@@ -25,27 +26,102 @@ <h1>
...
+                                <li>
+                                    <button class="btn" id="editor-add-qrcode-lead"
+                                            data-content="pseudonymization_id"
+                                            disabled>
+                                        {% trans "QR code for Lead Scanning" %}
+                                    </button>
+                                </li>
+                                <li>
+                                    <button class="btn" id="editor-add-qrcode-other"
+                                            data-content="secret"
```
raphaelm: Not a new bug, but I find this confusing as a user, as it adds the same QR code than the "QR code for Check In" option. Should we set "other" here?

wiffbi:
I probably added the `.btn-sm` to at least somewhat get the panel-head the same height as its counterpart next to it. But as I needed to anyways change padding-top/bottom, I can change it event more and get rid of `.btn-sm` – Good idea!

src/pretix/control/templates/pretixcontrol/pdf/index.html
wiffbi: Yes, make sense!

PR #4978, opened 2025-04-02
Fix datepicker date display when using min/max (Z#23185544)

wiffbi:
When using min/max dates in datepicker (either one or both) when opening, closing and then opening the datepicker again, the currentDate is now and datepicker-nav is disabled. This PR fixes the behaviour, ~~but changes a depency – although an old one. I was not able to find a way to fix this from the outside, so changing it in the dependency was the only way. Actually, there might be a way~~: register an event-listener to `dp.hide` and set the `viewDate(viewDate)` manually to a correct one inside the boundaries of minDate/maxDate. This needs to be done everywhere, where we instantiate datepicker though, but that is currently only in one spot.

### DIFF
--- a/src/pretix/static/datetimepicker/bootstrap-datetimepicker.js
+++ b/src/pretix/static/datetimepicker/bootstrap-datetimepicker.js
@@ -947,7 +947,7 @@
 
                 input.blur();
 
-                viewDate = date.clone();
+                //viewDate = date.clone();
 
                 return picker;
             },
--- a/src/pretix/static/datetimepicker/bootstrap-datetimepicker.js
+++ b/src/pretix/static/datetimepicker/bootstrap-datetimepicker.js
@@ -947,7 +947,6 @@
 
                 input.blur();
 
-                // custom edit: do not date.clone() as this breaks when minDate/maxDate is outside of date's month
                 //viewDate = date.clone();
 
                 return picker;
--- a/src/pretix/static/datetimepicker/bootstrap-datetimepicker.js
+++ b/src/pretix/static/datetimepicker/bootstrap-datetimepicker.js
@@ -947,6 +947,7 @@
 
                 input.blur();
 
+                // custom edit: do not date.clone() as this breaks when minDate/maxDate is outside of date's month
                 //viewDate = date.clone();
 
                 return picker;
--- a/src/pretix/static/datetimepicker/bootstrap-datetimepicker.js
+++ b/src/pretix/static/datetimepicker/bootstrap-datetimepicker.js
@@ -947,7 +947,7 @@
 
                 input.blur();
 
-                //viewDate = date.clone();
+                viewDate = date.clone();
 
                 return picker;
             },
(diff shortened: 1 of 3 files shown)


src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -82,7 +82,13 @@ var form_handlers = function (el) {
                     Math.abs(+new Date(opts.minDate) - new Date()) < Math.abs(+new Date(opts.maxDate) - new Date())
             ) ? opts.minDate : opts.maxDate;
         }
-        $(this).datetimepicker(opts);
+        $(this).datetimepicker(opts).on("dp.hide", function() {
```
raphaelm, line 85: Can you add a small comment on why this code is necessary so we remember in a few years when we refactor this file?

src/pretix/static/datetimepicker/bootstrap-datetimepicker.js
```diff
@@ -947,7 +947,7 @@
 
                 input.blur();
 
-                viewDate = date.clone();
+                //viewDate = date.clone();
```
raphaelm, line 950: The PR body says that the dependency does not need to be modified, so is this change unintentionally kept?

src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -82,7 +82,13 @@ var form_handlers = function (el) {
...
+            var $dtp = $(this).data("DateTimePicker");
+            var currentViewDate = $dtp.viewDate();
+            window.setTimeout(function () {
+                $dtp.viewDate(currentViewDate);
+            }, 500);
```
raphaelm, line 90: I am quite capable of still triggering the bug by clicking faster than 500ms, so I suggest 50ms. (Note that I tested with the bootstrap-datetimepicker.js change reverted, otherwise the issue does not occur.)

```suggestion
            }, 50);
```

src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -82,7 +82,13 @@ var form_handlers = function (el) {
...
+        });
         if ($(this).parent().is('.splitdatetimerow')) {
```
raphaelm, line 92: We also have support for data-min and data-max in pretixcontrol and we also render QuestionForm there, so I suggest to make the same change in pretixcontrol/ui/main.js

raphaelm (changes requested):

src/pretix/static/pretixpresale/js/ui/main.js
wiffbi: Good point! I added the fix (with a comment why this fix is needed) to pretixcontrol/ui/main.js as well.

src/pretix/static/pretixpresale/js/ui/main.js
wiffbi: Well, you have fast fingers. :) I changed it to 50ms, not sure why I chose 500 in the first place.

src/pretix/static/datetimepicker/bootstrap-datetimepicker.js
wiffbi: Yup, sorry, that is not intentional.

src/pretix/static/pretixpresale/js/ui/main.js
wiffbi: Done.

raphaelm (approved):

PR #4979, opened 2025-04-02
Fix datepicker locale when using min/max

wiffbi:
When using min/max with datepicker, the minDate/maxDate is parsed with moment() and replaces the viewDate inside datepicker. We have three options to fix this bug:

1. Change bootstrap-datepicker dependency to apply `options.locale` when parsing a date.
2. Add our own parseInputDate to `options` when instantiating datepickers
3. Globally tell `moment()` to use `body[data-datetimelocale]`

I did not want to change a dependency if not needed (1.). Providing a custom `options.parseInputDate` seemed error-prone as we need to provide it everywhere where we instantiate a datepicker (2.). So I chose 3, which IMHO anyways feels like the right way to do to make sure all dates on the page are used with the correct locale.

### DIFF
--- a/src/pretix/static/pretixpresale/js/ui/cart.js
+++ b/src/pretix/static/pretixpresale/js/ui/cart.js
@@ -69,8 +69,6 @@ var cart = {
 $(function () {
     "use strict";
 
-    moment.locale($("body").attr("data-locale").substr(0, 2));
-
     if ($("#cart-deadline").length) {
         cart.init();
     }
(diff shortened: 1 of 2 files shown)


raphaelm (approved):
I agree that it is probably the best solution and it is also consistent with what we already do in pretix/static/pretixcontrol/js/ui/main.js, but we should remove the moment.locale call from pretix/static/pretixpresale/js/ui/cart.js then so they don't "compete".

PR #4986, opened 2025-04-03
Fix checkin-list-filter test randomly matching secret

wiffbi:
`test_checkins_list_filter` happened to randomly falsely match op.secret with its short attendee_name „a4“. This PR changes the test to use a longer attendee_name.

### DIFF
--- a/src/tests/control/test_checkins.py
+++ b/src/tests/control/test_checkins.py
@@ -221,7 +221,7 @@ def checkin_list_env():
         item=item_ticket,
         variation=None,
         price=Decimal("23"),
-        attendee_name_parts={'full_name': "a4"},  # a3 attendee is a4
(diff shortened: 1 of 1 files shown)


raphaelm (approved):
I mean … yes, but I would have changed the A1, A2 and A3 as well, because why wouldn't they trigger the same issues?

wiffbi:
We never match for A1, A2 or A3 in a query – a4 was the only query-param short enough. The others are a3dummy and a3company – both should be long enough. The attendee_names A1, A2 and A4 are onyl used in `test_checkins_attendee_name_from_addon_available` which does not match in a query, but only assert attendee_name combined with order-code and item.name. So there should not be any other accidental matches in the tests.

PR #4989, opened 2025-04-04
Allow changing ticket secret via API (Z#23188201)

luelista:


src/pretix/base/services/orders.py
```diff
@@ -1563,7 +1563,8 @@ class OrderChangeManager:
     FeeValueOperation = namedtuple('FeeValueOperation', ('fee', 'value', 'price_diff'))
     AddFeeOperation = namedtuple('AddFeeOperation', ('fee', 'price_diff'))
     CancelFeeOperation = namedtuple('CancelFeeOperation', ('fee', 'price_diff'))
-    RegenerateSecretOperation = namedtuple('RegenerateSecretOperation', ('position',))
+    RegenerateSecretOperation = namedtuple('RegenerateSecretOperation', ('position', 'force_secret'))
```
raphaelm, line 1566: I think force_secret is a leftover from a previous attempt? It doesn't seem to be set anywhere and also breaks a test

src/pretix/base/services/orders.py
```diff
@@ -2428,7 +2432,7 @@ def _perform_operations(self):
             elif isinstance(op, self.SplitOperation):
                 split_positions.append(op.position)
             elif isinstance(op, self.RegenerateSecretOperation):
-                op.position.web_secret = generate_secret()
+                op.position.web_secret = op.force_secret or generate_secret()
```
raphaelm, line 2435: same here

raphaelm (approved):

src/pretix/base/services/orders.py
luelista: oops, yes
```suggestion
    RegenerateSecretOperation = namedtuple('RegenerateSecretOperation', ('position',))
```

src/pretix/base/services/orders.py
luelista:
```suggestion
                op.position.web_secret = generate_secret()
```

PR #4998, opened 2025-04-09
[WIP] Allow registering for all events in an event series

froggleston:
This PR constrains attendees to registering for all events in an event series if that series has the "bundle_event_series" meta_data tag set to "true". This is by no means optimal as it would be better to have an actual event setting to do this.

In essence, this is a hacky workaround, and still needs work:

- [ ] Implement voucher code processing properly
- [ ] Add a proper event setting instead of the meta_data tag


froggleston:
Created in error! (I meant to create this against our fork - apologies!)

PR #5020, opened 2025-04-21
Pluggable invoice transmission methods

raphaelm:
This PR does two main things:

1. It creates an abstract concept of invoice transmissions that allows us to add e-invoicing methods for other countries.
2. It moves the "send invoice to another address than order.email" from the banktransfer plugin to core. At the moment, it removes the functionality to request the invoice to be sent *after* the order was placed in self-service, because that doesn't really make sense with making the recipient address part of the invoice address and also doesn't make much sense for other transmission methods. If there is demand, we can add it back in later as a special case.

I have not yet created any transmission providers other than email because some other problems need to be solved first that are more related to invoice rendering and outside of the scope of this PR. For testing, you can use e.g. this stub plugin:
https://github.com/pretix/pretix-einvoice-italy/blob/main/pretix_einvoice_italy/provider.py

See also the dummy implementations in src/tests/testdummy/signals.py of this PR

---

- [x] Define transmission types
  - [x] email
  - [x] peppol
  - [x] SdI
- [x] Allow plugins to define transmission providers
- [x] Only show transmission types with an active providers
- [x] Render fields in invoice address form
  - [x] conditionally handle visibility and requirement of fields
  - [x] input validation
  - [x] test checkout
  - [x] test changing order in frontend
  - [x] test changing order in backend
  - [x] enforcement for specific provider (eg only SdI in Italy)
- [x] Render field contents when showing invoice address
  - [x] confirm step
  - [x] frontend order view
  - [x] backend order view
- [x] Addresses in API
- [x] Check integration with saved addresses in customer account
- [x] Simplify form by hiding transmission method choice if only email is available
- [x] Backend settings
  - [x] Choice of transmission types
  - [x] Choice of transmission providers
- [x] Actually transmit invoices
- [x] Show transmission status in backend
- [x] Allow manual retransmission
- [x] How to deal with invoice_email_attachment=False setting
- [x] backwards compatibility with new email field for people who don't use pretix for invoice sending
- [x] Lots of tests
  - [x] New API behavior
    - [x] Setting invoice address
    - [x] (Re)transmission
  - [x] (Re)transmission in backend
  - [x] Form rendering and validation
  - [x] JS helper
  - [x] Emailing of invoices to other addresses during purchase etc
    - [x] order
    - [x] approve
    - [x] paid
    - [x] cancel
    - [x] change
  - [x] No duplicate emails during purchase etc
  - [x] Invoices being marked as sent after email is out
- [x] Transmission status in API
- [x] Retransmission in API
- [x] Update API docs
- [x] Add plugin docs
- [x] Resolve TODOs in code
- [x] Handle test mode invoices
- [x] Remove old mechanism from banktransfer
  - [x] Remove code
  - [x] Migrate texts from old to new mechanism
- [ ] update plugins that use addressform.js
  - [ ] resellers
  - [ ] shipping

later

- [ ] Figure out how to correctly handle PDF vs non-PDF invoices (stamp/text on PDF, add xml field)
- [ ] Allow buyers to retransmit again?
- [x] Check dealing with optional invoice addresses and transmission types that require one
   - [ ] Tests


src/pretix/base/services/invoices.py
```diff
@@ -551,3 +575,99 @@ def send_invoices_to_organizer(sender, **kwargs):
...
+            invoice.save(update_fields=["transmission_status", "transmission_date"])
+            transmit_invoice.apply_async(args=(invoice.event_id, invoice.pk, True))
+
+
+@receiver(signal=periodic_task)
+@scopes_disabled()
+def send_pending_invoices(sender, **kwargs):
+    with transaction.atomic():
+        # Transmit all invoices that have not been transmitted by another process within
+        # the last hour (TODO: consider backwards-compatibility of this, maybe exclude email)
```
raphaelm: Happy about feedback here. In the past it was possible to create a invoice in the backend and only send it out manually but never have the system send it. Should we retain that behavior and exclude email here?

src/pretix/base/forms/questions.py
```diff
@@ -1250,6 +1258,51 @@ def __init__(self, *args, **kwargs):
...
                     v.widget.attrs['autocomplete'] = 'section-invoice billing ' + autocomplete
 
+        self.fields['is_business'].widget.attrs['data-trigger-address-info'] = 'on'
+        self.fields['transmission_type'] = forms.ChoiceField(
+            label=_('Invoice transmission method'),
+            choices=[
+                (t.identifier, t.public_name) for t in get_transmission_types()
+            ]
+        )
+        self.fields['transmission_type'].widget.attrs['data-trigger-address-info'] = 'on'
```
luelista: as this field is implicitly marked as required, invoice address fields will be switched to required by addressform.js (isAnyRequired will be true).

easiest fix is to move this block before line 1215 `if not self.address_required or self.all_optional:` so it gets affected by the `required = False` magic.

doc/api/resources/invoices.rst
```diff
@@ -119,6 +128,79 @@ internal_reference                    string                     Customer's refe
...
+transmission_email_address            string                     Required when ``transmission_email_other`` is set,
+                                                                 ignored otherwise. The email address that the invoice
+                                                                 should be sent to.
+===================================== ========================== =======================================================
+
+PEPPOL
+""""""
+
+The identifier ``"peppol"`` represents the transmission of XML invoices through the `PEPPOL`_ network.
+This is only availble for business addresses.
```
luelista:
```suggestion
This is only available for business addresses.
```

doc/api/resources/invoices.rst
```diff
@@ -119,6 +128,79 @@ internal_reference                    string                     Customer's refe
...
+Field                                 Type                       Description
+===================================== ========================== =======================================================
+transmission_peppol_participant_id    string                     Required. The PEPPOL participant ID of the recipient.
+===================================== ========================== =======================================================
+
+Italian Exchange System
+"""""""""""""""""""""""
+
+The identifier ``"it_sdi"`` represents the transmission of XML invoices through the `Sistema di Interscambio`_ network used in Italy.
+This is only availble for addresses with country ``"IT"``.
```
luelista:
```suggestion
This is only available for addresses with country ``"IT"``.
```

src/pretix/base/forms/questions.py
```diff
@@ -1303,6 +1356,36 @@ def clean(self):
...
+
+                self.instance.transmission_type = transmission_type.identifier
+                self.instance.transmission_info = {
+                    k: data.get(k) for k in transmission_type.invoice_address_form_fields
+                }
+            elif transmission_type.exclusive:
+                if transmission_type.is_available(self.event, data.get("country"), data.get("is_business")):
+                    raise ValidationError({
+                        "transmission_type": "The transmission type '%s' must be used for this country or address type." % (
+                            transmission_type.identifier,
```
luelista:
```suggestion
                            transmission_type.public_name,
```

luelista (commented):
A few remarks from a first look into it (mostly into the TransmissionType / TransmissionProvider structure, and the UI – not so much into the changes to email sending etc.)

src/pretix/control/templates/pretixcontrol/event/invoicing.html
```diff
@@ -59,6 +60,99 @@ <h1>{% trans "Invoice settings" %}</h1>
                 {% bootstrap_field form.invoice_renderer_highlight_order_code layout="control" %}
                 {% bootstrap_field form.invoice_eu_currencies layout="control" %}
             </fieldset>
+            <fieldset>
+                <legend>{% trans "Invoice transmission" %}</legend>
+                <p class="help-block">
```
luelista:
```suggestion
                <p>
```
The gray text color makes it a bit hard to read

src/pretix/control/templates/pretixcontrol/event/invoicing.html
```diff
@@ -59,6 +60,99 @@ <h1>{% trans "Invoice settings" %}</h1>
...
+                    {% blocktrans trimmed %}
+                        pretix can transmit invoices using different transmission methods. Different transmission methods
+                        might be required depending on country and industry. By default, sending invoices as PDF files
+                        via email is always available. Other types of transmission can be added by plugins.
+                    {% endblocktrans %}
+                </p>
+                <p class="help-block">
```
luelista:
```suggestion
                <p>
```

doc/api/resources/invoices.rst
```diff
@@ -119,6 +128,79 @@ internal_reference                    string                     Customer's refe
...
+
+===================================== ========================== =======================================================
+Field                                 Type                       Description
+===================================== ========================== =======================================================
+transmission_email_other              boolean                    Optional. Whether the invoice should be transmitted to
+                                                                 an email address other than the order email address.
+                                                                 Business customers only.
+transmission_email_address            string                     Required when ``transmission_email_other`` is set,
+                                                                 ignored otherwise. The email address that the invoice
+                                                                 should be sent to.
```
luelista: This feels a bit redundant, why not just check whether `transmission_email_address` is set or not?

Edit: At second glance, it's probably to get the checkbox + textbox UI? It might be better to add the checkbox in some other way (e.g. using Javascript in the frontend? or a Form Field type with template which contains a checkbox toggling the field?). I think it complicates the logic in other places, because we need to make sure the two fields fit correctly. Also, it's a bit unwieldy in the Order details in backend:
image

src/pretix/base/services/invoices.py
luelista: "within the last hour" - not sure if the comment is or wrong, or just somehow confusing to me, but the code below transmits only invoices that are older than an hour, not the invoices created within the last hour?

luelista (changes requested):
Now I also went through the email logic, there are a few inconsistencies left around reissuing/sending mails through the backend:

- wenn man bei einer Rechnung "Retransmit" klickt und keine abweichende Rechnungs-Email angegeben ist, geht trotzdem die "wir darum gebeten wurden, Ihnen die Rechnung weiterzuleiten" Nachricht raus (kann ggf etwas verwirrend sein)

  ...evtl reicht Nachricht anpassen und etwas allgemeiner formulieren? (auf Deutsch fehlt in der Nachricht außerdem ein komma vor dem "weil", aber die scheint schon älter zu sein)

- wenn man "Email invoices" klickt, geht die Nachricht immer an die Bestellungs-Email (auch wenn abweichende Rechnungs-Email gesetzt), und die Rechnung wird trotzdem als "Transmitted" gesetzt

- Die Rechnungs-E-Mails aus dem EmailTransmissionProvider tauchen nicht unter "View email history" auf

src/pretix/base/services/invoices.py
raphaelm: I think you overread the "not"

src/pretix/static/pretixbase/js/addressform.js
```diff
@@ -1,83 +1,119 @@
...
+                var selected_state = dependents.state.prop("data-selected-value");
+                if (selected_state) dependents.state.prop("data-selected-value", "");
+                dependents.state.find("option:not([value=''])").remove();
+                if (data.data.length > 0) {
+                    $.each(data.data, function (k, s) {
+                        var o = $("<option>").attr("value", s.code).text(s.name);
+                        if (selected_state === s.code) o.prop("selected", true);
+                        dependents.state.append(o);
+                    });
                 }
```
luelista:
```suggestion
                $.each(data.data, function (k, s) {
                    var o = $("<option>").attr("value", s.code).text(s.name);
                    if (selected_state === s.code) o.prop("selected", true);
                    dependents.state.append(o);
                });
```

src/pretix/static/pretixbase/js/addressform.js
```diff
@@ -1,83 +1,119 @@
...
-                    if (selected_value) dependents.state.prop("data-selected-value", "");
-                    dependents.state.find("option:not([value=''])").remove();
-                    if (data.data.length > 0) {
-                        $.each(data.data, function (k, s) {
+
+                if (dependents.transmission_type) {
+                    var selected_transmission_type = dependents.transmission_type.prop("data-selected-value");
+                    if (selected_transmission_type) dependents.transmission_type.prop("data-selected-value", "");
+                    dependents.transmission_type.find("option:not([value=''])").remove();
+                    if (data.transmission_types.length > 0) {
```
luelista:
```suggestion
```

src/pretix/static/pretixbase/js/addressform.js
```diff
@@ -1,83 +1,119 @@
...
+                        $.each(data.transmission_types, function (k, s) {
                             var o = $("<option>").attr("value", s.code).text(s.name);
-                            if (selected_value == s.code) o.prop("selected", true);
-                            dependents.state.append(o);
+                            if (selected_transmission_type === s.code) {
+                                o.prop("selected", true);
+                            }
+                            dependents.transmission_type.append(o);
                         });
                     }
```
luelista, line 65:
```suggestion
```

src/pretix/base/services/invoices.py
```diff
@@ -649,26 +656,52 @@ def transmit_invoice(sender, invoice_id, allow_retransmission=True, **kwargs):
...
+                data={
+                    "full_invoice_no": invoice.full_invoice_no,
+                    "transmission_provider": None,
+                    "transmission_type": invoice.transmission_type,
+                    "data": {
+                        "reason": "no_provider",
+                    },
+                }
+            )
+        return
```
luelista: I don't like the idea of choosing a different provider in test-mode vs. production, because that somewhat defeats testing. I'd rather use the same selection logic anyway, and just fail if the provider that would be selected in production doesn't support test mode:
```suggestion
            provider = p
            break

    if not provider:
        invoice.transmission_status = Invoice.TRANSMISSION_STATUS_FAILED
        invoice.transmission_date = now()
        invoice.save(update_fields=["transmission_status", "transmission_date"])
        invoice.order.log_action(
            "pretix.event.order.invoice.sending_failed",
            data={
                "full_invoice_no": invoice.full_invoice_no,
                "transmission_provider": None,
                "transmission_type": invoice.transmission_type,
                "data": {
                    "reason": "no_provider",
                },
            }
        )
        return

    if invoice.order.testmode and not provider.testmode_supported:
        invoice.transmission_status = Invoice.TRANSMISSION_STATUS_TESTMODE_IGNORED
        invoice.transmission_date = now()
        invoice.save(update_fields=["transmission_status", "transmission_date"])
        invoice.order.log_action(
            "pretix.event.order.invoice.testmode_ignored",
            data={
                "full_invoice_no": invoice.full_invoice_no,
                "transmission_provider": None,
                "transmission_type": invoice.transmission_type,
            }
        )
        return

```

src/pretix/base/invoicing/transmission.py
```diff
@@ -119,6 +127,13 @@ def verbose_name(self):
         """
         raise NotImplementedError
 
+    @property
+    def testmode_supported(self) -> bool:
+        """
+        Whether testmode invoices may be passed to this provider.
+        """
+        return True
```
luelista:
```suggestion
        return False
```
I think providers should have to explicitly declare support, either by defaulting to False or raising an exception here.

src/pretix/base/forms/questions.py
raphaelm: Good catch, unfortunately with some bad implications. I now needed to implement something I wanted to do post-PR: Solving the conflict between invoice_address_required=False and "it's impossible to create a Peppol or Italien invoice without the address". However, with email it might be in the interest of the organizer to create invoices even without address.

I don't think there's a perfect solution. I've now implemented a "empty" option for the transmission type that is only selectable when the address is completely empty (and treated like "email" by the backend for backwards compatibility). The UI isn't perfect, but I think it's the best we can do with the different customs in different countries.

In Italy, it would be common to first ask the user "Do you need an invoice?" and if they say no, don't ask for the form at all. In Germany, however, we want to be able to create invoices even when the customer doesn't add an address, so it's weird to ask "Do you need an invoice?" and when they say no, send them one anyways.

I think it's "acceptable" now and an area that we'll need to improve on in the future once we know real-world cases better. Might even need additional plugin hooks to decide when an invoice is creatable.

doc/api/resources/invoices.rst
raphaelm: I reduced it to an implementation detail on the form layer. The order details still look a bit unwieldly, but I don't have a quick idea for that that generalizes well.

raphaelm:
Hosted migration plan:

Before code deploy
```

BEGIN;

ALTER TABLE "pretixbase_invoice"
    RENAME COLUMN "sent_to_customer" TO "transmission_date";
ALTER TABLE "pretixbase_invoice"
    ADD COLUMN "sent_to_customer" timestamp with time zone DEFAULT '1970-01-01T00:00:00+00:00'::timestamptz NULL;
ALTER TABLE "pretixbase_invoice"
    ADD COLUMN "invoice_to_transmission_info" jsonb NULL;
ALTER TABLE "pretixbase_invoice"
    ADD COLUMN "transmission_info" jsonb NULL;
ALTER TABLE "pretixbase_invoice"
    ADD COLUMN "transmission_provider" varchar(255) NULL;
ALTER TABLE "pretixbase_invoice"
    ADD COLUMN "transmission_status" varchar(255) DEFAULT 'unknown' NOT NULL;
ALTER TABLE "pretixbase_invoice"
    ADD COLUMN "created" timestamp with time zone NULL;
ALTER TABLE "pretixbase_invoice"
    ADD COLUMN "invoice_to_is_business" boolean NULL;
ALTER TABLE "pretixbase_invoice"
    ADD COLUMN "transmission_type" varchar(255) DEFAULT 'email' NOT NULL;
ALTER TABLE "pretixbase_invoiceaddress"
    ADD COLUMN "transmission_info" jsonb NULL;
ALTER TABLE "pretixbase_invoiceaddress"
    ADD COLUMN "transmission_type" varchar(255) DEFAULT 'email' NOT NULL;

INSERT INTO pretixbase_event_settingsstore (object_id, key, value)
SELECT object_id, 'mail_text_order_invoice', value
FROM pretixbase_event_settingsstore
WHERE key = 'payment_banktransfer_invoice_email_text';

INSERT INTO pretixbase_event_settingsstore (object_id, key, value)
SELECT object_id, 'mail_subject_order_invoice', value
FROM pretixbase_event_settingsstore
WHERE key = 'payment_banktransfer_invoice_email_subject';

INSERT INTO django_migrations (app, name, applied) VALUES ('pretixbase', '0288_invoice_transmission', NOW());

COMMIT;
```

Then carefully, before code deploy:
```
UPDATE pretixbase_invoice
SET transmission_status = 'completed'
WHERE id IN (SELECT id from pretixbase_invoice WHERE transmission_date IS NOT NULL AND transmission_status != 'completed' LIMIT 50000);
```

After code deploy
```
BEGIN;
UPDATE pretixbase_invoice
SET transmission_status = 'completed', transmission_date = sent_to_customer
WHERE transmission_date IS NULL AND sent_to_customer > '2025-08-19T00:00:00Z'::timestamptz;

ALTER TABLE "pretixbase_invoice"
    DROP COLUMN "sent_to_customer";
COMMIT;

DELETE
FROM pretixbase_event_settingsstore
WHERE key = 'payment_banktransfer_invoice_email_subject';

DELETE
FROM pretixbase_event_settingsstore
WHERE key = 'payment_banktransfer_invoice_email_text';

```

After: delete hierarkey caches

```
from django.core.cache import cache

for event in Event.objects.only("id"):
    cache.delete('hierarkey_{}_{}'.format('event', event.id))
```

PR #5021, opened 2025-04-21
Add CSS for right-to-left language input fields

raphaelm:
See discussion in https://github.com/raphaelm/django-i18nfield/issues/39


wiffbi (approved):
As discussed seems to be the best compromise regarding handling of languages and scripts in pretix vs. i18nField. :)

PR #5034, opened 2025-04-24
[A11y] Improve customer account forms

wiffbi:
The forms for customer accounts (create, login, reset-password, etc.) all were set using the `DefaultRenderer`, which did not include any of the a11y-improvements we made with `CheckoutFieldRenderer`. This PR changes the customer forms to use the `CheckoutFieldRenderer`, which causes the forms to be a horizontal layout. It is not ideal for every form, especially the ones with longer labels (e.g. in the register form with „What is the result of X +Y?"), but IMHO it is a good compromise and now the forms are visually more like other presale-checkout forms.

Before:
Bildschirmfoto 2025-04-24 um 08 00 33

After:
Bildschirmfoto 2025-04-24 um 13 32 52

### DIFF
--- a/src/pretix/presale/forms/customer.py
+++ b/src/pretix/presale/forms/customer.py
@@ -32,7 +32,6 @@ from django.contrib.auth.password_validation import (
 )
 from django.contrib.auth.tokens import PasswordResetTokenGenerator
 from django.core import signing
-from django.urls import reverse
 from django.utils.functional import cached_property
 from django.utils.html import escape
 from django.utils.translation import gettext_lazy as _
@@ -45,6 +44,7 @@ from pretix.base.forms.questions import (
 from pretix.base.i18n import get_language_without_region
 from pretix.base.models import Customer
 from pretix.helpers.http import get_client_ip
+from pretix.multidomain.urlreverse import build_absolute_uri
 
 
 class TokenGenerator(PasswordResetTokenGenerator):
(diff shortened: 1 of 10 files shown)


raphaelm (changes requested):
I think it's weird to have a form in a panel and then the buttons outside of the panel. Looks like the buttons don't belong to the login form, especially with the spacing or when SSO buttons are added:

image

Also, the "Create account" button feels very invisible when SSO is active, which is probably not a new problem but is emphasized by moving the password reset:

What do you think about this:

image

raphaelm:
> Hm, we do not have the submit-button-row inside the panel anywhere in presale (and in control AFAIK). I would like to place them consistently, although especially with these smaller panels and short forms, it looks a bit to far from the form.

We also have primary action buttons in panels, e.g. in the cart:

image

In the checkout process, the panels function as something like a fieldset, grouping various fields of a big form together. In almost all cases, there are at least theoretically multiple panels above the button group.

Here, the panel does not really function as a grouping of fields, as there are only two fields which are grouped together. The panel looks a lot more like a border around the entire form and then… the buttons are missing.

I'm okay with taking the SSO buttons below the panel, but I really feel that the login button belongs within the panel, grouped with the username/password fields.

wiffbi:
I will move the buttons inside, I think with providers being `.btn-default` and outside of the panel that might be the perfect solution! Thanks for pointing it out! :) I will also look into the other forms in the customer area, they all have the buttons outside the panel and probably should have them inside.

wiffbi:
Just for it not being over-read: regarding your suggestion for the „create account“ link having a `.btn-default` style, I think it is misleading styling the link inside a form context as a button, as it suggests, it is a secondary submit-button and creates the account with the email and password from above. That is why I tried to keep it as a text-link.

raphaelm:
Okay, fine with me!

PR #5044, opened 2025-04-27
Add option to send emails to all attendees

juelg:
Let's the admin configure whether emails should be sent to all attendees including attendees that have the same email as the order. This allows the usage of placeholders that are position specific e.g. position secrets to be sent to all including the person that ordered the ticket.

For context:
We are currently integrating pretix as our presale solution into [StuStaPay](https://github.com/stustapay/stustapay), an opensource festival payment solution. We would like to allow our customers to be able to login to our customer portal to charge their payment account right after they have payed their ticket. For that we link the pretix position secret with the customer accounts in StuStaPay and allow our customers to login with it. A [special plugin](https://github.com/stustapay/pretix_placeholders) has been developed to access position secrets in email placeholders. The problem is that placeholders only have access to the position attributes in the attendee emails and not the normal confirmation mails. Activating attendee emails and configuring our instance to require the email address from each attendee lead emails only been sent to attendees that have a different email address than the purchaser. This option is probably implemented to reduce the amount of emails, but in our case the purchaser themself would not receive the login link.

Therefore, we decided to implement a new option that allows to sent the attendee emails to all attendees including the purchaser. We only require this option for the payment conformation, but implemented it also for they other emails for consistency.


raphaelm:
Thank you for the PR! We've discussed this in the team and are undecided whether we want to merge this into mainline pretix. On one hand, we see that it is useful, on the other hand, we are quite unhappy about the user experience of adding more fine-grained checkboxes into this already complex settings view. We feel that this would more belong towards the top of the page as a general selection affecting all emails, but that is in contradiction with "send emails to attendees" being an per-template choice in the first place. I currently do not have a good idea on how to best proceed.

juelg:
Thanks for the reply. I understand your point with the user experience. How about we replace both checkboxes with a new single dropdown menu which would have the options "no email to attendees", "email to attendees with different email address" and "email to all attendees"? The use experience would be similar to before and the unnecessary case where the "email_all_attendees" is set to true but "email_attendees" is set to false would then no longer be possible.

juelg:
@raphaelm any opinion on my suggestion or updates on your team discussion? If you agree, I'd be happy to implemented the required changes :)

raphaelm:
I think something like this might happen some day, but the implications are too unclear. Closing this PR because we are drowning a little in open PRs and it's not a priority. It's still around if we need it.

PR #5056, opened 2025-04-29
Double item thumbnail size for high-density screens

fronbasal:
On high-pixel density screens (such as modern smartphones), the 60x60 thunbnail size looks kind of grainy, especially with text embedded. 
This PR suggests doubling the thumbnail resolution and preserving the size of the img tag via the width attribute.
Naturally this increases traffic / decreases load speed, thus I totally understand if this is not a feasible option.
Maybe adding this as a configuration option could make sense?

Thank you kindly for considering and reviewing! I have sent the CLA to the support email as outlined in the contributor's guide.

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/fragment_addon_choice.html
+++ b/src/pretix/presale/templates/pretixpresale/event/fragment_addon_choice.html
@@ -47,8 +47,9 @@
                                         data-title="{{ item.name|force_escape|force_escape }}"
                                         {# Yes, double-escape to prevent XSS in lightbox #}
                                         data-lightbox="{{ item.id }}">
-                                    <img src="{{ item.picture|thumb:'60x60^' }}"
-                                            alt="{{ item.name }}"/>
+                                    <img src="{{ item.picture|thumb:'120x120^' }}"
+                                         width="60"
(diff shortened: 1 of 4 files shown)


raphaelm:
@cla-bot[bot] check

raphaelm:
Hmm, this would indeed be the easiest solution, at the expense of more traffic and load speed.

The "correct" option would be using `srcset` instead. However, the problem is then when a user actually uploads a 60x60 picture, we'd either need to detect that and skip the srcset (hard) or articially scale the picture up (useless and bad). :\

fronbasal:
Wholeheartedly agree.. on a related note (let me know if this derails this conversation too much): has there been discussion on implementing WebP compression for images? Or are there security concerns? 

If I find some time I'll attempt to implement the first outlined option (detect initial size & compile srcset from there), thanks for the review!

raphaelm:
Just after sending my comment, I had an idea how to implement this with little effort and overhead and got nerdsniped ;) See #5059!

We have not discussed webp yet, I don't think I see a significant security concern or downside but I haven't looked into it deeper.

fronbasal:
Magic! Thank you! 🚀😍

PR #5059, opened 2025-04-30
Provide high-res versions of product and event images

raphaelm:
Replaces previous attempts #3235 and #5056, see also #3506

### DIFF
--- a/src/pretix/helpers/templatetags/thumb.py
+++ b/src/pretix/helpers/templatetags/thumb.py
@@ -65,7 +65,7 @@ def thumbset(source, arg):
     if not cached_thumbset:
         for thumbsize, factor in get_srcset_sizes(arg):
             try:
-                t = get_thumbnail(source, thumbsize, formats=formats, only_if_resized=True)
+                t = get_thumbnail(source, thumbsize, formats=formats, skip_if_limited_by_input=True)
                 if t:
                     srcs.append(f"{t.thumb.url} {factor}")
             except:
--- a/src/pretix/helpers/templatetags/thumb.py
+++ b/src/pretix/helpers/templatetags/thumb.py
@@ -22,10 +22,11 @@
 import logging
 
 from django import template
+from django.core.cache import cache
 from django.core.files.storage import default_storage
 
 from pretix import settings
-from pretix.helpers.thumb import get_thumbnail
+from pretix.helpers.thumb import get_srcset_sizes, get_thumbnail
 
 register = template.Library()
 logger = logging.getLogger(__name__)
@@ -46,3 +47,30 @@ def thumb(source, arg):
         # default_storage.url works for all files in NanoCDNStorage. For others, this may return an invalid URL.
         # But for a fallback, this can probably be accepted.
         return source.url if hasattr(source, 'url') else default_storage.url(str(source))
+
+
+@register.filter
+def thumbset(source, arg):
+    cache_key = f"thumbset:{source}:{arg}"
+    cached_thumbset = cache.get(cache_key)
+    if cached_thumbset is not None:
+        return cached_thumbset
+    formats = list(set().union(
+        settings.PILLOW_FORMATS_IMAGE,
+        settings.PILLOW_FORMATS_QUESTIONS_FAVICON,
+        settings.PILLOW_FORMATS_QUESTIONS_IMAGE
+    ))
+
+    srcs = []
+    if not cached_thumbset:
+        for thumbsize, factor in get_srcset_sizes(arg):
+            try:
+                t = get_thumbnail(source, thumbsize, formats=formats, only_if_resized=True)
+                if t:
+                    srcs.append(f"{t.thumb.url} {factor}")
+            except:
+                logger.exception(f'Failed to create thumbnail of {source} at {thumbsize}')
+
+    srcset = ", ".join(srcs)
+    cache.set(cache_key, srcset, timeout=3600)
+    return srcset
--- a/src/pretix/helpers/thumb.py
+++ b/src/pretix/helpers/thumb.py
@@ -114,6 +114,9 @@ def get_srcset_sizes(size):
 
 
 def get_sizes(size, imgsize):
+    """
+    :return: Tuple of (new_size, crop_box, size_limited_by_input)
+    """
     crop = False
     if size.endswith('^'):
         crop = True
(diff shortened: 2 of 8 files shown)


src/pretix/helpers/templatetags/thumb.py
```diff
@@ -46,3 +47,30 @@ def thumb(source, arg):
...
+        settings.PILLOW_FORMATS_IMAGE,
+        settings.PILLOW_FORMATS_QUESTIONS_FAVICON,
+        settings.PILLOW_FORMATS_QUESTIONS_IMAGE
+    ))
+
+    srcs = []
+    if not cached_thumbset:
+        for thumbsize, factor in get_srcset_sizes(arg):
+            try:
+                t = get_thumbnail(source, thumbsize, formats=formats, only_if_resized=True)
```
wiffbi, line 68: `only_if_resized=True` creates the problem, that if I e.g. upload a product image sized 180x180, which would be perfect for 3x is discarded from srcset as it is not resized. So we would need something like `only_if_smaller_or_same`?

wiffbi (changes requested):
For images with a fixed display-width, the approach using x2, x3 is sufficient – one caveat though regarding uploading perfectly sized images for e.g. 3x or even 2x (the latter would currently result in only 1x images). Therefore my request for changes.

A sidenote: If an image is used as a higher-density replacement (ie. x2 or x3), they can be compressed far more before have visible artefacts – especially when using jpg. For our small product images in 60x60 or 180x180 I think it is negligible though.

For our dynamic sized event_logo and organizer_logo, the density-approach does not work that well as the images are resized on smaller screens and automatically get a higher density. That actually creates requests for larger files than needed especially on mobile. For these dynamic sized images we would need srcset with w-notation combined with sizes-attribute to fully exploit the power of srcset. As an pragmatic approach to support high-density displays though, I think it is okay for now, but we should look into supporting srcset with sizes.

src/pretix/helpers/templatetags/thumb.py
raphaelm: This was harder to fix than expected :sweat_smile:  Fixed now, happy for another look

raphaelm:
> A sidenote: If an image is used as a higher-density replacement (ie. x2 or x3), they can be compressed far more before have visible artefacts – especially when using jpg. For our small product images in 60x60 or 180x180 I think it is negligible though.

Interesting! Does that hold even for JPGs that include text? Because that's where the missing hidensity is usually visible.

> For our dynamic sized event_logo and organizer_logo, the density-approach does not work that well as the images are resized on smaller screens and automatically get a higher density. That actually creates requests for larger files than needed especially on mobile. For these dynamic sized images we would need srcset with w-notation combined with sizes-attribute to fully exploit the power of srcset. As an pragmatic approach to support high-density displays though, I think it is okay for now, but we should look into supporting srcset with sizes.

Hmm, good point. I'm not sure if we should continue this way then. However, I don't recall why I decided against the "w" approach, I looked at it first and something about it did not work for us.

raphaelm:
I think the main issue is that with the "w" approach, we need to know the *actual* size of the image, not the intended one, which we don't store, at least not for existing thumbnails.

Since this mostly concerns the event header, what are your thoughts on keeping the 2x,3x approach but use different base sizes on mobile and desktop?

src/pretix/helpers/thumb.py
```diff
@@ -112,37 +132,49 @@ def get_sizes(size, imgsize):
     else:
         size = [int(size), int(size)]
 
+    wfactor = min(1, size[0] / imgsize[0])
+    hfactor = min(1, size[1] / imgsize[1])
+    limited_by_input = size[1] / imgsize[1] > 1.0 and size[0] / imgsize[0] > 1.0
```
wiffbi: What is the reason behind comparing the ratio instead of directly comparing the numbers?

```suggestion
    limited_by_input = size[1] > imgsize[1] and size[0] > imgsize[0]
```

src/pretix/helpers/thumb.py
```diff
@@ -112,37 +132,49 @@ def get_sizes(size, imgsize):
...
+                (int((imgsize[0] * hfactor - size[0]) / 2), 0, int((imgsize[0] * hfactor + size[0]) / 2), size[1]),
+                limited_by_input
+            )
     else:
-        wfactor = min(1, size[0] / imgsize[0])
-        hfactor = min(1, size[1] / imgsize[1])
-
         if wfactor == hfactor:
-            return (int(imgsize[0] * hfactor), int(imgsize[1] * wfactor)), None
+            return (int(imgsize[0] * hfactor), int(imgsize[1] * wfactor)), None, limited_by_input
```
wiffbi, line 162: It does not make a difference, but semantically wouldn’t it be more correct to write:

```
            return (int(imgsize[0] * wfactor), int(imgsize[1] * hfactor)), None, limited_by_input
```

And couldn’t we just skip the whole multiply as both dimensions match the same factor and we can just use `size` – and to not upscale use imgsize if factor >= 1? I know, thanks to `min` above, it can never exceed 1, but to my head it reads easier and makes it clear, that we want no upscaling.

```suggestion
            return imgsize if wfactor >= 1 else size, None, limited_by_input
```

src/pretix/helpers/thumb.py
```diff
@@ -112,37 +132,49 @@ def get_sizes(size, imgsize):
...
     if crop:
         # currently crop and min-size cannot be combined
-        wfactor = min(1, size[0] / imgsize[0])
-        hfactor = min(1, size[1] / imgsize[1])
         if wfactor == hfactor:
-            return (int(imgsize[0] * wfactor), int(imgsize[1] * hfactor)), \
-                   (0, int((imgsize[1] * wfactor - imgsize[1] * hfactor) / 2),
-                    imgsize[0] * hfactor, int((imgsize[1] * wfactor + imgsize[1] * wfactor) / 2))
+            return (
+                (int(imgsize[0] * wfactor), int(imgsize[1] * hfactor)),
```
wiffbi, line 143: See comment below, but this makes it more clear to me that „resize to size, but do not upscale“:
```suggestion
                imgsize if wfactor >= 1 else size,
```

src/pretix/helpers/thumb.py
```diff
@@ -112,37 +132,49 @@ def get_sizes(size, imgsize):
...
+                (0, int((imgsize[1] * wfactor - imgsize[1] * hfactor) / 2),
+                 imgsize[0] * hfactor, int((imgsize[1] * wfactor + imgsize[1] * wfactor) / 2)),
+                limited_by_input
+            )
         elif wfactor > hfactor:
-            return (int(size[0]), int(imgsize[1] * wfactor)), \
-                   (0, int((imgsize[1] * wfactor - size[1]) / 2), size[0], int((imgsize[1] * wfactor + size[1]) / 2))
+            return (
+                (int(size[0]), int(imgsize[1] * wfactor)),
```
wiffbi, line 150: Size already has ints.

```suggestion
                (size[0], int(imgsize[1] * wfactor)),
```

src/pretix/helpers/thumb.py
```diff
@@ -112,37 +132,49 @@ def get_sizes(size, imgsize):
...
+                (0, int((imgsize[1] * wfactor - size[1]) / 2), size[0], int((imgsize[1] * wfactor + size[1]) / 2)),
+                limited_by_input
+            )
         else:
-            return (int(imgsize[0] * hfactor), int(size[1])), \
-                   (int((imgsize[0] * hfactor - size[0]) / 2), 0, int((imgsize[0] * hfactor + size[0]) / 2), size[1])
+            return (
+                (int(imgsize[0] * hfactor), int(size[1])),
```
wiffbi, line 156:
```suggestion
                (int(imgsize[0] * hfactor), size[1]),
```

src/pretix/helpers/thumb.py
```diff
@@ -112,37 +132,49 @@ def get_sizes(size, imgsize):
     else:
         size = [int(size), int(size)]
 
+    wfactor = min(1, size[0] / imgsize[0])
+    hfactor = min(1, size[1] / imgsize[1])
+    limited_by_input = size[1] / imgsize[1] > 1.0 and size[0] / imgsize[0] > 1.0
+    print(size, imgsize, size[1] / imgsize[1], size[0] / imgsize[0], limited_by_input)
```
wiffbi:
```suggestion
```

wiffbi (approved):
Sorry it tooke me so long … just some minor comments regarding better understanding of the code.

Regarding the event-header. That is a tough one. I think your original approach with x2 and x3 is actually good enough. The case I described is actually just an edge-case: a header-image that exceeds the device-width and gets scaled down on the client and works as an kind of x2 (probably just an x1.5, so the x2 would even be better here). Also x2 does not mean twice the filesize. So I’d argue, it is good enough to use x2 and x3 on the header as well. The header image is so flexible depending on its size, it would get rather complex to improve the image selection – IMHO not worth the small gain.

wiffbi:
One last suggestion: for the product images, we should probably either add CSS to define at least the width (and maybe height) and/or even add width/height attributes to the ``, so the browser can better decide, which image to download. It is not needed for the pixel-density srcset to work, but helps the browser to prepare space when layouting the product-list even before the image is loaded. Also adding width/height helps to keep the layout intact should images fail to load.

PR #5060, opened 2025-04-30
[A11y] fix tablist issues

wiffbi:
This PR fixes some a11y with the old boostrap-tabs implementation. It adds a new implementation loosely based on https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/

In presale, tabs are currently only used for payment qr-codes. This PR uses a simpler nav-pills-like tab-navigation instead of the previous box-inside-a-box tab-interface – on the hand, UI is lighter (alhtough the btn-primary might be a bit bold, but btn-default does not feel light a „selected“-state) and on the other hand, the original panel-style tabs was not possible to implement with the current set of styles due to not allowed HTML-nesting of li-elements inside an element with role=tablist.

Bildschirmfoto 2025-04-30 um 12 47 40

### DIFF
--- a/src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/pending.html
+++ b/src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/pending.html
@@ -126,7 +126,7 @@ SCT
             {% endif %}
             </div>
             <p class="text-muted" id="banktransfer_qrcodes_label">
-                {% trans "Scan the qr-code with your banking app" %}
+                {% trans "Scan the QR code with your banking app" %}
             </p>
         </div>
     {% endif %}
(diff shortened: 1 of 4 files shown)


src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/pending.html
```diff
@@ -85,20 +79,55 @@
...
+                    id="banktransfer_qrcodes_bezahlcode_tab"
+                    type="button"
+                    role="tab"
+                    aria-controls="banktransfer_qrcodes_bezahlcode"
+                    aria-selected="false"
+                    tabindex="-1">BezahlCode</button>
+            {% endif %}
+            </div>
+            <p class="text-muted" id="banktransfer_qrcodes_label">
+                {% trans "Scan the qr-code with your banking app" %}
```
raphaelm, line 129: Not new, but still wrong
```suggestion
                {% trans "Scan the QR code with your banking app" %}
```

raphaelm (approved):

PR #5061, opened 2025-04-30
[A11y] fix datepicker rejecting without error message

wiffbi:
This PR adds a `strong[role=alert]` to every label of a CheckoutFieldRenderer. Every element with a role=alert needs to be present in the DOM onload, otherwise it does not get read out. The „live“ error-message is displayed at the label and by that differs from the normal server-side errors, which are displayed after the input with a aria-describedby for several reasons:

1. The alert is only read out when used inside the label. When used inside a help-text linked with aria-describedby (like all server-side errors), it is not read by VoiceOver. One could add a help-text element after the input with role=alert, that is not linked by aria-describedby, but then the error is only read once the error-message was changed, not again when one leaves the date-input and enters it again. So the error-message is not that helpful.
2. It does not make the form „jump“ so much on showing the error message as it is either inline after the label (smaller screens) or below the left-floated label.
3. In this special case, the calendar-widget opens on focus and usually overlays the help-text following the input – exception being on smaller screens where the label is above the input and when the user has not scrolled as far so that the calendar-widget is opened above the input. We might hide the calender-widget once we encountered a dp.error and focussed the input again as dp.error is usually only triggered when a date is manually entered not using the widget.

Bildschirmfoto 2025-05-06 um 12 27 59

Bildschirmfoto 2025-05-06 um 12 28 19

### DIFF
--- a/src/pretix/presale/forms/renderers.py
+++ b/src/pretix/presale/forms/renderers.py
@@ -65,6 +65,8 @@ def render_label(content, label_for=None, label_class=None, label_title='', labe
     elif not optional:
         opt += '<i class="label-required">{}</i>'.format(pgettext('form', 'required'))
 
+    opt += '<strong class="label-alert" role="alert"></strong>'
+
     builder = '<{tag}{attrs}>{content}{opt}</{tag}>'
     return format_html(
         builder,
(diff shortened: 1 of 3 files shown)


wiffbi:
As discussed, I choose the more robust solution to just validate server-side and just use `keepInvalid` on datetimepicker.

PR #5087, opened 2025-05-12
[A11y] add skip-link to main content

wiffbi:
This PR adds a skip-link as the first element on the page to skip to the main-area on the page. JavaScript makes sure, there will be an element `*#content` available on the page, that can be linked to. The skip-link is only is visible when it has focus.

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/base.html
+++ b/src/pretix/presale/templates/pretixpresale/base.html
@@ -44,7 +44,7 @@
 </head>
 <body class="nojs" data-locale="{{ request.LANGUAGE_CODE }}" data-now="{% now "U.u" %}" data-datetimeformat="{{ js_datetime_format }}" data-timeformat="{{ js_time_format }}" data-dateformat="{{ js_date_format }}" data-datetimelocale="{{ js_locale }}" data-currency="{{ request.event.currency }}">
 {{ html_page_header|safe }}
-<nav id="skip-to-main" role="navigation" aria-label="{% trans "Skip-Link" %}" class="sr-only on-focus-visible">
+<nav id="skip-to-main" role="navigation" aria-label="{% trans "Skip link" context "skip-to-main-nav" %}" class="sr-only on-focus-visible">
   <p><a href="#content">{% trans "Skip to main content" %}</a></p>
 </nav>
 <header>
--- a/src/pretix/presale/templates/pretixpresale/base.html
+++ b/src/pretix/presale/templates/pretixpresale/base.html
@@ -44,6 +44,9 @@
 </head>
 <body class="nojs" data-locale="{{ request.LANGUAGE_CODE }}" data-now="{% now "U.u" %}" data-datetimeformat="{{ js_datetime_format }}" data-timeformat="{{ js_time_format }}" data-dateformat="{{ js_date_format }}" data-datetimelocale="{{ js_locale }}" data-currency="{{ request.event.currency }}">
 {{ html_page_header|safe }}
+<nav id="skip-to-main" role="navigation" aria-label="{% trans "Skip-Link" %}" class="sr-only on-focus-visible">
+  <p><a href="#content">{% trans "Skip to main content" %}</a></p>
+</nav>
 <header>
 {% if ie_deprecation_warning %}
     <div class="old-browser-warning">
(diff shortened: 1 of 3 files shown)


src/pretix/presale/templates/pretixpresale/base.html
```diff
@@ -44,6 +44,9 @@
 </head>
 <body class="nojs" data-locale="{{ request.LANGUAGE_CODE }}" data-now="{% now "U.u" %}" data-datetimeformat="{{ js_datetime_format }}" data-timeformat="{{ js_time_format }}" data-dateformat="{{ js_date_format }}" data-datetimelocale="{{ js_locale }}" data-currency="{{ request.event.currency }}">
 {{ html_page_header|safe }}
+<nav id="skip-to-main" role="navigation" aria-label="{% trans "Skip-Link" %}" class="sr-only on-focus-visible">
```
raphaelm, line 47:
```suggestion
<nav id="skip-to-main" role="navigation" aria-label="{% trans "Skip link" %}" class="sr-only on-focus-visible">
```

src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -263,6 +263,10 @@ function setup_basics(el) {
 
     el.find(".js-only").removeClass("js-only");
     el.find(".js-hidden").hide();
+    // make sure to always have a #content for skip-link to work
+    if (!el.find("#content").length) {
+        el.find("main").first().attr("id", "content");
```
raphaelm, line 268: The comment says "always", but we do not have a main element on every page. For example, I found no #content on the customer account login page, the waiting list page, pages generated by pretix-exhibitor or pretix-pages, etc.

Either we need to add main elements everywhere, or hack in another fallback like this:

```
$(".page-header").first().next()
```

src/pretix/static/pretixpresale/scss/main.scss
```diff
@@ -69,6 +69,17 @@ html {
...
+.on-focus-visible:focus, .on-focus-visible:focus-within {
+    height: auto;
+    width: auto;
+    overflow: auto;
+    clip: auto;
+}
+#skip-to-main {
+    /* padding is needed to make focus-outline visible */
+    padding: 6px;
+}
```
raphaelm, line 81: If it's on-focus-visible, it probably needs a color with enough contrast to the background, doesn't it?

image

raphaelm (changes requested):

src/pretix/presale/templates/pretixpresale/base.html
wiffbi: Hm, you are right, wikipedia doesn’t use a hyphen. For me though „Skip link“ reads like a call to action to skip the link, whereas „skip-link“ (maybe all lowercase?) is the name of a link, that skips something. When thinking of translation, in german I would translate „Skip link“ with „Link überspringen“, whereas „skip-link“ would either stay „skip-link“ (sorry, denglisch) or „Überspringen-Link“. So we would somehow make sure, what is meant here. Should we add a context for translation? Something like `context "a11y-skip-to-nav"`?

src/pretix/static/pretixpresale/js/ui/main.js
wiffbi: Good catch again, the pages I checked all had `main`. I also stumbled about the „Best-practice“ recommendation in our a11y-report that every page should have a main landmark. So I guess, at some point we should add a main-landmark to all pages, but for a robust solution, we should use your suggestion as a fallback, when no main is found. Although currently for my install this jumps to an alert-warning, but that is actually a good idea to not skip an alert-warning. I will add this.

src/pretix/static/pretixpresale/scss/main.scss
wiffbi: Fixed it with `--pretix-link-contrast-color`. I also added a background-color to overlay the back-to-organizer-link, if the viewport is to narrow.

src/pretix/static/pretixpresale/js/ui/main.js
wiffbi: I used `document.querySelector('main') || document.querySelector('.page-header + *')` as a fallback, if no `#content` is available in the document.

raphaelm (approved):

PR #5090, opened 2025-05-12
Allow to set multiple email invoice recipients (Z#23190766)

raphaelm:


wiffbi (approved):
Lloks good. We might add a hint in the help-text, that multiple email-addesses are possible?

PR #5097, opened 2025-05-13
Fix weird markup on "resend link" page

raphaelm:
Before:

image

After:

image


wiffbi (approved):
Form-Input needs a label, but I will fix that in another PR.

PR #3092, opened 2023-02-07
Event index: Increase white-space for products and headlines

wiffbi:
This PR tries to improve the visual appearance by increasing the white-space around products and headlines. It furthermore harmonizes line-width between products to 1px as used in other places for lines. Furthermore it fixes a minor padding-top issue with cart where the first item of the cart is too far away from the top/panel-title.

PR #5098, opened 2025-05-13
Adjust headlines aside from front page

raphaelm:
In #3092, we increased the margin of headlines to make the frontpage look cleaner. We never noticed that we not only affected the frontpage, and I think on the other pages it looks weirdly like too much space.

Before:

image

After:

image

It's even more accentuated with e.g. banner text. Before:

image

After:

image

@wiffbi If you disagree this is an improvement, feel free to close


wiffbi (approved):
The changes in [resend_link.html](https://github.com/pretix/pretix/pull/5098/files#diff-a98b0614d9a26a8e40eb15617d6ffb8d8178689b1d082cbce467edb2715f7cd4) are part of #5097, so I think they can ge discarded?

PR #5099, opened 2025-05-13
Metrics: Fix Content Type header

stranck:
Apparently (https://github.com/prometheus/prometheus/issues/15777) starting from Prometheus v3 the ContentType header of /metrics responses is now properly validated and, if it fails, the service is by default detected as offline. Without this fix Pretix returns `text/html; charset=utf-8` as ContentHeader for /metrics requests, which makes Prometheus' parser fail (By its logs, with error: `level=ERROR source=scrape.go:1609 msg="Failed to determine correct type of scrape target." component="scrape manager" scrape_pool=pretix target=http://xxxx/metrics content_type="text/html; charset=utf-8" fallback_media_type="" err="received unsupported Content-Type \"text/html; charset=utf-8\" and no fallback_scrape_protocol specified for target"`)

The list of correct ContentType headers can be found in the prometheus's source code (Linked as reply in this issue: https://github.com/prometheus/prometheus/issues/15830#issuecomment-2602669034)


stranck:
(also sorry for accidentally spamming the various PR I have immediately closed)

raphaelm:
This is a weird content type, but … well, if they say so …

src/pretix/base/views/metrics.py
```diff
@@ -80,4 +80,4 @@ def serve_metrics(request):
 
     content = "\n".join(output) + "\n"
 
-    return HttpResponse(content)
+    return HttpResponse(content, content_type="text/plain;version=1.0.0;escaping=allow-utf-8")
```
raphaelm, line 83:
```suggestion
    return HttpResponse(content, content_type="text/plain;version=1.0.0;escaping=allow-utf-8")

```

PR #5101, opened 2025-05-13
Add common modal support for no-product-selected-error

wiffbi:
This PR explores ways to add common modal-dialogs which make it easier to create accessible modals.

The current idea is to use events `pretix:alert` or  `pretix:info` to trigger a listener, that fills the dialogs with the info and does showModal() on the correct one. In the current example, cart triggers a pretix:alert to have a role=alertdialog open and show a confirmation-button. The async-task could e.g. trigger pretix:info, which opens a normal dialog without the option of a confirm-button.

The current layout of the overlay is based on the #loadingmodal. An alternative might be to use a style similar to the existing `alert-*` boxes, which allow for different colors depending on the level (e.g. info or warning or even danger). Another alternative might be a panel.

Currently no returnValue is provided when the alertdialog is closed as there is no need for currently. It could be done in a number of ways. Easiest would be a callback in the triggered event-details – although combining an event with a callback seems to be a strange combination. Another way could be another event, e.g. `document.addListener("pretix:alert:close", function() …, { once: true })`. Last, but not least, when e.g. not using events at all, we could use a global function `pretix.alert()` instead, that returns a Promise, which we could then act upon.

### DIFF
--- a/src/pretix/static/pretixbase/js/asynctask.js
+++ b/src/pretix/static/pretixbase/js/asynctask.js
@@ -225,8 +225,7 @@ $(function () {
         async_task_old_url = location.href;
         $("body").data('ajaxing', true);
 
-        const ac = new AbortController();
-        window.pretix.dialog({
+        window.pretix.dialog.open({
             label: this.getAttribute("data-asynctask-headline") || gettext("We are processing your request …"),
             message: (this.getAttribute("data-asynctask-text") || "") + gettext(
                 'We are currently sending your request to the server. If this takes longer ' +
@@ -234,10 +233,10 @@ $(function () {
                 'this page and try again.'
             ),
             icon: 'cog',
-        }, ac.signal);
+        });
 
         window.setTimeout(function() {
-            ac.abort();
+            window.pretix.dialog.close();
         }, 2000);
         return false;
 
--- a/src/pretix/static/pretixbase/js/asynctask.js
+++ b/src/pretix/static/pretixbase/js/asynctask.js
@@ -117,7 +117,6 @@ function async_task_callback(data, jqXHR, status) {
     $("body").data('ajaxing', false);
     if (data.redirect) {
         if (async_task_is_download && data.success) {
-            waitingDialog.hide();
             if (location.href.indexOf("async_id") !== -1) {
                 history.replaceState({}, "pretix", async_task_old_url);
             }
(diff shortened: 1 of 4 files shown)


src/pretix/static/pretixbase/js/asynctask.js
```diff
@@ -131,22 +130,21 @@ function async_task_callback(data, jqXHR, status) {
 
     if (async_task_is_long) {
         if (data.started) {
-            $("#loadingmodal p.status").text(gettext(
-                'Your request is currently being processed. Depending on the size of your event, this might take up to ' +
-                'a few minutes.'
-            ));
+            window.pretix.dialog.open({
```
raphaelm, line 133: Hmm this is a pointer to me that the API is not great yet. This is not an opening action, this is a status update for an ongoing operation. The backend also includes dialogs that show a process bar or a multi-step process, would these need to be closed and opened every few hundred milliseconds?

wiffbi: That is a good point. So one way could be, that we get access to the dialog and the promise. What about instead of just returning a Promise, we return an object containing the Promise and the dialog (could be the HTMLElement or IMHO better something encapsulating it). That way we could do things like:

```
const { listen, dialog } = window.pretix.dialog.open(…);
listen.then(/* do stuff with promise */);
$("h2", dialog).text("Update content …");
```

If we do not need to update the content, but just want to wait for the response:

`window.pretix.dialog.open(…).listen.then(/* do stuff with promise */)`

Or if we do not care about a response as it is just a normal status-dialog, we can do:

```
const d = window.pretix.dialog.open(…).dialog;
// doing fetch, etc.
$(".progressbar", d).data("progress", progress);
```

I would even go so far as to instead of returning the HTMLElement directly, encapsluate it in a another object, so we could e.g. do something like `d.update({label: 'Update …'})`. If we need access to the dialog HTMLElement directly, we could provide it through `d.el`or something similar – although I’d strongly advise against accessing the HTMLElement dialog directly as it needs to be reused.

One big plus using a container object is, that as we need to reuse the HTMLElement, one needs to be very careful when getting access to the HTMLElement directly. So the containing object itself could e.g. listen to the close-event of the dialog and once it has been closed, disable access to the actual HTMLElement through `d.update(…)`.

If we really need direct DOM-access of the dialog for the example for the loading-modal, I’d suggest we add the option to use one’s own HTMLdialog in `window.pretix.dialog.open` like so:

```
HTML:
<dialog id="loading-modal">… our super custom HTML here …</dialog>

JS:
const loadingDialog = document.getElementById("loading-modal");
window.pretix.dialog.open(loadingDialog).listen.then(/* do promise stuff */);
```

So instead of passing an options object, we just pass a HTMLDialogElement that basically gets wrapped in our promise. The only question is, what do we do with the wrapping dialog-container and a call to `d.update(…)`? Similar to `window.pretix.dialog.open` we could accept a HTMLElement, that we just replace the HTML-content of the HTMLDialogElement? Or instead of returning a wrapping dialog-container, when using a HTMLDialogElement instead of an options-object, we could return the passed HTMLDialogElement directly. So like so:

```
HTML:
<dialog id="loading-modal">… our super custom HTML here …</dialog>

JS:
const loadingDialog = document.getElementById("loading-modal");
const { listen, dialog } = window.pretix.dialog.open(loadingDialog);
if (dialog === loadingDialog) {}// always true
```

PR #5107, opened 2025-05-14
[A11y] Widget fix event status color contrast

wiffbi:
This one is ugly, we need to meet contrast value of 4.5 for texts in calendar view in the widget. This PR uses the colors used for text in alert-boxes. They have a slightly larger contrast than necessary as they do not work against white but a slight shade of the respective alert-color.

The super ugly problem is, what do we do with yellow or in this case: brown? There is no way we can meet the needed contrast with the current layout which is white text on a solid color.

I guess we could try an make it similar to the alert-boxes and the calendar on the main shop-page.


wiffbi:
Fixed in v2

PR #5108, opened 2025-05-14
[A11y] widget fix filter dropdown auto-submit not allowed

wiffbi:
This PR tries to fix the a11y issues with the filter dropdowns (meta-data) in the widget that does not have a submit button and auto-submits onchange (which is not allowed).


wiffbi:
As discussed on-hold

PR #5109, opened 2025-05-14
Cart renewal

luelista:


src/pretix/base/models/orders.py
```diff
@@ -3098,7 +3098,10 @@ class CartPosition(AbstractPosition):
         verbose_name=_("Expiration date"),
         db_index=True
     )
-
+    max_extend = models.DateTimeField(
```
raphaelm, line 3101: Only 80% happy with the naming, I would have chosen a more direct parallel to "expires", but "max_expires" can also be confusing… Not something I have strong feelings about

src/pretix/base/services/cart.py
```diff
@@ -286,11 +291,22 @@ def __init__(self, event: Event, cart_id: str, sales_channel: SalesChannel,
...
-        self._explicit_expiry = expiry
         self.invoice_address = invoice_address
         self._widget_data = widget_data or {}
         self._sales_channel = sales_channel
+        self.num_extended_positions = 0
+
+        if expiry and reservation_time:
+            raise TypeError('Cannot specify both expiry and reservation_time')
+        elif expiry:
+            warnings.warn('CartManager(expiry=...) is deprecated, use reservation_time=... instead')
```
raphaelm, line 302: if we keep this (see slack), we can properly classify it:

```suggestion
            warnings.warn('CartManager(expiry=...) is deprecated, use reservation_time=... instead', DeprecationWarning)
```

src/pretix/base/services/cart.py
```diff
@@ -590,6 +612,7 @@ def extend_expired_positions(self):
                 self._voucher_use_diff[cp.voucher] += 2
 
             self._operations.append(op)
+            self.num_extended_positions += 1
```
raphaelm, line 615: Would it be more "correct" to increase this only after *execution* of the ExtendOperation? Otherwise the `num_extended_positions` of a CartManager with a failed commit has a weird semantic.

src/pretix/base/services/cart.py
```diff
@@ -1416,14 +1443,11 @@ def recompute_final_prices_and_taxes(self):
     def commit(self):
         self._check_presale_dates()
         self._check_max_cart_size()
-        self._calculate_expiry()
 
         err = self._delete_out_of_timeframe()
         err = self.extend_expired_positions() or err
         err = err or self._check_min_per_voucher()
 
-        self.real_now_dt = now()
```
raphaelm, line 1425: Not really scope of this PR but since you touched it: This was there because when `_extend_expiry_of_valid_existing_positions`, it is *technically* possible that it validates cart positions that are not actually valid any more, because whenever  this code executes, `real_now_dt` is already in the past. So this way my paranoid try to make this error scenario less likely. Being aware there is no chance to fully avoid the error scenario without going deep into clock synchronization (and transaction isolation)… should we maybe just add a small timedelta (eg 5s) to the `expires__gt=self.real_now_dt` to avoid the risk, i.e. only extending the expiry if the current one is still a few seconds in the future? Selling too many tickets is always worse than to little…

src/pretix/base/services/cart.py
luelista: Removed the argument altogether

src/pretix/base/models/orders.py
luelista: I also considered `max_expires` but I thought it's too confusing. Do you think `max_expires` is better? Or maybe `extend_expiry_limit`? I'm not really attached to `max_extend` either, and I think it's unique enough the I can easily change the name with a replace-all :D

src/tests/presale/test_cart.py
```diff
@@ -1225,22 +1228,94 @@ def test_subevent_quota_partly(self):
...
+            self.client.post('/%s/%s/cart/add' % (self.orga.slug, self.event.slug), {
+                'item_%d' % (self.ticket.id,): '1'
+            }, follow=True)
+            cp.refresh_from_db()
+            with scopes_disabled():
+                positions = list(CartPosition.objects.filter(cart_id=cp.cart_id))
+            self.assertEqual(len(positions), 4)
+            for pos in positions:
+                self.assertEqual(pos.expires, start_time + timedelta(days=1) + self.cart_reservation_time)
+                self.assertEqual(pos.max_extend, start_time + timedelta(days=1) + 11 * self.cart_reservation_time)
```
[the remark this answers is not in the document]
luelista: is this the type of test you have in mind? https://github.com/pretix/pretix/pull/5109/commits/f31a62e84dd07340d14667cdc764052c8a2b1f34#diff-668d43a3117e5f4bd1ff3651d35a9fd7f0330d5ab171391ff600535999658f80R1376

src/tests/presale/test_cart.py
```diff
@@ -1225,22 +1228,94 @@ def test_subevent_quota_partly(self):
...
+            }, follow=True)
+            cp.refresh_from_db()
+            self.assertEqual(cp.expires, max_extend)
+            self.assertEqual(cp.max_extend, max_extend)
+        with freezegun.freeze_time(start_time + timedelta(minutes=65)):
+            self.client.post('/%s/%s/cart/extend' % (self.orga.slug, self.event.slug), {
+            }, follow=True)
+            cp.refresh_from_db()
+            self.assertEqual(cp.expires, start_time + timedelta(minutes=65) + self.cart_reservation_time)
+            self.assertEqual(cp.max_extend, start_time + timedelta(minutes=65) + 11 * self.cart_reservation_time)
```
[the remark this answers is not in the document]
luelista: i'm not sure how that would be different than above?

src/tests/presale/test_cart.py
```diff
@@ -1289,6 +1373,48 @@ def test_renew_expired_failed(self):
...
+                event=self.event, cart_id=self.session_key, item=self.shirt, variation=self.shirt_blue,
+                price=23, expires=now() - timedelta(minutes=10), max_extend=now() - timedelta(minutes=10)
+            )
+        response = self.client.post('/%s/%s/cart/extend' % (self.orga.slug, self.event.slug), {
+        }, follow=True)
+        doc = BeautifulSoup(response.rendered_content, "lxml")
+        self.assertIn('no longer available', doc.select('.alert-danger')[0].text)
+        with scopes_disabled():
+            self.assertFalse(CartPosition.objects.filter(id=cp1.id).exists())
+            self.assertTrue(CartPosition.objects.filter(id=cp2.id).exists())
```
raphaelm, line 1416: Exists yes, but could you also assert that the expires date was updated?

raphaelm (approved):

PR #5110, opened 2025-05-14
[A11y] Adjust thresholds of color picker

raphaelm:


wiffbi (approved):
Looks good. Do we want to fix the typo for `luminanace` to `luminance`? Nothing in particular with this PR, I just stumbled upon it reviewing this PR.

PR #5126, opened 2025-05-19
Connect giftcards with customer accounts

ser8phin:
Steps:

- [x] Add fk field to giftcard
- [x] Connect if logged in/ordered with account and refunded as giftcard
- [ ] Show customer account's giftcards (if there are any -> link back)
- [ ] Offer giftcard use


ser8phin:
These options coexist while it's WIP and we need to decide which one is the better one. Tending towards the second, where the list of gift cards is within the gift card panel.
screenschot_2025061602

wiffbi:
I prefer the 2nd option too. The only thing might be the discoverability, if we have gift-cards available … if I am a logged-in customer and get to the payment step and have not selected any payment method … when I have giftcards in my account, I only see them, when I select gift-card as the payment option.

Would it make sense to pre-select the gift-card payment option, when gift-cards are availible? And maybe we could add info to the label, like so:

Bildschirmfoto 2025-06-17 um 09 05 46

IMHO that does not necessarily need to be done server-side, but could also be added through JavaScript.

ser8phin:
> Would it make sense to pre-select the gift-card payment option, when gift-cards are availible? And maybe we could add info to the label, like so:

That, I already did that in that change, but it's easy to overlook, since it was just another if-branch :smile: 

The (X available) is a neat idea :+1: will see how to get that in there

src/pretix/presale/templates/pretixpresale/organizers/customer_giftcards.html
```diff
@@ -0,0 +1,78 @@
...
+                                        {% if gc.expires %}
+                                        {% trans " since" %}
+                                        {{ gc.expires|date:"SHORT_DATETIME_FORMAT" }}
+                                        {% endif %}
+                                    </time>
+                                {% elif gc.expires %}
+                                    {% icon "check" %}
+                                    {% trans "Valid until" %}
+                                    <time datetime="{{ gc.date_end|date:"Y-m-d H:i" }}">
+                                        {{ gc.date_end|date:"SHORT_DATETIME_FORMAT" }}
```
ser8phin:
```suggestion
                                        {{ gc.date_end|date:"SHORT_DATETIME_FORMAT" }}
```
```suggestion
                                        {{ gc.date_end|date:"SHORT_DATETIME_FORMAT" }}
```
```suggestion
                                        {{ gc.expires|date:"SHORT_DATETIME_FORMAT" }}
```

src/pretix/base/services/orders.py
```diff
@@ -2995,6 +2995,9 @@ def _try_auto_refund(order, auto_refund=True, manual_refund=False, allow_partial
                 currency=order.event.currency,
                 testmode=order.testmode
             )
+            if order.customer:
+                giftcard.customer = order.customer
+                giftcard.save()
```
raphaelm: This creates an additional UPDATE right after the insert, could you integrate this into the create() call?

src/pretix/control/templates/pretixcontrol/giftcards/checkout.html
```diff
@@ -1,7 +1,35 @@
 {% load i18n %}
 {% load bootstrap3 %}
+{% load money %}
 {% load rich_text %}
 
 {{ request.event.settings.payment_giftcard_public_description|rich_text }}
-
+{% if customer_gift_cards %}
+    <p><strong>
+        <span class="sr-only">Info:</span>
```
raphaelm: Info also needs translation

src/pretix/base/models/giftcards.py
```diff
@@ -80,6 +80,13 @@ class GiftCard(LoggedModel):
         null=True, blank=True,
         verbose_name=_('Owned by ticket holder')
     )
+    customer = models.ForeignKey(
```
raphaelm, line 83: Should we allow to manually set this field when editing the gift card?

src/pretix/control/views/orders.py
```diff
@@ -1234,6 +1234,9 @@ def perform_refund(self, comment, full_refund, payments):
                     currency=self.request.event.currency,
                     testmode=order.testmode
                 )
+                if order.customer:
+                    giftcard.customer = order.customer
+                    giftcard.save()
```
raphaelm: see above, can you integrate this into the create() call?

src/pretix/presale/checkoutflow.py
```diff
@@ -1400,6 +1400,8 @@ def get_context_data(self, **kwargs):
             ctx['selected'] = self.request.POST['payment']
         elif self.single_use_payment:
             ctx['selected'] = self.single_use_payment['provider']
+        elif "giftcard" in self.request.event.get_payment_providers():
```
raphaelm: This unconditionally makes giftcards our default payment method, I don't think we want that?

src/pretix/presale/templates/pretixpresale/event/checkout_payment.html
```diff
@@ -76,6 +76,11 @@
                                            aria-controls="payment_{{ p.provider.identifier }}"
                                            data-wallets="{{ p.provider.walletqueries|join:"|" }}" />
                                     <strong class="accordion-label-text">{{ p.provider.public_name }}</strong>
+                                    {% if p.provider.identifier == 'giftcard' and p.provider.customer_gift_cards %}
+                                        <small>
+                                            ({{p.provider.customer_gift_cards|length }}{% trans " available" %})
```
raphaelm: That space will confuse translators a lot. Better:
```suggestion
                                            ({{p.provider.customer_gift_cards|length }} {% trans "available" %})
```
However, that still requires the translator to know that a number will be prepended. Some languages might want the number at the end? Some might have different forms of the adjective "available" depending on the number. So better would be:

```suggestion
                                            {% blocktrans trimmed count count=p.provider.customer_gift_cards|length %}
                                                {{ count }} available
                                            {% plural %}
                                                {{ count }} available
                                            {% endblocktrans %}
```

src/pretix/presale/templates/pretixpresale/event/order_pay_change.html
```diff
@@ -40,6 +40,11 @@ <h2>
                                         {% if selected == p.provider.identifier %}checked="checked"{% endif %}
                                         data-wallets="{{ p.provider.walletqueries|join:"|" }}"/>
                                 <strong class="accordion-label-text">{{ p.provider.public_name }}</strong>
+                                {% if p.provider.identifier == 'giftcard' and p.provider.customer_gift_cards %}
+                                    <small>
+                                        ({{p.provider.customer_gift_cards|length }}{% trans " available" %})
```
raphaelm: see checkout_payment.html

src/pretix/base/payment.py
```diff
@@ -1373,6 +1375,24 @@ class GiftCardPayment(BasePaymentProvider):
     payment_form_class = GiftCardPaymentForm
     payment_form_template_name = 'pretixcontrol/giftcards/checkout.html'
 
+    @cached_property
+    def customer_gift_cards(self):
+        if not self.request:
+            return None
+        cs = cart_session(self.request)
+        if cs.get('customer_mode', 'guest') == 'login':
```
raphaelm: That only works in the cart, not for existing orders. You should **also** check for `getattr(self.request, "customer", None)`

src/pretix/presale/templates/pretixpresale/organizers/customer_giftcards.html
```diff
@@ -0,0 +1,78 @@
...
+                    <article class="row">
+                        <div class="col-xs-6">
+                            <h4>
+                                {% icon "gift" %} {{ gc }}
+                            </h4>
+                            {% if gc.issuance %}
+                            <p class="text-muted">
+                                {% icon "calendar" %}
+                                <span>{% trans "Issued on" %}:</span>
+                                {{ gc.issuance }}
```
raphaelm: Do we need the span? Can you merge this into a {% blocktrans %} with variables in case other translations need the variable in the middle?

src/pretix/presale/templates/pretixpresale/organizers/customer_giftcards.html
```diff
@@ -0,0 +1,78 @@
...
+                                <small>
+                                {% if gc.expired %}
+                                    {% icon "clock-o" %}
+                                    {% trans "Expired" %}
+                                    <time datetime="{{ gc.date_end|date:"Y-m-d H:i" }}">
+                                        {% if gc.expires %}
+                                        {% trans " since" %}
+                                        {{ gc.expires|date:"SHORT_DATETIME_FORMAT" }}
+                                        {% endif %}
+                                    </time>
```
raphaelm: This needs to be a blocktrans with variables, not every language will allow to split this into "Expired", "since" and a datetime in this order.

src/pretix/presale/templates/pretixpresale/organizers/customer_giftcards.html
```diff
@@ -0,0 +1,78 @@
...
+                                {% elif gc.expires %}
+                                    {% icon "check" %}
+                                    {% trans "Valid until" %}
+                                    <time datetime="{{ gc.date_end|date:"Y-m-d H:i" }}">
+                                        {{ gc.expires|date:"SHORT_DATETIME_FORMAT" }}
+                                    </time>
```
raphaelm: Same here

src/pretix/presale/templates/pretixpresale/organizers/customer_giftcards.html
```diff
@@ -0,0 +1,78 @@
...
+                        </div>
+                        <div class="col-xs-6 text-right">
+                            {% trans "Remaining value" %}:
+                            <p class="text-right">{{ gc.value|money:gc.currency }}</p>
+                        </div>
+                    </article>
+                    {% endfor %}
+                </div>
+            {% else %}
+                <p class="text-center">{% trans "You don’t have any gift cards in your account currently." %}</p>
```
raphaelm: Should we add an explainer here that (currently) only gift cards resulting from refunds show up here, and purchased gift cards show up under "orders"?

src/pretix/base/models/giftcards.py
ser8phin: Sure. Done. Should we also make it possible to set this on manually creating a giftcard then, as we're already there?

raphaelm: Yes, please!

src/pretix/base/services/orders.py
```diff
@@ -2993,6 +2993,7 @@ def _try_auto_refund(order, auto_refund=True, manual_refund=False, allow_partial
                 expires=order.event.organizer.default_gift_card_expiry if giftcard_expires is _unset else giftcard_expires,
                 conditions=giftcard_conditions,
                 currency=order.event.currency,
+                customer=order.customer if order.customer else None,
```
raphaelm: This is the same as

```suggestion
                customer=order.customer,
```

so it could be simplified.

src/pretix/control/forms/organizer.py
```diff
@@ -729,6 +729,22 @@ def __init__(self, *args, **kwargs):
...
 
+        if self.organizer.settings.customer_accounts:
+            self.fields['customer'].queryset = self.organizer.customers.all()
+            self.fields['customer'].widget = Select2(
+                attrs={
+                    'data-model-select2': 'generic',
+                    'data-select2-url': reverse('control:organizer.customers.select2', kwargs={
+                        'organizer': self.organizer.slug,
+                    }),
+                    'data-placeholder': _('Customer')
```
raphaelm: We no longer set placeholders that repeat the label (see #5135)

```suggestion
```

src/pretix/control/forms/organizer.py
```diff
@@ -786,6 +804,22 @@ def __init__(self, *args, **kwargs):
...
 
+        if organizer.settings.customer_accounts:
+            self.fields['customer'].queryset = organizer.customers.all()
+            self.fields['customer'].widget = Select2(
+                attrs={
+                    'data-model-select2': 'generic',
+                    'data-select2-url': reverse('control:organizer.customers.select2', kwargs={
+                        'organizer': organizer.slug,
+                    }),
+                    'data-placeholder': _('Customer')
```
raphaelm:
```suggestion
```
see above

src/pretix/control/templates/pretixcontrol/giftcards/checkout.html
```diff
@@ -1,7 +1,35 @@
...
 {% load bootstrap3 %}
+{% load money %}
 {% load rich_text %}
 
 {{ request.event.settings.payment_giftcard_public_description|rich_text }}
-
+{% if customer_gift_cards %}
+    <p><strong>
+        <span class="sr-only">{% trans "Information" %}</span>
+        {% trans "You have gift cards available:" %}
```
raphaelm:
```suggestion
        {% trans "The following gift cards are available in your customer account:" %}
```

src/pretix/control/views/orders.py
```diff
@@ -1234,6 +1234,7 @@ def perform_refund(self, comment, full_refund, payments):
                 giftcard = self.request.organizer.issued_gift_cards.create(
                     expires=expires,
                     currency=self.request.event.currency,
+                    customer=order.customer if order.customer else None,
```
```suggestion
                    customer=order.customer,
```

src/pretix/presale/templates/pretixpresale/event/checkout_payment.html
```diff
@@ -9,6 +9,34 @@
 {% endblock %}
 {% block inner %}
     <h3 class="sr-only">{% trans "Payment" %}</h3>
+    {% if customer_gift_cards %}
+        <p><strong>
+            <span class="sr-only">Info:</span>
+            {% trans "You have gift cards available:" %}
```
raphaelm: For consistency with above
```suggestion
            <span class="sr-only">{% trans "Information" %}</span>
            {% trans "The following gift cards are available in your customer account:" %}
```

src/pretix/presale/templates/pretixpresale/organizers/customer_giftcards.html
```diff
@@ -0,0 +1,79 @@
...
+                    <article class="row">
+                        <div class="col-xs-6">
+                            <h4>
+                                {% icon "gift" %} {{ gc }}
+                            </h4>
+                            {% if gc.issuance %}
+                            <p class="text-muted">
+                                {% icon "calendar" %}
+                                {% blocktrans trimmed %}
+                                    Issued on {{ gc.issuance }}
```
raphaelm: Doesn't work for me, I'd recommend:

```suggestion
                                {% blocktrans trimmed with date=gc.issuance|date:"SHORT_DATE_FORMAT" %}
                                    Issued on {{ date }}
```
Also easier to understand for translators, plus explicit formatting

src/pretix/presale/templates/pretixpresale/organizers/customer_giftcards.html
```diff
@@ -0,0 +1,79 @@
...
+                                    {% trans "TEST MODE" %}
+                                {% endtextbubble %}
+                                </small>
+                            </p>
+                            {% endif %}
+                        </div>
+                        <div class="col-xs-6 text-right">
+                            {% blocktrans trimmed with value=gc.value|money:gc.currency %}
+                                Remaining value: <p class="text-right">{{ value }}</p>
+                            {% endblocktrans %}
```
raphaelm, line 68: I think this is a little too specific HTML to have in a translation. If we want this in separate rows, we can go for

```suggestion
                            {% blocktrans trimmed with value=gc.value|money:gc.currency %}
                                Remaining value:
                            {% endblocktrans %}
                            <p class="text-right">{{ value }}</p>
```

Otherwise, translators might end up scrambling the layout

src/pretix/presale/templates/pretixpresale/organizers/customer_giftcards.html
```diff
@@ -0,0 +1,79 @@
...
+                        </div>
+                    </article>
+                    {% endfor %}
+                </div>
+            {% else %}
+                <p class="text-center">{% trans "You don’t have any gift cards in your account currently. So far, only gift cards resulting from refunds show up here, any purchased gift cards show up under the orders tab." %}</p>
```
raphaelm:
```suggestion
                <p class="text-center">
                    {% trans "You don’t have any gift cards in your account currently." %}
                    {% trans "Currently, only gift cards resulting from refunds show up here, any purchased gift cards show up under the orders tab." %}
                </p>
```
makes it easier to change the second part in the future ;)

src/pretix/presale/checkoutflow.py
```diff
@@ -1400,6 +1400,10 @@ def get_context_data(self, **kwargs):
             ctx['selected'] = self.request.POST['payment']
         elif self.single_use_payment:
             ctx['selected'] = self.single_use_payment['provider']
+        elif ("giftcard" in self.request.event.get_payment_providers()
+              and self.cart_customer
+              and self.cart_customer.usable_gift_cards()):
+            ctx['selected'] = 'giftcard'
```
raphaelm: The logic is still not perfect. Situation:

- I'm logged in
- I have tickets for 17€ in my cart
- I have one gift card with 14€ balance in my account

Now, gift-card is pre-selected as a payment method (ok) and I click "Apply". The gift card is applied, and them … I'm again showing the form with the gift card pre-selected and offering me the same card (hmm).

image

Ideally, we would need to exclude cards already used in the current cart from `usable_gift_cards()`?

If that seems to complex, we should just remove this block and never pre-select the payment method. It's highlighted enough with the "1 available".

src/pretix/base/payment.py
```diff
@@ -1373,6 +1375,27 @@ class GiftCardPayment(BasePaymentProvider):
     payment_form_class = GiftCardPaymentForm
     payment_form_template_name = 'pretixcontrol/giftcards/checkout.html'
 
+    @cached_property
+    def customer_gift_cards(self):
+        if not self.request:
+            return None
+        cs = cart_session(self.request)
+        customer = getattr(self.request, "customer", None)
```
raphaelm, line 1387: Weird, during testing, I was (logged in, no cart) still able to trigger a state where this didn't show up:

image

Now I wanted to debug the issue a bit and it's gone… Works now every time:

image

Not sure what to do about it, but given that this seems to be risky logic (because we shouldn't be able to access self.request here, this is all a big hack), do you think we should also invest the time in adding tests for the frontend (both checkout + paying an order)?

raphaelm (changes requested):

src/pretix/base/payment.py
ser8phin: Weird caching issue (problem exists but hidden because of browser cache maybe)?

> given that this seems to be risky logic (because we shouldn't be able to access self.request here, this is all a big hack), do you think we should also invest the time in adding tests for the frontend (both checkout + paying an order)?

We better do, yes

src/pretix/presale/checkoutflow.py
ser8phin: It would still wrongly show the "1 available" even if we do not preselect, so we cannot get around excluding applied cards in some way

src/pretix/base/payment.py
ser8phin: Tests added :white_check_mark:

src/pretix/presale/checkoutflow.py
ser8phin: resolved: preselection removed, used cards get excluded from available-count and list

src/pretix/base/payment.py
```diff
@@ -1373,10 +1375,36 @@ class GiftCardPayment(BasePaymentProvider):
...
+        form = self.payment_form(request)
+        template = get_template(self.payment_form_template_name)
+        ctx = {'request': request, 'form': form, 'customer_gift_cards': form.customer_gift_cards, }
+        return template.render(ctx)
+
     def payment_form(self, request: HttpRequest) -> Form:
         # Unfortunately, in payment_form we do not know if we're in checkout
         # or in an existing order. But we need to do the validation logic in the
         # form to get the error messages in the right places for accessbility :-(
+        # used_cards = []
```
ser8phin:
```suggestion

```
leftover from debugging?

src/pretix/base/migrations/0289_giftcard_customer.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 4.2.19 on 2025-05-19 11:06
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('pretixbase', '0288_invoice_transmission'),
```
raphaelm: just as a reminder, before merge this PR needs to be rebased and the migration recreated against current master

PR #5130, opened 2025-05-20
[A11y] move timezone info inline instead of tooltip

wiffbi:
TODO: subevents


raphaelm:
> I put the new setting at the timezone-settings tab. Maybe it just feels right there, because currently we only use it for additional timezone-info. But maybe below location is a better place?

I would have also put it with the location

> Interestingly, dates in an event-series do not have timezone-info, but a location … which kind of goes hand in hand. Hm.

Yeah, we'll need to live with that. We're not going to do multiple timezones within an event series, that would make many things really crazy (such as what time zone to use for the payment deadline :D)

raphaelm (approved):

PR #5135, opened 2025-05-21
FormFieldRenderer: do not set placeholder automatically

wiffbi:
This PR removes the automatism to set the placeholder to the same value as label if not otherwise defined. This makes the forms IMHO way cleaner.

Before:
Bildschirmfoto 2025-05-21 um 13 24 44

After:
Bildschirmfoto 2025-05-21 um 13 24 11

### DIFF
--- a/src/pretix/control/forms/filter.py
+++ b/src/pretix/control/forms/filter.py
@@ -2444,7 +2444,7 @@ class CheckinFilterForm(FilterForm):
                     'event': self.event.slug,
                     'organizer': self.event.organizer.slug,
                 }),
-                'data-placeholder': _('Check-in list'),
+                'data-placeholder': _('All Check-in lists'),
             }
         )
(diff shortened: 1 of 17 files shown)


wiffbi:
This needs more work especially on name and phone fields.

src/pretix/control/forms/filter.py
```diff
@@ -2448,7 +2444,7 @@ def __init__(self, *args, **kwargs):
                     'event': self.event.slug,
                     'organizer': self.event.organizer.slug,
                 }),
-                'data-placeholder': _('Check-in list'),
+                'data-placeholder': _('All Check-in lists'),
```
raphaelm:
```suggestion
                'data-placeholder': _('All check-in lists'),
```

raphaelm (approved):

PR #5136, opened 2025-05-21
Widget: add versioning support and add v2 with improved a11y-support

wiffbi:


src/pretix/presale/views/widget.py
```diff
@@ -109,19 +109,24 @@ def widget_css_etag(request, **kwargs):
...
 @gzip_page
 @condition(etag_func=widget_css_etag)
 @cache_page(60)
-def widget_css(request, **kwargs):
+def widget_css(request, version, **kwargs):
     o = getattr(request, 'event', request.organizer)
 
+    # TODO: handle version deprecation
+    css_path = 'pretixpresale/scss/widget.v{}.scss'.format(version)
+    if not finders.find(css_path):
```
raphaelm, line 125: I think if the system just knows which version is the current one, we can avoid this code path. We'll hit it *a lot*.

src/pretix/presale/urls.py
```diff
@@ -172,7 +172,7 @@
 
     re_path(r'^widget/product_list$', pretix.presale.views.widget.WidgetAPIProductList.as_view(),
             name='event.widget.productlist'),
-    re_path(r'^widget/v1.css$', pretix.presale.views.widget.widget_css, name='event.widget.css'),
+    re_path(r'^widget/v(?P<version>[1,2]+).css$', pretix.presale.views.widget.widget_css, name='event.widget.css'),
```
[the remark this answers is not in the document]
wiffbi: Sorry, changed it to a `path(<int:version>)` which does the type-cast to int for me. See latest version. If 404 are now handled inside the view-function if version>version_max – less code changes, when increasing versions.

src/pretix/presale/views/widget.py
```diff
@@ -81,15 +83,22 @@
 # we never change static source without restart, so we can cache this thread-wise
 _source_cache_key = None
 
+version_min = 1
+version_max = 2
+version_default = 1  # used for output in widget-embed-code
```
raphaelm, line 88:
```suggestion
version_default = 2  # used for output in widget-embed-code
```

raphaelm (approved):

PR #5139, opened 2025-05-21
Close dialog onclick on backdrop

wiffbi:
TODO: Test how this interferes with cancel-suppressing the closing the dialog. Currently we do not suppress cancel, but would be good to know.

### DIFF
--- a/src/pretix/static/pretixpresale/js/ui/main.js
+++ b/src/pretix/static/pretixpresale/js/ui/main.js
@@ -390,6 +390,13 @@ $(function () {
         sessionStorage.removeItem('scrollpos');
     }
 
+    $("dialog").on("mousedown", function (e) {
+        if (e.target == this) {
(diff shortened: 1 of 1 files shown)


wiffbi:
This intereferes with with the cancel-event of a dialog as in no cancel-event is triggered. So that way, one cannot abort canceling the dialog with Escape. We currently do not have any cancel-suppression, but maybe we do need it at some point? But also IMHO clicking in the background is a different intention (ignore) vs. hitting e.g. Escape (abort/cancel the operation). Furthermore triggering a cancel through a click on the backdrop might be problematic with accessibility as it might be surprising.

The only dialog I found clicking in the backdrop would be useful (and kind of learned behaviour), is the lightbox-dialog. So maybe we could add it there? It would then behave differently than all other dialgos we have, but I think that’s okay.

PR #5140, opened 2025-05-21
Allow to add declaration of accessibility

raphaelm:


src/pretix/control/templates/pretixcontrol/organizers/edit.html
```diff
@@ -200,6 +200,19 @@ <h1>
...
+                        <div class="alert alert-legal">
+                            {% blocktrans trimmed with a_href="https://docs.pretix.eu/trust/accessibility/" %}
+                                Some jurisdictions, including the European Union, require you to publish information
+                                about the accessibility of your ticket shop. You can find a template in
+                                <a href="https://docs.pretix.eu/trust/accessibility/" target="_blank">our documentation</a>.
+                            {% endblocktrans %}
+                        </div>
+                        {% bootstrap_field sform.accessibility_url layout="control" %}
+                        {% bootstrap_field sform.accessibility_title layout="control" %}
+                        {% bootstrap_field sform.accessibility_text layout="control" %}
```
wiffbi, line 214: I’d remove the help-text and move it before the input, IMHO having it below the input does not help here.
```suggestion
                        <div class="row">
                            <div class="col-md-9 col-md-offset-3">
                                <p class="text-muted">{% trans "Instead of an URL, you can also configure a text that will be shown within pretix. This will be ignored if a URL is configured." %}</p>
                            </div>
                        </div>
                        {% bootstrap_field sform.accessibility_text layout="control" %}
```

src/pretix/control/templates/pretixcontrol/organizers/edit.html
```diff
@@ -200,6 +200,19 @@ <h1>
...
+                        <legend>{% trans "Accessibility" %}</legend>
+                        <div class="alert alert-legal">
+                            {% blocktrans trimmed with a_href="https://docs.pretix.eu/trust/accessibility/" %}
+                                Some jurisdictions, including the European Union, require you to publish information
+                                about the accessibility of your ticket shop. You can find a template in
+                                <a href="https://docs.pretix.eu/trust/accessibility/" target="_blank">our documentation</a>.
+                            {% endblocktrans %}
+                        </div>
+                        {% bootstrap_field sform.accessibility_url layout="control" %}
+                        {% bootstrap_field sform.accessibility_title layout="control" %}
```
wiffbi, line 213: I’d move the title to the top, not sure why. It feels that it makes it more clear, that the title is also the link’s label.
```suggestion
                        {% bootstrap_field sform.accessibility_title layout="control" %}
                        {% bootstrap_field sform.accessibility_url layout="control" %}
```

src/pretix/presale/templates/pretixpresale/event/base.html
```diff
@@ -224,6 +224,15 @@ <h1>
     {% if request.event.settings.privacy_url %}
         <li><a href="{% safelink request.event.settings.privacy_url %}" target="_blank" rel="noopener">{% trans "Privacy policy" %}</a></li>
     {% endif %}
+    {% if request.event.settings.accessibility_url %}
+        <li><a href="{% safelink request.event.settings.accessibility_url %}" target="_blank" rel="noopener">
+            {{ request.event.settings.accessibility_title }}
```
wiffbi, line 229: What if I leave the title blank? Should we fallback to a default value or shouldn’t we link to it? IMHO we should not link to it, if there is not title.

src/pretix/presale/templates/pretixpresale/event/base.html
```diff
@@ -224,6 +224,15 @@ <h1>
...
+        </a></li>
+    {% elif request.event.settings.accessibility_text %}
+        <li><a href="{% abseventurl request.organizer "presale:organizer.accessibility" %}" target="_blank" rel="noopener">
+            {{ request.event.settings.accessibility_title }}
```
wiffbi, line 233: See above

wiffbi (changes requested):
Looks good. If I remove the default title, then the link is empty and no headline. Should we use a default fallback or not link and 404 and output a warning in the settings-form?

PR #5144, opened 2025-05-22
Fix nothing-to-add condition

luelista:


src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -482,7 +482,7 @@ $(function () {
         if (
             (this.classList.contains("has-seating") && this.querySelector("pretix-seating-checkout-button button")) ||
             this.querySelector("input[type=checkbox]:checked") || 
-            [...this.querySelectorAll(".input-item-count[type=text]")].some(input => input.value && input.value !== "0") // TODO: seating hat noch einen seating-dummy-item-count, das ist Mist!
+            [...this.querySelectorAll(".input-item-count[type=text], .input-item-count[type=number]")].some(input => input.value && input.value !== "0") // TODO: seating hat noch einen seating-dummy-item-count, das ist Mist!
```
wiffbi, line 485: Seating adds a hidden input, which we should ignore, so maybe better change the selector to:

```suggestion
            [...this.querySelectorAll(".input-item-count:not([type=hidden])")].some(input => input.value && input.value !== "0") // TODO: seating adds a hidden seating-dummy-item-count, which is not useful and should at some point be removed
```

wiffbi (approved):
I’d change the selector to better reflect its intention – ignoring hidden item-counts

PR #5148, opened 2025-05-22
Dialog for cart renewal, async task without page refresh

luelista:
- Adds a warning dialog when the cart nears expiry, allowing to extend the cart reservation time
- In-place async tasks (without full page reload in response - use `data-asynctask-no-redirect` attribute and `pretix:async-task-success` event)
- Use in-place async task for cart extend form
- Change #loadingmodal to use new accessible dialog style

replaces #5117 and #5118

Cart expired dialog:
image

New AsyncTask dialog:
(frontend)
image

(backend)
image


src/pretix/static/pretixbase/scss/_dialogs.scss
```diff
@@ -0,0 +1,172 @@
...
+  dialog.modal-card:has(.modal-card-icon) .modal-card-inner {
+    flex-direction: row;
+    .modal-card-content {
+      padding: 2em;
+    }
+    .modal-card-icon {
+      font-size: 4em;
+      padding: 6px 16px;
+    }
+  }
```
wiffbi, line 76: .modal-card-content should always have a bigger padding on bigger screens (at least that is how it was in _theme.scss), not just when having an icon.

```suggestion
  dialog.modal-card:has(.modal-card-icon) .modal-card-inner {
    flex-direction: row;
  }
  .modal-card-content {
    padding: 2em;
  }
  .modal-card-icon {
    font-size: 4em;
    padding: 6px 16px;
  }
```

src/pretix/static/pretixbase/scss/_dialogs.scss
```diff
@@ -0,0 +1,172 @@
...
+  0% { transform: skewX(0deg); }
+  20% { transform: skewX(-5deg); }
+  40% { transform: skewX(5deg); }
+  60% { transform: skewX(-5deg); }
+  80% { transform: skewX(5deg); }
+  100% { transform: skewX(0deg); }
+}
+
+
+
```
wiffbi, line 95: Do we want to keep the animation in there? AFAIK we do not use it anywhere as we do allow to cancel any dialogs. Or maybe we should add it to the loading-modal if someone tries to cancel with escape?

src/pretix/static/pretixbase/js/asynctask.js
```diff
@@ -275,21 +247,65 @@ $(function () {
             }, 10);
         }
     }, false);
+
+    $("#ajaxerr").on("click", ".ajaxerr-close", ajaxErrDialog.hide);
+    $("#loadingmodal").on("cancel", function() {
+        return false;
+    });
```
wiffbi, line 254: This only works in Firefox, but causes a problem anyways: when I hit the Escape key, while the Ajax-request is still running, Firefox aborts any Ajax-requests running, but now does not close the modal either. Luckily the Ajax-request is still executing the success-callback when canceled through Firefox, but with different than expected results. So on success, we should always remove the loading-dialog.

In Chrome, blocking the Escape-key only works once. Hitting Escape a second time closes the dialog. It does not abort the Ajax-request though. So we need to check the `returnValue` on `dialog.close` whether the dialog was canceled and re-open the dialog if the Ajax-request is still loading.

src/pretix/static/pretixbase/js/asynctask.js
```diff
@@ -5,91 +5,105 @@ var async_task_check_url = null;
...
     }
-    async_task_timeout = window.setTimeout(async_task_check, 250);
 
+    if (typeof data.percentage === "number") {
+        waitingDialog.setProgress(data.percentage);
+    }
+    if (typeof data.steps === "object" && Array.isArray(data.steps)) {
+        waitingDialog.setSteps(data.steps);
+    }
+    async_task_schedule_check(this, 250);
```
wiffbi, line 68: I know originally it was checked every 250ms. But isn’t that a bit often? 4 times per second? Wouldn’t 1000ms be enough, maybe 500ms?

wiffbi (commented):
I just reviewd the loadingModal part, haven’t come to check different instances of of asynctask. Sorry!

wiffbi (approved):
Looks good to me. It As noted when I hit escape quickly while the loadingmodal is open, Firefox aborts any connection I get locked in the loading-modal. Only way out is to reload the whole page – which we would probably need to do anyways if someone aborted the request, to make sure we show the latest state on the page. So IMHO not a show-stopper for this PR.

PR #5149, opened 2025-05-22
[A11y] fix input border & focus style

wiffbi:
Changing the `$btn-default-border:` and `$input-border` in pretixbase instead of presale main.scss fixes contrast issues in all of pretix. This might be considered a breaking change with the widget – but we have already changed lots of colors in the past, which also came through to the widget. So I would say, it should be fine to keep using `$btn-default-border:` and `$input-border`, which then would tickle down into widget.v1.scss. Or should I hard-code the old values in widget.v1.scss?


src/pretix/static/pretixbase/scss/_bootstrap_vars.scss
```diff
@@ -72,7 +73,7 @@ $panel-default-heading-bg: #e5e5e5 !default;
 
 $link-hover-color: var(--pretix-brand-primary-darken-15);
 
-$btn-default-border: #CCCCCC;
+$btn-default-border: #949494;;
```
wiffbi, line 76:
```suggestion
$btn-default-border: #949494;
```

raphaelm (approved):

PR #5194, opened 2025-06-02
 [A11y] Do not set autocomplete section if autocomplete is off 

raphaelm:


src/pretix/base/forms/questions.py
```diff
@@ -182,7 +182,9 @@ def render(self, name: str, value, attrs=None, renderer=None) -> str:
                     if self.field.required:
                         these_attrs['required'] = 'required'
                     these_attrs.pop('data-no-required-attr', None)
-                these_attrs['autocomplete'] = (self.attrs.get('autocomplete', '') + ' ' + self.autofill_map.get(self.scheme['fields'][i][0], 'off')).strip()
+
+                autofill_prop = self.autofill_map.get(self.scheme['fields'][i][0], 'off')
+                these_attrs['autocomplete'] = (self.attrs.get('autocomplete', '') + ' ' + autofill_prop).strip() if autofill_prop != "off" else "off"
```
wiffbi: We currently do not do this, but what if at some point we set the autocomplete-attribute of the `NamePartsWidget` somewhere to `'off'`? Shouldn’t `self.attrs.get('autocomplete', 'off') == 'off'` override the autofill_map? So instead maybe do something like:

```suggestion
                autofill_prop = set(self.attrs.get('autocomplete', '').split() + self.autofill_map.get(self.scheme['fields'][i][0], '').split())
                these_attrs['autocomplete'] = autofill_prop if autofill_prop and 'off' not in autofill_prop else 'off'
```

wiffbi (commented):
Looks good. Just a small question regarding autocomplete on `NamePartsWidget`.

PR #5196, opened 2025-06-02
[A11y] fix organizer event view navigation missing aria-current

wiffbi:


src/pretix/presale/templates/pretixpresale/fragment_calendar_nav.html
```diff
@@ -2,22 +2,21 @@
 {% load eventurl %}
 {% load urlreplace %}
 
-<div class="btn-group" role="group">
-    <a href="?{% url_replace request "style" "list" "date" "" %}" type="button" 
-            class="btn btn-default{% if style == "list" %} active{% endif %}">
+<div class="btn-group" role="navigation" aria-label="{% trans "Event overview" %}">
```
wiffbi: Maybe we need to specifiy this a bit more. What do you think?
```suggestion
<div class="btn-group" role="navigation" aria-label="{% trans "Event overview by month, week, etc." %}">
```

PR #5202, opened 2025-06-02
[A11y] add inline alertidalog to confirm successful cart renewal

wiffbi:
When the cart reservation is renewed, we need to give feedback. This was originally done through a `role=alert` on `#cart-deadline`, but that did not work as intended – when the „renew reservation“-button was removed, focus moved to the focusable parent element and its aria-label and -description was read out. The confirmation message – although being `role=alert` was not read aloud. So then we added the renewal-confirmation message to the aria-describedby on the parent element, that received focus, when the „renew reservation“-button was removed. In my tests with VoiceOver on Safari, the message was read as the description, but after the label „Your cart“ – according to external a11y-tests this was not sufficient.

This PR adds an inline alertdialog with a confirmation button „reservation renewed“. That dialog/button gets focus on successfully renewing the reservation. Once it is confirmed (clicked) or loses focus, the dialog is closed.

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
+++ b/src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
@@ -513,10 +513,12 @@
         </form>
         <dialog role="alertdialog" id="cart-extend-feedback-dialog" aria-labelledby="cart-extend-feedback">
             <form method="dialog">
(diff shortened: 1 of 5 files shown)


raphaelm (approved):

wiffbi:
I think #5206 is the better solution.

PR #5206, opened 2025-06-03
[A11y] improve cart renew confirmation

wiffbi:
This PR simplifies the cart renew confirmation suggested in #5202 . It does so by registering one-time event listeners for the asny-renewal-task depending on which button was clicked. If the renewal-dialog was clicked, then we confirm in a dialog as well. If the inline-cart-form-renewal-button was clicked, we confirm inline in the cart-form. ~The inline-confirmation-button currently does nothing except being focusable so the success-message can be read. We will have to see, whether we need an inline alertdialog for this as suggested in #5202~ I added the inline-dialog so the confirmation-button actually has interactive meaning to close the dialog and disappear.

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
+++ b/src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
@@ -496,8 +496,7 @@
         <form class="text-muted" id="cart-extend-form" data-asynctask data-asynctask-no-redirect
                 method="post" action="{% eventurl request.event "presale:event.cart.extend" cart_namespace=cart_namespace %}">
             {% csrf_token %}
-            <p>
-                <span id="cart-deadline" data-expires="{{ cart.first_expiry|date:"Y-m-d H:i:sO" }}" data-max-expiry-extend="{{ cart.max_expiry_extend|date:"Y-m-d H:i:sO" }}">
+            <p id="cart-deadline" data-expires="{{ cart.first_expiry|date:"Y-m-d H:i:sO" }}" data-max-expiry-extend="{{ cart.max_expiry_extend|date:"Y-m-d H:i:sO" }}">
                 {% if cart.minutes_left > 0 or cart.seconds_left > 0 %}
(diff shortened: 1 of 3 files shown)


src/pretix/static/pretixpresale/js/ui/cart.js
```diff
@@ -74,13 +72,12 @@ var cart = {
...
             );
 
-            cart._renewed_message = "";
             cart._deadline_timeout = window.setTimeout(cart.draw_deadline, 500);
         }
         var already_expired = diff_total_seconds <= 0;
         var can_extend_cart = diff_minutes < 3 && (already_expired || cart._deadline < cart._max_extend);
         $("#cart-extend-button").toggle(can_extend_cart);
-        if (can_extend_cart && diff_total_seconds < 45) {
+        if (can_extend_cart && diff_total_seconds < 245) {
```
wiffbi:
```suggestion
        if (can_extend_cart && diff_total_seconds < 45) {
```

src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
```diff
@@ -495,18 +496,23 @@
...
                 {% endif %}
-            </p>
-            <p>
-                <button class="btn btn-default" type="submit" id="cart-extend-button" aria-describedby="cart-deadline">
-                    <i class="fa fa-refresh" aria-hidden="true"></i> {% trans "Renew reservation" %}
+                </span>
+                <button class="btn btn-default btn-xs" type="submit" id="cart-extend-button" aria-describedby="cart-deadline">
+                    {% icon "refresh" %} {% trans "Renew reservation" %}
+                </button>
+                <button class="btn btn-success btn-xs" type="button" id="cart-extend-confirmation-button" aria-describedby="cart-deadline">
```
raphaelm: Should we hide this using a HTML attribute or CSS class instead of only via JS? This button should not be visible to a no-javascript user.

PR #5207, opened 2025-06-03
Fix sneak-peek missing inert-attribute

wiffbi:
Since adding the renew-button to the sneak-peek we had an interactive element inside an aria-hidden element. This is not allowed, so this PR adds the inert-attribute to the container, so the button is not focusable any more.


raphaelm (approved):
Nice, TIL about "inert"

PR #5232, opened 2025-06-10
API: Add items__in filter for quotas (Z#23195926)

raphaelm:


doc/api/resources/quotas.rst
```diff
@@ -82,6 +82,7 @@ Endpoints
                            Default: ``position``
    :query integer subevent: Only return quotas of the sub-event with the given ID.
    :query integer subevent__in: Only return quotas of sub-events with one the given IDs (comma-separated).
+   :query integer items__in: Only return quotas that include a product with one the given IDs (comma-separated).
```
wiffbi:
```suggestion
   :query integer subevent__in: Only return quotas of sub-events with one of the given IDs (comma-separated).
   :query integer items__in: Only return quotas that include a product with one of the given IDs (comma-separated).
```

wiffbi (approved):
Looks good, just a tiny typo in the docs.

Should we change the `subevent__in`-filter from `Meta` to a `NumberInFilter` as well?

PR #5238, opened 2025-06-11
Subevent bulk editing: Warn about deleted quotas

raphaelm:


wiffbi (commented):
Should that dialog be shown even if all quotas can be edited as they are the same? Ie. like the message above the edit-block `{% if sampled_quotas|default_if_none:"NONE" == "NONE" %}`. Maybe we do not need that dialog (might be clicked away to quickly) and just need to remove the warning from above and actually replace it with the content from the dialog and place that inside the `.field-content`? Like so:

Bildschirmfoto 2025-06-12 um 13 07 32

raphaelm:
> Should that dialog be shown even if all quotas can be edited as they are the same? Ie. like the message above the edit-block `{% if sampled_quotas|default_if_none:"NONE" == "NONE" %}`. 

Oh yes of course, I somehow lost that condition in the final refactoring. Added back in.

> Maybe we do not need that dialog (might be clicked away to quickly) and just need to remove the warning from above and actually replace it with the content from the dialog and place that inside the `.field-content`? Like so:

I do believe we cannot do enough here. We've now had three incidents where customers caused multi-hour work on their or our side by not understanding the implications. I implemented your changed proposal in addition.

wiffbi (approved):
Perfect, good idea to warn in both ways!

PR #5239, opened 2025-06-12
Questions: Express percentage of tickets (Z#23196542)

raphaelm:


wiffbi (approved):
Looks good. I added the missing td in the tfoot as I could not add a comment onto not changed code.

PR #5244, opened 2025-06-13
Add data-article-id to reference cart-item in product-list

wiffbi:
This is a helper-id on the cart-row object to enable tracking of cart-interactions. Previously only add-operations were possible to track as remove-operations only had the op.id instead of the op.item.id

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
+++ b/src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
@@ -25,7 +25,7 @@
     </div>
     <div role="rowgroup" class="firstchild-in-panel">
     {% for line in cart.positions %}
-        <div role="row" class="row cart-row {% if hide_prices %}hide-prices{% endif %} {% if download %}has-downloads{% endif %}{% if editable %}editable{% endif %}">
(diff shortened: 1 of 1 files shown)


src/pretix/presale/templates/pretixpresale/event/fragment_cart.html
```diff
@@ -25,7 +25,7 @@
     </div>
     <div role="rowgroup" class="firstchild-in-panel">
     {% for line in cart.positions %}
-        <div role="row" class="row cart-row {% if hide_prices %}hide-prices{% endif %} {% if download %}has-downloads{% endif %}{% if editable %}editable{% endif %}">
+        <div role="row" class="row cart-row {% if hide_prices %}hide-prices{% endif %} {% if download %}has-downloads{% endif %}{% if editable %}editable{% endif %}" data-article-id="item-{{ line.item.id }}{% if line.variation %}-{{ line.variation.id }}{% endif %}">
```
raphaelm, line 28: Do we not need something similar for fees?

wiffbi: No, currently not, as one cannot add/remove fees manually from the cart – the ID is only needed to get the necessary information to create the add_to_cart and remove_from_cart tracking calls, when the user clicks the +/-/trash-buttons.

Also generally fees are not tracked at all currently as the tracking of cart interactions are based on user-interaction.

raphaelm: sounds good

PR #5248, opened 2025-06-16
Email: Remove more characters from sender name (Z#23197264)

raphaelm:


src/pretix/base/services/mail.py
```diff
@@ -96,12 +96,19 @@ class SendMailException(Exception):
...
+
     # Emails with @ in their sender name are rejected by some mailservers (e.g. Microsoft) because it looks like
     # a phishing attempt.
     sender_name = sender_name.replace("@", " ")
     # Emails with : in their sender name are treated by Microsoft like emails with no From header at all, leading
     # to a higher spam likelihood.
     sender_name = sender_name.replace(":", " ")
+    # Emails with , in their sender name look like multiple senders
+    sender_name = sender_name.replace(",", "")
+    # Emails with " in their sender name could be escaped, but somehow create issues in reailty
```
wiffbi:
```suggestion
    # Emails with " in their sender name could be escaped, but somehow create issues in reality
```

PR #5259, opened 2025-06-23
Questions: Validate type changes (Z#23197118)

raphaelm:


src/pretix/base/forms/questions.py
```diff
@@ -896,21 +896,35 @@ def __init__(self, *args, **kwargs):
                             'Please enter a date no later than {max}.',
                             max=date_format(q.valid_date_max, "SHORT_DATE_FORMAT"),
                         )
+                if initial and initial.anser:
```
wiffbi:
```suggestion
                if initial and initial.answer:
```

src/pretix/base/forms/questions.py
```diff
@@ -896,21 +896,35 @@ def __init__(self, *args, **kwargs):
...
-                    initial=dateutil.parser.parse(initial.answer).date() if initial and initial.answer else None,
+                    initial=_initial,
                     widget=DatePickerWidget(attrs),
                 )
                 if q.valid_date_min:
                     field.validators.append(MinDateValidator(q.valid_date_min))
                 if q.valid_date_max:
                     field.validators.append(MaxDateValidator(q.valid_date_max))
             elif q.type == Question.TYPE_TIME:
+                if initial and initial.anser:
```
wiffbi:
```suggestion
                if initial and initial.answer:
```

src/pretix/base/forms/questions.py
```diff
@@ -931,10 +945,19 @@ def __init__(self, *args, **kwargs):
                             'Please enter a date and time no later than {max}.',
                             max=date_format(q.valid_datetime_max, "SHORT_DATETIME_FORMAT"),
                         )
+
+                if initial and initial.anser:
```
wiffbi:
```suggestion
                if initial and initial.answer:
```

src/pretix/base/models/items.py
```diff
@@ -1925,6 +1925,25 @@ def clean(self):
...
+    def clean_type_change(self, old_type, new_type):
+        if old_type == new_type:
+            return True
+        if not self.answers.exists():
+            return True
+        if new_type == self.TYPE_TEXT and old_type != self.TYPE_FILE:
+            # All types can be converted to text except file
+            return True
+        if new_type == self.TYPE_STRING and old_type not in (self.TYPE_TEXT, self.TYPE_FILE):
+            # All types can be converted to string except for text except file
```
wiffbi:
```suggestion
            # All types can be converted to string except text or file
```

src/pretix/control/forms/item.py
```diff
@@ -201,6 +201,12 @@ def clean_show_during_checkin(self):
 
         return val
 
+    def clean_type(self):
+        val = self.cleaned_data.get('type')
+        if self.instance:
+            self.instance.clean_type_change(self.instance.type, val)
```
wiffbi, line 207: I re-read the docs, but could not find why there would not be a `self.instance` at this point. So:
```suggestion
        self.instance.clean_type_change(self.instance.type, val)
```

raphaelm: Generally, Django ModelForms are not guaranteed to have an instance available during clean step. For example, a CreateForm usually would have self.instance = None at this point.

In this case, the form does currently always have an instance due to
```
    def get_form_kwargs(self):
        kwargs = super().get_form_kwargs()
        kwargs['instance'] = Question(event=self.request.event)
        return kwargs
```
But that's an implementation detail of the view (the view could also create the event relation when calling save()) and I don't think the form should rely on it.

PR #5291, opened 2025-07-03
Voucher form: Field-specific error messages (Z#23199648)

raphaelm:


src/pretix/control/forms/vouchers.py
```diff
@@ -224,10 +227,17 @@ def clean(self):
...
-            data, self.instance.event
-        )
-        Voucher.clean_max_usages(data, self.instance.redeemed)
+
+        try:
+            Voucher.clean_subevent(
+                data, self.instance.event
+            )
+        except ValidationError as e:
+            raise ValidationError({"quota": e.message})
```
wiffbi: `clean_subevent` checks if block_quota is set, but the error still says 'If you want this voucher to block quota, you need to select a specific date.' – so IMHO we should assign the error to `subevent`.

```suggestion
            raise ValidationError({"subevent": e.message})
```

Also, if you still want to assign to quota, I guess you meant `block_quota` as quota is not a field on the form and fails assigning the error.

wiffbi (changes requested):
Looks good, just a typo in the quota => block_quota-error – which IMHO we should anyways assign to subevent – see comment.

PR #5301, opened 2025-07-10
FEAT: Support for binary ticket secrets and Aztec barcodes in PDFs

TheEnbyperor:
First of all; appologies for how large this PR is for what looks like a fairly small change.

The essence of this change is to allow ticket secret generators to return binary, as well at the existing text secrets.
As QR codes don't support the full range of binary data, it also adds support for rendering Aztec barcodes on the PDF to handle such binary data.

The external API won't change, unless there is binary data in use (data outside of the ASCII printable range), then secrets will be handleded with Base64.

If you're wondering "why on *earth* do you need this Q?!" - the answer is here: https://github.com/TheEnbyperor/pretix-uic-barcode


raphaelm:
First of all, thanks for sending the PR :blush:  To be quite honest, I'm currently not convinced we're going to merge it. (Generally, it's always a good idea to talk about features in discussions first before putting in the work, since there might be less-obvious things that make it more complex.)

### Why it's not trivial

Making sure that pretix is still good software in 5-10 years is, above all, a fight against complexity. This adds complexity and honestly, at the moment, I struggle to see the justification for it. Besides this general "complexity needs to be justified" argument, however, there are specific additional "hidden costs" for us to merging this. At least four come to mind immediately:

- All tickets generated by pretix should be readable by pretix as well. This would require us to check (and maintain) that Aztec codes with binary content work fine across all of our platforms, which includes not only our pretixSCAN apps for Android and iOS, but also scenarios where barcode scanners are connected as HID devices (e.g. pretixSCAN Desktop) or serial devices (e.g. pretixSCAN turnstile). Then, we'd need to make sure those clients all adhere to the "base64-if-binary" scheme (that I think is flawed, see below). I'm sure not all of them cope with binary content out of the box.
- All PDF layouts configured in pretix need to render the same when printed through as a badge through our SCAN or POS apps. This would require us to implement (and maintain) Aztec rendering for our Java-based rendering toolchain in libpretixprint as well.
- This would also require changes to pretix-passbook (and it's planned successor that includes Google Wallet)
- Changing the type of a field like OrderPosition.secret would require us to invest significant effort into a deployment for pretix Hosted. Our OrderPosition table has tens of millions of rows and changing this field type would require PostgreSQL to hold a lock on that table while the entire table is rewritten. We'd either need to announce a maintenance downtime for this (which we like to avoid) or invest significant effort into a slow-migration scheme that keeps two different columns and syncs over the data until they are in sync and we can drop the old field.

### Technical notes

> The external API won't change, unless there is binary data in use (data outside of the ASCII printable range), then secrets will be handleded with Base64.

I don't think it's that simple. First of all, for all we know people could currently use barcodes that include printable characters that are not ASCII, and second of all your SecretField.to_internal_value will consider anything decimal that looks like base64. Say I work with secrets of 8 random characters, then `ABCDasdf` would be converted to `\x00\x10\x83j\xc7_` because it's valid bas64. I don't think there's a way that doesn't change the API and is safe (but it would be possible to change the API in a backwards-compatible way).

### The big why

> If you're wondering "why on earth do you need this Q?!" - the answer is here: [TheEnbyperor/pretix-uic-barcode](https://github.com/TheEnbyperor/pretix-uic-barcode)

This doesn't really answer the why – it just transforms it to "why would you want UIC barcodes in your event ticketing" other than "because I can and it's fun". We have enough train nerds on the team that we totally relate to that idea – but given the complexities above, "for fun" will not cut it.

"I want to use pretix to sell train tickets" also probably wouldn't be good enough, since pretix isn't built for or focused in transport ticketing, that would bring a whole lot of other requirements that we do not want to touch (think about capacity management, pricing, etc.).

I think the only business case that would sell me for UIC support in pretix would be if we could offer an *actual* "combination ticket", meaning that the same ticket would be valid at the event as well as for the public transport of the city of the event. However, I *don't* think we'll find sufficient public transport operators willing to hand us the secret keys for generating arbitrary UIC barcodes, and even if they would, it would likely require come with compliance dances or security requirements we wouldn't want to comply with.

### Alternative options

With some additional effort, I think this could be entirely implemented in a plugin.

- Have the ticket secret generator just store base64 content in the existing secret field
- Implement the Aztec rendering in a custom subclass of the PDF layout renderer shipped with your plugin (same for Wallet?) *or* supply the Aztec code through the "dynamical picture" API of the PDF renderer
- Run a patched version of pretixSCAN that detects UIC barcodes and sends them to the server base64-encoded

TheEnbyperor:
Points entirely understood, and very valid.

I will have a poke at the alternatives yous suggested, and see if this achieves what I want.

PR #5304, opened 2025-07-11
Order import: Allow to create multiple multi-ticket orders

raphaelm:
Add-ons are left as a future improvement, let's go step by step ;)

### DIFF
--- a/src/pretix/base/modelimport_orders.py
+++ b/src/pretix/base/modelimport_orders.py
@@ -57,6 +57,7 @@ from pretix.base.signals import order_import_columns
 class EmailColumn(ImportColumn):
     identifier = 'email'
     verbose_name = gettext_lazy('Email address')
+    order_level = True
 
     def clean(self, value, previous_values):
         if value:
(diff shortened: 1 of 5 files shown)


src/pretix/base/modelimport_orders.py
```diff
@@ -67,9 +68,23 @@ def assign(self, value, order, position, invoice_address, **kwargs):
...
 
 
+class GroupingColumn(ImportColumn):
+    identifier = 'grouping'
+    verbose_name = gettext_lazy('Grouping')
+    help_text = gettext_lazy(
+        'Only applicable when "Import mode" is set to "Group multiple lines together...". Lines with the same grouping '
+        'value will be put in the same order, but MUST be consecutive lines of the input file.'
+    )
+    order_level = True
```
luelista, line 78: "Keep empty" seems a bit off here, I'd go for `"--"` or `_("No grouping")`
```suggestion
    order_level = True
    default_label = "--"
```

raphaelm:
```suggestion
    order_level = True
    default_label = "---"
```

PR #5305, opened 2025-07-12
Organizer-level plugins

raphaelm:
- [x] Plugin registration
- [x] Signal handling
- [x] Registry handling
- [x] New handling for is_available
- [x] Defaults for new organizers
  - [ ] create docs PR
- [x] Event-level UI
  - [x] Tests
- [x] Organizer-level UI
  - [x] Tests
- [x] Event-level API
  - [x] Tests
  - [x] Docs
- [x] Organizer-level API
  - [x] Tests
  - [x] Docs
 
Open questions
- [ ] How do we deal with existing organizers when migrating plugins from the old "active for everyone" to the new organizer-level-plugins?
- [x] Should LogEntryTypeRegistry really be an EventPluginRegistry? We also use it on organizer and global level…

### DIFF
--- a/src/pretix/base/migrations/0287_organizer_plugins.py
+++ b/src/pretix/base/migrations/0287_organizer_plugins.py
@@ -6,7 +6,7 @@ from django.db import migrations, models
 class Migration(migrations.Migration):
 
     dependencies = [
-        ("pretixbase", "0284_ordersyncresult_ordersyncqueue"),
+        ("pretixbase", "0286_settingsstore_unique"),
     ]
 
     operations = [


doc/development/api/plugins.rst
```diff
@@ -56,14 +56,28 @@ restricted         boolean (optional)   ``False`` by default, restricts a plugin
                                         for an event by system administrators / superusers.
 experimental       boolean (optional)   ``False`` by default, marks a plugin as an experimental feature in the plugins list.
 compatibility      string               Specifier for compatible pretix versions.
+level              string               System level the plugin can be activated at.
+                                        Set to ``pretix.base.plugins.PLUGIN_LEVEL_EVENT`` for plugins that can be activated
+                                        at event level and then be active for that event only.
+                                        Set to ``pretix.base.plugins.PLUGIN_LEVEL_ORGANIZER`` for plugins that can be
+                                        activated only for the organizer as a whole and are active for any event within
+                                        that organizer.
+                                        Set to ``pretix.base.plugins.PLUGIN_LEVEL_EVENT_ORGANIZER_HYBRID`` for events that
```
felixrindt:
```suggestion
                                        Set to ``pretix.base.plugins.PLUGIN_LEVEL_EVENT_ORGANIZER_HYBRID`` for plugins that
```

src/pretix/control/templates/pretixcontrol/organizers/plugin_events.html
```diff
@@ -0,0 +1,45 @@
...
+            {% blocktrans trimmed with name=plugin.name %}
+                The plugin "{{ name }}" is enabled for your organizer account,  but also needs to be enabled for the
+                specific events you want to use it with.
+            {% endblocktrans %}
+        </p>
+    {% endif %}
+    <p>
+        {% blocktrans trimmed %}
+            Using this form, you can quickly enable or disable it for many events. Note that the configuration of
+            the plugin might still be necessary for each event individually.
```
luelista:
```suggestion
            Using this form, you can quickly enable or disable it for many events. Note that it might still
            be necessary to configure the plugin for each event individually.
```

src/pretix/control/templates/pretixcontrol/organizers/plugins.html
```diff
@@ -0,0 +1,193 @@
+{% extends "pretixcontrol/organizers/base.html" %}
+{% load i18n %}
+{% load static %}
+{% load bootstrap3 %}
+{% block content %}
+    <h1>{% trans "Available plugins" %}</h1>
+    <p>
+        {% blocktrans trimmed %}
+            On this page, you can choose plugins you want to enable for your organizer. Plugins might bring additional
```
luelista:
```suggestion
            On this page, you can choose plugins you want to enable for your organizer account. Plugins might bring additional
```

src/pretix/control/templates/pretixcontrol/organizers/plugins.html
```diff
@@ -0,0 +1,193 @@
...
+            software functionality, connect your event to third-party services, or apply other forms of customizations.
```
luelista:
```suggestion
            software functionality, connect your events to third-party services, or apply other forms of customizations.
```

src/pretix/control/navigation.py
```diff
@@ -490,6 +490,13 @@ def get_organizer_navigation(request):
                     }),
                     'active': url.url_name == 'organizer.edit',
                 },
+                {
+                    'label': _('Plugins'),
+                    'url': reverse('control:organizer.settings.plugins', kwargs={
+                        'organizer': request.organizer.slug,
+                    }),
+                    'active': url.url_name == 'organizer.settings.plugins',
```
luelista:
```suggestion
                    'active': url.url_name == 'organizer.settings.plugins' or url.url_name == 'organizer.settings.plugin-events',
```

src/pretix/control/urls.py
```diff
@@ -114,6 +114,10 @@
     re_path(r'^organizers/select2$', typeahead.organizer_select2, name='organizers.select2'),
     re_path(r'^organizer/(?P<organizer>[^/]+)/$', organizer.OrganizerDetail.as_view(), name='organizer'),
     re_path(r'^organizer/(?P<organizer>[^/]+)/edit$', organizer.OrganizerUpdate.as_view(), name='organizer.edit'),
+    re_path(r'^organizer/(?P<organizer>[^/]+)/settings/plugins$',
+            organizer.OrganizerPlugins.as_view(), name='organizer.settings.plugins'),
+    re_path(r'^organizer/(?P<organizer>[^/]+)/settings/plugin-events/(?P<plugin>[^/]+)$',
+            organizer.OrganizerPluginEvents.as_view(), name='organizer.settings.plugin-events'),
```
luelista: I think this would represent the page structure better in the URL:
```suggestion
    re_path(r'^organizer/(?P<organizer>[^/]+)/settings/plugins/(?P<plugin>[^/]+)/events$',
            organizer.OrganizerPluginEvents.as_view(), name='organizer.settings.plugin-events'),
```

src/pretix/base/models/organizer.py
```diff
@@ -143,6 +154,52 @@ def cache(self):
...
+
+        self.plugins = ",".join(modules)
+
+    def enable_plugin(self, module, allow_restricted=frozenset()):
+        plugins_active = self.get_plugins()
+        if module not in plugins_active:
+            plugins_active.append(module)
+            self.set_active_plugins(plugins_active, allow_restricted=allow_restricted)
+
+    def disable_plugin(self, module):
```
luelista, line 197: In case of a hybrid plugin, we need to check whether the plugin is still enabled in some events (and either disable everywhere, or fail)

doc/api/resources/organizers.rst
```diff
@@ -87,14 +93,61 @@ Endpoints
...
+
+   Updates an organizer. Currently only the ``plugins`` field may be updated.
+
+   Permission required: "Can change organizer settings"
+
+   **Example request**:
+
+   .. sourcecode:: http
+
+      PATCH /api/v1/organizers/bigevents/events/sampleconf/ HTTP/1.1
```
wiffbi:
```suggestion
      PATCH /api/v1/organizers/bigevents/ HTTP/1.1
```

doc/development/api/plugins.rst
```diff
@@ -56,14 +56,28 @@ restricted         boolean (optional)   ``False`` by default, restricts a plugin
...
 compatibility      string               Specifier for compatible pretix versions.
+level              string               System level the plugin can be activated at.
+                                        Set to ``pretix.base.plugins.PLUGIN_LEVEL_EVENT`` for plugins that can be activated
+                                        at event level and then be active for that event only.
+                                        Set to ``pretix.base.plugins.PLUGIN_LEVEL_ORGANIZER`` for plugins that can be
+                                        activated only for the organizer as a whole and are active for any event within
+                                        that organizer.
+                                        Set to ``pretix.base.plugins.PLUGIN_LEVEL_EVENT_ORGANIZER_HYBRID`` for plugins that
+                                        can be activated at organizer level but are considered active only within events
+                                        for which they have also been specifically activated.
```
wiffbi, line 67: Should/Can we give an example here? I guess seating is something that falls into this category?

src/pretix/base/plugins.py
```diff
@@ -56,8 +63,29 @@ def get_all_plugins(event=None) -> List[type]:
...
+                    event_fallback_used = True
+                    if not event or not app.is_available(event):
+                        continue
+            elif level == PLUGIN_LEVEL_ORGANIZER:
+                if organizer and hasattr(app, 'is_available'):
+                    if not app.is_available(organizer):
+                        continue
+                elif event and hasattr(app, 'is_available'):
+                    if not app.is_available(event.organizer):
+                        continue
```
wiffbi, line 84: If I pass an `event`, don’t I expect to only get a list of `PLUGIN_LEVEL_EVENT || PLUGIN_LEVEL_EVENT_ORGANIZER_HYBRID` plugins? So I would not include `PLUGIN_LEVEL_ORGANIZER` plugins at all if `event` is set.

```suggestion
                elif event:
                    continue
```

src/pretix/base/plugins.py
```diff
@@ -43,11 +48,13 @@ class PluginType(Enum):
...
 
 
-def get_all_plugins(event=None) -> List[type]:
+def get_all_plugins(*, event=None, organizer=None) -> List[type]:
     """
     Returns the PretixPluginMeta classes of all plugins found in the installed Django apps.
     """
+    assert not event or not organizer
     plugins = []
+    event_fallback_used = False
```
wiffbi, line 58: See comment below for using `event_fallback` instead of reusing `event`:
```suggestion
    event_fallback = None
    event_fallback_used = False
```

src/pretix/base/plugins.py
```diff
@@ -56,8 +63,29 @@ def get_all_plugins(event=None) -> List[type]:
...
+                        continue
+                elif organizer and hasattr(app, 'is_available'):
+                    # TODO: Re-think this behaviour. I don't think we have a better one available that's backwards-compatible?
+                    if event_fallback_used:
+                        # No events exist, no need to run query again
+                        continue
+                    event = organizer.events.first()
+                    event_fallback_used = True
+                    if not event or not app.is_available(event):
+                        continue
```
wiffbi: Hm, setting `event = organizer.events.first()` is clever as it afterwards just reuses the is_availble-check above, but feels dangerous to me as it contradicts the `assert not event or not organizer` from the beginning of the method and will have probably unexpected effects on e.g. line 84 and line 88. What if we use an `event_fallback` instead. Something like:

```suggestion
                    if not event_fallback_used:
                        event_fallback = organizer.events.first()
                        event_fallback_used = True
                    if not event_fallback or not app.is_available(event_fallback):
                        continue
```

wiffbi (commented):
Looks good so far, just some minor comments. I will do some more testing though.

doc/development/api/plugins.rst
raphaelm: I'm not sure we have any open-source plugins that would be an useful example here.

src/pretix/base/plugins.py
raphaelm: I think we said in Ulm that we want to list all plugins on both levels, such that when you are looking for the DATEV plugin in an event, you see that it exists, but you need to go to organizer level to activate it. So I think it's correct this way. Or do you notice any following bugs?

src/pretix/base/models/organizer.py
raphaelm: We don't technically *need* to, because is_app_active will only be True if it is active on both levels, so leaving it enabled on the events is unclean, but harmless, isn't it?

How strongly do you feel about this? The problem is that it would require us to change the semantics of enable_plugin/disable_plugin: Currently, they only work on the object they are called on, but do not to any database work themselves. This would likely need to change then.

src/pretix/base/plugins.py
```diff
@@ -56,8 +64,29 @@ def get_all_plugins(event=None) -> List[type]:
...
+                elif organizer and hasattr(app, 'is_available'):
+                    # TODO: Re-think this behaviour. I don't think we have a better one available that's backwards-compatible?
+                    if event_fallback_used:
+                        # No events exist, no need to run query again
+                        continue
+                    event_fallback = organizer.events.first()
+                    event_fallback_used = True
+                    if not event_fallback or not app.is_available(event_fallback):
+                        continue
```
wiffbi: Hm, that is not the way I suggested it. You still need to check `event_fallback` for hasattr, otherwise you only check the first plugin, all others won’t be available. So either do it as I suggested with `if not event_fallback_used:` and set the event_fallback in that if-condition. Or check above with (IMHO not as readable/understandable as my original suggestion):
```
                if (event or event_fallback) and hasattr(app, 'is_available'):
                    if not app.is_available(event or event_fallback):
```

src/pretix/base/plugins.py
wiffbi: You are right, sorry!

src/pretix/base/models/organizer.py
raphaelm: To resolve this without a major refactoring, I moved the responsibility to the caller (UI + API) and implemented it the way you expected. I think there's an argument to be made for both possibilities (i.e. keeping it in the list of events can also be useful, with the organizer level being a quick on/off switch), but this approach is certainly "purer". (21f08ad06)

doc/development/api/plugins.rst
```diff
@@ -56,14 +56,28 @@ restricted         boolean (optional)   ``False`` by default, restricts a plugin
...
 
 A working example would be:
 
 .. code-block:: python
 
     try:
-        from pretix.base.plugins import PluginConfig
+        from pretix.base.plugins import PluginConfig, PLUGIN_LEVEL_EVENT
     except ImportError:
         raise RuntimeError("Please use pretix 2.7 or above to run this plugin!")
```
raphaelm:
```suggestion
        raise RuntimeError("Please use pretix 2025.7 or above to run this plugin!")
```

raphaelm:
Hosted migration plan:

```
ALTER TABLE "pretixbase_organizer" ADD COLUMN "plugins" text DEFAULT '' NOT NULL;
INSERT INTO django_migrations (app, name, applied) VALUES ('pretixbase', '0287_organizer_plugins',  NOW());
```

PR #5308, opened 2025-07-14
Add Spanish (LatAm) and improve how we count language coverage (Z#23200505)

raphaelm:
On request, this adds a language "Spanish (Latin America)". We're well aware that the Spanish spoken varies between different parts of Latin America, but this seems a decent compromise for now to avoid dozens of translations. 

This is our first locale that includes a number, which required some changes in pretix and some minimal monkeypatching for Django.

This is one of our first locales where we also support a "parent" locale, i.e. we now support ``es`` and ``es-419``. I verified that the fallback mechanism works, i.e. if a string does not exist in es-419, the string from es will be displayed, avoiding the need to copy translations over. I have therefore updated the `get_language_score` function to reflect this fallback, so es-419 isn't shown with <5% even though 80% will show up in Spanish. Of course, this is also somehow wrong, because the Spanish spoken in Spain might show up, but I think the 80% still reflects the reality more closely than the 5%, especially if we encourage translators only to translate strings that are actually different.


src/pretix/_base_settings.py
```diff
@@ -172,6 +173,12 @@
         'name': 'Norwegian Bokmal',
         'name_local': 'norsk (bokmål)',
     },
+    'es-419': {
+        'bidi': False,
+        'code': 'es-419',
+        'name': 'Spanish (Latin America)',
+        'name_local': 'Español (Hispanoamerica)',
```
wiffbi: Why do we make the distinction in the `name_local` here, but not for e.g. „Portuguese (Portugal)“? „Español (Hispanoamerica)“ in the language selector on pretix-presale IMHO looks a bit strange. Of course it makes sense, if one offers both variants of Spanish, but is that an even remotely common use-case?

I’d say we just use „Spanish“ and if the need really arrises, we look into distinguishing language variants only if other language variants with the same base-language are active?

```suggestion
        'name_local': 'Español',
```

src/pretix/base/management/commands/makemessages.py
```diff
@@ -0,0 +1,46 @@
...
+# Unless required by applicable law or agreed to in writing, software distributed under the Apache License 2.0 is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+# License for the specific language governing permissions and limitations under the License.
+import re
+
+from django.core.management.commands import makemessages
+
+
+def is_valid_locale(locale):
+    return re.match(r"^[a-z]+$", locale) or re.match(r"^[a-z]+_[A-Z0-9].*$", locale)
```
wiffbi, line 28: Not sure how strong we want to verify a valid locale, but the second re seems a bit broad with the trailing `.`. Looking at https://www.localeplanet.com/icu/index.html I would limit to numbers, letters and underscore.

```suggestion
    return re.match(r"^[a-z]+$", locale) or re.match(r"^[a-z]+_[A-Z0-9][a-zA-Z0-9_]*$", locale)
```

src/pretix/helpers/i18n.py
```diff
@@ -198,10 +197,16 @@ def get_language_score(locale):
...
-            else:
-                catalog.update(translation)
+
+            catalog.update(translation._catalog.copy())
+
+            # Also add fallback catalog (e.g. es for es-419, de for de-informal, …)
+            while translation._fallback:
+                if not locale.startswith(translation._fallback.info().get("language", "XX")):
+                    break
+                catalog.update(translation._fallback._catalog.copy())
```
wiffbi: Don’t line 207 and line 209 do the same thing?
```suggestion
```

src/pretix/helpers/i18n.py
```diff
@@ -214,10 +219,14 @@ def get_language_score(locale):
...
-        if catalog is None:
-            catalog = TranslationCatalog(translation)
-        else:
-            catalog.update(translation)
+        catalog.update(translation._catalog.copy())
+
+        while translation._fallback:
+            if not locale.startswith(translation._fallback.info().get("language", "XX")):
+                break
+            catalog.update(translation._fallback._catalog.copy())
```
wiffbi: Same as above, don’t line 227 and 229 do the same thing?
```suggestion
```

wiffbi (changes requested):
Looks good and works as expected. When calculating the score, the catalog will be updated twice with the same info. And an opinion regarding labelling the language variant in presale.

src/pretix/_base_settings.py
raphaelm: You are correct.

src/pretix/base/management/commands/makemessages.py
raphaelm: This is copy-pasted from Django, I only extended the 0-9 and would not extend more then necessary

PR #5311, opened 2025-07-15
Add a system-wide style for admin-only things

raphaelm:


src/pretix/static/pretixcontrol/scss/main.scss
```diff
@@ -490,6 +490,13 @@ td > p:last-child,
     }
 }
 
+.admin-only,
+.panel-default > .panel-heading.admin-only,
```
wiffbi, line 494: What about adding the stripes to the active admin-button at the top as well? One would need to change line 50 in this file from `background:` to `background-color:` and add `admin-only` CSS-class to `pretix/control/templates/pretixcontrol/base.html::201`

src/pretix/static/pretixcontrol/scss/main.scss
```diff
@@ -490,6 +490,13 @@ td > p:last-child,
...
+.panel-default > legend > .panel-heading.admin-only {
+    background-color: rgba(255, 231, 97, 0.1);
```
wiffbi: I am not super keen on the yellow here – it works well on the panel-heads, but on the small inspect-buttons to me it feels like an artefact of broken rendering on the screen, especially combined with the stripes. Or maybe it is just my tired eyes? But leaving it out doesn’t help on the panel-heads. So maybe just add the yellow only to the panel-heads and only the stripes to all .admin-only?

wiffbi (approved):
Just an idea to put stripes on the end-admin-session-button as well to make the connection a little bit clearer. And maybe not use the light yellow on the small inspect-buttons, in my (tired) eyes, it does not look good on those small parts of the UI.

src/pretix/static/pretixcontrol/scss/main.scss
raphaelm: Wanna try it out? ;) You have the better eye for things like this

wiffbi: That was a tough one. I was looking for something, that distinguishes enough from other controls, but can be applied to several different types of elements (buttons, panel-head, etc.), I even added a little dot with :after, but that was not really standing out. So I kept your stripes, made them a bit narrower (so the little buttons have more than one stripe) and tried a red-ish tone to match the color of the „end-admin“-button – which also got stripes, when active to make a visual connection with the red-ish striped elements.

Bildschirmfoto 2025-07-18 um 23 24 45

PR #5315, opened 2025-07-16
Generalize link in footer to "Contact" (Z#23200756)

raphaelm:


raphaelm:
ah, GitHub misclick. will do the other change tomorrow

PR #5320, opened 2025-07-21
OrderPayment.fail: Change race condition detection condition

stranck:
I have updated the race condition check of the method `OrderPayment.fail()` (inside `pretix/base/models/orders.py`) from:
```py=
if locked_instance.state not in (OrderPayment.PAYMENT_STATE_CREATED, OrderPayment.PAYMENT_STATE_PENDING):
    # abort fail
```
to:
```py=
if locked_instance.state in (OrderPayment.PAYMENT_STATE_CONFIRMED, OrderPayment.PAYMENT_STATE_FAILED, OrderPayment.PAYMENT_STATE_REFUNDED):
    # abort fail
```

This enables the state transition from `PAYMENT_STATE_CANCELED` to `PAYMENT_STATE_FAILED` (which was previously forbidden).

I personally have no idea if it was intended to forbid passing from canceled to failed, but from the description of the method (_[...] since we do not want to report a failure for an order that has just been marked as paid._) it seems acceptable

I'm requesting this change for a _strange_ reason: I have some payments which may be incorrectly marked as canceled due to a bad payment provider, which I need to refresh (using the runperiodic signal) until a specific timeout is reached. In this instance, the payment which was previously canceled should be now marked as failed (implying also we won't refresh it anymore in runperiodic to check for its status)

### DIFF
--- a/src/pretix/base/models/orders.py
+++ b/src/pretix/base/models/orders.py
@@ -1821,7 +1821,7 @@ class OrderPayment(models.Model):
 
     def fail(self, info=None, user=None, auth=None, log_data=None, send_mail=True):
         """
-        Marks the order as failed and sets info to ``info``, but only if the order is in ``created`` or ``pending``
+        Marks the order as failed and sets info to ``info``, but only if the order is in ``created``, ``pending`` or ``canceled``
         state. This is equivalent to setting ``state`` to ``OrderPayment.PAYMENT_STATE_FAILED`` and logging a failure,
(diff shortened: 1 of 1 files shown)


raphaelm:
Hmm, I mean, one could expect "canceled" also to be a "final" state that won't change any more, but we're usually not so strict about these at the moment…

raphaelm:
I'll merge it with the warning that we might revert this one day in the far future if we need to

PR #5323, opened 2025-07-23
Orders API: Add check_quotas to orders/change and PATCH/POST orderpositions query params

stranck:
While there's an option to ignore quota when adding/updating orderpositions from the /control/ pages in the UI, there is not via the REST API. This commit adds a `check_quota` boolean query parameter on the following endpoints:
- PATCH `/orderpositions/{id}/`
- POST `/orderpositions/`
- POST `/orders/{orderCode}/change/` (both on `patch_positions` and `create_positions`)

For the first two endpoints, with this PR we now add the `check_quota` variable to the context of the `OrderPositionViewSet` (in `get_serializer_context()`) which is later used inside the `OrderPositionChangeSerializer` and the `OrderPositionCreateForExistingOrderSerializer` by passing that value to the `OrderChangeManager.commit()` method.  
For the last one it's instead trivially used at the end of the `EventOrderViewSet.change()` method, again by passing it to the `commit()` call.  
I hope I got the implementation right! :D

I have also updated the docs and added some unit tests. I don't have a lot of experience with django, so I'm not sure I've used the `with scopes_disabled():` directive correctly in the unit tests I have written. I kindly ask to review them :)

### DIFF
--- a/src/pretix/api/serializers/orderchange.py
+++ b/src/pretix/api/serializers/orderchange.py
@@ -83,7 +83,7 @@ class OrderPositionCreateForExistingOrderSerializer(OrderPositionCreateSerialize
 
     def create(self, validated_data):
         ocm = self.context['ocm']
-        check_quotas = self.context['check_quotas'] if 'check_quotas' in self.context else True
+        check_quotas = self.context.get('check_quotas', True)
 
         try:
             ocm.add_position(
@@ -311,7 +311,7 @@ class OrderPositionChangeSerializer(serializers.ModelSerializer):
 
     def update(self, instance, validated_data):
         ocm = self.context['ocm']
-        check_quotas = self.context['check_quotas'] if 'check_quotas' in self.context else True
+        check_quotas = self.context.get('check_quotas', True)
         current_seat = {'seat_guid': instance.seat.seat_guid} if instance.seat else None
         item = validated_data.get('item', instance.item)
         variation = validated_data.get('variation', instance.variation)
--- a/src/pretix/api/serializers/orderchange.py
+++ b/src/pretix/api/serializers/orderchange.py
@@ -83,6 +83,7 @@ class OrderPositionCreateForExistingOrderSerializer(OrderPositionCreateSerialize
 
     def create(self, validated_data):
         ocm = self.context['ocm']
+        check_quotas = self.context['check_quotas'] if 'check_quotas' in self.context else True
 
         try:
             ocm.add_position(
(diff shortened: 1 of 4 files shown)


src/pretix/api/serializers/orderchange.py
```diff
@@ -83,6 +83,7 @@ def validate(self, data):
 
     def create(self, validated_data):
         ocm = self.context['ocm']
+        check_quotas = self.context['check_quotas'] if 'check_quotas' in self.context else True
```
raphaelm:
```suggestion
        check_quotas = self.context.get('check_quotas', True)
```

src/pretix/api/serializers/orderchange.py
```diff
@@ -310,6 +311,7 @@ def validate(self, data, instance=None):
 
     def update(self, instance, validated_data):
         ocm = self.context['ocm']
+        check_quotas = self.context['check_quotas'] if 'check_quotas' in self.context else True
```
raphaelm:
```suggestion
        check_quotas = self.context.get('check_quotas', True)
```

src/pretix/api/views/order.py
```diff
@@ -1087,6 +1088,7 @@ def get_serializer_context(self):
         ctx = super().get_serializer_context()
         ctx['event'] = self.request.event
         ctx['pdf_data'] = self.request.query_params.get('pdf_data', 'false') == 'true'
+        ctx['check_quotas'] = self.request.query_params.get('check_quotas', 'true') == 'true'
```
raphaelm:
```suggestion
        ctx['check_quotas'] = self.request.query_params.get('check_quotas', 'true').lower() == 'true'
```
alternatively: fail with an error 400 if the value is neither true nor false

raphaelm (commented):
Implementation looks good, tests also look good. Two minor inline comments on style or robustness.

I'm not 100% happy with it being a query parameter yet, as for the /change/ endpoint I would have expected it in the same place as e.g. "reissue_invoice", which is the body, but I see that that doesn't make much sense for e.g. the PATCH endpoint, so query parameter might be best. Happy for additional input from the @pretix/dev team.

src/pretix/api/views/order.py
stranck: Unrelated from this PR, shouldn't the .lower() be added also to the other request params (like the `pdf_data` on line 1089)? 
I'm going to commit everything soon btw!

raphaelm: Yes… agreed, we should have that for the existing ones as well :D Or ignore it here for now and add it later

stranck: It's fine :D
Maybe tho going through all possible occurrence of this kind of error needs a separate pr?

raphaelm: Yes, agreed

raphaelm (approved):

crackwitz:
Thanks for this feature!

By the way, the docs don't say when it was introduced.

PR #5326, opened 2025-07-25
API/Organizer: Allow Device-Token access to Organizer settings; expose mf0aes_random_uid

pc-coholic:
:exclamation: Check for security implications of this change

As of right now, none of our apps would be using this endpoint - but I feel like exposing it to the unrestricted SecurityProfile shouldn't do any harm.


raphaelm (approved):

pc-coholic:
ToDo: Fix tests

PR #5327, opened 2025-07-25
Order change: Add setting for preventing users from removing positions

stranck:
This PR adds a new option in the _Event > Settings > Cancellation > Order Changes_ to specifically disable users from removing positions from their order. This includes both item with variations and add-ons. I haven't found an existing setting to prevent this beside the price contraints which were flawed in some cases:

Let's say we have two addons for preordering items before an event: A shirt $S$ which costs 20€ and a mug $M$ which instead costs 5€ and it's personalized with the name of the owner of an order. Let's also define the price $p$ of the order, a user $u$ and an organization $o$. If:
- $u$ places an order with $M$ as an addon ($p'$ = $p$ + 5)
- $o$ physically buys a mug which should be given to $u$ at event time. The mug has $u$ name on it
- Now $u$ removes $M$ from its order but at the same time adds $S$ ($p''$ = $p'$ - 5 + 20 = $p'$ + 15)
- Since trivially $p'' > p'$ all of the options under _Event > Settings > Cancellation > Order Changes > Requirement for changed prices_ will allow this order change to proceed
- This means that now $o$ has an extra personalized mug which cannot give to someone else

With this new setting I aim to prevent this exact scenario, while still allowing order changes which instead add items (We want that $u$ can buy $S$ at any time)

This PR includes also unit tests. Like the previous PR I did (#5323) I'm still not sure if I've used the `with scopes_disabled():` directive correctly, so please review the unit tests I've added hahahha

Currently I've implemented the validation only on the backend side: Like for order changes price constraints, an user will still be able to deselect positions from the presale order change's page and will be stopped with an alert only after they clicked on _continue_. This can also be theoretically enforced on the frontend (maybe by disabling buttons for already selected max-1 items and by faking min-count=current-count for items with max-count > 1), but like for an old PR I did (and I still have to work on, #4662) I don't have a clever idea of how to properly implement it with a good UX

### DIFF
--- a/src/pretix/api/serializers/event.py
+++ b/src/pretix/api/serializers/event.py
@@ -848,6 +848,7 @@ class EventSettingsSerializer(SettingsSerializer):
         'cancel_terms_paid',
         'cancel_terms_unpaid',
         'change_allow_user_variation',
+        'change_allow_user_remove_positions',
         'change_allow_user_addons',
         'change_allow_user_until',
         'change_allow_user_price',
--- a/src/pretix/base/settings.py
+++ b/src/pretix/base/settings.py
@@ -1783,6 +1783,15 @@ DEFAULTS = {
             label=_("Customers can change their selected add-on products"),
         )
     },
+    'change_allow_user_remove_positions': {
+        'default': 'False',
+        'type': bool,
+        'form_class': forms.BooleanField,
+        'serializer_class': serializers.BooleanField,
+        'form_kwargs': dict(
+            label=_("Customers can remove positions from their order"),
+        )
+    },
     'change_allow_user_price': {
         'default': 'gte',
         'type': str,
--- a/src/pretix/control/forms/event.py
+++ b/src/pretix/control/forms/event.py
@@ -758,6 +758,7 @@ class CancelSettingsForm(SettingsForm):
         'change_allow_user_price',
         'change_allow_user_until',
         'change_allow_user_addons',
+        'change_allow_user_remove_positions',
         'change_allow_user_if_checked_in',
         'change_allow_attendee',
         'tax_rule_cancellation',
--- a/src/pretix/control/templates/pretixcontrol/event/cancel.html
+++ b/src/pretix/control/templates/pretixcontrol/event/cancel.html
@@ -68,6 +68,7 @@
                 </div>
                 {% bootstrap_field form.change_allow_user_variation layout="control" %}
                 {% bootstrap_field form.change_allow_user_addons layout="control" %}
+                {% bootstrap_field form.change_allow_user_remove_positions layout="control" %}
                 {% bootstrap_field form.change_allow_user_until layout="control" %}
                 {% bootstrap_field form.change_allow_user_price layout="control" %}
                 {% bootstrap_field form.change_allow_user_if_checked_in layout="control" %}
--- a/src/pretix/presale/views/order.py
+++ b/src/pretix/presale/views/order.py
@@ -1544,12 +1544,12 @@ class OrderChangeMixin:
                         messages.error(request, e.message % e.params if e.params else e.message)
                         return self.get(request, *args, **kwargs)
 
-                    for (i, v), (c, price) in selected.items():
+                    for (i, v), (count, price) in selected.items():
                         addons_data.append({
                             'addon_to': p.pk,
                             'item': i.pk,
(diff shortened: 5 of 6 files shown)


stranck:
Sorry for the spam of unit test fix. Sadly on my local setup pytest is not working: When I try to run it with Pretix's test suite it fails like a 20% of the tests for various reasons (I'd like to open a issue/discussion in the future to seek for some help) + wsl is slow as hell with IO operations, so I had no proper way tu run the tests and debug. I know this is not professional and I'd like to apologize for that

raphaelm:
Wouldn't it be sufficient in your setup to set the flag in the mug product that it is not cancellable? I think an additional setting is not necessary

raphaelm:
Granted, currently setting the flag on one product prevents changes to the entire order so it wouldn't help you. But then making that logic more precise would be the harder, but better change. The bar for introducing new settings is quite high and I don't think this meets it.

stranck:
Yep, sadly disabling the _Product > Availability > Allow product to be canceled or changed_ setting will totally prevent an order with that item from being changed :(
I felt like this option I introduced is similar to the already present ones for preventing users from changing/canceling both addons and variations, this is why I intruduced it as a "global" setting. However, I have to agree that the ability to specify which items can be canceled and which not it's definitely more useful

I don't think tho it requires much extra labor: To check whenever we're canceling something, I already iterate through the changes inside a OCM and if I find an instance of CancelOperation I quit with an error. Having this setting on a item-basis instead of globally for an event, wuold just mean that for whenever CancelOperation found, we just retrieve the Item settings, check if the flag for preventing cancelation is set and, in case, quit with an error. The extra labor for me would simply be understanding what I have to do to add a new option to an order (I think I also have to write a migration?), but I'll eventually manage to do it, maybe I'll ask for a bit of help hahahah

Anyway, as a pretix user I'd honestly won't change the previous "Allow product to be canceled or changed" option, but rename it AND add this other one, to keep compatibility with previous event settings. Like having:
- "Allow orders with this item to be canceled or changed" (Which is the "legacy" option we have now, simply renamed)
- "Allow this item to be canceled" (This new option)
I honestly won't "incorporate" together disabling the cancelation of an item and changing its variation. There may be cases where we want people to change variation, but not cancel the position, EG (from the kind of evets I manage): Sponsorship item for people who wants to donate towards the organization with 2 variation: "Normal sponsor" (which includes irl gadgets A, B and C), "Super sponsor" (which includes gadgets A, B, C, but also D and E). In this instance, I want people who already bought a normal sponsor to upgrade to a supersponsor, but I don't want to cancel their sponsorship position nor passing from supersponsor to the normal one (this last thing it's already prevented using the order change price constraints)

raphaelm:
> I don't think tho it requires much extra labor: To check whenever we're canceling something, I already iterate through the changes inside a OCM and if I find an instance of CancelOperation I quit with an error. Having this setting on a item-basis instead of globally for an event, wuold just mean that for whenever CancelOperation found, we just retrieve the Item settings, check if the flag for preventing cancelation is set and, in case, quit with an error. The extra labor for me would simply be understanding what I have to do to add a new option to an order (I think I also have to write a migration?), but I'll eventually manage to do it, maybe I'll ask for a bit of help hahahah

A migration is not necessary, the hard part is probably in the UI around showing when an order is change-able and how to show which aprts are change-able.

> Anyway, as a pretix user I'd honestly won't change the previous "Allow product to be canceled or changed" option, but rename it AND add this other one, to keep compatibility with previous event settings. Like having:
> "Allow orders with this item to be canceled or changed" (Which is the "legacy" option we have now, simply renamed)
>"Allow this item to be canceled" (This new option)

I disagree. What is the scenario where you want one item to block all changes to an order?

> I honestly won't "incorporate" together disabling the cancelation of an item and changing its variation. There may be cases where we want people to change variation, but not cancel the position, EG (from the kind of evets I manage): Sponsorship item for people who wants to donate towards the organization with 2 variation: "Normal sponsor" (which includes irl gadgets A, B and C), "Super sponsor" (which includes gadgets A, B, C, but also D and E). In this instance, I want people who already bought a normal sponsor to upgrade to a supersponsor, but I don't want to cancel their sponsorship position nor passing from supersponsor to the normal one (this last thing it's already prevented using the order change price constraints)

We'll not be able to accommodate *every* situation. I don't think we want separate options for this.

PR #5333, opened 2025-07-29
Pass more objects to order secret creator

TheEnbyperor:
It would be useful to have access to the order position and the order in some secret generators. This PR makes Pretix pass them onwards, if requested.

### DIFF
--- a/src/pretix/base/secrets.py
+++ b/src/pretix/base/secrets.py
@@ -259,6 +259,12 @@ def assign_ticket_secret(event, position, force_invalidate_if_revokation_list_us
         kwargs['valid_until'] = position.valid_until
     if 'order_datetime' in params:
         kwargs['order_datetime'] = position.order.datetime
+    if 'order_position' in params:
+        kwargs['order_position'] = position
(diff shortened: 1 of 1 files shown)


raphaelm:
Nope, sorry. This is very intentionally not the case. Will reply with more detailed reason after my vacation.

raphaelm:
Okay, so here's the thing: The current parameters to generate secret (item, variation, subevent, attendee_name, valid_from, valid_untril, order_datetime) are either immutable or things where we strongly control any change. Specifically, the core system guarantees the plugin that whenever one of these values is changed, the secret generator is called again and can make a decision on whether to generate a new secret – which might be necessary when e.g. the valid_from value becomes part of the barcode.

Passing full position/order objects is obviously convenient, but kills this guarantee. Now, a QR code might be generated that includes OrderPosition.seat. However, core does not include mechanisms to trigger a new QR code whenever a seat is reassigned, including options to send new ticket codes via email etc. Giving this guarantee for the entire model does not seem realistic.

(side note: organizer is already available through the event)

PR #5336, opened 2025-07-31
Clarify checkout success message

waldyrious:
Since both emails and links can be said to have addresses, the previous formulation of these messages could be confusing and cause a bit of a mental hiccup. Specifically, the previous construct could be interpreted as "we sent you [...] the link to the address you specified" (i.e. "address" could be interpreted as an URL).

By swapping the order of the phrasing, the sentence structure becomes less ambiguous. With the proposed order, it is clear that we mean "sent you an email to the [email] address you specified".


waldyrious:
I assume this change qualifies for the CLA exception, but just in case I sent the signed document via email.

Also, just for the sake of record-keeping, these messages were last edited in https://github.com/pretix/pretix/pull/832.

pc-coholic:
@cla-bot check

PR #5352, opened 2025-08-07
Allow users to run sync jobs immediately

luelista:


src/pretix/base/services/datasync.py
```diff
@@ -52,6 +52,22 @@ def periodic_reset_in_flight(sender, **kwargs):
         sq.set_sync_error('timeout', [], 'Timeout')
 
 
+def run_sync(queue):
+        grouped = groupby(sorted(queue, key=lambda q: (q.sync_provider, q.event.pk)), lambda q: (q.sync_provider, q.event))
```
raphaelm: Too much indentation? Legal, but weird.

raphaelm (approved):
Yes, but: https://github.com/pretix/pretix/pull/5355

src/pretix/base/services/datasync.py
luelista: oops 🫣

PR #5356, opened 2025-08-08
Order overview: Try to make linked filters behave as expected for line-level cancellations (Z#23203500)

raphaelm:


luelista:
The footnote will be outdated with this change, right?

https://github.com/pretix/pretix/blob/e2b3cd748ffc763485d515908b8da1729d71c440/src/pretix/control/templates/pretixcontrol/orders/overview.html#L188-L190

luelista (approved):

raphaelm:
This footnote was apparently so well hidden that even I did not see it when confronted with the "bug" :joy:

PR #5360, opened 2025-08-08
Webhooks: Add vouchers (Z#23203072)

raphaelm:
This also requires more consistent usage of webhook types to avoid vouchers not being known to the external system.

### DIFF
--- a/src/pretix/api/webhooks.py
+++ b/src/pretix/api/webhooks.py
@@ -78,6 +78,13 @@ class WebhookEvent:
         """
         raise NotImplementedError()  # NOQA
 
+    @property
+    def help_text(self) -> str:
+        """
+        A human-readable description
+        """
+        return ""
+
 
 def get_all_webhook_events():
     global _ALL_EVENTS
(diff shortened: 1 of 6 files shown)


src/pretix/api/webhooks.py
```diff
@@ -381,6 +407,19 @@ def register_default_webhook_events(sender, **kwargs):
...
             _('Waiting list entry received voucher'),
         ),
+        ParametrizedVoucherWebhookEvent(
+            'pretix.voucher.added',
+            _('Voucher added'),
+        ),
+        ParametrizedVoucherWebhookEvent(
+            'pretix.voucher.changed',
+            _('Voucher changed'),
+            _('Only includes explicit to the voucher, not e.g. an increase of the number of redemptions.')
```
luelista:
```suggestion
            _('Only includes explicit changes to the voucher, not e.g. an increase of the number of redemptions.')
```
changes?

luelista (approved):

luelista:
`test_waitinglist_shredder` seems to fail because the log entries are not in the expected order (pretix.voucher.added.waitinglist is first, pretix.voucher.added is last) and therefore the email dict entry is not found there

PR #5364, opened 2025-08-11
Add program times for items (Z#23178639)

ser8phin:
~~Todo~~ Done:
---
General
- [x] datetime widget validation
- [x] adding code paths for copying events and copying items 
- [x] checking/refactoring/bug fixing/tests
---
ical
- [x] refactoring incl commenting (branches and deduplication)
---
API
- [x] extending API with updating and reading of program times
- [x] extending of API docs
- [x] extending of API tests
---
PDF placeholders
- [x] Placeholder for program times
- [x] making it work for addons with program times (like workshops that are addons)

### DIFF
--- a/src/pretix/presale/ical.py
+++ b/src/pretix/presale/ical.py
@@ -174,7 +174,7 @@ def get_private_icals(event, positions):
                 url = build_absolute_uri(event, 'presale:event.index')
 
             if event.settings.mail_attach_ical_description:
-                ctx = get_email_context(event=event, event_or_subevent=p.subevent)
+                ctx = get_email_context(event=event, event_or_subevent=ev)
                 description = format_map(str(event.settings.mail_attach_ical_description), ctx)
             else:
                 # Default description


src/pretix/base/models/items.py
```diff
@@ -2294,3 +2294,30 @@ class ItemVariationMetaValue(LoggedModel):
...
+    :type item: Item
+    :param start: The date and time this program time starts
+    :type start: datetime
+    :param end: The date and time this program time ends
+    :type end: datetime
+    """
+    item = models.ForeignKey('Item', related_name='item_program_times', on_delete=models.CASCADE)
+    start = models.DateTimeField(
+        verbose_name=_("Start"),
+        null=True, blank=True
```
raphaelm: Why are these nullable? Having program times only makes sense when you actually enter them, I think.

src/pretix/control/views/item.py
```diff
@@ -1559,6 +1567,16 @@ def formsets(self):
                 queryset=ItemBundle.objects.filter(base_item=self.get_object()),
                 event=self.request.event, item=self.item, prefix="bundles"
             )),
+            ('program_times', inlineformset_factory(
```
raphaelm, line 1574: The formset must not be shown in event series, where program times make no sense.

src/pretix/control/templates/pretixcontrol/item/include_program_times.html
```diff
@@ -0,0 +1,68 @@
+{% load i18n %}
+{% load bootstrap3 %}
+{% load formset_tags %}
+<p>
+    {% blocktrans trimmed %}
+        With program times, you can set specific dates and times for this product.
```
raphaelm, line 6:
```suggestion
        With program times, you can set specific dates and times for this product.
        This is useful if this product represents access to parts of your event that happen at different times than your event in general.
        This will not affect access control, but will affect calendar invites and ticket output. 
```

src/pretix/presale/ical.py
```diff
@@ -122,61 +123,109 @@ def get_private_icals(event, positions):
...
-
-    evs = set(p.subevent or event for p in positions)
-    for ev in evs:
-        if isinstance(ev, Event):
+    calentries = set()  # using set for automatic deduplication of CalEntries
+    CalEntry = namedtuple('CalEntry', ['summary', 'description', 'location', 'dtstart', 'dtend', 'uid'])
+
+    # collecting the positions' calendar entries, preferring the most exact date and time available (positions > subevent > event)
+    for p in positions:
+        program_times = p.item.item_program_times.all()
```
raphaelm: You can prefetch this with https://docs.djangoproject.com/en/5.1/ref/models/querysets/#django.db.models.prefetch_related_objects

src/pretix/presale/ical.py
```diff
@@ -122,61 +123,109 @@ def get_private_icals(event, positions):
...
+        if program_times:
+            # if program times have been configured, they are preferred for the position's calendar entries
             url = build_absolute_uri(event, 'presale:event.index')
+            for index, pt in enumerate(program_times):
+                summary = _('Program time {number} for {item}').format(number=index + 1, item=p.item.name)
```
raphaelm: I think it's a little weird to show the buyer of a workshop something like "Program time 1 for Workshop ABCDE". I think just the title of the item is sufficient? Maybe "{event} – {item}" for more context?

src/pretix/presale/ical.py
```diff
@@ -122,61 +123,109 @@ def get_private_icals(event, positions):
...
+                if event.settings.mail_attach_ical_description:
+                    ctx = get_email_context(event=event, event_or_subevent=event)
+                    description = format_map(str(event.settings.mail_attach_ical_description), ctx)
+                else:
+                    # Default description
+                    descr = []
+                    descr.append(_('Program time {index}: {url}').format(index=index, url=url))
```
raphaelm: I don't think we need to expose the "program time" wording and the index to the buyer

src/pretix/presale/ical.py
```diff
@@ -122,61 +123,109 @@ def get_private_icals(event, positions):
...
+                        datetime=date_format(pt.start.astimezone(tz), 'SHORT_DATETIME_FORMAT')
+                    ))
+                    descr.append(str(_('End: {datetime}')).format(
+                        datetime=date_format(pt.end.astimezone(tz), 'SHORT_DATETIME_FORMAT')
+                    ))
+                    # Actual ical organizer field is not useful since it will cause "your invitation was accepted" emails to the organizer
+                    descr.append(_('Organizer: {organizer}').format(organizer=event.organizer.name))
+                    description = '\n'.join(descr)
+                location = None
+                dtstart = pt.start.astimezone(tz)
```
raphaelm, line 157: This will crash if pt.start is None, but I think we should forbid that anyways, see above.

src/pretix/presale/ical.py
```diff
@@ -122,61 +123,109 @@ def get_private_icals(event, positions):
...
+                dtend = pt.end.astimezone(tz)
+                uid = 'pretix-{}-{}-{}-{}@{}'.format(
+                    event.organizer.slug,
+                    event.slug,
+                    'programtime',
+                    index,
```
raphaelm, line 163: This will generate the same UID for the first program time of each product, so if I have two products in my order with differing program times, I will get a calendar file with a duplicate UID, which is not good.

We could fix this by including the item ID, which will make for less grouping, but is probably the best ID. We can't re-use the deduplication (like a hash over CalEntry), as we do *not* want the UID to change if the start/end time changes.

src/pretix/api/serializers/item.py
```diff
@@ -364,6 +385,7 @@ def create(self, validated_data):
         variations_data = validated_data.pop('variations') if 'variations' in validated_data else {}
         addons_data = validated_data.pop('addons') if 'addons' in validated_data else {}
         bundles_data = validated_data.pop('bundles') if 'bundles' in validated_data else {}
+        program_times_data = validated_data.pop('program_times') if 'program_times' in validated_data else {}
```
raphaelm, line 407: This requires extension of API docs and teste

raphaelm, line 407: This requires extension of API docs and tests, also am I reading this correctly that this allows creating program times through the API, but neither updating nor reading them is possible?

src/pretix/base/models/items.py
```diff
@@ -2294,3 +2294,30 @@ class ItemVariationMetaValue(LoggedModel):
 
     class Meta:
         unique_together = ('variation', 'property')
+
+
+class ItemProgramTime(models.Model):
```
raphaelm, line 2299: This requires addition to the code paths that are executed when copying events and copying items so the information is carried over.

src/pretix/api/serializers/item.py
```diff
@@ -50,6 +50,7 @@
     Item, ItemAddOn, ItemBundle, ItemCategory, ItemMetaValue, ItemVariation,
     ItemVariationMetaValue, Question, QuestionOption, Quota, SalesChannel,
 )
+from pretix.base.models.items import ItemProgramTime
```
wiffbi: As you already import ItemProgramTime to the imports in pretix/base/models/__init__.py you can move that import into the imports above from pretix.base.models

src/pretix/api/serializers/item.py
wiffbi: As @raphaelm points out, IMHO this need an `InlineItemProgramTimeSerializer` similar to variations’ `InlineItemVariationSerializer`, etc. Also this then needs a `validate_program_times`.

src/pretix/api/views/item.py
```diff
@@ -46,14 +46,15 @@
...
-    QuestionSerializer, QuotaSerializer,
+    ItemProgramTimeSerializer, ItemSerializer, ItemVariationSerializer,
+    QuestionOptionSerializer, QuestionSerializer, QuotaSerializer,
 )
 from pretix.api.views import ConditionalListView
 from pretix.base.models import (
     CartPosition, Item, ItemAddOn, ItemBundle, ItemCategory, ItemVariation,
     Question, QuestionOption, Quota,
 )
+from pretix.base.models.items import ItemProgramTime
```
wiffbi: see comment above regarding importing from pretix.base.models

src/pretix/api/views/item.py
```diff
@@ -279,6 +280,57 @@ def perform_destroy(self, instance):
...
+        ctx = super().get_serializer_context()
+        ctx['event'] = self.request.event
+        ctx['item'] = self.item
+        return ctx
+
+    def perform_create(self, serializer):
+        item = get_object_or_404(Item, pk=self.kwargs['item'], event=self.request.event)
+        serializer.save(item=item)
+        item.log_action(
+            'pretix.event.item.item_program_times.added',
```
wiffbi: You probably should add a translation to logdisplay.py – and also don’t forget to change to `pretix.event.item.program_times.added`

src/pretix/api/views/item.py
```diff
@@ -279,6 +280,57 @@ def perform_destroy(self, instance):
...
+            user=self.request.user,
+            auth=self.request.auth,
+            data=merge_dicts(self.request.data, {'id': serializer.instance.pk})
+        )
+
+    def perform_update(self, serializer):
+        serializer.save(event=self.request.event)
+        serializer.instance.item.log_action(
+            'pretix.event.item.item_program_times.changed',
```
wiffbi: You probably should add a translation to logdisplay.py.

src/pretix/api/views/item.py
```diff
@@ -279,6 +280,57 @@ def perform_destroy(self, instance):
...
+            user=self.request.user,
+            auth=self.request.auth,
+            data=merge_dicts(self.request.data, {'id': serializer.instance.pk})
+        )
+
+    def perform_destroy(self, instance):
+        super().perform_destroy(instance)
+        instance.item.log_action(
+            'pretix.event.item.item_program_times.removed',
```
wiffbi: You probably should add a translation to logdisplay.py.

src/pretix/base/models/items.py
```diff
@@ -2294,3 +2294,24 @@ class ItemVariationMetaValue(LoggedModel):
...
+    This model can be used to add a program time to an item.
+
+    :param item: The item the program time applies to
+    :type item: Item
+    :param start: The date and time this program time starts
+    :type start: datetime
+    :param end: The date and time this program time ends
+    :type end: datetime
+    """
+    item = models.ForeignKey('Item', related_name='item_program_times', on_delete=models.CASCADE)
```
wiffbi: I don’t think we need the `item_` prefix here. Of course you need to replace that in several places.

```suggestion
    item = models.ForeignKey('Item', related_name='program_times', on_delete=models.CASCADE)
```

src/pretix/api/serializers/item.py
```diff
@@ -398,6 +425,8 @@ def create(self, validated_data):
             ItemAddOn.objects.create(base_item=item, **addon_data)
         for bundle_data in bundles_data:
             ItemBundle.objects.create(base_item=item, **bundle_data)
+        for program_times_data in program_times_data:
+            ItemProgramTime.objects.create(item=item, **program_times_data)
```
wiffbi: It’s not wrong, but IMHO unusual to use the same name for the local var as the looped over var, so I would change it according to the scheme from bundles and addons:

```suggestion
        for program_time_data in program_times_data:
            ItemProgramTime.objects.create(item=item, **program_time_data)
```

src/pretix/api/urls.py
```diff
@@ -111,6 +111,7 @@
 item_router.register(r'variations', item.ItemVariationViewSet)
 item_router.register(r'addons', item.ItemAddOnViewSet)
 item_router.register(r'bundles', item.ItemBundleViewSet)
+item_router.register(r'programtimes', item.ItemProgramTimeViewSet)
```
wiffbi: As the key in the JSON-object is `program_times`, I would use the same name in the URL. We have different styles with checkinlists, cartpositions, but on the other hand scheduled_exports, item_meta_properties, so I guess there is no right or wrong here.

```suggestion
item_router.register(r'program_times', item.ItemProgramTimeViewSet)
```

src/pretix/base/models/event.py
```diff
@@ -990,6 +990,11 @@ def copy_data_from(self, other, skip_meta_data=False):
                 ia.bundled_variation = variation_map[ia.bundled_variation.pk]
             ia.save(force_insert=True)
 
+        for ipt in ItemProgramTime.objects.filter(item__event=other).prefetch_related('item'):
+            ipt.pk = None
+            ipt.item = item_map[ipt.item.pk]
```
wiffbi, line 995: I know ItemBundle and ItemAddOn does it similarly, but I wonder, whether we could get away with not prefetching the related item und just do:

```suggestion
            ipt.item = item_map[ipt.item_id]
```

src/pretix/base/pdf.py
```diff
@@ -734,6 +740,20 @@ def get_seat(op: OrderPosition):
     return None
 
 
+def get_program_times(op: OrderPosition, ev: Event):
+    program_times = op.item.item_program_times.all()  # todo: get this to work with addons as well
```
wiffbi: Is that todo still applicable? AFAIK addons are their own orderposition (same with bundles) so no need for that todo?

src/pretix/base/pdf.py
```diff
@@ -734,6 +740,20 @@ def get_seat(op: OrderPosition):
...
+    pt_list = []
+    pt_str = ''
+    if program_times:
+        for pt in program_times:
+            pt_list.append(str(_('{start_datetime} – {end_datetime}')).format(
+                start_datetime=date_format(pt.start.astimezone(ev.timezone), 'SHORT_DATETIME_FORMAT'),
+                end_datetime=date_format(pt.end.astimezone(ev.timezone), 'SHORT_DATETIME_FORMAT')
+            ))
```
wiffbi: Hm, what about pretix.helpers.daterange.datetimerange? Might be more consistent what we have already. But might create differently looking results in one list.

src/pretix/control/forms/item.py
```diff
@@ -1321,3 +1325,76 @@ class Meta:
...
+            if self.can_delete:
+                if self._should_delete_form(form):
+                    # This form is going to be deleted so any of its errors
+                    # should not cause the entire formset to be invalid.
+                    try:
+                        dts.remove(form.cleaned_data['start'])
+                        dts.remove(form.cleaned_data['end'])
+                    except KeyError:
+                        pass
+                    continue
```
wiffbi: I am not sure I understand this code, how can I ever remove something from an always empty set? Maybe you need to add the combination of `(start, end)` into dts if the form does not need to be deleted and check if that already exists? I guess it is helpful to not be able to define the same program times twice?

src/pretix/api/serializers/item.py
ser8phin: this wasn't even intended :laughing: nice to know it works, but i didn't even want to use the same var name, so thanks for noticing

src/pretix/base/pdf.py
ser8phin:
```suggestion
    program_times = op.item.item_program_times.all()
```
nope, that had been solved, the todo comment was just leftover m)

src/pretix/control/forms/item.py
ser8phin: Removed for now, for now we do not check for overlapping program times (might add, tbd)

src/pretix/base/models/event.py
ser8phin: I don't think so, because we have different items in different events here, so this might then reference the wrong one. But now I'm not quite sure anymore.

src/pretix/base/pdf.py
ser8phin: It does create different looking results, but in a good way, if the day doesn't change, it isn't repeated, but is, if it changes. Example in the editor now shows both.

src/pretix/base/models/event.py
wiffbi: It should be the same, but just ignore it as the similar code-paths for addons, etc. do the same. And copying events isn’t a hot code-path anyways.

src/pretix/control/views/item.py
```diff
@@ -1497,6 +1506,11 @@ def form_valid(self, form):
                     'bundles', 'bundles', 'base_item', order=False,
                     serializer=ItemBundleSerializer
                 )
+            elif k == 'program_times':
+                self.save_formset(
+                    'program_times', 'program_times', 'item', order=False,
+                    serializer=ItemProgramTimeSerializer
+                )
```
wiffbi, line 1506: Instead of invalidating the cache in save_formset (which is used for several formsets here), I would just check this formset whether any form in there is either deleted and its instance has a pk or has_changed(). Also checking for change_data makes sure, we do not invalidate_cache twice, when the cache was invalidated anyways due to changes on the item.

```suggestion
                )
                if not change_data:
                    for form in v.forms:
                        if (form in v.deleted_forms and form.instance.pk) or form.has_changed():
                            invalidate_cache.apply_async(kwargs={'event': self.request.event.pk, 'item': self.object.pk})
                            break
```

src/pretix/control/views/item.py
```diff
@@ -1442,6 +1446,11 @@ def save_formset(self, key, log_base, attr='item', order=True, serializer=None,
                     'pretix.event.item.{}.added'.format(log_base),
                     user=self.request.user, data=change_data
                 )
+                if key == 'program_times':
+                    program_times_changed = True
+        if program_times_changed:
+            # a change in program times warrants rebuilding tickets
+            invalidate_cache.apply_async(kwargs={'event': self.request.event.pk, 'item': self.object.pk})
```
wiffbi: See suggestion below
```suggestion
```

src/pretix/control/views/item.py
```diff
@@ -1413,6 +1415,8 @@ def save_formset(self, key, log_base, attr='item', order=True, serializer=None,
             self.get_object().log_action(
                 'pretix.event.item.{}.{}'.format(log_base, rm_verb), user=self.request.user, data=d
             )
+            if key == 'program_times':
+                program_times_changed = True
```
wiffbi: See suggestion below
```suggestion
```

src/pretix/control/views/item.py
```diff
@@ -1402,6 +1403,7 @@ def post(self, request, *args, **kwargs):
 
     def save_formset(self, key, log_base, attr='item', order=True, serializer=None,
                      rm_verb='removed'):
+        program_times_changed = False
```
wiffbi: See suggestion below
```suggestion
```

src/pretix/control/views/item.py
ser8phin: this was my first idea, that I threw away after thinking about it too much, but you actually made it work (and better). thank you!

src/pretix/presale/ical.py
```diff
@@ -122,61 +124,110 @@ def get_private_icals(event, positions):
...
+
+    # collecting the positions' calendar entries, preferring the most exact date and time available (positions > subevent > event)
+    prefetch_related_objects(positions, 'item__program_times')
+    for p in positions:
+        program_times = p.item.program_times.all()
+        if program_times:
+            # if program times have been configured, they are preferred for the position's calendar entries
             url = build_absolute_uri(event, 'presale:event.index')
+            for index, pt in enumerate(program_times):
+                summary = _('{event} - {item}').format(event=event, item=p.item.name)
```
wiffbi: I think we want use the subevent if available?

```suggestion
                summary = _('{event} - {item}').format(event=ev, item=p.item.name)
```

src/pretix/presale/ical.py
```diff
@@ -122,61 +124,110 @@ def get_private_icals(event, positions):
...
+                if event.settings.mail_attach_ical_description:
+                    ctx = get_email_context(event=event, event_or_subevent=event)
```
wiffbi:
```suggestion
                    ctx = get_email_context(event=event, event_or_subevent=ev)
```

src/pretix/presale/ical.py
```diff
@@ -122,61 +124,110 @@ def get_private_icals(event, positions):
...
-    evs = set(p.subevent or event for p in positions)
-    for ev in evs:
-        if isinstance(ev, Event):
+    calentries = set()  # using set for automatic deduplication of CalEntries
+    CalEntry = namedtuple('CalEntry', ['summary', 'description', 'location', 'dtstart', 'dtend', 'uid'])
+
+    # collecting the positions' calendar entries, preferring the most exact date and time available (positions > subevent > event)
+    prefetch_related_objects(positions, 'item__program_times')
+    for p in positions:
+        program_times = p.item.program_times.all()
```
wiffbi, line 134: There are several places where we need the subevent, if any. So previously we had `ev`, which I think is harder to remember than necessary, that `ev` can be a subevent as well. So:
```suggestion
        ev = p.subevent or event
        program_times = p.item.program_times.all()
```

src/pretix/presale/ical.py
```diff
@@ -122,61 +124,110 @@ def get_private_icals(event, positions):
...
             url = build_absolute_uri(event, 'presale:event.index')
+            for index, pt in enumerate(program_times):
+                summary = _('{event} - {item}').format(event=event, item=p.item.name)
+                if event.settings.mail_attach_ical_description:
+                    ctx = get_email_context(event=event, event_or_subevent=event)
+                    description = format_map(str(event.settings.mail_attach_ical_description), ctx)
+                else:
+                    # Default description
+                    descr = []
+                    descr.append(_('{event} - {item}: {url}').format(event=event, item=p.item.name, url=url))
```
wiffbi:
```suggestion
                    descr.append(_('{event} - {item}: {url}').format(event=ev, item=p.item.name, url=url))
```

src/pretix/presale/ical.py
```diff
@@ -122,61 +124,110 @@ def get_private_icals(event, positions):
...
+                ev = p.subevent
+                url = build_absolute_uri(event, 'presale:event.index', {
+                    'subevent': p.subevent.pk
+                })
+            else:
+                ev = event
+                url = build_absolute_uri(event, 'presale:event.index')
 
+            if event.settings.mail_attach_ical_description:
+                ctx = get_email_context(event=event, event_or_subevent=p.subevent)
```
wiffbi:
```suggestion
                ctx = get_email_context(event=event, event_or_subevent=ev)
```

wiffbi (approved):

PR #5367, opened 2025-08-12
PDF: Add font fallback on a pragraph level (Z#23203886)

raphaelm:
reportlab unfortunately does not support TTF fonts to automatically fall back to other fonts for individual characters they can't render. However, we want rendering of non-Latin names to "just work" in our default reports. Therefore, this PR implements a paragraph-level flag that – whenever a text is not renderable in the default font – searches for *any* available font that can handle it. This PR applies this to all core PDF reports as well as invoices.

This requires registering all fonts with reportlab, but since font registry are thread-scoped, this should have no significant performance burden. For the actual fallback, I'm a little more concerned but in the places we use it, a little performance downside is acceptible.

### DIFF
--- a/src/pretix/base/invoice.py
+++ b/src/pretix/base/invoice.py
@@ -48,7 +48,7 @@ from reportlab.pdfbase.ttfonts import TTFont
 from reportlab.pdfgen.canvas import Canvas
 from reportlab.platypus import (
     BaseDocTemplate, Flowable, Frame, KeepTogether, NextPageTemplate,
-    PageTemplate, Paragraph, Spacer, Table, TableStyle,
+    PageTemplate, Spacer, Table, TableStyle,
 )
 
 from pretix.base.decimal import round_decimal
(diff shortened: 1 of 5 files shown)


src/pretix/helpers/reportlab.py
```diff
@@ -59,3 +68,35 @@ def _read_image(self, fp):
...
+class FontFallbackParagraph(Paragraph):
+    def __init__(self, text, style=None, *args, **kwargs):
+        if style is None:
+            style = ParagraphStyle(name='paragraphImplicitDefaultStyle')
+
+        if not self._font_supports_text(text, style.fontName):
+            newFont = self._find_font(text, style.fontName)
+            if newFont:
+                logger.debug(f"replacing {style.fontName} with {newFont} for {text!r}")
+                style = style.clone(name=style.name + '_' + newFont, fontName=newFont)
```
wiffbi, line 82: What if instead of trying to set the whole text in one font, we take `text` and replace any character, that is not supported with `<font name="{nextBestFontThatSupportsThatCharacter}">{c}</font>`? That way we could support e.g. mixed names with Latin and Non-Latin chars.

src/pretix/base/invoice.py
```diff
@@ -48,15 +48,17 @@
...
 )
 
 from pretix.base.decimal import round_decimal
 from pretix.base.models import Event, Invoice, Order, OrderPayment
 from pretix.base.services.currencies import SOURCE_NAMES
 from pretix.base.signals import register_invoice_renderers
 from pretix.base.templatetags.money import money_filter
-from pretix.helpers.reportlab import ThumbnailingImageReader, reshaper
+from pretix.helpers.reportlab import (
+    FontFallbackParagraph, ThumbnailingImageReader, reshaper,
```
wiffbi, line 60: Wouldn’t it jhave been easier to just import FontFallbackParagraph as Paragraph? Or is there some place where we do not use FontFallbackParagraph and want to use Paragraph explicitly?

```suggestion
    FontFallbackParagraph as Paragraph, ThumbnailingImageReader, reshaper,
```

PR #5374, opened 2025-08-14
Subevents: set inactive if non-batch deletion of subevent fails (Z#23204183)

raphaelm:


src/pretix/control/views/subevents.py
```diff
@@ -184,11 +184,25 @@ def delete(self, request, *args, **kwargs):
...
+                self.object.delete()
+                messages.success(request, pgettext_lazy('subevent', 'The selected date has been deleted.'))
+            except ProtectedError:
+                self.object.log_action(
+                    'pretix.subevent.changed', user=self.request.user, data={
+                        'active': False
+                    }, save=False,
+                )
+                self.object.active = False
+                self.object.save(update_fields=['active'])
```
wiffbi: Should we set active=False, only if subevent is active? Not sure, if it is helpful to have the log_action here, if actually nothing changed on the object. On the other hand, it might actually be helpful as an indicator of a failed attempt to delete. Hm, I am undecided.

raphaelm: yes, fair enough

src/pretix/control/views/subevents.py
```diff
@@ -184,11 +184,26 @@ def delete(self, request, *args, **kwargs):
...
+                CartPosition.objects.filter(addon_to__subevent=self.object).delete()
+                self.object.cartposition_set.all().delete()
+                self.object.delete()
+                messages.success(request, pgettext_lazy('subevent', 'The selected date has been deleted.'))
+            except ProtectedError:
+                if self.object.active:
+                    self.object.log_action(
+                        'pretix.subevent.changed', user=self.request.user, data={
+                            'active': False
+                        }, save=False,
```
wiffbi: I guess this was left in there from copying the change-log_action from the bulk_creation.
```suggestion
                        },
```
Or if we decide to log the failed attempt to delete the subevent, we could batch-create two logs: `pretix.subevent.changed` from above and before that a new `pretix.subevent.delete_failed`? I am fine with leaving it like you suggested though and just have both logs deleted and changed in there – not immediately understandable, but might be good enough.

src/pretix/control/views/subevents.py
```diff
@@ -184,11 +184,26 @@ def delete(self, request, *args, **kwargs):
...
-            CartPosition.objects.filter(addon_to__subevent=self.object).delete()
-            self.object.cartposition_set.all().delete()
-            self.object.delete()
-            messages.success(request, pgettext_lazy('subevent', 'The selected date has been deleted.'))
+            try:
+                self.object.log_action('pretix.subevent.deleted', user=self.request.user)
+                CartPosition.objects.filter(addon_to__subevent=self.object).delete()
+                self.object.cartposition_set.all().delete()
+                self.object.delete()
+                messages.success(request, pgettext_lazy('subevent', 'The selected date has been deleted.'))
```
wiffbi: Hm, it feels a bit strange to me to have a deleted-log-entry shown to me when editing a failed-to-be-deleted subevent. I know it means, that someone attempted to delete the subevent, but still feels weird. Also on the eventseries-dashboard, the log-entry shows up right before an changed-log for the same subevent. I think we should log the delete only if it succeeds – which means, we should probably add a delete_failed-log as described above though. So:

```suggestion
                CartPosition.objects.filter(addon_to__subevent=self.object).delete()
                self.object.cartposition_set.all().delete()
                self.object.delete()
                self.object.log_action('pretix.subevent.deleted', user=self.request.user)
                messages.success(request, pgettext_lazy('subevent', 'The selected date has been deleted.'))
```

wiffbi (changes requested):

src/pretix/control/views/subevents.py
raphaelm: ~I don't think we need a delete_Failed, but reversing the order makes total sense~ is what I thought, but "delete first, log second" is not possible because then self.object no longer has an ID that the log can be stored with

wiffbi: Oh, makes total sense. Never mind, then it’s fine like it is!

PR #5396, opened 2025-08-20
Organizer plugins: Do not show plugins as active if they are inactive on org-level

luelista:


src/pretix/control/templates/pretixcontrol/organizers/plugins.html
```diff
@@ -69,34 +69,36 @@ <h4>
...
-                                                            Active ({{ count }} events)
-                                                        {% endblocktrans %}
-                                                    </span>
+                                                {% if is_active or plugin.level == "event" %}
+                                                    {% if plugin.level == "organizer" %}
+                                                        <span class="label label-success" data-is-active>
+                                                            <span class="fa fa-check" aria-hidden="true"></span>
+                                                            {% trans "Active" %}
+                                                        </span>
+                                                    {% elif events_counter == events_total %}
```
wiffbi: This is not really part of what the PR wants to change, but if we have an organizer with no events, doesn’t this show event-level plugins active? So shouldn’t we add either a check for events_counter or events_total? Probably the latter better illustrates the itent as in „only if there are events“, so:

```suggestion
                                                    {% elif events_total and events_counter == events_total %}
```

raphaelm: Agreed

PR #5401, opened 2025-08-22
asynctask.js: Fix gettext being used before translations are loaded (Z#23204825)

raphaelm:


src/pretix/static/pretixbase/js/asynctask.js
```diff
@@ -73,12 +74,12 @@ function async_task_check_callback(data, textStatus, jqXHR) {
 function async_task_update_status(data) {
     if (async_task_is_long) {
         if (data.started) {
-            waitingDialog.setStatus(async_task_status_messages.long_task_started);
+            waitingDialog.setStatus(async_task_status_messages.long_task_started());
```
wiffbi, line 77: As the strings in `async_task_status_messages` are each only used in one place once, wouldn’t it be easier to move the whole `gettext()` call with the string into the `waitingDialog.setStatus` and remove the global `var async_task_status_messages` completely?

```suggestion
            waitingDialog.setStatus(gettext(
                'Your request is currently being processed. Depending on the size of your event, this might take up to ' +
                'a few minutes.'
            ));
```

wiffbi (approved):
Looks good, just an idea to remove the whole global `var async_task_status_messages` and use `gettext`-call directly in `waitingDialog.setStatus`.

src/pretix/static/pretixbase/js/asynctask.js
raphaelm: Or, alternatively, move all the strings in the variable for a little bit cleaner code. I'm undecided and would go just for the fix for now

PR #5407, opened 2025-08-26
Fix incorrect placement of background when merging PDFs

TheEnbyperor:
PDFs have this neat/annoying (delete as appropriate) feature where the origin point of content for each page can be adjusted by one of several boxes, e.g. the media box.

Pretix uses `pypdf` to merge backgrounds into generated PDFs for ticket output. `pypdf`'s docs states than when merging pages "the mediabox, cropbox, etc of this page are not altered"; [1] thus Pretix needs to handle these itself.

All pages Pretix generates have content boxes at the `(0,0)` origin, but not all source PDFs for the background will have such origins. Without correct handling, the background is misplaced, resulting in elements being in the wrong location, white bars appearing on coloured backgrounds, and some of the backgrounds being cut off; e.g.:
Bildschirmfoto 2025-08-26 um 11 17 52

[1]: https://pypdf.readthedocs.io/en/stable/modules/PageObject.html

### DIFF
--- a/src/pretix/base/pdf.py
+++ b/src/pretix/base/pdf.py
@@ -1189,26 +1189,7 @@ class Renderer:
 
             for i, page in enumerate(fg_pdf.pages):
                 bg_page = self.bg_pdf.pages[i]
-                if bg_page.rotation != 0:
-                    bg_page.transfer_rotation_to_content()
-                bg_media_box = bg_page.mediabox
-                trsf = pypdf.Transformation()
-                if bg_media_box.bottom != 0:
-                    trsf = trsf.translate(0, -bg_media_box.bottom)
-                if bg_media_box.left != 0:
-                    trsf = trsf.translate(-bg_media_box.left, 0)
-                bg_page.add_transformation(trsf, False)
-                for b in ["/MediaBox", "/CropBox", "/BleedBox", "/TrimBox", "/ArtBox"]:
-                    if b in bg_page:
-                        rr = pypdf.generic.RectangleObject(bg_page[b])
-                        pt1 = trsf.apply_on(rr.lower_left)
-                        pt2 = trsf.apply_on(rr.upper_right)
-                        bg_page[pypdf.generic.NameObject(b)] = pypdf.generic.RectangleObject((
-                            min(pt1[0], pt2[0]),
-                            min(pt1[1], pt2[1]),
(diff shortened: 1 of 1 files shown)


raphaelm:
Internal note: if we merge this, someone has to do it for libpretixprint 👀

TheEnbyperor:
Given the way libpretixprint inputs the background - by creating a new document from the background and drawing on top of that, rather than merging [1]; I don't believe this is an issue with libpretixprint. Testing required though.

[1]: https://github.com/pretix/libpretixprint/blob/8943abe1a005e712c8c1b7d1633edc0e10b33b78/libpretixprint/src/main/java/eu/pretix/libpretixprint/templating/Layout.java#L424

raphaelm:
Interesting. @wiffbi can you weigh in of this will affect the work you did with mediaboxes on the badges n-up module?

raphaelm:
@TheEnbyperor have you tried using the pdftk implementation path to see if its currently doing a different or the same thing? We use it in prod for better performance

TheEnbyperor:
No, pdftk is not affected by this; presumably it has similar logic internally to handle offset media boxes.

wiffbi:
> Interesting. @wiffbi can you weigh in of this will affect the work you did with mediaboxes on the badges n-up module?

Generally badges need the same adjustment: as `render_background` and method below `merge_background` share very similar code, the latter which is used in badge-export will need the same adjustments.

The actual n-up in `plugins.badges.exporters._render_nup_page` should not be affected as it uses the mediabox of the finished page, where the background is already correctly placed.

TLDR; n-up is fine, but `merge_background` (right below `render_background`) needs the same adjustments. Could you add those as well @TheEnbyperor ?

TheEnbyperor:
> Could you add those as well @TheEnbyperor?

Will do.

TheEnbyperor:
I've refactored the correction code out to its own function, and applied it to the `merge_background` function as well.

wiffbi:
Attached small PDF to test this manually – the background is red, the rectangleobject is offset by A6 and the mediabox is then transformed by A6 to make the background fullpage again.

[bg-test-mediabox-offset-by-a6.pdf](https://github.com/user-attachments/files/22170154/bg-test-mediabox-offset-by-a6.pdf)

PR #5409, opened 2025-08-27
Detect empty data export

dannyadair:
The csv/xlsx renderers never return None so EmptyExportError was never raised. Detect whether data consists only of the header.

Note: This will lead to scheduled exports no longer sending an email when there is no data.

### DIFF
--- a/src/pretix/base/exporter.py
+++ b/src/pretix/base/exporter.py
@@ -217,6 +217,7 @@ class ListExporter(BaseExporter):
             writer = csv.writer(output_file, **kwargs)
             total = 0
             counter = 0
+            rows_written = 0
             for line in self.iterate_list(form_data):
                 if isinstance(line, self.ProgressSetTotal):
                     total = line.total
@@ -230,12 +231,18 @@ class ListExporter(BaseExporter):
                     if counter % max(10, total // 100) == 0:
                         self.progress_callback(counter / total * 100)
                 writer.writerow(line)
+                rows_written += 1
+            
+            if rows_written <= 1:  # Only header or no rows
+                return None
+            
             return self.get_filename() + '.csv', 'text/csv', None
         else:
             output = io.StringIO()
             writer = csv.writer(output, **kwargs)
             total = 0
             counter = 0
+            rows_written = 0
             for line in self.iterate_list(form_data):
                 if isinstance(line, self.ProgressSetTotal):
                     total = line.total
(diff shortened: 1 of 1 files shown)


dannyadair:
Emailed signed CLA

dannyadair:
No test yet for ExportEmptyError in https://github.com/pretix/pretix/blob/master/src/tests/control/test_export.py Sorry didn't add any in this PR, should probably confirm that this continues to work. (I see this PR more of a bugfix than a new feature)

dannyadair:
Sorry there seems to be one more thing to this:
https://github.com/pretix/pretix/blob/master/src/pretix/base/services/export.py#L230
```
        except ExportEmptyError as e:
            _handle_error(str(e), soft=True)
```

Even though it's soft, it will still send an "Export failed" email - in my humble opinion that should simply not happen.
I would just `pass` in this case but maybe that's where a checkbox for the user to decide could come in?

pc-coholic:
@cla-bot check

wiffbi:
I think this is a pragmatic approach to checking whether any results are in the export and is probably good enough. As a side note though: I have recently seen an – albeit custom – export that has two header-rows, so that check would fail there. IMHO not that bad in that case as at least the report is sent. I have not seen an export with no header-row, but if there is, it would fail the test here – but only on the edge-case of one (or of course no) result, which might come as a suprise to some users at some random point in the future. So I am a bit torn … It doesn’t feel like the „pure right way“ to do it, but the alternative isn’t really practical as AFAICT it would mean to not yield the header-row before we check if there is any result – shouldn’t be to hard to do for one export, but we have a lot of exports.

dannyadair:
> So I am a bit torn … It doesn’t feel like the „pure right way“ to do it, but the alternative isn’t really practical as AFAICT it would mean to not yield the header-row before we check if there is any result – shouldn’t be to hard to do for one export, but we have a lot of exports.

I was torn, too, but I couldn't find a nice central place to inject this. Headers and data should be separate to begin with but then this would quickly escalate.
That "1" for the check could be made a default argument, and somehow custom reports could indicate how many header rows they have and override. My starting point was that this never got detected at all.

raphaelm:
Yes, exports with no headers might exist, if an external system requires a specific fixed format with header rows, e.g. accounting systems :(

raphaelm:
We could have a

header_rows = 1

on the base class that subclasses can override?

dannyadair:
> We could have a
> 
> header_rows = 1
> 
> on the base class that subclasses can override?

Yeah that's what I meant
> somehow custom reports could indicate how many header rows they have and override

The people who can write custom reports can also adjust that :)

Sorry when I find more time I'm happy to help a bit better on something like this, like a proper PR that separates headers and data. Bit under the pump atm, stumbled over this as an end user. (problem see idea link)

raphaelm:
I think I'll close this then for now, feel free to reopen when you get the time :)

PR #5433, opened 2025-09-04
Model-based mail queuing

raphaelm:
This allows:

- Outbox view for organizers
- Future notification features about email failures
- Cleaner code in services/mail.py
- More reliability in sending if a celery task gets lost
- Rate-limit implementation for pretix Hosted

Todo list:

- [x] Data model
- [x] Code refactoring
- [x] Deprecate useless SendMailException
- [x] Outbox viewer
- [x] Manual retry mechanism
- [x] Decide on attachment handling
  - [x] Make source files persist if possible
- [x] Decide on if this replaces storing mail text in logentries
- [x] Decide on correct on_delete settings for ForeignKeys
  - [x] Tests
- [x] Data shredder integration
- [x] Cleanup of old data that is not part of an event and should still be shredded
- [x] Cleanup task for stuck inflight state
- [x] Bounce handling 
- [x] Compatibility of mail_send
- [x] Remove old usages of mail_send_task
  - [x] pretixeu
  - [x] pretix-vrt
- [x] Signals for pretixeu handling
  - [x] Rate limiting
- [x] view ins search and permission Tests
- [x] Rate limit state
- [x] rate Limit cost per recipient
- [x] Code comment und race condition
- [x] abort queue button

Post-Merge:
- [ ] Notify organizers about increase in failures

### DIFF
--- a/src/pretix/base/services/mail.py
+++ b/src/pretix/base/services/mail.py
@@ -749,7 +749,7 @@ def mail_send(to: List[str], subject: str, body: str, html: Optional[str], sende
     Low-level function to send mails, kept for backwards-compatibility. You should usually use mail() instead.
     """
     m = OutgoingMail.objects.create(
-        organizer_id=organizer.pk if isinstance(organizer, Organizer) else None,
+        organizer_id=organizer.pk if isinstance(organizer, Organizer) else organizer,
         event_id=event.pk if isinstance(event, Event) else event,
         order_id=order.pk if isinstance(order, Order) else order,
         orderposition_id=position.pk if isinstance(position, OrderPosition) else position,


src/pretix/base/models/mail.py
```diff
@@ -0,0 +1,222 @@
...
+
+from django.core.mail import get_connection
+from django.db import models
+from django.utils.translation import gettext_lazy as _
+from django_scopes import scope, scopes_disabled
+
+
+def CASCADE_IF_QUEUED(collector, field, sub_objs, using):
+    # If the email is still queued and the thing it is related to vanishes, the email can vanish as well
+    print(sub_objs)
```
pc-coholic:
```suggestion
```

src/pretix/base/models/mail.py
```diff
@@ -0,0 +1,222 @@
...
+
+    error = models.TextField(null=True, blank=True)
+    error_detail = models.TextField(null=True, blank=True)
+
+    # There is a conflict here between the different purposes of the model. As a system administrator,
+    # one wants *all* emails to be persisted as long as possible to debug issues. This means that if
+    # e.g. the event or order is deleted, we want SET_NULL behavior. However, in that case, the email
+    # would be an "orphan" forever and there's no way to remove the personal information.
+    # We try to find a middle-ground with the following behaviour:
+    # - The email is always deleted if the entire organize or user is deleted
```
pc-coholic:
```suggestion
    # - The email is always deleted if the entire organizer or user is deleted
```

src/pretix/base/services/mail.py
```diff
@@ -91,9 +97,18 @@ def __missing__(self, key):
 
 
 class SendMailException(Exception):
+    """
+    Deprecated, not thrown any more.
+    """
     pass
```
pc-coholic, line 103: Instead of just `pass`ing, should we log an error to make it more obvious if a plugin is still using this Exception that me might perhaps drop at some point in the future?

src/pretix/base/services/mail.py
```diff
@@ -866,3 +877,203 @@ def normalize_image_url(url):
...
+def _retry_strategy(e: Exception):
+    if isinstance(e, (smtplib.SMTPResponseException, smtplib.SMTPSenderRefused)):
+        if e.smtp_code == 432:
+            # This is likely Microsoft Exchange Online which has a pretty bad rate limit of max. 3 concurrent
+            # SMTP connections which is *easily* exceeded with many celery threads. Just retrying with exponential
+            # backoff won't be good enough if we have a lot of emails, instead we'll need to make sure our retry
+            # intervals scatter such that the email won't all be retried at the same time again and cause the
+            # same problem.
+            # See also https://docs.microsoft.com/en-us/exchange/troubleshoot/send-emails/smtp-submission-improvements
+            return "microsoft_concurrncy"
```
pc-coholic:
```suggestion
            return "microsoft_concurrency"
```

Unrelated - but another one of "this has never worked"

src/pretix/base/services/notifications.py
```diff
@@ -153,16 +157,26 @@ def send_notification_mail(notification: Notification, user: User):
...
+        headers={
+            'X-Auto-Response-Suppress': 'OOF, NRN, AutoReply, RN',
+            'Auto-Submitted': 'auto-generated',
+            'X-Mailer': 'pretix',
+            'X-PX-Correlation': str(guid),
+        },
+    )
+    mail_send_task.apply_async(kwargs={
+        'outgoing_mail': m.pk,
     })
```
pc-coholic, line 182: Can this not go through `pretix.base.services.mail.mai()`?

If not, at least the correlation-headers, etc. should already be set through the `mail_send_task`, I think...

pc-coholic (changes requested):

src/pretix/control/templates/pretixcontrol/organizers/outgoing_mails.html
```diff
@@ -0,0 +1,182 @@
...
+                                    {{ m.subject }}
+                                </a>
+                            </td>
+                            <td>
+                                {{ m.to|join:", " }}
+                                {% if m.cc %}
+                                    <br><small class="text-muted">{% trans "Cc" context "email" %}: {{ m.cc|join:", " }}</small>
+                                {% endif %}
+                                {% if m.bcc %}
+                                    <br><small class="text-muted">{% trans "Bcc" context "email" %} {{ m.bcc|join:", " }}</small>
```
pc-coholic:
```suggestion
                                    <br><small class="text-muted">{% trans "Bcc" context "email" %}: {{ m.bcc|join:", " }}</small>
```

src/pretix/base/services/mail.py
```diff
@@ -392,194 +367,239 @@ def _create_mime_attachment(self, content, mimetype):
...
+                        args.append((name, content, ct.type))
+                        attach_size += len(content)
+                    except Exception:
+                        # This sometimes fails e.g. with FileNotFoundError. We haven't been able to figure out
+                        # why (probably some race condition with ticket cache invalidation?), so retry later.
+                        try:
+                            self.retry(max_retries=5, countdown=60)
+                        except MaxRetriesExceededError:
+                            # Well then, something is really wrong, let's send it without attachment before we
+                            # don't sent at all
```
luelista:
```suggestion
                            # don't send at all
```

src/pretix/base/services/mail.py
raphaelm: I have never seen a plugin *throw* this exceptuon, only *catch* it, so that shouldn't make a difference?

src/pretix/base/services/mail.py
raphaelm: Na, this is new code only :)

src/pretix/base/services/notifications.py
raphaelm: > Can this not go through pretix.base.services.mail.mai()?

No, because mail() always renders the HTML from the text, wheras notifications are much richer in HTML and only have minimal text, rendered separately.

> If not, at least the correlation-headers, etc. should already be set through the mail_send_task, I think...

If they are set in mail_send_task, they do not show up in the outbox viewer. mail_send_task has a fallback to  set it, if missing, though.

src/pretix/control/templates/pretixcontrol/organizers/outgoing_mail.html
```diff
@@ -0,0 +1,222 @@
...
+                </div>
+            {% endif %}
+            {% if mail.body_html %}
+                <div role="tabpanel"
+                     class="tab-pane {% if not mail.is_failed %}active{% endif %}"
+                     id="tab-html">
+                    {% if mail.sensitive %}
+                        <div class="empty-collection">
+                            <p>
+                                {% icon "eye-slash 4x" %}
```
wiffbi: icon template tag does not add „fa-“ to every class in its param. It is probably kind of a hack to use a classlist as a param, but one that is okayish IMHO. Nonetheless, for t4x to work, you need to write:
```suggestion
                                {% icon "eye-slash fa-4x" %}
```

src/pretix/control/templates/pretixcontrol/organizers/outgoing_mail.html
```diff
@@ -0,0 +1,222 @@
...
+                    {% endif %}
+                </div>
+            {% endif %}
+            <div role="tabpanel"
+                 class="tab-pane {% if not mail.is_failed and not mail.body_html %}active{% endif %}"
+                 id="tab-text">
+                {% if mail.sensitive %}
+                    <div class="empty-collection">
+                        <p>
+                            {% icon "eye-slash 4x" %}
```
wiffbi:
```suggestion
                            {% icon "eye-slash fa-4x" %}
```

src/pretix/control/templates/pretixcontrol/organizers/outgoing_mails.html
```diff
@@ -0,0 +1,185 @@
...
+                                       class="batch-select-label"><input type="checkbox" name="outgoingmail"
+                                                                         class="batch-select-checkbox"
+                                                                         value="{{ m.pk }}"/></label>
+                            </td>
+                            <td>
+                                <a href="{% url "control:organizer.outgoingmail" organizer=request.organizer.slug mail=m.id %}">
+                                    {{ m.subject }}
+                                </a>
+                                {% if m.sensitive %}
+                                    <span class="text-muted">{% icon "eye-slash" %}</span>
```
wiffbi, line 100:
```suggestion
                                    {% icon "eye-slash text-muted" %}
```

src/pretix/base/services/mail.py
```diff
@@ -730,20 +741,20 @@ def mail_send_task(self, *args, outgoing_mail: int) -> bool:
...
     Low-level function to send mails, kept for backwards-compatibility. You should usually use mail() instead.
     """
     m = OutgoingMail.objects.create(
-        organizer=organizer,
-        event=event,
-        order=order,
-        orderposition=position,
-        customer=customer,
-        user=user,
+        organizer_id=organizer.pk if isinstance(organizer, Organizer) else None,
```
wiffbi: Why is it `None` instead of `organizer`? According to the params-list, organizer can be `int`.

src/pretix/control/templates/pretixcontrol/organizers/outgoing_mails.html
raphaelm: I didn't want to abuse the class list hack beyond classes defined by fontawesome, because it feld even dirtier then :see_no_evil:

src/pretix/base/services/mail.py
raphaelm: bad autocompletion…

PR #5436, opened 2025-09-05
Event calendar: Only show "waiting list" if products allow it (Z#23205941)

raphaelm:
Before:

image

After:

image

### DIFF
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -396,7 +396,6 @@ class EventMixin:
                     ),
                     has_active_variations_with_waitinglist=Exists(
                         sq_active_variation.filter(item__allow_waitinglist=True),
-                        output_field=models.TextField()
                     ),
                 ).exclude(
                     Q(active_items="") & Q(active_variations="")
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -393,7 +393,6 @@ class EventMixin:
                         output_field=models.TextField()),
                     has_active_items_with_waitinglist=Exists(
                         sq_active_item.filter(allow_waitinglist=True),
-                        output_field=models.TextField()
                     ),
                     has_active_variations_with_waitinglist=Exists(
                         sq_active_variation.filter(item__allow_waitinglist=True),
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -391,17 +391,12 @@ class EventMixin:
                             items=GroupConcat('pk', delimiter=',')
                         ).values('items'),
                         output_field=models.TextField()),
-                    active_items_with_waitinglist=Subquery(
-                        sq_active_item.filter(allow_waitinglist=True).order_by().values_list('quotas__pk').annotate(
-                            items=GroupConcat('pk', delimiter=',')
-                        ).values('items'),
+                    has_active_items_with_waitinglist=Exists(
+                        sq_active_item.filter(allow_waitinglist=True),
                         output_field=models.TextField()
                     ),
-                    active_variations_with_waitinglist=Subquery(
-                        sq_active_variation.filter(item__allow_waitinglist=True).order_by().values_list(
-                            'quotas__pk').annotate(
-                            items=GroupConcat('pk', delimiter=',')
-                        ).values('items'),
+                    has_active_variations_with_waitinglist=Exists(
+                        sq_active_variation.filter(item__allow_waitinglist=True),
                         output_field=models.TextField()
                     ),
                 ).exclude(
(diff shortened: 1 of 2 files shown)


src/pretix/base/models/event.py
```diff
@@ -481,7 +514,22 @@ def best_availability(self):
...
+                worst_state_for_ticket = min(r[q][0] for q in quota_list)
+                if worst_state_for_ticket < Quota.AVAILABILITY_OK and k in items_with_waiting_list:
+                    waiting_list_found = True
+                    break
+            else:
+                for k, quota_list in quotas_for_variation.items():
+                    worst_state_for_ticket = min(r[q][0] for q in quota_list)
+                    if worst_state_for_ticket < Quota.AVAILABILITY_OK and k in vars_with_waiting_list:
+                        waiting_list_found = True
+                        break
```
wiffbi: I am probably missing something here, but when the best availability state of all items and variations combined is already `< Quota.AVAILABILITY_OK`, then the worst_state of any individual item/variation will always be `< Quota.AVAILABILITY_OK`? So as every item/variation is not available, why check if it is in `items_with_waiting_list`  as we anyway only are interested if any item/variation has a waiting_list. So IMHO it should be enough to just check, if `items_with_waiting_list` and `vars_with_waiting_list` aren’t empty.

Maybe we could go even further as to not even fetch all items’/variations’ IDs with `GroupConcat('pk', delimiter=',')`, but only use a `When(…)` to mark a boolean, that at least one item/variation has the wiating_list enabled.

src/pretix/base/models/event.py
```diff
@@ -376,8 +380,30 @@ def annotated(cls, qs, channel, voucher=None):
...
+                        ).values('items'),
+                        output_field=models.TextField()
+                    ),
+                    active_variations_with_waitinglist=Subquery(
+                        sq_active_variation.filter(item__allow_waitinglist=True).order_by().values_list(
+                            'quotas__pk').annotate(
+                            items=GroupConcat('pk', delimiter=',')
+                        ).values('items'),
+                        output_field=models.TextField()
+                    ),
```
wiffbi: See below for comment regarding whether we actually need to fetch individual IDs instead of just a `When(…)` boolean to indicate that at least one item/variation has the waitinglist active.

Also maybe we should only run these potentially heavy DB-queries only for events, where the waitinglist is active at all.

src/pretix/base/models/event.py
```diff
@@ -481,7 +514,22 @@ def best_availability(self):
...
+                    waiting_list_found = True
+                    break
+            else:
+                for k, quota_list in quotas_for_variation.items():
+                    worst_state_for_ticket = min(r[q][0] for q in quota_list)
+                    if worst_state_for_ticket < Quota.AVAILABILITY_OK and k in vars_with_waiting_list:
+                        waiting_list_found = True
+                        break
+
+        return best_state_found, num_tickets_found, num_tickets_possible, waiting_list_found
```
wiffbi, line 508: Very nitkpicky: this always returns `waiting_list_found=False` for `best_state_found==Quota.AVAILABILITY_OK`. I know we do not need that info if the event is available, but it might be confusing lateron. Maybe instead return None, if we have not checked? And add a comment in the doc-string, that if `best_state_found==Quota.AVAILABILITY_OK` the returned value of `waiting_list_found` cannot be relied upon?

raphaelm: With the refactoring for the comment above, this is no longer a problem

src/pretix/base/models/event.py
raphaelm: Your logic is logicing, I think, yes.

src/pretix/base/models/event.py
raphaelm: > Also maybe we should only run these potentially heavy DB-queries only for events, where the waitinglist is active at all.

That would however, at least on the organizer-level, require yet another query to access the settings store.

src/pretix/base/models/event.py
```diff
@@ -376,8 +380,25 @@ def annotated(cls, qs, channel, voucher=None):
...
+                        output_field=models.TextField()
+                    ),
+                    active_variations=Subquery(
+                        sq_active_variation.order_by().values_list('quotas__pk').annotate(
+                            items=GroupConcat('pk', delimiter=',')
+                        ).values('items'),
+                        output_field=models.TextField()),
+                    has_active_items_with_waitinglist=Exists(
+                        sq_active_item.filter(allow_waitinglist=True),
+                        output_field=models.TextField()
```
wiffbi: What is the reasoning behind using an output_field other than the default BooleanField? Or maybe just an artefact from the previous solution using GroupConcat-Query?

```suggestion
```

src/pretix/base/models/event.py
```diff
@@ -376,8 +380,25 @@ def annotated(cls, qs, channel, voucher=None):
...
+                    ),
+                    has_active_variations_with_waitinglist=Exists(
+                        sq_active_variation.filter(item__allow_waitinglist=True),
+                        output_field=models.TextField()
```
wiffbi: See above
```suggestion
```

PR #5437, opened 2025-09-05
Fix waitingDialog being shown on browser history back

wiffbi:
fixes #5362

### DIFF
--- a/src/pretix/static/pretixbase/js/asynctask.js
+++ b/src/pretix/static/pretixbase/js/asynctask.js
@@ -42,8 +42,8 @@ function async_task_schedule_check(context, timeout) {
 
 function async_task_on_success(data) {
     "use strict";
+    waitingDialog.hide();
     if ((async_task_is_download && data.success) || async_task_dont_redirect) {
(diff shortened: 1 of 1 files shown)


luelista:
I'd like it better if we keep it visible during the redirect, which might be slow on high server load, and only hide it if the user returns back to the cached page. This should be possible using the [pageshow event](https://webkit.org/blog/516/webkit-page-cache-ii-the-unload-event/):

```js
function async_task_on_success(data) {
    "use strict";
    if ((async_task_is_download && data.success) || async_task_dont_redirect) {
        waitingDialog.hide();
        if (location.href.indexOf("async_id") !== -1) {
            history.replaceState({}, "pretix", async_task_old_url);
        }
    }
    if (!async_task_dont_redirect) {
        $(window).one("pageshow", function (e) {
            waitingDialog.hide();
        });
        location.href = data.redirect;
    }
    $(this).trigger('pretix:async-task-success', data);
}
```

wiffbi:
Thanks for the suggestion, works wonderfully!

PR #5445, opened 2025-09-08
Invoicing: Configurable service date

raphaelm:
In many countries, such as Germany, the "date of delivery or service execution" is a mandatory element of an invoice (§ 14 UStG). For pretix, this used to be not so much of a problem, because the date of service execution for an event is quite trivially the event date which is printed on top of invoices by default. However, there are two issues:

- We need the date to be an explicit and immutable part of the invoice in order to consistently render it in non-PDF formats such as ZUGFeRD XML or Peppol BIS
- We need to account for shops where the event date is not the service date, such as when tickets are issued with the ``vaild_from``/``valid_until`` mechanism or non-ticket products are sold. pretix-zugferd already contains a custom toggle that will now no longer be needed.

pretix already contains ``InvoiceLine.event_date_from/event_date_to``, which we are hereby reusing to make sense of the old data, but which get new semantics when e.g. tickets with a validity timeframe are involved.

- [x] Data model change
- [x] API compatibility
- [x] Set dates in build_invoice correcly
- [x] Render on invoices
- [x] Tests
- [x] Update pretix-zugferd
- [x] Warn users that will be affected by the migration of pretix-zugferd who might not even be aware of their current settings because they got auto-migrated


src/pretix/base/invoicing/pdf.py
```diff
@@ -657,6 +664,12 @@ def _get_intro(self):
 
     def _get_story(self, doc):
         has_taxes = any(il.tax_value for il in self.invoice.lines.all()) or self.invoice.reverse_charge
+        header_dates = self._date_range_in_header()
+        tz = self.invoice.event.timezone
+        has_multiple_service_dates = len(set(
+            (il.period_start, il.period_end) for il in self.invoice.lines.all()
```
wiffbi: Doesn’t `all()` here and in line 726 trigger two DB-calls with potentially different results? Wouldn’t it make sense to put the returned queryset from `self.invoice.lines.all()` into a variable and re-use that in line 726 to make use of djangos caching? Alternatively we could filter for uniques on the database-level, but that is probably overkill and has the same problem regarding differing result (or maybe I missed where we lock the DB rows here).

doc/api/resources/invoices.rst
```diff
@@ -80,17 +80,12 @@ lines                                 list of objects            The actual invo
...
-                                                                 cancellation fees).
-├ event_date_to                       datetime                   End date of the (sub)event this line was created for as it
-                                                                 was set during invoice creation. Can be ``null`` for all invoice
-                                                                 lines created before this was introduced as well as for lines in
-                                                                 an event series not created by a product (e.g. shipping or
-                                                                 cancellation fees) as well as whenever the respective (sub)event
-                                                                 has no end date set.
+├ period_start                        datetime                   Start date of the service or delivery period of the invoice line.
+                                                                 Can be ``null`` if not known.
+├ period_end                        datetime                     End date of the service or delivery period of the invoice line.
```
wiffbi:
```suggestion
├ period_end                          datetime                   End date of the service or delivery period of the invoice line.
```

src/pretix/base/invoicing/pdf.py
```diff
@@ -700,15 +713,73 @@ def _get_story(self, doc):
...
             self.invoice.lines.all(),
             key=_group_key,
             is_addon=lambda l: l.description.startswith("  +"),
         ):
+            # Try to be clever and figure out when organizers would want to show the period. This heuristic is
+            # not perfect and the only "fully correct" way would be to include the period on every line always,
+            # however this will cause confusion (a) due to useless repetition of the same date all over the invoice
+            # (b) due to not respecting the show_date_to setting of events in cases where we could have respected it.
+            # Still, we want to show the date explicitly if its different to the event or invoice date.
+            if period_start and period_end and day(period_end) != day(period_start):
```
wiffbi: Just a refactoring idea as this code often calls `day()` on `period_start` and `period_end`. How about:

```suggestion
            period_start_day = day(period_start)
            period_end_day = day(period_end)
            if period_start_day and period_end_day and period_start_day != period_end_day:
```

Of course, one needs to adapt the other places this is used in.

src/pretix/base/invoicing/pdf.py
```diff
@@ -700,15 +713,73 @@ def _get_story(self, doc):
...
+                    period_line = ""
+
+                else:
+                    period_line = f"{date_format(day(period_start), 'SHORT_DATE_FORMAT')} – {date_format(day(period_end), 'SHORT_DATE_FORMAT')}"
+
+            elif period_start or period_end:
+                # It's a single-day period
+
+                delivery_day = day(period_end or period_start)
+                if delivery_day in (header_dates[0], header_dates[1]):
```
wiffbi:
```suggestion
                if delivery_day in header_dates:
```

src/pretix/base/invoicing/pdf.py
```diff
@@ -700,15 +713,73 @@ def _get_story(self, doc):
...
+                    # This is the event date we already printed in the header, no need to repeat it.
+                    period_line = ""
+
+                elif self.event.has_subevents and subevent and delivery_day in (day(subevent.date_from), day(subevent.date_to)):
```
wiffbi, line 790: What if delivery_day is in between both dates? Then the date would be printed. But what makes that date different from when it matches e.g. the date_from? As it is potentially a daterange, matching either one of the subevent, does not make it more clear what the delivery_date of the line is.

I know the heuristic is not perfect as described in the comment at the top, just wanted to make sure, we are aware of that edge-case.

src/pretix/base/invoicing/pdf.py
```diff
@@ -700,15 +713,73 @@ def _get_story(self, doc):
...
+                    period_line = date_format(delivery_day, 'SHORT_DATE_FORMAT')
+            else:
+                # No period known
+                period_line = ""
+
+            if not has_multiple_service_dates and period_line:
+                # Group together at the end of the invoice
+                request_show_service_date = period_line
+            else:
+                description += "\n" + period_line
```
wiffbi, line 813: If no period is known, then this adds an empty line. Otherwise there is no empty line between period_line and single_price_line (if this is set at all). Is that on purpose?

src/pretix/base/services/invoices.py
```diff
@@ -351,6 +377,49 @@ def build_cancellation(invoice: Invoice):
     return invoice
 
 
+def _service_period_for_position(invoice, position, invoice_dt):
+    if invoice.event.settings.invoice_period in ("auto", "auto_no_event"):
+        if position.valid_from or position.valid_until:
+            period_start = position.valid_from or now()
```
wiffbi: Could it happen that there is a position with a `valid_from == None`, which then results in `valid_from` being `now()`, but having a valid_until in the past? I haven’t checked it thorougly, but this would result in strange dateranges, I guess?

src/pretix/base/invoicing/pdf.py
raphaelm: Invoices shouldn't change once created, but yes, it makes sense to move this into one query

src/pretix/base/invoicing/pdf.py
raphaelm: It's a weird edge case and will likely never happen. But suppose these situations:

- `Subevent(date_from=Monday, date_to=Wednesday)` Ticket with no special properties will have `delivery_date = Wednesday` and will show no delivery date because the invoice already shows "Monday – Wednesday" :heavy_check_mark: 
- `Subevent(date_from=Monday, date_to=None)` Ticket with no special properties will have `delivery_date = Monday` and will show no delivery date because the invoice already shows "Monday" :heavy_check_mark: 
- `Subevent(date_from=Monday, date_to=None)` Ticket with `valid_from=Monday, valid_until=Monday` will have `delivery_date = Monday` and will show no delivery date because the invoice already shows "Monday" :heavy_check_mark: 
- `Subevent(date_from=Monday, date_to=Tuesday)` Ticket with `valid_from=Tuesday, valid_until=Tuesday` will have `delivery_date = Tuesday` and will show a delivery date because it differs from what the invoice already shows :heavy_check_mark: 

I think that makes sense

src/pretix/base/invoicing/pdf.py
raphaelm: No, thanks

src/pretix/base/services/invoices.py
raphaelm: yes, we need to handle these cases explicitly

src/pretix/base/services/invoices.py
```diff
@@ -351,6 +380,55 @@ def build_cancellation(invoice: Invoice):
...
 
 
+def _service_period_for_position(invoice, position, invoice_dt):
+    if invoice.event.settings.invoice_period in ("auto", "auto_no_event"):
+        if position.valid_from and position.valid_until:
+            period_start = position.valid_from
+            period_end = position.valid_until
+        elif position.valid_from:
+            period_start = position.valid_from
+            period_end = position.valid_from  # weird, but we have nothing else ot base this on
```
wiffbi:
```suggestion
            period_end = position.valid_from  # weird, but we have nothing else to base this on
```

src/pretix/base/migrations/0289_invoiceline_period.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def store():
+        EventSettingsStore.objects.bulk_create(
+            insert_queue,
+            update_conflicts=True,
+            update_fields=["value"],
+            unique_fields=["object", "key"],
+        )
+        for f in flush_queue:
+            cache.delete(f)
```
wiffbi: Not sure about the underlying optimization, but maybe we should use this?

```suggestion
        cache.delete_many(flush_queue)
```

src/pretix/base/migrations/0289_invoiceline_period.py
```diff
@@ -0,0 +1,80 @@
...
+        )
+        ev_seen.add(setting.object_id)
+
+        if len(insert_queue) > 1000:
+            store()
+
+    # Existing events that previously hid their date on invoices
+    for setting in EventSettingsStore.objects.filter(key="show_dates_on_frontpage", value="False"):
+        if setting.object_id in ev_seen:
+            continue
```
wiffbi, line 41: Would it make sense to move that to the `filter`-call above? Of course that only works, if we have distinct values for event/setting.object_id … See comment below

src/pretix/base/migrations/0289_invoiceline_period.py
```diff
@@ -0,0 +1,80 @@
...
+
+        flush_queue.append("hierarkey_{}_{}".format("event", setting.object_id))
+        insert_queue.append(
+            EventSettingsStore(
+                object_id=setting.object_id,
+                key="invoice_period",
+                value="auto_no_event",
+            )
+        )
+        ev_seen.add(setting.object_id)
```
wiffbi: Could it happen, that we get mutliple `setting` for the same object_id? Otherwise this is superfluous. Or could we just append a `.values('object_id').distinct()` to the filter-call in the for-loop?

raphaelm:
Migration plan for hosted:

```
-- Before code deployment
BEGIN;

ALTER TABLE "pretixbase_invoiceline" ADD COLUMN "period_start" timestamp with time zone;
ALTER TABLE "pretixbase_invoiceline" ADD COLUMN "period_end" timestamp with time zone;

CREATE OR REPLACE FUNCTION update_invoiceline_period() RETURNS TRIGGER AS $BODY$
BEGIN
    NEW.period_start=COALESCE(NEW.period_start, NEW.event_date_from);
    NEW.period_end=COALESCE(NEW.period_end, NEW.event_date_to);
    NEW.event_date_from=COALESCE(NEW.event_date_from, NEW.period_start);
    NEW.event_date_to=COALESCE(NEW.event_date_to, NEW.period_end);
    RETURN NEW;
END
$BODY$ LANGUAGE PLPGSQL;

CREATE TRIGGER pretixbase_invoiceline
    BEFORE INSERT
    ON pretixbase_invoiceline
    FOR EACH ROW EXECUTE PROCEDURE update_invoiceline_period();

ALTER TABLE "pretixbase_invoice" ADD COLUMN "plugin_data" jsonb DEFAULT '{}' NOT NULL;

INSERT INTO django_migrations (app, name, applied) VALUES ('pretixbase', '0289_invoiceline_period', NOW());
INSERT INTO django_migrations (app, name, applied) VALUES ('pretixbase', '0290_invoice_plugin_data', NOW());

COMMIT;

-- Before code deployment, in multiple iterations
UPDATE pretixbase_invoiceline
SET period_start = event_date_from,
    period_end   = event_date_to
WHERE id IN (SELECT id
             FROM pretixbase_invoiceline
             WHERE period_start IS NULL
               AND period_end IS NULL
               AND (event_date_from IS NOT NULL OR event_date_to IS NOT NULL)
    LIMIT 5000
    );


-- After code deployment
BEGIN;
DROP TRIGGER pretixbase_invoiceline ON pretixbase_invoiceline;
DROP FUNCTION update_invoiceline_period;
ALTER TABLE pretixbase_invoiceline DROP COLUMN event_date_from;
ALTER TABLE pretixbase_invoiceline DROP COLUMN event_date_to;
COMMIT;
```

src/pretix/base/migrations/0289_invoiceline_period.py
raphaelm: not every database (*cough* SQLite) does well with filter calls with IN (…) with possibly more than 1000 entries

src/pretix/base/migrations/0289_invoiceline_period.py
raphaelm: na, it can't happen any more, so yes, we can drop it

PR #5449, opened 2025-09-08
Add support for New Zealand (en-nz) date/time formats

dannyadair:
- Implement en-NZ specific formatting in daterange.py
- Create en_NZ/formats.py with NZ-compliant formats

### DIFF
--- a/src/pretix/helpers/daterange.py
+++ b/src/pretix/helpers/daterange.py
@@ -64,6 +64,16 @@ def daterange(df, dt, as_html=False):
             return format_html(base_format, _date(df, "j."), mark_safe(until.strip()), _date(dt, "j. F Y"))
         elif df.year == dt.year:
             return format_html(base_format, _date(df, "j. F"), until, _date(dt, "j. F Y"))
+    elif lng == "en-nz":
+        if df.year == dt.year and df.month == dt.month and df.day == dt.day:
+            # Mon, 15 January 2024
+            return format_html(base_format, _date(df, "D, j F Y"))
+        elif df.year == dt.year and df.month == dt.month:
+            # 1 – 3 January 2024
+            return format_html(base_format, _date(df, "j"), until, _date(dt, "j F Y"))
+        elif df.year == dt.year:
+            # 1 January – 3 April 2024
+            return format_html(base_format, _date(df, "j F"), until, _date(dt, "j F Y"))
     elif lng.startswith("en"):
         if df.year == dt.year and df.month == dt.month and df.day == dt.day:
             return format_html(base_format, _date(df, "D, N jS, Y"))
--- a/src/pretix/helpers/formats/en_NZ/__init__.py
+++ b/src/pretix/helpers/formats/en_NZ/__init__.py
@@ -1,0 +1,21 @@
+#
+# This file is part of pretix (Community Edition).
+#
+# Copyright (C) 2014-2020 Raphael Michel and contributors
+# Copyright (C) 2020-2021 rami.io GmbH and contributors
+#
+# This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General
+# Public License as published by the Free Software Foundation in version 3 of the License.
+#
+# ADDITIONAL TERMS APPLY: Pursuant to Section 7 of the GNU Affero General Public License, additional terms are
+# applicable granting you additional permissions and placing additional restrictions on your usage of this software.
+# Please refer to the pretix LICENSE file to obtain the full terms applicable to this work. If you did not receive
+# this file, see <https://pretix.eu/about/en/license>.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Affero General Public License for more
+# details.
+#
+# You should have received a copy of the GNU Affero General Public License along with this program.  If not, see
+# <https://www.gnu.org/licenses/>.
+#
--- a/src/pretix/helpers/formats/en_NZ/formats.py
+++ b/src/pretix/helpers/formats/en_NZ/formats.py
@@ -1,0 +1,50 @@
+#
+# This file is part of pretix (Community Edition).
+#
+# Copyright (C) 2014-2020 Raphael Michel and contributors
+# Copyright (C) 2020-2021 rami.io GmbH and contributors
+#
+# This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General
+# Public License as published by the Free Software Foundation in version 3 of the License.
+#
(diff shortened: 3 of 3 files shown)


dannyadair:
I'd like to note the non-intrusiveness of this change:
While it adds region support (`daterange` will now see `en-ca` instead of just `en`), only `en-nz` is actually handled so nobody else will see any difference (yet).

dannyadair:
P.S.: According to
https://www.digital.govt.nz/standards-and-guidance/design-and-ux/content-design-guidance/writing-style/numbers
"12:00p.m." should be "12 noon" but I think Django localisation just doesn't have that granularity (and don't want to monkeypatch it for that). Nobody will care :)

dannyadair:
```
        request_lang = get_language()
        if '-' not in request_lang and event.settings.region:
            request_lang += '-' + event.settings.region.lower()
        with language(request_lang):
 ```
 
 This pattern was used six times and probably deserves its own function, or maybe event method?
`Event.get_full_locale` ?

So you could just have `with language(event.get_full_locale())` - happy to improve it that way but I'll wait for feedback first.

**Comment by raphaelm** [comment]
> This pattern was used six times and probably deserves its own function, or maybe event method?
> Event.get_full_locale ?
> 
> So you could just have with language(event.get_full_locale()) - happy to improve it that way but I'll wait for feedback first.

I think we need no additional function, as it can be simplified to:

```
with language(get_language(), event.settings.region):
```

File: src/pretix/base/models/event.py [hunk 1]
```diff
@@ -191,7 +192,12 @@ def get_date_range_display(self, tz=None, force_show_end=False, as_html=False, t
                 (dt.astimezone(tz).date() == df.astimezone(tz).date() + timedelta(days=1) and
                  dt.astimezone(tz).time() < df.astimezone(tz).time())
             )
-        d = daterange(df.astimezone(tz), dt.astimezone(tz), as_html)
+
+        request_lang = get_language()
+        if '-' not in request_lang and self.settings.region:
+            request_lang += '-' + self.settings.region.lower()
+        with language(request_lang):
+            d = daterange(df.astimezone(tz), dt.astimezone(tz), as_html)
```
raphaelm: Hmm, I don't really think get_date_range_display should need to call with language(). The idea of the context manager is not to call it on the lowest possible levels. I think we should figure out why it isn't called in the context where get_range_display is called and where the language is activated and fix it there.

src/pretix/control/views/dashboards.py
```diff
@@ -522,13 +525,21 @@ def widgets_for_event_qs(request, qs, user, nmax, lazy=False):
...
+                            (event.max_fromto or event.max_to or event.max_from).astimezone(tz)
+                        )
             else:
                 if event.date_to:
-                    dr = daterange(event.date_from.astimezone(tz), event.date_to.astimezone(tz))
+                    request_lang = get_language()
+                    if '-' not in request_lang and event.settings.region:
+                        request_lang += '-' + event.settings.region.lower()
+                    with language(request_lang):
+                        dr = daterange(event.date_from.astimezone(tz), event.date_to.astimezone(tz))
```
raphaelm: See above, I don't think this belongs on this layer

src/pretix/control/views/typeahead.py
```diff
@@ -89,10 +90,14 @@ def serialize_event(e):
...
-                (e.max_fromto or e.max_to or e.max_from).astimezone(tz)
-            )
+            request_lang = get_language()
+            if '-' not in request_lang and e.settings.region:
+                request_lang += '-' + e.settings.region.lower()
+            with language(request_lang):
+                dr = _('Series:') + ' ' + daterange(
+                    e.min_from.astimezone(tz),
+                    (e.max_fromto or e.max_to or e.max_from).astimezone(tz)
+                )
```
raphaelm: See above, I don't think this belongs on this layer

src/pretix/presale/views/organizer.py
```diff
@@ -479,11 +480,15 @@ def get_context_data(self, **kwargs):
...
-                )
+                request_lang = get_language()
+                if '-' not in request_lang and event.settings.region:
+                    request_lang += '-' + event.settings.region.lower()
+                with language(request_lang):
+                    event.daterange = daterange(
+                        event.min_from.astimezone(event.tzname),
+                        (event.max_fromto or event.max_to or event.max_from).astimezone(event.tzname),
+                        as_html=True,
+                    )
```
raphaelm: See above, I don't think this belongs on this layer

src/pretix/presale/views/widget.py
```diff
@@ -693,10 +693,14 @@ def _get_event_list(self, request, **kwargs):
...
-                            (event.max_fromto or event.max_to or event.max_from).astimezone(tz)
-                        )
+                        request_lang = get_language()
+                        if '-' not in request_lang and event.settings.region:
+                            request_lang += '-' + event.settings.region.lower()
+                        with language(request_lang):
+                            dr = daterange(
+                                event.min_from.astimezone(tz),
+                                (event.max_fromto or event.max_to or event.max_from).astimezone(tz)
+                            )
```
raphaelm: See above, I don't think this belongs on this layer

raphaelm (changes requested):
Could you create the PR with *only* the changes to daterange.py and formats/en_NZ/? I'd merge that right away, even if it is apparently not used where you would like it to be used.

Then, let's continue the discussion on where it should be used.

The idea behind our language system is that the locale should be active for the biggest scope possible, i.e. the entire request handling (as done by LocaleMiddleware), not that we switch locales on every low-level function call.

dannyadair:
> > This pattern was used six times and probably deserves its own function, or maybe event method?
> > Event.get_full_locale ?
> > So you could just have with language(event.get_full_locale()) - happy to improve it that way but I'll wait for feedback first.
> 
> I think we need no additional function, as it can be simplified to:
> 
> ```
> with language(get_language(), event.settings.region):
> ```

Oh, haha, shoulda looked at language() - much better

dannyadair:
> Could you create the PR with _only_ the changes to daterange.py and formats/en_NZ/? I'd merge that right away, even if it is apparently not used where you would like it to be used.
> 

Done.

> Then, let's continue the discussion on where it should be used.
> 
> The idea behind our language system is that the locale should be active for the biggest scope possible, i.e. the entire request handling (as done by LocaleMiddleware), not that we switch locales on every low-level function call.

Thank you, that makes sense. I'll have a look at the call stack for each of these calls.
What would you expect a list of events and their dates to look like if they come from mixed regions?

raphaelm:
>  What would you expect a list of events and their dates to look like if they come from mixed regions?

I wouldn't mix formats in the same view. Date formats are not an inherent property of the event but should gather to the user reading them.

As for multi-event views:

- If we're in pretix.presale, I'd expect organizer.settings.region.
- If we're in pretix.control on a /organizers/… path, I'd also expect organizer.settings.region (since we do not have a user.region - field).
- If we're in pretix.control on a global path, I say that for now there is possible solution since we do not have a user.region field and we need to show system default.

PR #5457, opened 2025-09-09
Allow plugins to filter subevents in the public calendar

raphaelm:


src/pretix/presale/views/organizer.py
```diff
@@ -561,16 +563,56 @@ def add_events_for_days(request, baseqs, before, after, ebd, timezones):
             })
 
 
-def add_subevents_for_days(qs, before, after, ebd, timezones, event=None, cart_namespace=None, voucher=None):
+def filter_subevents_with_plugins(subevents, sales_channel=None):
```
wiffbi, line 566: Should we add a comment, that this function uses some IMHO internals of django’s [Signal.send](https://github.com/django/django/blob/main/django/dispatch/dispatcher.py#L179)? At least TIL about `sender_receivers_cache` and `_live_receivers` – actually from this [SO-answer](https://stackoverflow.com/a/76282087)

raphaelm: I mean, there is a comment referring to this being a special case of GlobalSignal.send_chained(), and all our signal code is full of django internals?

raphaelm: i'll move the comment up a bit

PR #5471, opened 2025-09-17
Scheduled exports: Check permissions on creation

raphaelm:
This is especially important in admin mode, because currently you can setup scheduled exports in admin mode, but they will never run if your user does not have the permissions to do so outside of admin mode.

### DIFF
--- a/src/pretix/control/views/orders.py
+++ b/src/pretix/control/views/orders.py
@@ -2832,11 +2832,25 @@ class ExportView(EventPermissionRequiredMixin, ExportMixin, ListView):
     def get_queryset(self):
         return self.get_scheduled_queryset()
 
+    def has_permission(self):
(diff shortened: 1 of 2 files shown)


src/pretix/control/views/orders.py
```diff
@@ -2832,11 +2832,25 @@ def schedule_form(self):
     def get_queryset(self):
         return self.get_scheduled_queryset()
 
+    def has_permission(self):
+        if not self.request.user.has_event_permission(self.request.organizer, self.request.event, "can_view_orders"):
+            return False
+        return True
```
wiffbi:
```suggestion
        return self.request.user.has_event_permission(self.request.organizer, self.request.event, "can_view_orders")
```

wiffbi (approved):

PR #5477, opened 2025-09-19
Fix #1907 – Obfuscate contact email addresses in public HTML

### LINKED ISSUE #1907: hide contact e-mail address in source code of a shop
If I enter a contact e-mail address for an event in the general settings the address is embedded directly into the source code of the shops website. It would be great if I could insert some random html comments into the e-mail address field or even better if pretix would insert them automatically.
I want to protect the e-mail address against additional spam mails by doing this.

ananyatimalsina:
This change obfuscates contact email addresses in public shop HTML to reduce exposure to harvesting bots, as requested in #1907.

Contact email addresses in relevant templates are now rendered with `@` replaced by `[at]` and `.` replaced by `[dot]`, using a new `anon_email` template filter. A small JavaScript file is included to automatically restore the email addresses for users after page load, ensuring usability while reducing the risk of simple scraping.

This approach is limited to web-facing templates; email templates and backend logic remain unchanged, as JavaScript-based obfuscation would not be effective there.

By combining template-level obfuscation and client-side deanonymization, this change balances accessibility for users and basic protection against email scraping bots.


wiffbi:
Thanks, we added HTML entities encoding as well

PR #5499, opened 2025-09-30
Add refund comment to gift card transaction text (Z#23208349)

raphaelm:


src/pretix/base/models/giftcards.py
```diff
@@ -195,20 +195,21 @@ def display(self, customer_facing=True):
...
+                    reverse(
+                        "control:event.order",
+                        kwargs={
+                            "event": self.order.event.slug,
+                            "organizer": self.order.event.organizer.slug,
+                            "code": self.order.code,
+                        }
+                    ),
+                    self.order.full_code,
+                    self.text,
```
luelista:
```suggestion
                    self.text or "",
```
otherwise it says "EVENT-12345 None" if no text is provided

PR #5500, opened 2025-09-30
Run exporters in repeatable read by default (Z#23173095)

raphaelm:


src/pretix/helpers/database.py
```diff
@@ -62,6 +63,39 @@ def casual_reads():
...
+    of isolation, but we NEED to use it: Otherwise e.g. our quota logic breaks, because we need to be able to get the
+    *current* number of tickets sold at any time in a transaction, not the number of tickets sold *before* our transaction
+    started.
+
+    However, this isolation mode has drawbacks, for example during reporting. When a user retrieves a report from the
+    system, it should return numbers that are consistent with each other. However, if the report makes multiple SQL
+    queries in READ COMMITTED mode, the results might be different for each query, causing numbers to be inconsistent
+    with each other.
+
+    This decorator creates a transaction that is running in REPEATABLE READ mode to avoid this problem.
```
wiffbi:
```suggestion
    This context manager creates a transaction that is running in REPEATABLE READ mode to avoid this problem.
```

src/pretix/base/services/quotas.py
```diff
@@ -64,7 +64,8 @@ class QuotaAvailability:
     * count_cart (dict mapping quotas to ints)
     """
 
-    def __init__(self, count_waitinglist=True, ignore_closed=False, full_results=False, early_out=True):
+    def __init__(self, count_waitinglist=True, ignore_closed=False, full_results=False, early_out=True,
+                 ignore_repeatable_read=False):
```
wiffbi: For me, calling the param `do_not_allow_repeatable_read` – or IMHO better invert the logic to ``allow_repeatable_read`` with a default `True` is easier to get my head around, what this does. 'ignore' somehow does not feel right to me, but I may be wrong. So feel free to ignore (haha, sorry, pun not intended) this suggestion.

src/pretix/base/services/quotas.py
```diff
@@ -119,6 +123,10 @@ def compute(self, now_dt=None, allow_cache=False, allow_cache_stale=False):
         Compute the queued quotas. If ``allow_cache`` is set, results may also be taken from a cache that might
         be a few minutes outdated. In this case, you may not rely on the results in the ``count_*`` properties.
         """
+        if not self._ignore_repeatable_read and getattr(connection, "tx_in_repeatable_read", False):
```
wiffbi: I know it is a basic line of code, but should we add a test for this situation, just to make sure, that that warning does not somehow get lost? It is a rather important feature on the quota calc to not allow the repeatable read if not stated explicitly IMHO.

src/pretix/base/services/quotas.py
raphaelm: I think you mean allow_repeatable_read with a default False?

PR #5510, opened 2025-10-02
Log entries: Check config before bulk-scheduling

raphaelm:


raphaelm:
I wrote all this and then noticed that it is useless… it's only just one task prevented anyways. So I'm just keeping the PR for archeological purposes in case we ever need it and close it again

PR #5512, opened 2025-10-02
Notification queues: Optimize order for less queries

raphaelm:


src/pretix/api/webhooks.py
```diff
@@ -450,7 +454,7 @@ def notify_webhooks(logentry_ids: list):
         if not notification_type:
             break  # Ignore, no webhooks for this event type
 
-        if _org != logentry.organizer or _at != logentry.action_type or webhooks is None:
+        if _org != logentry.organizer or _at != logentry.action_type or webhooks is None or _ev != logentry.event_id:
```
luelista:
```suggestion
        if _org != logentry.organizer or _at != logentry.action_type or _ev != logentry.event_id or webhooks is None:
```
to keep the "grouping" conditions next to each other for readability

luelista (approved):

PR #5513, opened 2025-10-02
Add prioritization to webhook/notifications queue

raphaelm:


src/pretix/helpers/celery.py
```diff
@@ -0,0 +1,63 @@
...
+        return settings.PRIORITY_CELERY_HIGH
+
+    # We use redis directly instead of the Django cache API since the Django cache API does not support INCR for
+    # nonexistant keys
+    rc = get_redis_connection("redis")
+
+    cache_key = f"pretix:task_priority:{shard}:{organizer_id}"
+
+    new_counter = rc.incr(cache_key)
+    print("counter", new_counter)
```
luelista:
```suggestion

```

src/pretix/api/webhooks.py
```diff
@@ -470,7 +471,10 @@ def notify_webhooks(logentry_ids: list):
                 )
 
         for wh in webhooks:
-            send_webhook.apply_async(args=(logentry.id, notification_type.action_type, wh.pk))
+            send_webhook.apply_async(
+                args=(logentry.id, notification_type.action_type, wh.pk),
+                priority=get_task_priority("notifications", logentry.organizer_id),
```
luelista, line 480: is it on purpose that this is in shard="notifications"? (not e.g. shard="webhooks" ?)

src/pretix/base/models/base.py
```diff
@@ -131,9 +132,15 @@ def log_action(self, action, data=None, user=None, api_token=None, auth=None, sa
...
-                notify.apply_async(args=(logentry.pk,))
+                notify.apply_async(
+                    args=(logentry.pk,),
+                    priority=get_task_priority("notifications", logentry.organizer_id),
+                )
             if logentry.webhook_type:
-                notify_webhooks.apply_async(args=(logentry.pk,))
+                notify_webhooks.apply_async(
+                    args=(logentry.pk,),
+                    priority=get_task_priority("notifications", logentry.organizer_id),
```
luelista, line 142: (as above, is same shard correct?)

src/pretix/base/models/log.py
```diff
@@ -186,7 +189,19 @@ def bulk_postprocess(cls, objects):
...
+                ),
+            )
         to_wh = [o.id for o in objects if o.webhook_type]
         if to_wh:
-            notify_webhooks.apply_async(args=(to_wh,))
+            organizer_ids = set(o.organizer_id for o in objects if o.webhook_type)
+            notify_webhooks.apply_async(
+                args=(to_wh,),
+                priority=settings.PRIORITY_CELERY_HIGHEST_FUNC(
+                    get_task_priority("notifications", oid) for oid in organizer_ids
```
luelista, line 205: (as above, is same shard correct?)

src/pretix/helpers/celery.py
```diff
@@ -0,0 +1,63 @@
...
+    if new_counter == 1:
+        # Make sure counters expire after a while, but only do so when they are newly set, to avoid additional EXPIRE
+        # calls to cache.
```
luelista: highly unlikely case, but if the process were killed or lost connection to redis between `incr` and `expire`, an organizer might get caught in prio low forever (until redis loses the key somehow). not sure how we could be prevent that though.

src/pretix/api/webhooks.py
raphaelm: Yes-ish, as these currently share the same Celery queue

src/pretix/helpers/celery.py
raphaelm: Pushed something which is not fully equivalent, but probably good enough

PR #5519, opened 2025-10-06
Improve interface to create and edit subevents (Z#23207890)

raphaelm:


src/pretix/control/templates/pretixcontrol/subevents/detail.html
```diff
@@ -130,30 +130,32 @@ <h4 class="panel-title">
                         {% trans "These settings are optional, if you leave them empty, the default values from the product settings will be used." %}
```
wiffbi, line 130: Should we add a hint, that only products that are assigned to a quota are shown here? Might be confusing why I do not see all products as it used to be. Could be made fancier with that hint only shown, when at least one product is hidden due to not being assigned to a quota. Also, if no product is shown (as no quota is defined), we might show a hint? A rather edge case though. Very likely that it is good enough as it is now, so fell free to ignore that comment.

`{% trans "These settings are optional, if you leave them empty, the default values from the product settings will be used. Only products assigned to a quota are shown." %}`

src/pretix/static/pretixcontrol/js/ui/subevent.js
```diff
@@ -167,13 +175,45 @@ $(document).on("pretix:bind-forms", function () {
...
+            el.prop("hidden", !itemvars.includes(el.attr("data-itemvar")) && !el.find(".has-error, .alert-danger").length);
+        });
+    }
 
-    $("#rrule-formset").on("formAdded", "div", function (event) { rrule_bind_form($(event.target)); });
+    $('[data-formset-prefix="quotas"]').on("formDeleted", "div", () => {
+        update_item_visibility();
+    }).on("formAdded", "div", (event) => {
+        quota_form_handlers($(event.target));
+        update_item_visibility();
```
wiffbi, line 211: I don’t think we need to `update_item_visibility` on `formAdded` as that quota-form is empty as default so no changes to items in a quota? Of course, if we add pre-defined values for that quota-form, we need to have it in there. So again, feel free to ignore.

```suggestion
```

src/pretix/control/templates/pretixcontrol/subevents/detail.html
raphaelm: I think this will only confuse further… Why would one be looking to change the price of a product that's not available on the subevent.

src/pretix/static/pretixcontrol/js/ui/subevent.js
raphaelm: You're technically correct, but I think it makes sense to keep this for future-safety, as it doesn't cost us much

PR #5532, opened 2025-10-09
Fix RRules with "until"-date (Z#23210685)

luelista:


luelista:
Fixed with https://github.com/pretix/pretix/pull/5538

PR #5535, opened 2025-10-09
Only store transmission email if invoice receiver is a business (Z#23210414)

luelista:


src/pretix/base/forms/questions.py
```diff
@@ -1394,9 +1396,7 @@ def clean(self):
                         raise ValidationError({r: _("This field is required for the selected type of invoice transmission.")})
 
                 self.instance.transmission_type = transmission_type.identifier
-                self.instance.transmission_info = {
-                    k: data.get(k) for k in transmission_type.invoice_address_form_fields
-                }
+                self.instance.transmission_info = transmission_type.form_data_to_transmission_info(data)
```
raphaelm, line 1401: This is not an equivalent, since the default implementation is

```
    def form_data_to_transmission_info(self, form_data: dict) -> dict:
        return form_data
```

So for any TransmissionType *other* than email, this change will store the full form content into transmission_info, including e.g. company name, address, etc. which is redundant and unexpected.

I think your change is still correct but we do need to change the default implementation as well, do you agree?

luelista: Yes, I agree :+1:

src/pretix/base/forms/questions.py
```diff
@@ -77,7 +77,7 @@
 from pretix.base.i18n import (
     get_babel_locale, get_language_without_region, language,
 )
-from pretix.base.invoicing.transmission import get_transmission_types
+from pretix.base.invoicing.transmission import get_transmission_types, transmission_types
```
luelista:
```suggestion
from pretix.base.invoicing.transmission import (
    get_transmission_types, transmission_types,
)
```

PR #5545, opened 2025-10-14
Invoice-PDF: split line.description into multiple rows so table can pagebreak

wiffbi:
When including questions/answers on the invoice, this can cause a failure to create the PDF when the description of a line inside the invoice-table’s cell exceeds the page size. ~This PR adds a basic char-limit on setting the description, so it always fits on the default renderers page (tested with hundreds of letter M as a question).~

UPDATE: This PR splits the description into chunks which then can be pagebroken by reportlab. It also lets the additional description (everything except the first line) span multiple columns.

Bildschirmfoto 2025-10-16 um 16 29 23

### DIFF
--- a/src/pretix/base/invoicing/pdf.py
+++ b/src/pretix/base/invoicing/pdf.py
@@ -903,7 +903,7 @@ class ClassicInvoiceRenderer(BaseReportlabInvoiceRenderer):
                     )
                     description_p_list.append(FontFallbackParagraph(
                         single_price_line,
-                        self.stylesheet['Normal']
+                        self.stylesheet['Fineprint']
                     ))
                 tdata.append((
                     description_p_list.pop(0),


src/pretix/base/invoicing/pdf.py
```diff
@@ -752,11 +753,60 @@ def day(dt: datetime.datetime) -> datetime.date:
...
+                if not description_p_list:
+                    # first "manual" line is larger than 5 "real" lines => only allow one line and set rest in Fineprint
+                    max_height = self.stylesheet['Normal'].leading
+
+                if h > max_height * 1.1:
+                    # quickly bring the text-length down to a managable length to then stepwise reduce
+                    wrap_to = math.ceil(len(curr_description) * max_height * 1.1 / h)
+                else:
+                    # long trimming long texts by percent creates to big jumps, but trimming short texts
+                    # by e.g. static 10 chars might be to much also
```
raphaelm: I was not able to understand what this comment says

src/pretix/base/invoicing/pdf.py
```diff
@@ -752,11 +753,60 @@ def day(dt: datetime.datetime) -> datetime.date:
...
+            # normalize linebreaks to newlines instead of HTML so we can safely substring
+            description = description.replace('<br>', '<br />').replace('<br />\n', '\n').replace('<br />', '\n')
+
+            # start first line with different settings than the rest of the description
+            curr_description = description.split("\n", maxsplit=1)[0]
+            cellpadding = 6  # default cellpadding is only set on right side of column
+            max_width = colwidths[0] - cellpadding
+            max_height = self.stylesheet['Normal'].leading * 5
+            p_style = self.stylesheet['Normal']
+            while True:
```
raphaelm: I spent 15min reading the code in this while-loop, and it took me the full 15min to figure out why it is very unlikely that this will ever be an infinite loop (but not completely guaranteed, either).

Infinite loops are a horrible problem to have in production. 

If we go with:

```suggestion
            for __ in range(1000):
```

We have limited the risk, also if it breaks because of a future change that is not done carefully enough

src/pretix/base/invoicing/pdf.py
```diff
@@ -819,13 +872,13 @@ def day(dt: datetime.datetime) -> datetime.date:
                         net_price=money_filter(net_value, self.invoice.event.currency),
                         gross_price=money_filter(gross_value, self.invoice.event.currency),
                     )
-                    description = description + "\n" + single_price_line
+                    description_p_list.append(FontFallbackParagraph(
+                        single_price_line,
```
raphaelm, line 875: I think this needs padding adjustments.
Before:
image

After:
image

Before is clearer what it refers to

src/pretix/base/invoicing/pdf.py
```diff
@@ -837,23 +890,52 @@ def day(dt: datetime.datetime) -> datetime.date:
...
+                    (0, len(tdata) - len(description_p_list)),
+                    (-1, len(tdata) - 2),
+                    0
+                ),
+                (
+                    'TOPPADDING',
+                    (0, len(tdata) - len(description_p_list) + 1),
+                    (-1, len(tdata) - 1),
+                    0
+                ),
```
raphaelm, line 936: Is the padding here expected like this? The gap between first and second like feels too big.
image

raphaelm: Also, should period_line and single_price_line also get the smaller font in this case? Otherwise they look like they are the next line a little bit

src/pretix/base/invoicing/pdf.py
wiffbi: You are right, I improved the comment to say: trim to 95% length, but at most 10 chars“

src/pretix/base/invoicing/pdf.py
wiffbi: Good idea! We could even go lower (e.g. 100) as even with very long texts in my tests it took at most 25 iterations to split the text. But let’s be safe and use 1000.

src/pretix/base/invoicing/pdf.py
wiffbi: You are right! Sorry, my test-invoice didn’t have the single-price line. I made the period_line and single_price_line in Fineprint-size as well. I also reduced the padding between the first and second line/block. It looks like this now:

Bildschirmfoto 2025-11-06 um 08 36 39

src/pretix/base/invoicing/pdf.py
wiffbi: See below regarding Fineprint-size for period_line and single_price_line.

PR #5555, opened 2025-10-18
Add option to show confirmation texts on order detail page

jasle:
Implements [showing confirmation texts on order details page](https://github.com/pretix/pretix/discussions/5023)

### DIFF
--- a/src/pretix/presale/views/order.py
+++ b/src/pretix/presale/views/order.py
@@ -92,7 +92,10 @@ from pretix.helpers.safedownload import check_token
 from pretix.multidomain.urlreverse import build_absolute_uri, eventreverse
 from pretix.presale.forms.checkout import InvoiceAddressForm, QuestionsForm
 from pretix.presale.forms.order import OrderPositionChangeForm
-from pretix.presale.signals import question_form_fields_overrides
+from pretix.presale.signals import (
+    checkout_all_optional, checkout_confirm_messages,
+    question_form_fields_overrides,
+)
 from pretix.presale.views import (
     CartMixin, EventViewMixin, iframe_entry_view_wrapper,
 )
@@ -237,6 +240,23 @@ class TicketPageMixin:
 class OrderDetails(EventViewMixin, OrderDetailMixin, CartMixin, TicketPageMixin, TemplateView):
     template_name = "pretixpresale/event/order.html"
 
+    @cached_property
+    def all_optional(self):
+        for recv, resp in checkout_all_optional.send(sender=self.request.event, request=self.request):
+            if resp:
+                return True
+        return False
+
+    @cached_property
+    def confirm_messages(self):
+        if self.all_optional:
+            return {}
+        msgs = {}
+        responses = checkout_confirm_messages.send(self.request.event)
+        for receiver, response in responses:
+            msgs.update(response)
+        return msgs
+
     def get(self, request, *args, **kwargs):
         self.kwargs = kwargs
         if not self.order:
@@ -257,6 +277,8 @@ class OrderDetails(EventViewMixin, OrderDetailMixin, CartMixin, TicketPageMixin,
         )
         ctx = super().get_context_data(**kwargs)
 
+        ctx['confirm_messages'] = self.confirm_messages
+        ctx['show_confirm_texts'] = self.request.event.settings.show_confirm_texts_on_order_detail
         ctx['can_download_multi'] = any([b['multi'] for b in self.download_buttons]) and (
             [p.generate_ticket for p in ctx['cart']['positions']].count(True) > 1
         )
--- a/src/pretix/presale/views/order.py
+++ b/src/pretix/presale/views/order.py
@@ -92,10 +92,7 @@ from pretix.helpers.safedownload import check_token
 from pretix.multidomain.urlreverse import build_absolute_uri, eventreverse
 from pretix.presale.forms.checkout import InvoiceAddressForm, QuestionsForm
 from pretix.presale.forms.order import OrderPositionChangeForm
-from pretix.presale.signals import (
-    checkout_all_optional, checkout_confirm_messages,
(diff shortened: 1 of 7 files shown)


src/pretix/presale/views/order.py
```diff
@@ -237,6 +240,23 @@ def download_buttons(self):
...
+
+    @cached_property
+    def confirm_messages(self):
+        if self.all_optional:
+            return {}
+        msgs = {}
+        responses = checkout_confirm_messages.send(self.request.event)
+        for receiver, response in responses:
+            msgs.update(response)
+        return msgs
```
jasle: Is here a good place for this? Or should I put it somewhere else?

raphaelm:
I don't think this is accurate. If the organizer adds a new confirm text, it will show on all old orders as confirmed even though it never was, wouldn't it?

jasle:
> I don't think this is accurate. If the organizer adds a new confirm text, it will show on all old orders as confirmed even though it never was, wouldn't it?

Yes, I didn't think of this. I have switched to display the confirm messages stored in the order information. So also when they get changed, the actual confirmed texts are shown.

One other question: Currently the confirm messages are displayed as `panel-primary` on the order details page. Should I switch to display it as `panel-default`?

src/pretix/presale/views/order.py
jasle: Switched to display the confirm message stored in the order and removed this.

raphaelm:
Okay, so now it is accurate, thanks for adjusting.

> One other question: Currently the confirm messages are displayed as panel-primary on the order details page. Should I switch to display it as panel-default?

Yes, it is certainly not one of the import things on the page and should be a panel-default.

I'm wondering if we can make it *even less dominant*, and then drop the config option and just show it on all order pages. Arguably, pretix has enough config options already. Maybe not even its own panel but somehow part of the "Your information" panel?

@wiffbi Happen to have any ideas? ;)

wiffbi:
I like @raphaelm’s idea moving it to the information-panel to make it less prominent. I tried several ways of offsetting it, moving it into the `<dl>` with the checkbox being in the `<dt>` as the „label“, but IMHO that does not look that good and kind of draws more attention than necessary. Also, currently linebreaks on mobile are awkward.

Bildschirmfoto 2025-10-22 um 12 11 11
Bildschirmfoto 2025-10-22 um 12 10 46

So I would go with a simple list with check-icons or probably even just multiple `<p>` – that way we circumvent the situation of potentially having a list with only one item in it.

Bildschirmfoto 2025-10-22 um 12 12 07
Bildschirmfoto 2025-10-22 um 12 11 56

I treid `.small` and/or `.text-muted` on that as well, but that again kind of drew attention to it. I also tried different check-icons (check, check-square, check-square-o), but check was IMHO the best, most clear one – the others look like an interative checkbox, but strangly off/different.

From a a11y point of view, I would not use disabled checkbox-inputs with labels to mark that list. Depending on the label of the checkbox, we could leave the check-icon invisible to screen-readers, but to make absolutely sure, I would markup the check-icon with a role=img and an alt-text "Checked checkbox" or something similar.


Arguably, with only one short checkbox, the `<dl>`-version looks a bit better, but overall, I think the `<p>` version is more flexible.

Bildschirmfoto 2025-10-22 um 12 24 42
Bildschirmfoto 2025-10-22 um 12 24 00


So, if we decide to go with the `<p>` version, would you be okay, if I adapt your HTML-templates? Of course you can do it yourself as well if you want!

jasle:
I really like the idea of integrating the confirmation texts in a less prominent way and always showing them, instead of an additional configuration option.

But maybe would put them in an own panel still a good idea?
If there are multiple confirmation texts or longer texts, they take up a lot of the space inside the information panel. When the information panel is only half-width (`col-xs-12` e.g. when there is an invoice) it could get very long. Also, it could be confusing that there is a `Change details`, but only for a part in the panel.

@wiffbi have you also tried the different options with confirmation texts that use markdown?

I am totally fine, if you adjust my PR.

wiffbi:
I actually haven’t tried markdown in the confirmation texts yet, but will. I also must note, that I would not recommend using these confirmation texts for anything long-ish form as they are markuped as the `<label>` for the checkbox – which isn’t ideal regarding accessibility: anything mid-long form should be aria-descibedby (which we currently do not support with these confirmation texts), anything longer should be moved to its own page – which brings me to the pretix-pages feature to having to confirm the contents of the pages (which currently isn’t ideal either as it packs `<a>`s inside the `<label>`, but that is a different problem not for this PR).

What if we just do it as it is on the checkout-confirm-step – full-width panel, but not primary and don’t mark it up as disabled inputs (not ideal for accessibility). Looking at order-confirm, we probably should split „your information“ and adapt it to „invoice information“ and „contact information“ as individual panels like on checkout-confirm. That way both pages, which come after each other, look the same. Also „Change details“ on the „Your information“ panel only allows me to change the invoice address, not my email address, which can be confusing, when I only gave my email address and want to change that.

Sooo, this PR unveiled some other issues, that are not to be handled in this PR, but TLDR; I guess putting them in their own panel is indeed the best way. I would move it below the „Your information“ panel – same position as it is on checkout-confirm.

Bildschirmfoto 2025-10-23 um 12 59 53

It looks okayish even if there is only one confirmation:

Bildschirmfoto 2025-10-23 um 13 06 17

I will clean my code for the screenshot above up a bit and commit it to this PR. Not sure though, if we need the checkbox in settings to show confirmations or just alsways show them with this panel-default.

jasle:
I really like the solution in the last screenshots. Would go for this solution.
I think, with this solution we could remove the option in the settings again and avoid having even more options.

jasle:
Hi @wiffbi,

I saw your draft for the layout changes and wanted to check in: are you planning to implement it yourself, or would it be okay if I take this over?

Happy to proceed either way.

Issue #5548, opened 2025-10-15
Possible race condition in api.views.OrderPositionViewSet.create / api.serializers.OrderPositionCreateForExistingOrderSerializer.create


stranck:
While I was working on a private plugin, I was taking a look at the codebase to understand how to properly create new positions with extra information fields. 

I stumbled upon the following code:
https://github.com/pretix/pretix/blob/2e674916c22dbd3ea8c1837bdd6f506204efc40f/src/pretix/api/views/order.py#L1459-L1482

While the OrderPosition instance coming from OrderPositionCreateForExistingOrderSerializer is obtained with:
https://github.com/pretix/pretix/blob/2e674916c22dbd3ea8c1837bdd6f506204efc40f/src/pretix/api/serializers/orderchange.py#L100-L101

My question is simple: Isn't this a race condition, where we can patch extra information (including answers) on another, unrelated, position?
In the previous code, if someone adds a new position $P_2$ while the transaction.atomic() block of the ocm.commit() has already ended (which created $P_1$), but before we hit the return of the OrderPositionCreateForExistingOrderSerializer.create() function, isn't `validated_data['order'].positions.order_by('-positionid').first()` returning $P_2$ (IE the last created position) instead of $P_1$?

I know that this is under a transaction, but from what I can understand from this PR (https://github.com/pretix/pretix/pull/2408) pretix runs in `READ COMMITTED` postgres mode, which doesn't protect from nonrepetable reads (https://www.postgresql.org/docs/current/transaction-iso.html)

Sorry if my question may sound dumb, I'm still trying to understand how to properly work with the codebase! :D

raphaelm:
Yes, this looks like an (unlikely but possible) race condition

stranck:
What's generally the approach on the codebase regarding this kind of not critical race conditions? Since they're not critical nor may bring to big vulnerabilities, do you usually live with them?

I'm asking just to understand how much "paranoic" should I be inside my plugin with races that does not touch quota/vouchers/seat/memberships

raphaelm:
We tend to fix them if we either (a) get bitten by them or similar ones in production, a typical case is a payment provider sending a confirmation webhook and a confirmation redirect at the same time or (b) find them by accident like you just did and they are fixable with reasonable effort or have the risk of severe consequences.

stranck:
Actually this is not fixed yet, my previous PR only created the proper way to fix this issue. Fix is in #5756

PR #5557, opened 2025-10-20
OrderChangeManager: add_position() returns a handle to the newly created position

stranck:
This PR changes OrdereChangeManager in two ways:
- It adds an abstract class for custom operations. In this way plugins which extend OrderChangeManager have the chance of implementing their own operations. I've originally planned of allowing this by creating a single `Operation` abstract class extended by all of the existing operations (+ the eventual custom operations implemented by plugins) and moving their logic from the `_perform_operations()` method to their own class. In this way the `_perform_operations()` method would become a simple `for op in self._operations: op.execute(...)`. However this way required "major" code changes and dismissing the use of `namedtuple` (which may (? I honestly have no idea tbf, I'm saying what docs say) be more lightweight than standard classes annotated with `@dataclass`), so I opted to essentially add a new case to the switch of that method. If the way I suggested is preferred I can change my PR without problems
- It adds a callback argument for `add_position()` which will be called at the end of the related `AddOperation` with the original operation and the newly created `OrderPosition`. This is a way for the OCM's caller to do operations on the exact created position in case there are multiple `add_position()` in a single commit. This could also help solving #5548: Instead of retrieving the last `OrderPosition` of the order (which allows for race conditions), we can do something like `created_pos = []; ocm.add_position(item=item, ..., callback=lambda op, pos: created_pos.append(pos)); return created_pos[0];`

### DIFF
--- a/src/pretix/base/services/orders.py
+++ b/src/pretix/base/services/orders.py
@@ -1671,19 +1671,15 @@ class OrderChangeManager:
     RemoveBlockOperation = namedtuple('RemoveBlockOperation', ('position', 'block_name', 'ignore_from_quota_while_blocked'))
 
     class AddPositionResult:
-        operation: 'OrderChangeManager.AddOperation'
-        original_order_change_manager: 'OrderChangeManager'
         _position: Optional[OrderPosition]
 
-        def __init__(self, order_change_manager: 'OrderChangeManager'):
-            self.original_order_change_manager = order_change_manager
-            self.operation = None
+        def __init__(self):
             self._position = None
 
         @property
         def position(self) -> OrderPosition:
             if self._position is None:
-                raise ValueError("OrderPosition has not been created yet. Call commit() first on the original OrderChangeManager")
+                raise RuntimeError("Order position has not been created yet. Call commit() first on OrderChangeManager.")
             return self._position
 
     def __init__(self, order: Order, user=None, auth=None, notify=True, reissue_invoice=True, allow_blocked_seats=False):
@@ -1950,10 +1946,9 @@ class OrderChangeManager:
         if seat:
             self._seatdiff.update([seat])
 
-        result = self.AddPositionResult(self)
-        op = self.AddOperation(item, variation, price, addon_to, subevent, seat, membership, valid_from, valid_until, is_bundled, result)
-        result.operation = op
-        self._operations.append(op)
+        result = self.AddPositionResult()
+        self._operations.append(self.AddOperation(item, variation, price, addon_to, subevent, seat, membership,
+                                                  valid_from, valid_until, is_bundled, result))
         return result
 
     def split(self, position: OrderPosition):
--- a/src/tests/base/test_orders.py
+++ b/src/tests/base/test_orders.py
@@ -2469,9 +2469,7 @@ class OrderChangeManagerTests(TestCase):
     def test_add_item_result_value(self):
         res_shirt = self.ocm.add_position(self.shirt, None, None, None)
         res_ticket2 = self.ocm.add_position(self.ticket2, None, None, None)
-        assert res_shirt.original_order_change_manager == self.ocm
-        assert res_shirt.operation.item == self.shirt
-        with self.assertRaises(ValueError):
+        with self.assertRaises(RuntimeError):
             _ = res_ticket2.position
         self.ocm.commit()
         assert res_shirt.position.item == self.shirt


stranck:
Not of my bingo card doing a PR while AWS or something is dead
image

Edit: I will push the checkstyle fix commit once aws is back to also trigger a new checks round

raphaelm:
Hey @stranck, thanks for the PR!

My first gut reaction and after talking in the team is that we might not want to have this. The idea and appeal of the OrderChangeManager is that it is a self-contained, well-tested, reliable engine that prevents you from screwing up because all the different combination of operations are thought through, tested and cared for by the classe.

Having plugins hook into that undermines this property significantly. Any moderately advanced CustomOperation will need to interact with the internal state of the OrderChangeManager (such as _quotadiff, _totaldiff, _invoice_dirty, etc). This not only makes it hard to develop the OrderChangeManager any further without breaking plugins, it also makes it much harder to test and understand all the ways that things go wrong. Any really complex Custom Operation might also need other hooks in various stages of commit(), which makes it even more complex.

The callback feature feels a little less dangerous, but raises some of the same questions. Also, depending on the use case one would want the callback to be called in different phases of the operation, such as inside or after the transaction.atomic() statement, etc.

So: Most likely we do not want this. But can you detail your use case for both features a bit further so we understand the reasoning behind it and maybe have other ideas to solve it?

stranck:
Hi! First of all, thanks for reviewing my PR! 

### Custom Operation

My thought behind allowing to extend ocm is that since python doesn't really have a way to prevent class hierarchy extension (like `final class` in java), if someone really want to, it still can hook ocm in worst and less reliable way. I actually have done it on a private plugin for really small things (EG disable locking in the commit phase, so they can be acquired once outside instead of calling `lock_objects` many times). I understand though that extending ocm without support makes the developer fully responsible of his errors, while if ocm "opens" even a bit, the responsibility of eventual breaking changes will lie upon you, making everything harder to update

Speaking about a usecase for this and why I started thinking about this in the first place: One of the features which my plugin allows is exchanging items between different orders. I managed to do everything and it works fine, but the only missing piece is moving also the `is_bundle` status: 
- Let's say order $O_{src}$ has two positions, with item $a$ and $b$, $b$ being bundled with $a$ (`is_bundled = True`)
- Let's say order $O_{dst}$ has two positions as well, $c$ and $d$, but $d$ is NOT bundled with $c$ (`is_bundled = False`)

Now I want to exchange these items between the two orders. To update the positions I will use various `ocm.change_XXX()` calls for each and I will end up in a status like this:
- $O_{src}$ has $c$ and $d$ with $d$ being bundled with $c$ (`is_bundled = True`)
- $O_{dst}$ has $a$ and $b$, but $b$ is NOT bundled with $a$ (`is_bundled = False`)

IE: The bundled status hasn't moved and I didn't had a way to forcefully move it, like a `ocm.change_bundle(orderPosition)` function

Summarizing, my real reason of why I came out with this pr is the following:
Let's consider the topic "maybe `ocm.change_item()` should set `is_bundle` as well" out of the current scope _(also because, like `ocm.add_position()` which can't set the bundle status if you're creating the `addon_to` position in the same commit "round", an updated change_item implementation would not as well except if he actually considers the others change operations inside the ocm)_, the reason why I've chosen to implement a `CustomOperation` is to have a way to actually implement something like a `ocm.change_bundle()`, or any other feature I might need _(which maybe shouldn't be included at all in the "plain" pretix)_ without doing major "copy paste with minor changes" of OCM methods which needs to always reflect the changes to the original one, or not using ocm at all. Both of which are even more error prone (but the responsibility will be on me and not on you)

### Callback

That callback is meant to be a simple way to "deterministically" retrieve the correct OrderPosition after it's created, nothing more in fact. I originally thought about doing an internal `ocm.created_positions` map, but the doubts were regarding what would be a good key for that map. In fact, we cannot use EG just the item to search for the created position, if we have two `ocm.add_position()` calls with the same item but with different prices? One might think "let's use a tuple (item, price) as a key instead!". But also this present problems: what if the price is non specified to `ocm.add_position()` and instead is retrieved internally?
This is why I thought to a callback, as the easier solution to solve that problem

BUT, but... While writing this comment I thought to two other possible solutions. Tell me if you're eventually "more ok" with the current callback implementation or which one of the following twos you might like more, I'm happy to change my PR to what you prefer:
1) Use a `ocm.created_positions` map and use a random value as the key for the positions. That value could be returned by the `ocm.add_position()` as a "ticket" to later access the related position (tbh I don't like this solution)
2) Make `ocm.add_position()` return OR take as optional input, a sort of "Promise" object: that object would have just one field originally set to null which will eventually populate once `ocm.commit()` is called, so the caller can retrieve the position. This is in fact what I suggested to do at the start of the PR and what I actually do "externally" as the unit test for that feature. In this way we don't expose a full callback, but just a way to get the object, moving all the logic "internally"

stranck:
I'm adding that:
While creating a `CustomOperator` class might be a clear signal for everyone who looks at the ocm class "hey, you can add your own operator! Please do!!!", the other solution I suggested in the original post (IE creating a `Operator` abstract class and moving the operation logic inside) could be an in between: While it doesn't scream "we are happy to support custom operators" it still allows that, leaving your team without the responsibility of breaking changes to the internals of ocm which might disrupt plugins. Like someone who extends OCM and overrides its methods even if he's not meant to, someone who implement its own `Operator` (again, without the codebase explicitly telling you that you can do it) should and (hopefully) will take care of eventual updates to the ocm class

I'm waiting for your feedback also for this, for me it's not a problem changing again the PR :D

luelista:
Hi @stranck !

For passing the created `OrderPosition` to the calling code, we like the idea of returning a promise/future/handle-like object from `add_position` a lot. If you could adapt your PR to that end, we'd be happy to merge that.

For the custom operation – I see the appeal of refactoring all `Operation`s to self-contained classes, containing the data currently stored in the named tuple as well as an `execute` method performing the actual operation. However, we don't want do that big a change in OCM right now, and are also still not sure it's a good idea to make it extensible at all.

Best way going forward is to split this PR into two, so we could merge the "OrderPosition result promise" part relatively soon, and can think a bit more about what's the best way to fix your `is_bundle` problem.

stranck:
Perfect! I will work on this the upcoming week! :D

stranck:
As requested, I removed from this PR the changes regarding the ocm operation refactor and changed how a user can obtain an OrderPosition after using `add_position`. Tell me if everything is now ok or if there's something else I should change!

For the ocm refactor, should I do something, like creating an issue?

src/pretix/base/services/orders.py
```diff
@@ -1670,6 +1670,22 @@ class OrderChangeManager:
...
     RemoveBlockOperation = namedtuple('RemoveBlockOperation', ('position', 'block_name', 'ignore_from_quota_while_blocked'))
 
+    class AddPositionResult:
+        operation: 'OrderChangeManager.AddOperation'
+        original_order_change_manager: 'OrderChangeManager'
+        _position: Optional[OrderPosition]
+
+        def __init__(self, order_change_manager: 'OrderChangeManager'):
+            self.original_order_change_manager = order_change_manager
+            self.operation = None
```
luelista: The caller will have a reference to `OrderChangeManager` anyway. The operation tuples are more of an internal structure of `OrderChangeManager`, so I think we should not expose it to the caller without good reason.

Do you have any special use cases in mind for these attributes? Otherwise, I think they should be removed and we should just keep `_position` / `position`.

src/pretix/base/services/orders.py
```diff
@@ -1670,6 +1670,22 @@ class OrderChangeManager:
...
+            self._position = None
+
+        @property
+        def position(self) -> OrderPosition:
+            if self._position is None:
+                raise ValueError("OrderPosition has not been created yet. Call commit() first on the original OrderChangeManager")
```
luelista:
```suggestion
                raise RuntimeError("Order position has not been created yet. Call commit() first on OrderChangeManager.")
```

src/pretix/base/services/orders.py
```diff
@@ -1933,8 +1949,12 @@ def add_position(self, item: Item, variation: ItemVariation, price: Decimal, add
...
         if seat:
             self._seatdiff.update([seat])
-        self._operations.append(self.AddOperation(item, variation, price, addon_to, subevent, seat, membership,
-                                                  valid_from, valid_until, is_bundled))
+
+        result = self.AddPositionResult(self)
+        op = self.AddOperation(item, variation, price, addon_to, subevent, seat, membership, valid_from, valid_until, is_bundled, result)
+        result.operation = op
+        self._operations.append(op)
+        return result
```
luelista, line 1952:
```suggestion
        result = self.AddPositionResult()
        self._operations.append(self.AddOperation(item, variation, price, addon_to, subevent, seat, membership,
                                                  valid_from, valid_until, is_bundled, result))
        return result
```

luelista:
Sorry, it took a while to get around to this – I added some feedback to the PR. Apart from those, I think it looks good now :)

Feel free to open a [discussion](https://github.com/pretix/pretix/discussions) about making OrderChangeManager more extensible, as I think that's something where we need to find out whether we should really do that, and even if so, it might take quite a while until we get around to that. :/ 

If you need help regarding your specific problem of shuffling around items between orders, you could create a separate issue or discussion for that. Does it cause the order total and taxes to change? If not, you might be able to get away with just changing the OrderPositions manually (in a transaction, with locks on both orders), without even using OrderChangeManager. Maybe you can share your code, so we can see what exactly you want to do.

src/pretix/base/services/orders.py
stranck: Nop, actually I don't have any use case crossing my mind. I will remove them

stranck:
I'll open a discussion post for the OCM soon then, thank you all for accepting the PR btw!

After this is merged I'm going to do another PR to fix #5548. I have already looked through the codebase and it seems to be the only case where you do a select like this to obtain a newly created position

For the shuffling items in various orders, I don't want to open a full new extra discussion thread just to send a message, sorry if I'll write it down here even if it's off topic:
Yeah, I also make total and taxes changes, this is why I thought about using OCM: it makes everything easier, by not only doing everything by its own, but also by having a batch preparation/commit paradigm which simplifies stuff a lot. I was previously doing this by making multiple calls to the API but it was a pain, slow and not atomic, so I'm really happy to have implemented this inside "pretix" itself. If you're curious to take a look at my ugly and non-django-compliant code: https://github.com/APSfurizon/pretix-fzbackend-utils/blob/master/pretix_fzbackend_utils/views/exchange_rooms.py

Be aware that this code makes assumptions on some stuff which are specific for our kind of events, it's not something generalizable at all. The only thing I'm unsure about this code is if using `select_for_update` is enough to gather locks

stranck:
Bumping this! (Sorry 😭)

raphaelm:
All good, got it on my todo list now again!

PR #5563, opened 2025-10-21
Markdown link parser: Fix fediverse URLs and URLs with user or path (Z#23211629)

raphaelm:


src/pretix/base/templatetags/rich_text.py
```diff
@@ -54,6 +54,19 @@
...
 register = template.Library()
 
+
+def build_fediverse_re(tlds):
+    return re.compile(
+        r"""\(*  # Match any opening parentheses.
+        @[^@]+@
+        ([\w-]+\.)+(?:{0})(?:\:[0-9]+)?(?!\.\w)\b   # xx.yy.tld(:##)?
+        """.format(
+            "|".join(sorted(tlds))
```
wiffbi, line 64: Is there a reason we sort the tlds again? They get sorted when passed via the lambda-call.

src/pretix/base/templatetags/rich_text.py
```diff
@@ -144,20 +159,32 @@ def truelink_callback(attrs, new=False):
...
             text = 'https://' + text
 
+        text_url = urllib.parse.urlparse(text)
+        if text_url.netloc.split("@")[-1] != href_url.netloc.split("@")[-1] or not href_url.path.startswith(text_url.path):
+            # link text contains an URL that has a different base than the actual URL
+            attrs['_text'] = attrs[None, 'href']
+
+    # Verify server name of mastodon display names (@name@server.tld)
+    if (None, 'href') in attrs and FEDIVERSE_RE.match(text):
+        parts = text.split('@')
```
wiffbi, line 181: Not a huge deal, but I can create a markdown-link to another user like so:

`[@good@mastodon.social](https://mastodon.social/@bad)`

Is that something we want to allow?

src/pretix/base/templatetags/rich_text.py
raphaelm: This is copied from Django's code :shrug:

src/pretix/base/templatetags/rich_text.py
raphaelm: I think that's "safe" enough as in "safe against malware" etc since the server is fixed

PR #5565, opened 2025-10-21
Event cancellation: Add safety and security checks

raphaelm:
When cancelling an event, a large sum of money might be refunded instantly. This PR adds safety features around this by

- doing a dry-run first that shows a preview of the expected refund sum

- sending a confirmation code via email for any automatic refunds of more than 100 currency units

- keeping a more detailed log of the settings this was executed with

### DIFF
--- a/src/pretix/control/views/orders.py
+++ b/src/pretix/control/views/orders.py
@@ -2994,11 +2994,17 @@ class EventCancel(EventPermissionRequiredMixin, AsyncAction, FormView):
                      'check all uncanceled orders.').format(count=value)
 
     def get_success_url(self, value):
-        return reverse('control:event.cancel.confirm', kwargs={
-            'organizer': self.request.organizer.slug,
-            'event': self.request.event.slug,
-            'task': value["id"],
-        })
+        if settings.HAS_CELERY:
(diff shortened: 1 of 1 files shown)


src/pretix/control/views/orders.py
```diff
@@ -2975,10 +2976,93 @@ def form_valid(self, form):
...
+        else:
+            return _('The orders have been canceled. An error occurred with {count} orders, please '
+                     'check all uncanceled orders.').format(count=value)
+
+    def get_success_url(self, value):
+        return reverse('control:event.cancel.confirm', kwargs={
+            'organizer': self.request.organizer.slug,
+            'event': self.request.event.slug,
+            'task': value["id"],
+        })
```
luelista: we shouldn't redirect to the confirm page in case we don't have celery, should we?
```suggestion
        if settings.HAS_CELERY:
          return reverse('control:event.cancel.confirm', kwargs={
              'organizer': self.request.organizer.slug,
              'event': self.request.event.slug,
              'task': value["id"],
          })
        else:
          return reverse('control:event.cancel', kwargs={
              'organizer': self.request.organizer.slug,
              'event': self.request.event.slug,
          })
```

PR #5598, opened 2025-11-10
Turn attendee emails on by default for new events (Z#23213656)

raphaelm:
I think the thing that makes me most unhappy is that *most* organizers will probably want to turn off mail_send_order_paid_attendee when they set ticket_download_pending and I don't think organizers will remember that, but it also seems complex and weird to create an automatism for it?

### DIFF
--- a/src/pretix/base/models/event.py
+++ b/src/pretix/base/models/event.py
@@ -715,6 +715,13 @@ class Event(EventMixin, LoggedModel):
         self.settings.name_scheme = 'given_family'
         self.settings.payment_banktransfer_invoice_immediately = True
         self.settings.low_availability_percentage = 10
+        self.settings.mail_send_order_free_attendee = True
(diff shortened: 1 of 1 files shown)


src/pretix/base/models/event.py
```diff
@@ -715,6 +715,13 @@ def set_defaults(self):
         self.settings.name_scheme = 'given_family'
         self.settings.payment_banktransfer_invoice_immediately = True
         self.settings.low_availability_percentage = 10
+        self.settings.mail_send_order_free_attendee = True
+        self.settings.mail_send_order_placed_attendee = True
+        self.settings.mail_send_order_placed_attendee = True
```
pc-coholic: Duplicate
```suggestion
        self.settings.mail_send_order_placed_attendee = True
```

pc-coholic (changes requested):

PR #5599, opened 2025-11-10
Product list: Show number of items currently in cart (Z#23212546)

raphaelm:
image

### DIFF
--- a/src/pretix/static/pretixpresale/scss/main.scss
+++ b/src/pretix/static/pretixpresale/scss/main.scss
@@ -574,6 +574,7 @@ h2 .label {
 
 
 .textbubble-success, .textbubble-success-warning, .textbubble-info, .textbubble-warning, .textbubble-danger {
+  display: inline-block;
   padding: 0 .4em;
   border-radius: $border-radius-base;
   font-weight: bold;


src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
```diff
@@ -119,7 +123,19 @@
                                 data-price="{% if event.settings.display_net_prices %}{{ var.display_price.net|unlocalize }}{% else %}{{ var.display_price.gross|unlocalize }}{% endif %}"
                             {% endif %}>
                                 <div class="col-md-8 col-sm-6 col-xs-12">
-                                    <h{{ headline_level|default:3|add:2 }} class="h5" id="{{ form_prefix }}item-{{ item.pk }}-{{ var.pk }}-legend">{{ var }}</h{{ headline_level|default:3|add:2 }}>
+                                    <h{{ headline_level|default:3|add:2 }} class="h5" id="{{ form_prefix }}item-{{ item.pk }}-{{ var.pk }}-legend">
+                                        {{ var }}
+                                        {% if cart.itemvarsums|getitem:var %}
+                                            <span class="badge badge-success">
```
wiffbi: In the customer account pages I added `textbubble-*` classes, which make sure, that all badges meet a11y-criteria for contrast. So for better a11y-support and to make the bubbles look the same, I’d use these:
Bildschirmfoto 2025-11-10 um 14 36 18

```suggestion
                                            <span class="textbubble-success">
```

src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
```diff
@@ -264,7 +280,19 @@
                             </a>
                         {% endif %}
                         <div class="product-description {% if item.picture %}with-picture{% endif %}">
-                            <h{{ headline_level|default:3|add:1 }} class="h4" id="{{ form_prefix }}item-{{ item.pk }}-legend">{{ item.name }}</h{{ headline_level|default:3|add:1 }}>
+                            <h{{ headline_level|default:3|add:1 }} class="h4" id="{{ form_prefix }}item-{{ item.pk }}-legend">
+                                {{ item.name }}
+                                {% if cart.itemvarsums|getitem:item %}
+                                    <span class="badge badge-success">
```
wiffbi:
```suggestion
                                    <span class="textbubble-success">
```

src/pretix/static/pretixpresale/scss/_event.scss
```diff
@@ -164,6 +164,9 @@ article.item-with-variations .product-row:last-child:after {
         display: block;
         line-height: 1;
     }
+    h2 .badge, h3 .badge, h4 .badge, h5 .badge, h6 .badge {
+        vertical-align: top;
+    }
```
wiffbi: Not needed when using `.textbubble-*`. But I realized in testing, that `.textbubble-*` should have `display: inline-block` to make sure, that not event the start-padding is line-broken – I originally through nowrap would havce been enough, but testing revealed, it is not.

Could you add `display: inline-block` to pretix/static/pretixpresale/scss/main.scss:576 or should make a PR for that fix?

```suggestion
```

raphaelm:
> Looks good. I am not sure, whether directly after the item’s name in the heading is an ideal place – from a a11y-pespective it feels like it should belong to a description-element (e.g. a following `<p>`). 

Visually, I like mine a little more, but don't have strong opinions. Do you think I should change it?

> Adding it below the quantity input might be an alternative, though I am not sure really – especially depending on the translation the width might vary and not look good at all.

Agreed, that's too risky.

wiffbi:
> Visually, I like mine a little more, but don't have strong opinions. Do you think I should change it?

No, I also think yours looks better and I can find arguments for and against putting that info into the headline. So I’d leave it.
It just crossed my mind during review and because I was not sure wanted to check with what you think of it.

PR #5600, opened 2025-11-10
Questions: Allow type country code as dependency

raphaelm:
- [ ] Check android compatibility

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/items/question_edit.html
+++ b/src/pretix/control/templates/pretixcontrol/items/question_edit.html
@@ -143,6 +143,7 @@
                     </div>
                     <div class="col-md-5">
                         <script type="text/plain" id="dependency_value_val">{{ form.instance.dependency_values|escapejson_dumps }}</script>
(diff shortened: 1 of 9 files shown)


src/pretix/control/templates/pretixcontrol/items/question_edit.html
```diff
@@ -143,6 +143,7 @@ <h3>{% trans "Answer options" %}</h3>
                     </div>
                     <div class="col-md-5">
                         <script type="text/plain" id="dependency_value_val">{{ form.instance.dependency_values|escapejson_dumps }}</script>
+                        <script type="text/plain" id="countries">{{ countries|escapejson_dumps }}</script>
```
wiffbi, line 146: Is there a reason other than `dependency_value_val` using escapejson_dumps not to use djangos `json_script`? In my tests the encoding/output was the same.
```suggestion
                       {{ countries|json_script:"countries" }}
```

raphaelm:
We have no pressure here currently, let's wait on the questions rewrite to not create two tasks for android

PR #5602, opened 2025-11-10
Peppol: Live ID validation

raphaelm:


src/pretix/base/invoicing/peppol.py
```diff
@@ -136,6 +144,33 @@ def __call__(self, value):
...
+            smp_id = base64.b32encode(hashlib.sha256(value.lower().encode()).digest()).decode().rstrip("=")
+            found = False
+            for base_hostname in base_hostnames:
+                smp_domain = f'{smp_id}.iso6523-actorid-upis.{base_hostname}'
+                resolver = dns.resolver.Resolver()
+                try:
+                    answers = resolver.resolve(smp_domain, 'NAPTR', lifetime=1.0)
+                    if answers:
+                        found = True
+                        break
```
wiffbi: Wouldn’t it just make sense to directly return value similar to when a general Exception occurs? That way we can even skip the variable `found`
```suggestion
                        return value
```

src/pretix/base/invoicing/peppol.py
```diff
@@ -136,6 +144,33 @@ def __call__(self, value):
             raise ValidationError(_("The Peppol participant ID does not match the validation rules for the prefix "
                                     "%(number)s. Please reach out to us if you are sure this ID is correct."),
                                   params={"number": prefix})
+
+        if self.validate_online:
+            base_hostnames = ['edelivery.tech.ec.europa.eu']
+            if self.allow_testmode:
+                base_hostnames.append('acc.edelivery.tech.ec.europa.eu')
+            smp_id = base64.b32encode(hashlib.sha256(value.lower().encode()).digest()).decode().rstrip("=")
+            found = False
```
wiffbi:
```suggestion
```

src/pretix/base/invoicing/peppol.py
```diff
@@ -136,6 +144,33 @@ def __call__(self, value):
...
+                except Exception:  # noqa
+                    # Error likely on our end or infrastructure is down, allow user to proceed
+                    return value
+
+            if not found:
+                raise ValidationError(
+                    _("The Peppol participant ID is not registered on the Peppol network."),
+                )
+
         return value
```
wiffbi:
```suggestion
            raise ValidationError(
                _("The Peppol participant ID is not registered on the Peppol network."),
            )
```

PR #5603, opened 2025-11-10
Invoice issuer address: Add state field

raphaelm:


src/pretix/base/models/invoices.py
```diff
@@ -230,10 +244,22 @@ def full_invoice_from(self):
...
+                            code='{}-{}'.format(self.invoice_from_country, self.invoice_from_state)
+                        ).name
+                    except:
+                        pass
+
         parts = [
             self.invoice_from_name,
             self.invoice_from,
-            (self.invoice_from_zipcode or "") + " " + (self.invoice_from_city or ""),
+            ((self.invoice_from_zipcode or "") + " " + (self.invoice_from_city or "") + " " + (state_name or "")).strip(),
```
wiffbi: Very nitpicky, but if we have a zipcode and a state, but no city, the strip does not prevent the two spaces in the middle. We probably always have a city, so practically it never makes a difference and two spaces inbetweens is not a big deal either, but maybe do:

```suggestion
            " ".join(s for s in [self.invoice_from_zipcode, self.invoice_from_city, state_name] if s)
```

src/pretix/base/models/invoices.py
```diff
@@ -218,10 +219,23 @@ def full_invoice_from(self):
...
+            state_name = self.invoice_from_state
+            if str(self.invoice_from_country) in COUNTRIES_WITH_STATE_IN_ADDRESS:
+                if COUNTRIES_WITH_STATE_IN_ADDRESS[str(self.invoice_from_country)][1] == 'long':
+                    try:
+                        state_name = pycountry.subdivisions.get(
+                            code='{}-{}'.format(self.invoice_from_country, self.invoice_from_state)
+                        ).name
+                    except:
+                        pass
+
```
wiffbi, line 234: Wouldn’t it make sense to just reuse `self.address_invoice_from` and append `self.invoice_from_vat_id` and `taxidrow`. Either add an internal helper for the basic invoice_from that returns a list or just add to the string. Something like:

```
addr_str = self.address_invoice_from
if self.invoice_from_vat_id:
    addr_str += "\n" + pgettext("invoice", "VAT-ID: %s") % self.invoice_from_vat_id
addr_str += "\n" + taxidrow
return addr_str
```

src/pretix/base/settings.py
```diff
@@ -3971,6 +3991,16 @@ def validate_event_settings(event, settings_dict):
...
         })
+    if settings_dict.get('invoice_address_from_state') and settings_dict.get('invoice_address_from_country'):
+        cc = str(settings_dict.get('invoice_address_from_country'))
+        if cc not in COUNTRIES_WITH_STATE_IN_ADDRESS:
+            raise ValidationError(
+                {'invoice_address_from_state': ['States are not supported in country "{}".'.format(cc)]}
+            )
+        if not pycountry.subdivisions.get(code=cc + '-' + settings_dict.get('invoice_address_from_state')):
+            raise ValidationError(
+                {'state': ['"{}" is not a known subdivision of the country "{}".'.format(settings_dict.get('invoice_address_from_state'), cc)]}
```
wiffbi:
```suggestion
                {'invoice_address_from_state': ['"{}" is not a known subdivision of the country "{}".'.format(settings_dict.get('invoice_address_from_state'), cc)]}
```

src/pretix/static/pretixbase/js/addressform.js
```diff
@@ -5,6 +5,14 @@ $(function () {
     // to prevent fetching the same thing many times.
     var responseCache = {};
 
+    const cleanName = (name) => {
+        // Remove form prefix
+        name = name.split("-").pop();
+        // Remove settings prefix
+        name = name.replace(/^invoice_address_from_/, "");
```
wiffbi, line 12: I could not test a case, where this was necessary. Do I need to install pretix_peppol?

wiffbi (commented):
Looks good, just minor comments and one typo in prop-name when validating.

src/pretix/base/models/invoices.py
raphaelm: This feels outside the scope of the PR and I think it makes sense to keep them separate for the future as one should tend to "shorter" forms than the other (I was on the edge on handling the state differently but then didn't)

PR #5609, opened 2025-11-12
Auto-verify user email addresses on accepting invites

raphaelm:


src/pretix/control/views/auth.py
```diff
@@ -274,7 +277,8 @@ def invite(request, token):
                 user = User.objects.create_user(
                     form.cleaned_data['email'], form.cleaned_data['password'],
                     locale=request.LANGUAGE_CODE,
-                    timezone=request.timezone if hasattr(request, 'timezone') else settings.TIME_ZONE
+                    timezone=request.timezone if hasattr(request, 'timezone') else settings.TIME_ZONE,
+                    is_verified=form.cleaned_data['email'] == inv.email.lower()
```
wiffbi: Don’t we need a `.lower()` on the cleaned data as well? 
```suggestion
                    is_verified=form.cleaned_data['email'].lower() == inv.email.lower()
```

PR #5616, opened 2025-11-14
Accounting report: Allow subclasses to skip tables

raphaelm:


wiffbi (approved):
Definitions of `c_head` could be moved inside the `if s:` block, but that is a micro-optimization.

PR #5621, opened 2025-11-14
Bank transfer: Do not show reference before it is as complete as possible (fixes #5296)

### LINKED ISSUE #5296: Invoice Number Not Shown in Payment Reference Until "Continue"
### Problem and impact

The invoice number is not shown in the payment reference for Bank Transfer on the payment selection screen. It only appears after clicking "Continue".

Impact: low.

### Expected behaviour

The invoice number should appear immediately in the payment reference when the details are shown/expanded, next to the event shortcode and order number. It only appears after clicking "continue".

### Steps to reproduce

0) Enable setting "Include invoice number in the payment reference" (Event Settings > Payment > Bank Transfer)
0) Enable Bank Transfer and SEPA as payment options.
1) Place order that needs manual approval.
2) Approve order in backend.
3) Go to payment where to choose between SEPA and Bank Transfer ("Zahlungsmethode auswählen")
4) Select Bank Transfer.
5) Do not klick "Continue" and observe the expanded payment details:
The invoice number is not shown/contained in the reference initially. It only appears after clicking "Continue".

### Screenshots

_No response_

### Link

_No response_

_No response_

_No response_

### Version

2025.6.0

raphaelm:


src/pretix/plugins/banktransfer/payment.py
```diff
@@ -271,7 +271,7 @@ def payment_form_render(self, request, total=None, order=None) -> str:
             'request': request,
             'event': self.event,
             'settings': self.settings,
-            'code': self._code(order) if order else None,
+            'code': self._code(order, force=False) if order else None,
```
wiffbi, line 274: Isn’t `order` missing in this context? When I setup my invoicing and banktransfer so that it matches the late code-generating, create the order with a different payment method and then change to banktransfer, I still get the message „We will assign you a personal reference code to use after you completed the order.“ But my order already exists! Adding `order`  to the context here, fixes it for me.

```suggestion
            'code': self._code(order, force=False) if order else None,
            'order': order,
```

wiffbi (changes requested):

src/pretix/plugins/banktransfer/payment.py
```diff
@@ -313,11 +314,56 @@ def order_pending_mail_render(self, order, payment) -> str:
...
+            '',  # debtor address
+            'NON',
+            '',  # structured reference
+            self._code(payment.order, force=True),
+            'EPD',
+        ]
+
+        data_fields = [unidecode(d or '') for d in data_fields]
+        return '\r\n'.join(data_fields)
+
```
raphaelm:
```suggestion
```

PR #5625, opened 2025-11-14
Bank transfer: Consistancy in order of fields (fixes #5577)

### LINKED ISSUE #5577: Unify displaying of bank account details
With #5562 the fields for the bank details have been reordered on the payment pending page, but not on the other pages that display them (`src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/checkout_payment_form.html` and `src/pretix/plugins/banktransfer/templates/pretixplugins/banktransfer/checkout_confirm.html`).

I think it would be better to have the bank details displayed everywhere in the same way and would prefer having always the new layout.

raphaelm:


raphaelm:
It actually appears to be unused :eye: :wastebasket:

PR #5636, opened 2025-11-18
Import large package lazily to speed up startup

raphaelm:


src/pretix/base/models/tax.py
```diff
@@ -298,6 +297,8 @@ def cc_to_vat_prefix(country_code):
 @deconstructible
 class CustomRulesValidator:
     def __call__(self, value):
+        import jsonschema
```
wiffbi, line 300: src/pretix/base/pdf.py and src/pretix/base/models/seating.py both import jsonschema as well. Should we replace it there as well?

src/pretix/control/views/dashboards.py
```diff
@@ -341,6 +340,8 @@ def welcome_wizard_widget(sender, **kwargs):
 
 
 def event_index(request, organizer, event):
+    from pretix.control.forms.event import CommentForm
```
wiffbi, line 343: CommentForm is imported in src/pretix/control/views/event.py as well. Is it that big that it slows down startup-time?

raphaelm: Ah, so: This does not help for runserver, when runserver starts, all the views and forms are loaded anywhere.

However, dashboard.py is ran during plugin collection as well, since it contains signals. Hence, this import pulls in the entire form/view stack on *any* management command (like migrate). Moving this import solves that.

wiffbi (approved):

PR #5639, opened 2025-11-18
Add field length validation for invoice settings (Z#23215182)

ser8phin:


src/pretix/base/settings.py
```diff
@@ -725,8 +726,9 @@ def unserialize(cls, s):
                     message=lazy(lambda *args: _('Please only use the characters {allowed} in this field.').format(
                         allowed='A-Z, a-z, 0-9, -./:#'
                     ), str)()
-                )
+                ),
             ],
+            max_length=160,
```
wiffbi: When in TEST-mode, the prefix is extended with `"TEST-"`, so we probably need to make this 155 to really make sure, it works even in TEST-mode.

src/pretix/base/settings.py
```diff
@@ -747,8 +749,9 @@ def unserialize(cls, s):
...
```
wiffbi: Same as above, in TEST-mode `"TEST-"` is added at the end of the prefix.

PR #5640, opened 2025-11-18
Invoices: Increase retry interval

raphaelm:
e.g. Invopop states that receipt confirmation in italy can take 24h

### DIFF
--- a/src/pretix/base/services/invoices.py
+++ b/src/pretix/base/services/invoices.py
@@ -696,7 +696,7 @@ def retry_stuck_invoices(sender, **kwargs):
     with transaction.atomic():
         qs = Invoice.objects.filter(
             transmission_status=Invoice.TRANSMISSION_STATUS_INFLIGHT,
-            transmission_date__lte=now() - timedelta(hours=24),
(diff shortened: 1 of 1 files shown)


raphaelm:
What do you mean by "slip through"?

wiffbi:
What I mean by „slip through“ is, that if we have lots and lots of stuck invoices and also get lots of new ones quickly and we process them with the default ordering, we might never retry old stuck invoices as they are never read in the 5000-batch and get older and older and eventually pass the 48h mark. Not a very realistic scenario though.

raphaelm:
> What I mean by „slip through“ is, that if we have lots and lots of stuck invoices and also get lots of new ones quickly and we process them with the default ordering, we might never retry old stuck invoices as they are never read in the 5000-batch and get older and older and eventually pass the 48h mark. Not a very realistic scenario though.

Uh yes. But with reverse order, the other scenario would be a batch of old invoices that gets retried indefinitely but never succeeds. I think the only safe solution is dropping the batch.

However, the job runs every 10 minutes, so with the new 48h window there would need to be 144000 stuck invoices for this to happen, that is really unlikely to go unnoticed.

PR #5647, opened 2025-11-19
Invoice address: Improve VAT ID input

raphaelm:
Four related commits that make our VAT ID handling logic even more complex, but that's reality's fault, not ours:

- Replace dependency vat_moss with vendoring some code. We only use two functions of it, it is unmaintained for a long time, and it breaks on Python 3.14. Not worth it.

- Dynamic, country-specific labels for the VAT ID, since awareness around VAT IDs differs by EU country. For example, in Germany the VAT ID is assigned separately to each company and only used in cross-country transactions. Therefore, it makes sense to call it just "VAT ID" on the form, and people will either know their VAT ID or they don't. In contrast, in Italy the EU-compatible VAT ID is not separately assigned, but is just "IT" + the national tax number (Partita IVA) and also used on domestic transactions. So someone who never purchased something international for their company, might still know the value, if we call it the right way and not just "VAT ID".

- For the same reason, allow users to enter their ID without the national prefix, e.g. allow them to enter just 1234567890 and we'll autocomplete it to IT1234567890 if it looks like a plausible Italian VAT ID and the county is set to Italy. We always store the EU-normalized version with the country code. (I'm not quite sure we previously stored the normalized version in other cases as well, there might have been a bug, but now we do.)

- Allow organizers to switch the field to required for specific countries. This is necessary, because e.g. in Germany it makes never sense to have it required, since you can be a German business (or gov institution) without a VAT ID (e.g. small business). Also, a domestic German invoice must not have the VAT ID of the buyer on it. In contrast, in Italy or Greece you are simply not a business if you do not have a VAT ID, and the VAT ID is required to be printed on all invoices. For Italy, this is already solved if you turn on a plugin that implements SdI invoice transmission, however this PR will still make a customer very happy (Z#23203448). Also, in Greece, there is (for now) no national transmission type, but there is a legal requirement to have the TIN (= VAT ID minus the prefix) on every invoice ([source](https://www.gov.gr/en/sdg/taxes/vat/general/invoices)). Therefore this PR is a prerequisite for proper invoicing in Greece with pretix.

### DIFF
--- a/src/pretix/base/forms/questions.py
+++ b/src/pretix/base/forms/questions.py
@@ -1357,6 +1357,7 @@ class BaseInvoiceAddressForm(forms.ModelForm):
                 )
 
         vat_id_applicable = (
+            'vat_id' in self.fields and
             data.get('is_business') and
             ask_for_vat_id(data.get('country'))
         )
--- a/src/pretix/base/settings.py
+++ b/src/pretix/base/settings.py
@@ -659,7 +659,7 @@ DEFAULTS = {
             ),
             widget=forms.CheckboxSelectMultiple(attrs={
                 "class": "scrolling-multiple-choice",
-                'data-display-dependency': '#invoice_address_vatid,#id_invoice_address_asked'
+                'data-display-dependency': '#id_invoice_address_vatid'
             }),
         )
     },


src/pretix/base/settings.py
```diff
@@ -629,13 +629,40 @@ def unserialize(cls, s):
...
+        ),
+        'form_kwargs': dict(
+            label=_("Require VAT ID in"),
+            choices=lazy(
+                lambda *args: sorted([(cc, gettext(Country(cc).name)) for cc in VAT_ID_COUNTRIES], key=lambda c: c[1]),
+                list
+            )(),
+            help_text=format_lazy(
+                _("VAT ID is optional by default, because not all businesses are assigned a VAT ID in all countries. "
+                  "VAT ID will be required for all busines addresses in the selected countries."),
```
luelista:
```suggestion
                  "VAT ID will be required for all business addresses in the selected countries."),
```

src/pretix/base/forms/questions.py
```diff
@@ -1358,20 +1356,33 @@ def clean(self):
                                             "transmission method.")}
                 )
 
+        vat_id_applicable = (
+            data.get('is_business') and
```
luelista, line 1361: this lacks a check whether the field is visible at all, doesn't it?
```suggestion
            self.event.settings.invoice_address and
            data.get('is_business') and
```

src/pretix/base/settings.py
```diff
@@ -629,13 +629,40 @@ def unserialize(cls, s):
...
+            ),
+            widget=forms.CheckboxSelectMultiple(attrs={
+                "class": "scrolling-multiple-choice",
+                'data-display-dependency': '#invoice_address_vatid,#id_invoice_address_asked'
```
luelista:
```suggestion
                'data-display-dependency': '#id_invoice_address_vatid'
```

PR #5648, opened 2025-11-20
Add "Pay by bank" option for UK customers via Stripe

14praveenk:
This pull request adds support for the "Pay by bank (UK)" payment method via Stripe, specifically for GBP transactions and UK customers. The changes include the introduction of a new payment provider class, updates to the settings form to allow enabling/disabling this method, and registration of the provider so it is available in the system.

**New payment method integration:**

* Added the `StripePayByBank` class in `payment.py`, implementing the logic for the "Pay by bank (UK)" payment method, including eligibility checks and payment intent configuration.

**Settings and configuration:**

* Updated the `settings_form_fields` method in `payment.py` to include a new boolean field for enabling/disabling "Pay by bank (UK)", with appropriate help text and GBP currency restriction.

**Provider registration:**

* Registered the new `StripePayByBank` provider in `signals.py` so it becomes available as a payment option in the system.

### DIFF
--- a/src/pretix/plugins/stripe/payment.py
+++ b/src/pretix/plugins/stripe/payment.py
@@ -412,15 +412,15 @@ class StripeSettingsHolder(BasePaymentProvider):
                                  'before they work properly.'),
                      required=False,
                  )),
-
                 ('method_pay_by_bank',
                  forms.BooleanField(
-                     label=_('Pay by bank (UK)'),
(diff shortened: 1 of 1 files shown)


raphaelm (approved):
I loosened the texts and currency restrictions a little since apparently Stripe already has a private beta running for Germany and France: https://docs.stripe.com/payments/pay-by-bank

PR #5649, opened 2025-11-20
Hide adress info from invisible fields in confirmation step (Z#23212149)

ser8phin:


ser8phin:
Maybe we can set visibility for each and not only for all types?

raphaelm (changes requested):
See thread on Slack

src/pretix/base/models/orders.py
```diff
@@ -3526,11 +3526,30 @@ def for_js(self):
...
 
-        t, __ = transmission_types.get(identifier=self.transmission_type)
+        t, m = transmission_types.get(identifier=self.transmission_type)
+
+        d_event = self.order.event if self.order else event
+        # reusing hack from questions.py for default transmission -- this should at least be fast
+        if d_event:
+            if (
+                t.identifier == "email" and
+                m in ("transmission_email_other", "transmission_email_address") and
```
raphaelm: I don't think m is a field name here?

src/pretix/base/models/orders.py
```diff
@@ -3526,11 +3526,30 @@ def for_js(self):
...
+                (
+                    d_event.settings.invoice_generate == "False" or
+                    not d_event.settings.invoice_email_attachment
+                )
+            ):
+                return data
+
+        # we also don't show transmission data if we never showed the corresponding form fields
+        if not t.invoice_address_form_fields_visible(country=self.country, is_business=self.is_business):
```
raphaelm: This works if the TransmissionType is an all-or-nothing thing, but what if it returns only one of the fields? Also, this still can't really work with the dependency between the two fields in the email transmission type. I'd really recommend adding a TransmissionType.transmission_info_to_display_data or even TransmissionType.describe_transmission to allow this to be implemented properly.

raphaelm (changes requested):

PR #5666, opened 2025-11-28
Add support for multiple linked_orderpositions on reusable media

wiffbi:
This PR adds a ManyToManyField on ReusableMedium for linked_orderpositions. It keeps linked_orderposition in the API for backwards-compatibility. If the medium has more than one linked_orderposition, then the API does not output property `linked_orderposition` for that medium and blocks PATCH requests with `linked_orderposition` as invalid.

It also adds a field secret to the medium, so in the future, the medium can be claimed by a customer through the identifier and secret.

internal note: update & merge pretixpos-android!361 after merging this pr


doc/api/resources/reusablemedia.rst
```diff
@@ -21,12 +21,15 @@ id                                    integer                    Internal ID of
...
+secret                                string                     Secret to identify the medium (or ``null``)
+label                                 string                     Label to identify the medium, usually something human readable (or ``null``)
 active                                boolean                    Whether this medium may be used.
 created                               datetime                   Date of creation
 updated                               datetime                   Date of last modification
 expires                               datetime                   Expiry date (or ``null``)
 customer                              string                     Identifier of a customer account this medium belongs to.
-linked_orderposition                  integer                    Internal ID of a ticket this medium is linked to.
+linked_orderpositions                 list of integer            Internal IDs of tickets this medium is linked to.
+linked_orderposition                  integer                    Internal ID of a ticket this medium is linked to. (**deprecated**, do not use, will be removed)
```
raphaelm: For consistency with other API docs:

```suggestion
linked_orderposition                  integer                    **DEPRECATED.** ID of one the ticket the medium is linked to, if it is linked to only one ticket. ``null``, if the medium is linked to none or multiple tickets.
```

doc/api/resources/reusablemedia.rst
```diff
@@ -21,12 +21,15 @@ id                                    integer                    Internal ID of
 type                                  string                     Type of medium, e.g. ``"barcode"``, ``"nfc_uid"`` or ``"nfc_mf0aes"``.
 organizer                             string                     Organizer slug of the organizer who "owns" this medium.
 identifier                            string                     Unique identifier of the medium. The format depends on the ``type``.
+secret                                string                     Secret to identify the medium (or ``null``)
```
raphaelm: That appears confusing. When comparing this with OrderPosition or GiftCard together with this description, one could think that `secret` is the thing that is printed on the medium and scanned, such as the barcode content. However, that is not what this `secret` is for. If I recall correctly, this is more like a "token to add this to a customer account". 

At the minimum, the description here should make that clearer, but ideally we find another name than "secret" that makes the purpose more clear. Could very well be that we need other kinds of secret things on this model in the future. Maybe `claim_token` or `claim_secret`?

src/pretix/api/serializers/media.py
```diff
@@ -97,6 +110,19 @@ def __init__(self, *args, **kwargs):
 
     def validate(self, data):
         data = super().validate(data)
+        linked_orderposition = data.pop('linked_orderposition', None)
+        if linked_orderposition:
```
raphaelm: I think you should actually check `"linked_orderposition" in data` without the fallback to None.

Your current code would allow PATCH/POST with
```
{
"linked_orderposition": null,
"linked_orderpositions": [3]
}
```
to get interpreted, but I think it should be considered invalid and return a 400. Especially on PATCH it's really ambiguous otherwise.

src/pretix/api/serializers/order.py
```diff
@@ -1655,10 +1655,9 @@ def create(self, validated_data):
                         answ.options.add(*options)
 
                 if use_reusable_medium:
-                    use_reusable_medium.linked_orderposition = pos
-                    use_reusable_medium.save(update_fields=['linked_orderposition'])
+                    use_reusable_medium.linked_orderpositions.add(pos)
```
raphaelm: This is in conflict with the API docs:

> ``use_reusable_medium`` (optional, causes the new ticket to take over the given reusable medium, identified by its ID)

API consumers expect ``use_reusable_medium`` to *replace*, not *add*. I think we might need a new ``add_to_reusable_medium`` next to it as well so API consumers can make that choice conciously?

src/pretix/api/views/checkin.py
```diff
@@ -636,9 +636,11 @@ def _make_context(context, event):
...
                     'list': MiniCheckinListSerializer(checkinlists[0]).data,
                 }, status=404)
-            op_candidates = [media.linked_orderposition]
-            if list_by_event[media.linked_orderposition.order.event_id].addon_match:
-                op_candidates += list(media.linked_orderposition.addons.all())
+            op_candidates = []
+            for op in media.linked_orderpositions.all().prefetch_related("order"):
+                op_candidates.append(op)
+                if list_by_event[op.order.event_id].addon_match:
+                    op_candidates += list(op.addons.all())
```
raphaelm: As discussed this morning, we should also check the following case: If we have multiple op_candidates, but only one is active with respect to op.valid_from/op.valid_until, we should auto-select that one.

This should also be something with good test coverage.

src/pretix/base/exporters/reusablemedia.py
```diff
@@ -58,17 +61,28 @@ def iterate_list(self, form_data):
...
+                    medium.type,
+                    medium.identifier,
+                    _('Yes') if medium.active else _('No'),
+                    date_format(medium.expires, 'SHORT_DATETIME_FORMAT') if medium.expires else '',
+                    medium.customer.identifier if medium.customer_id else '',
+                    f"{op.order.code}-{op.positionid}",
+                    medium.linked_giftcard.secret if medium.linked_giftcard_id else '',
+                    medium.notes,
+                ]
+            else:
```
raphaelm: This appears as you want this to run if the loop is empty, however that is not how [for-else](https://book.pythontips.com/en/latest/for_-_else.html#else-clause) works in Python

src/pretix/base/exporters/reusablemedia.py
```diff
@@ -58,17 +61,28 @@ def iterate_list(self, form_data):
...
-                medium.identifier,
-                _('Yes') if medium.active else _('No'),
-                date_format(medium.expires, 'SHORT_DATETIME_FORMAT') if medium.expires else '',
-                medium.customer.identifier if medium.customer_id else '',
-                f"{medium.linked_orderposition.order.code}-{medium.linked_orderposition.positionid}" if medium.linked_orderposition_id else '',
-                medium.linked_giftcard.secret if medium.linked_giftcard_id else '',
-                medium.notes,
-            ]
-            yield row
+            for op in medium.linked_orderpositions:
```
raphaelm: This loop is surprising if you are expecting a list of reusable media. A comma-separated value of order codes might be more expected for many use-cases?

src/pretix/base/migrations/0301_reusablemedium_remove_orderposition.py
```diff
@@ -0,0 +1,44 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ("pretixbase", "0297_add_reusablemedium_secret"),
+    ]
+
+    operations = [
+        # according to the docs, UPDATE FROM should run similarly on sqlite and postgres, but I could not get it to work
+        # so roll back the data migration with code before deleting data from through-table in 0297
+        migrations.RunPython(migrations.RunPython.noop, reverse),
```
raphaelm, line 31: The rollback-only migration adds complexity; I'm fine with it here, but generally as a tip if you only need this during dev to test the migration: I tend do just `cp` my SQLite file to snapshot the database before the migration and then revert back to that :)

src/pretix/base/models/media.py
```diff
@@ -72,6 +72,16 @@ class ReusableMedium(LoggedModel):
         max_length=200,
         verbose_name=pgettext_lazy('reusable_medium', 'Identifier'),
     )
+    secret = models.CharField(
+        max_length=200,
+        verbose_name=pgettext_lazy('reusable_medium', 'Secret'),
+        null=True, blank=True
+    )
```
raphaelm: See comment on naming in reusablemedia.rst

src/pretix/base/models/media.py
```diff
@@ -89,12 +99,10 @@ class ReusableMedium(LoggedModel):
...
         verbose_name=_('Customer account'),
     )
-    linked_orderposition = models.ForeignKey(
+    linked_orderpositions = models.ManyToManyField(
         OrderPosition,
-        null=True, blank=True,
         related_name='linked_media',
-        on_delete=models.SET_NULL,
-        verbose_name=_('Linked ticket'),
+        verbose_name=_('Linked tickets'),
```
raphaelm, line 105: Maybe a help text that explains that one should only link tickets that have no overlap in validity to avoid problems at check-in?

src/pretix/control/logdisplay.py
```diff
@@ -733,6 +733,7 @@ class CoreUserImpersonatedLogEntryType(UserImpersonatedLogEntryType):
     'pretix.reusable_medium.created': _('The reusable medium has been created.'),
     'pretix.reusable_medium.created.auto': _('The reusable medium has been created automatically.'),
     'pretix.reusable_medium.changed': _('The reusable medium has been changed.'),
+    'pretix.reusable_medium.linked_orderposition.added': _('A new ticket has been added to the medium.'),
```
raphaelm, line 746: Should we also add a text for removed as well?

src/pretix/control/templates/pretixcontrol/organizers/reusable_medium.html
```diff
@@ -49,13 +49,13 @@ <h3 class="panel-title">
...
+                                {% for op in medium.linked_orderpositions.all %}
                                     <span class="helper-display-block">
                                     <span class="fa fa-ticket fa-fw"></span>
-                                    <a href="{% url "control:event.order" event=medium.linked_orderposition.order.event.slug organizer=request.organizer.slug code=medium.linked_orderposition.order.code %}">
-                                        {{ medium.linked_orderposition.order.code }}</a>-{{ medium.linked_orderposition.positionid }}
+                                    <a href="{% url "control:event.order" event=op.order.event.slug organizer=request.organizer.slug code=op.order.code %}">
+                                        {{ op.order.code }}</a>-{{ op.positionid }}
                                 </span>
-                                {% endif %}
+                                {% endfor %}
```
raphaelm: Indentation is broken here

src/tests/api/test_checkinrpc.py
```diff
@@ -286,12 +286,12 @@ def test_by_secret_special_chars(token_client, organizer, clist, event, order):
 @pytest.mark.django_db
 def test_by_medium(token_client, organizer, clist, event, order):
```
raphaelm, line 287: This file needs tests for the interesting cases with multiple positions on the medium

src/tests/api/test_order_create.py
```diff
@@ -3086,7 +3086,7 @@ def test_order_create_use_medium(token_client, organizer, event, item, quota, qu
     with scopes_disabled():
         o = Order.objects.get(code=resp.data['code'])
         medium.refresh_from_db()
-        assert o.positions.first() == medium.linked_orderposition
+        assert o.positions.first() == medium.linked_orderpositions.first()
```
raphaelm, line 3124: This might need more tests for the new cases that might arise, see comment on the API implementation

src/tests/api/test_reusable_media.py
```diff
@@ -419,7 +422,7 @@ def test_medium_lookup_cross_organizer(token_client, organizer, organizer2, org2
         ticket = org2_event.items.create(name='Early-bird ticket', category=None, default_price=23, admission=True,
                                          personalized=True)
         op = o.positions.create(item=ticket, price=Decimal("14"))
-        medium2.linked_orderposition = op
+        medium2.linked_orderpositions.add(op)
         medium2.linked_giftcard = giftcard2
         medium2.save()
```
raphaelm, line 712: This requires tests for all the combinations of linked_orderposition and linked_orderpositions to make sure the fallbacks work

doc/api/resources/reusablemedia.rst
wiffbi: I like `claim_token`!

doc/api/resources/reusablemedia.rst
wiffbi: I took the format from from carts includes_tax :) will update that there as well

src/pretix/api/serializers/media.py
wiffbi: Good catch! I just thought about when it is set with a value, but setting it to None means clearing it, so yes. I will fix it.

src/pretix/base/exporters/reusablemedia.py
wiffbi: Oh, you are right! I just tested `for i in []:else:` which only seems to work as expected.

src/pretix/base/exporters/reusablemedia.py
wiffbi: Makes sense!

src/pretix/api/views/checkin.py
wiffbi: I added this filter later in line 660 – I did not want to change the orignal behaviour and only fall back to valid_from if needed. I added test_by_medium_multiple_orderpositions.

src/pretix/base/migrations/0301_reusablemedium_remove_orderposition.py
wiffbi: Okay, I thought it might be helpful to have a safe way back – assuming the rollback would be executed immediately after an update fail. But it’s a good tip to shapshot the sqlite-file, thanks!

src/pretix/base/models/media.py
wiffbi: Good idea, I added a help text.

src/pretix/control/logdisplay.py
wiffbi: Hm, might be helpful. We haven’t had it before and currently there is no code to handle removing linked_orderpositions other than fully overwriting the set. Do you think it is worth the effort?

src/pretix/control/templates/pretixcontrol/organizers/reusable_medium.html
wiffbi: Ah, good catch! It’s been broken before and I haven’t noticed when updating to the loop. Is fixed.

robbi5:
Screenshot 2026-01-22 at 14 13 22

Currently, when editing a reusable medium, the order code `3UYXK` and position `-1` is only shown for things I'm adding fresh to the select. The already linked tickets are only displayed with their position, but the order code would be great to know to differentiate them :)

src/pretix/control/logdisplay.py
wiffbi: I added `pretix.reusable_medium.linked_orderposition.removed` when creating an order through the API and `use_reusable_medium` (instead of `add_to_reusable_medium`) is used. Currently, when changing the reusable medium directly through the API, there is a `pretix.reusable_medium.changed` with all new data. No explicit removed-logentry though. Good enough?

wiffbi:
> Screenshot 2026-01-22 at 14 13 22
> 
> Currently, when editing a reusable medium, the order code `3UYXK` and position `-1` is only shown for things I'm adding fresh to the select. The already linked tickets are only displayed with their position, but the order code would be great to know to differentiate them :)

Good idea, when using it as a single-choice field, it already had a label_from_instance that does this. Added this to the multiple-choice version as well.

src/tests/api/test_checkinrpc.py
wiffbi: added test_by_medium_multiple_orderpositions

src/tests/api/test_reusable_media.py
wiffbi: Done

src/pretix/control/templates/pretixcontrol/organizers/reusable_medium.html
```diff
@@ -41,34 +53,34 @@ <h3 class="panel-title">
...
+                            {% for op in medium.linked_orderpositions.all %}
                                 <span class="helper-display-block">
-                                <span class="fa fa-ticket fa-fw"></span>
-                                <a href="{% url "control:event.order" event=medium.linked_orderposition.order.event.slug organizer=request.organizer.slug code=medium.linked_orderposition.order.code %}">
-                                    {{ medium.linked_orderposition.order.code }}</a>-{{ medium.linked_orderposition.positionid }}
-                            </span>
-                            {% endif %}
+                                    <span class="fa fa-ticket fa-fw"></span>
+                                    <a href="{% url "control:event.order" event=op.order.event.slug organizer=request.organizer.slug code=op.order.code %}">
+                                        {{ op.order.code }}</a>-{{ op.positionid }}
```
wiffbi, line 70: This leaks order-code info to users that potentially do not have access to orders on event the event. In the API we output just the ID of the orderposition. Is the order-code worth hiding and should we show some kind of "****** (you do not have sufficient permissions")

wiffbi: As descussed: leave as is.

doc/api/resources/reusablemedia.rst
```diff
@@ -39,6 +43,14 @@ Existing media types are:
 - ``nfc_uid``
 - ``nfc_mf0aes``
 
+
+.. versionchanged:: 2025.11
```
raphaelm:
```suggestion
.. versionchanged:: 2026.5
```

doc/api/resources/reusablemedia.rst
```diff
@@ -21,12 +21,16 @@ id                                    integer                    Internal ID of
...
 identifier                            string                     Unique identifier of the medium. The format depends on the ``type``.
+claim_token                           string                     Secret token to claim ownership of the medium (or ``null``)
+label                                 string                     Label to identify the medium, usually something human readable (or ``null``)
 active                                boolean                    Whether this medium may be used.
 created                               datetime                   Date of creation
 updated                               datetime                   Date of last modification
 expires                               datetime                   Expiry date (or ``null``)
 customer                              string                     Identifier of a customer account this medium belongs to.
-linked_orderposition                  integer                    Internal ID of a ticket this medium is linked to.
+linked_orderpositions                 list of integer            Internal IDs of tickets this medium is linked to.
```
raphaelm:
```suggestion
linked_orderpositions                 list of integers           Internal IDs of tickets this medium is linked to.
```

doc/api/resources/reusablemedia.rst
```diff
@@ -92,10 +105,13 @@ Endpoints
    :query string customer: Only show media linked to the given customer.
    :query string created_since: Only show media created since a given date.
    :query string updated_since: Only show media updated since a given date.
+   :query integer linked_orderpositions: Only show media linked to the given tickets. Note: you can pass multiple ticket IDs by passing
+                         ``linked_orderpositions`` multiple times. Any medium matching any linked orderposition will be returned.
```
raphaelm:
```suggestion
                                         ``linked_orderpositions`` multiple times. Any medium matching any linked orderposition will be returned.
```

src/pretix/api/serializers/media.py
```diff
@@ -106,6 +117,21 @@ def __init__(self, *args, **kwargs):
...
+        if 'linked_orderposition' in data:
+            linked_orderposition = data['linked_orderposition']
+            # backwards-compatibility
+            if 'linked_orderpositions' in data:
+                raise ValidationError({
+                    'linked_orderposition': _('You cannot use linked_orderposition and linked_orderpositions at the same time.')
+                })
+            if self.instance and self.instance.linked_orderpositions.count() > 1:
+                raise ValidationError({
+                    'linked_orderposition': _('There are more than one linked_orderposition. You need to use linked_orderpositions.')
```
raphaelm: We typically do not translate API error messages that are very technical

```suggestion
                    'linked_orderposition': 'You cannot use linked_orderposition and linked_orderpositions at the same time.'
                })
            if self.instance and self.instance.linked_orderpositions.count() > 1:
                raise ValidationError({
                    'linked_orderposition': 'There are more than one linked_orderposition. You need to use linked_orderpositions.'
```

src/pretix/api/serializers/media.py
```diff
@@ -121,14 +147,28 @@ def validate(self, data):
...
+        ops = r.get('linked_orderpositions', [])
         # late permission evaluations for checks that depend on the actual linked events
         expand_nested = self.context['request'].query_params.getlist('expand')
         perm_holder = request.auth if isinstance(request.auth, (Device, TeamAPIToken)) else request.user
-        if 'linked_orderposition' in expand_nested:
-            if instance.linked_orderposition is not None:
-                event = instance.linked_orderposition.order.event
+        if ops and 'linked_orderposition' in expand_nested or 'linked_orderpositions' in expand_nested:
+            ops_noperm = []
+            for lop in instance.linked_orderpositions.all().prefetch_related('order__event', 'order__event__organizer'):
```
raphaelm: Shouldn't this prefetch happen on the ViewSet level to prevent N+1 query problem on a list of media?

src/pretix/api/serializers/order.py
```diff
@@ -1700,15 +1715,30 @@ def create(self, validated_data):
...
+                        )
+                    use_reusable_medium.linked_orderpositions.set([pos])
                     use_reusable_medium.log_action(
                         'pretix.reusable_medium.linked_orderposition.changed',
                         data={
                             'by_order': order.code,
                             'linked_orderposition': pos.pk,
                         }
                     )
+                if add_to_reusable_medium:
```
raphaelm:
```suggestion
                elif add_to_reusable_medium:
```
to further highlight that these are alternative, not additive code paths?

src/pretix/api/serializers/order.py
```diff
@@ -1700,15 +1715,30 @@ def create(self, validated_data):
...
+                    for op_pk in use_reusable_medium.linked_orderpositions.values_list('pk', flat=True):
+                        use_reusable_medium.log_action(
+                            'pretix.reusable_medium.linked_orderposition.removed',
+                            data={
+                                'linked_orderposition': op_pk,
+                            }
+                        )
+                    use_reusable_medium.linked_orderpositions.set([pos])
                     use_reusable_medium.log_action(
                         'pretix.reusable_medium.linked_orderposition.changed',
```
raphaelm: Should we use `added` always in the future instead of `changed`?

src/pretix/api/views/checkin.py
```diff
@@ -628,7 +629,8 @@ def _make_context(context, event):
                     'list': MiniCheckinListSerializer(list_by_event[revoked_matches[0].event_id]).data,
                 }, status=400)
         else:
-            if media.linked_orderposition.order.event_id not in list_by_event:
+            linked_event_ids = media.linked_orderpositions.values_list("order__event_id", flat=True).order_by().distinct()
```
raphaelm: In the most likely case, we'll query the actual orderposition objects down below anyways, so we could do it here already and avoid this query?

src/pretix/api/views/checkin.py
```diff
@@ -654,21 +656,34 @@ def _make_context(context, event):
                     'checkin_texts': [],
                     'list': MiniCheckinListSerializer(checkinlists[0]).data,
                 }, status=404)
-            op_candidates = [media.linked_orderposition]
-            if list_by_event[media.linked_orderposition.order.event_id].addon_match:
-                op_candidates += list(media.linked_orderposition.addons.all())
+            op_candidates = []
+            for op in media.linked_orderpositions.all().select_related("order"):
```
raphaelm:
```suggestion
            for op in media.linked_orderpositions.all().select_related("order").prefetch_related("addons"):
```
maybe?

src/pretix/api/views/checkin.py
```diff
@@ -654,21 +656,34 @@ def _make_context(context, event):
...
+                )
+            ]
+        else:
+            op_candidates_matching_product = [
+                op for op in op_candidates
+                if (
+                    (not op.valid_from or op.valid_from < now()) and
+                    (not op.valid_until or op.valid_until > now())
+                )
+            ]
```
raphaelm: I think the assumption that we never want to have a "mix" of these two features is a little dangerous, some day someone will mix addons with validity. Would something prevent us from applying both filters?

(if we change this, we also need the change on the android side for offline scaning)

src/pretix/base/migrations/0300_add_reusablemedium_label.py
```diff
@@ -0,0 +1,35 @@
+# Generated by Django 4.2.26 on 2025-11-24 11:32
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
raphaelm, line 6: Just a note for us that we run this migration manually through SQL before deploying this PR

src/tests/api/test_checkinrpc.py
```diff
@@ -301,6 +301,48 @@ def test_by_medium(token_client, organizer, clist, event, order):
...
+    with freeze_time("2020-01-01 10:45:00"):
+        resp = _redeem(token_client, organizer, clist, "abcdef", {"source_type": "barcode"})
+        assert resp.status_code == 201
+        assert resp.data['status'] == 'ok'
+
+    with freeze_time("2020-01-01 15:45:00"):
+        resp = _redeem(token_client, organizer, clist, "abcdef", {"source_type": "barcode"})
+        assert resp.status_code == 400
+        assert resp.data['status'] == 'error'
+        assert resp.data['reason'] == 'already_redeemed'
```
raphaelm, line 344: Could you also check the case where NO ticket is valid that it gives the expected error message?

src/tests/api/test_reusable_media.py
```diff
@@ -352,6 +355,90 @@ def test_medium_create(token_client, organizer, giftcard):
...
+
+    payload = dict(TEST_MEDIUM_CREATE_PAYLOAD)
+
+    # wrong organizer for orderposition
+    resp = token_client.post(
+        '/api/v1/organizers/{}/reusablemedia/'.format(organizer2.slug),
+        payload,
+        format='json'
+    )
+    assert resp.status_code == 403
```
raphaelm: This is not really testing the validation in the serializer, is it? Looks like it is testing the access control level of the API endpoint, which is not interesting here. TEST_MEDIUM_CREATE_PAYLOAD does not even contain a linked_orderposition.

src/pretix/control/views/organizer.py
```diff
@@ -3466,10 +3472,13 @@ def get_object(self, queryset=None):
...
     def form_valid(self, form):
         if form.has_changed():
-            self.object.log_action('pretix.reusable_medium.changed', user=self.request.user, data={
+            data = {
                 k: getattr(self.object, k)
                 for k in form.changed_data
-            })
+            }
+            if "linked_orderpositions" in data:
+                data["linked_orderpositions"] = data["linked_orderpositions"].values_list("pk", flat=True)
```
raphaelm: This does not seem to work as expected.

I tested with a medium that currently has no tickets attached. Then I linked two ticket, and the log entry showed `"linked_orderpositions": []`. Then I removed one of them, and now it showed `  "linked_orderpositions": [434, 17368]`. I suspect this is using the "value before change" instead of "value after change"?

src/pretix/api/serializers/media.py
wiffbi: Good catch, I will look into this!

src/pretix/control/views/organizer.py
```diff
@@ -3466,10 +3472,13 @@ def get_object(self, queryset=None):
...
+            self.object.log_action('pretix.reusable_medium.changed', user=self.request.user, data=data)
```
raphaelm: I don't like how "I link a ticket to a medium" can result in three different log entry types:

- pretix.reusable_medium.changed if I do it in the UI
- pretix.reusable_medium.linked_orderposition.added if I do it through the order API with add_to
- pretix.reusable_medium.linked_orderposition.changed if I do it through the order API with use

If we ever add webhooks for this, this is unexpected and weird for API consumers. I think we should normalize it:

- Always use pretix.reusable_medium.linked_orderposition.added and pretix.reusable_medium.linked_orderposition.removed
- Drop pretix.reusable_medium.linked_orderposition.changed (except for historical reasons)
- Use pretix.reusable_medium.changed only for other changes of the medium.

src/pretix/api/serializers/order.py
wiffbi: Hm, semantically it is different, but it probably does not help much when debugging something. So probably yes. As we explicitly log the removal anyways, then we can just use `added` for both cases. I will added that.

wiffbi: Should I leave a comment in logdisplay.py to make sure, we keep `changed` around to properly display old logs?

src/tests/api/test_reusable_media.py
wiffbi: You are right, this just tests access control. According to the comment, it should test for adding a linked_orderposition that does not belong to the organizer. Will fix.

src/tests/api/test_checkinrpc.py
wiffbi: Added tests for when both tickets are out of the selected timerange and another one for when both ops are canceled.

src/tests/api/test_reusable_media.py
wiffbi: I added tests for mixing ops from different organizers.

src/pretix/control/views/organizer.py
wiffbi: Good catch! Fixed it by running `super().form_valid(form)` before the logging code.

src/pretix/control/views/organizer.py
wiffbi: Agreed. How do we handle `ReusableMediaViewSet`? It does have `perform_update`, which can result in added linked_ops – I guess I need to add special handling there if the request contains orderpositions. What about `perform_create` when media is added in the creation? Or do we just make everything `pretix.reusable_medium.changed` and let the client decide what to do with it? Especially the remove-log makes it a bit harder to code.

wiffbi: Logging this isn’t the most straight-forward thing, but IMHO okay if we really want to make that distinciton – see commit https://github.com/pretix/pretix/pull/5666/changes/d0b94332db34084f43d3905eac239fdfbe143d3f

src/pretix/api/serializers/media.py
wiffbi: `ReusableMediaViewSet.get_queryset` already does the prefetch – or to be more precise, does a `select_related`. Not sure it would help much to prefetch instead of select e.g. the organizer as I would assume the list will be only of one organizer most of the time. But I guess that would be a premature optimization.

src/pretix/api/views/checkin.py
wiffbi: Good idea, fetched to ops here and create a set of event_ids from it.

src/pretix/api/views/checkin.py
wiffbi: Good idea, will reduce N+1 if addon_match is set on the event.

src/pretix/api/views/checkin.py
wiffbi: Should be possible. I am actually not sure why there is this condition that we only check addons if there is at most one non-addon product. We might have multiple linked_orderpositions and for both of them addon_match could be active. I will update accordingly.

wiffbi: I updated the check here to combine time-based validity with addons-product-matching. I actually had a hard time to get my head around it and whether checkin does the right thing now (according to the tests, it does, but I am unsure). When multiple op_candidates are available this PR basically does a pre-checkin-check based on validity-time and product – before this was done only in the addons-case, but due to multiple linked_orderpositions this needs to be done now whenever there are multiple op_candidates. It also skips the check on clist.addon_match here as addons through matched ops are already only added if clist.addon_match is True. Also otherwise op_candidates aquired through multiple linked_orderpositions would now fall through as their secret does not match the barcode.

PR #5680, opened 2025-12-02
Refactor payment QR code generation code and add SPAYD format

luelista:
https://en.wikipedia.org/wiki/Short_Payment_Descriptor

### DIFF
--- a/src/tests/helpers/test_payment.py
+++ b/src/tests/helpers/test_payment.py
@@ -122,3 +122,20 @@ def test_payment_qr_codes_swiss(env):
         "TESTVERANST-12345",
         "EPD",
     ])
+
+
+@pytest.mark.django_db
+def test_payment_qr_codes_spayd(env):
+    o, event = env
+    codes = generate_payment_qr_codes(
+        event=event,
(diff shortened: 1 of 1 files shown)


luelista:
We could then use the new method in other plugins like pretix-mollie or pretix-secuconnect, where the QR code templates are currently duplicated.

src/pretix/helpers/payment.py
```diff
@@ -0,0 +1,208 @@
...
+
+
+def euro_bezahlcode(
+        event,
+        code,
+        amount,
+        bank_details_sepa_bic,
+        bank_details_sepa_name,
+        bank_details_sepa_iban,
+):
```
raphaelm, line 105: Not really part of this PR, but while we're at it, we could probably limit this to IBANs starting with DE. It's a very German thing and more or less on its way out.

src/tests/helpers/test_payment.py
bockstaller, line 1: I might add a test case for SPAYD as well, just because its tedious to verify it with manual testing

bockstaller (approved):

PR #5684, opened 2025-12-03
Fix N+1 queries in API

wiffbi:
Reduces number of SQL-queries in my modest test-event for quotas from 33 to 7, for membership from 10 to 4, etc.

~~In the third commit, this PR adds a prefetch for Voucher.budget_used, which is basically copied from `Voucher.annotate_budget_used_orders`. I tried to use `Voucher.annotate_budget_used_orders` instead and just rename the annotated `budget_used_orders` to `budget_used` like so (I need to rename it so VoucherSerializer recognizes it and doesn’t add its N+1-queries for budget_used):~~

```
        return Voucher.annotate_budget_used_orders(
            self.request.event.vouchers
        ).annotate(
            budget_used=F('budget_used_orders')
        ).select_related(
            'item', 'quota', 'seat', 'variation'
        )
```

~~But this added the following to the SQL-query, which is an exact copy from what `annotate_budget_used_orders` does:~~
```
CAST(COALESCE((SELECT CAST(SUM(U0."voucher_budget_use") AS NUMERIC) AS "s" FROM "pretixbase_orderposition" U0 INNER JOIN "pretixbase_order" U1 ON (U0."order_id" = U1."id") WHERE (NOT U0."canceled" AND U1."status" IN (\'p\', \'n\') AND U0."voucher_budget_use" IS NOT NULL AND U0."voucher_id" = ("pretixbase_voucher"."id")) GROUP BY U0."voucher_id"), CAST(\'0.00\' AS NUMERIC)) AS NUMERIC) AS "budget_used", 
```

~~So reusing `Voucher.annotate_budget_used_orders` would create an extra subquery, which is not necessary – as it is the exact same copy of budget_used_orders. Maybe the database optimizes this. Or instead of copying the whole annotate code, I could add a param `name` to `Voucher.annotate_budget_used_orders` like so?~~

```
        return Voucher.annotate_budget_used_orders(
            self.request.event.vouchers,
            name='budget_used'
        ).select_related(
            'item', 'quota', 'seat', 'variation'
        )
```

~~That way, we could reuse `Voucher.annotate_budget_used_orders` here.~~

### DIFF
--- a/src/pretix/api/views/voucher.py
+++ b/src/pretix/api/views/voucher.py
@@ -20,11 +20,8 @@
 # <https://www.gnu.org/licenses/>.
 #
 
-from decimal import Decimal
-
-from django.db import models, transaction
-from django.db.models import F, OuterRef, Q, Subquery, Sum
-from django.db.models.functions import Coalesce
+from django.db import transaction
+from django.db.models import F, Q
 from django.utils.timezone import now
 from django_filters.rest_framework import (
     BooleanFilter, CharFilter, DjangoFilterBackend, FilterSet,
@@ -37,7 +34,7 @@ from rest_framework.response import Response
 
 from pretix.api.pagination import TotalOrderingFilter
 from pretix.api.serializers.voucher import VoucherSerializer
-from pretix.base.models import Order, OrderPosition, Voucher
+from pretix.base.models import Voucher
 
 with scopes_disabled():
     class VoucherFilter(FilterSet):


raphaelm:
> So reusing Voucher.annotate_budget_used_orders would create an extra subquery, which is not necessary – as it is the exact same copy of budget_used_orders. Maybe the database optimizes this. Or instead of copying the whole annotate code, I could add a param name to Voucher.annotate_budget_used_orders like so?

That seems overkill, given that annotate_budget_used_orders is used in only one place currently. Can you just statically rename it to annotate_budget_used? I think I added the orders back then since it does not account for budget claimed in **carts**, but I don't think we strongly need to have that in the name.

raphaelm:
Ohterwise PR looks good

PR #5691, opened 2025-12-03
Update pricing.rst to fix number typo

felixrindt:
I think you meant to point out the difference to the values in the table above...


wiffbi:
Thanks! :)

PR #5699, opened 2025-12-04
Sendmail: Chunk query to prevent high memory load (Z#23217167)

raphaelm:


src/pretix/plugins/sendmail/tasks.py
```diff
@@ -179,6 +186,11 @@ def send_mails_to_orders(event: Event, user: int, subject: dict, message: dict,
             except SendMailException:
                 failures.append(o.email)
 
+    for chunk in _chunks(objects, 1000):
```
luelista, line 189:
```suggestion
    for chunk in itertools.batched(objects, 1000):
```

src/pretix/plugins/sendmail/tasks.py
```diff
@@ -47,6 +47,14 @@
...
 
 
+def _chunks(lst, n):
+    """
+    Yield successive n-sized chunks from lst.
+    """
+    for i in range(0, len(lst), n):
+        yield lst[i:i + n]
+
+
```
luelista, line 57:
```suggestion
```

src/pretix/plugins/sendmail/tasks.py
```diff
@@ -179,6 +186,11 @@ def send_mails_to_orders(event: Event, user: int, subject: dict, message: dict,
             except SendMailException:
                 failures.append(o.email)
 
+    for chunk in _chunks(objects, 1000):
+        orders = Order.objects.filter(pk__in=chunk, event=event)
+        for o in orders:
```
luelista, line 191: as far as I understand we could prevent Django from keeping all orders matched by the QuerySet in memory by using [QuerySet.iterator()](https://docs.djangoproject.com/en/dev/ref/models/querysets/#iterator)?
```suggestion
        for o in orders.iterator():
```

raphaelm: behind the scenes this is doing batching again as soon as some prefetching gets involved so… not worth it if we already batch?

raphaelm: there's no prefetching here (yet) though

PR #5706, opened 2025-12-09
Questions: filter answers by dateFrame (Z#23216406)

bockstaller:
Allows users to filter answers for events with subevents by subevent as well as time range in which the subevent will happen.
Simpler version of https://github.com/pretix/pretix/pull/5689

- converts the html form from `src/pretix/control/templates/pretixcontrol/items/question.html` into a Django Form
- adds a `DateFrameField` to the form to filter by `date_range`
- prevents combinations of `subevents` and `date_range` that would result in an empty query

### DIFF
--- a/src/pretix/control/forms/filter.py
+++ b/src/pretix/control/forms/filter.py
@@ -1296,14 +1296,10 @@ class QuestionAnswerFilterForm(forms.Form):
         date_range = cleaned_data.get('date_range')
 
         if subevent is not None and date_range is not None:
-            start_d, end_d = resolve_timeframe_to_datetime_start_inclusive_end_exclusive(
-                now(),
-                date_range,
-                self.event.timezone
-            )
+            d_start, d_end = resolve_timeframe_to_datetime_start_inclusive_end_exclusive(now(), date_range, self.event.timezone)
             if (
-                    (start_d and not (start_d <= subevent.date_from)) or
-                    (end_d and not (subevent.date_from < end_d))
+                (d_start and not (d_start <= subevent.date_from)) or
+                (d_end and not (subevent.date_from < d_end))
             ):
                 self.add_error('subevent', pgettext_lazy('subevent', "Date doesn't start in selected date range."))
         return cleaned_data


src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,126 @@ class Meta:
         ]
 
 
+class QuestionFilterForm(forms.Form):
```
wiffbi: I would name it `QuestionAnswerFilterForm`, because if we ever want to add a filterform to the question-listview …

```suggestion
class QuestionAnswerFilterForm(forms.Form):
```

src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,126 @@ class Meta:
...
+        ("pv", _("Paid or confirmed")),
+        ("n", _("Pending")),
+        ("np", _("Pending or paid")),
+        ("o", _("Pending (overdue)")),
+        ("e", _("Expired")),
+        ("ne", _("Pending or expired")),
+        ("c", _("Canceled"))
+    ]
+
+    status = forms.ChoiceField(
```
wiffbi: Please add labels to all the inputs.

src/pretix/control/templates/pretixcontrol/items/question.html
```diff
@@ -14,40 +14,31 @@ <h1>
...
-                    <select name="item" class="form-control">
-                        <option value="">{% trans "All products" %}</option>
-                        {% for item in items %}
-                            <option value="{{ item.id }}"
-                                    {% if request.GET.item|add:0 == item.id %}selected="selected"{% endif %}>
-                                {{ item.name }}
-                            </option>
-                        {% endfor %}
-                    </select>
+                    {% bootstrap_field form.date_range layout='inline' %}
```
wiffbi: Especially date_range is very ambigous what is affected, so I would display all inputs with our default rendering with visible label. Do not forget to update the others as well. Also show the the date_range on normal events as well, as I think date_range should filter on date ordered, not on the subevent-date.
```suggestion
                    {% bootstrap_field form.date_range %}
```

src/pretix/control/templates/pretixcontrol/items/question.html
```diff
@@ -14,40 +14,31 @@ <h1>
         </a>
     </h1>
 
+    {% for e in form.errors.values %}
+        <div class="alert alert-danger has-error">
+            {{ e }}
+        </div>
+    {% endfor %}
+
```
wiffbi: IIRC we never show form-errors like this – especially on a filterform, we just show it on the input+label (see comment below regarding labels).

```suggestion
```

src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,126 @@ class Meta:
...
+        date_range = fdata.get('date_range', None)
+
+        if subevent is not None:
+            opqs = opqs.filter(order__event__subevents=subevent)
+
+        if date_range is not None:
+            d_start, d_end = resolve_timeframe_to_dates_inclusive(now(), date_range, timezone.utc)
+            opqs = opqs.filter(
+                order__event__subevents__date_from__gte=d_start,
+                order__event__subevents__date_from__lte=d_end
```
wiffbi: I think what Z#23216406 wants, is a date_range for when the order was placed.

src/pretix/control/views/item.py
```diff
@@ -660,46 +660,25 @@ def get_context_data(self, **kwargs):
...
-class QuestionView(EventPermissionRequiredMixin, QuestionMixin, ChartContainingView, DetailView):
+class QuestionView(EventPermissionRequiredMixin, ChartContainingView, DetailView):
     model = Question
     template_name = 'pretixcontrol/items/question.html'
     permission = 'can_change_items'
     template_name_field = 'question'
 
+    @cached_property
+    def filter_form(self):
+        return QuestionFilterForm(event=self.request.event, data=self.request.GET)
```
wiffbi: See above:

```suggestion
        return QuestionAnswerFilterForm(event=self.request.event, data=self.request.GET)
```

src/pretix/control/views/item.py
```diff
@@ -660,46 +660,25 @@ def get_context_data(self, **kwargs):
...
+
     def get_answer_statistics(self):
-        opqs = OrderPosition.objects.filter(
-            order__event=self.request.event,
-        )
+        if not self.filter_form.is_valid():
+            return
+
+        opqs = self.filter_form.filter_qs()
+
```
wiffbi: I cannot say why, but for me it feels wrong to return the whole queryset instead of filtering an existing one. So I would instead do:

```suggestion
        opqs = OrderPosition.objects.filter(
            order__event=self.event,
        )
        if self.filter_form.is_valid():
            opqs = self.filter_form.filter_qs(opqs)

```

This way, we display non-filtered stats if the filter-form is not valid. We might want this, or maybe not. @raphaelm What do you think? IMHO this follows our existing filter-form listviews.

src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,126 @@ class Meta:
...
+            if not (start_d <= subevent.date_from.date() <= end_d):
+                raise forms.ValidationError(_("Selected subevent isn't part of the date range."))
+        return cleaned_data
+
+    def filter_qs(self):
+        fdata = self.cleaned_data
+
+        opqs = OrderPosition.objects.filter(
+            order__event=self.event,
+        )
```
wiffbi: See other comment regarding filter_qs, I would filter an existing queryset like so:

```suggestion
    def filter_qs(self, opqs):
        fdata = self.cleaned_data
```

src/pretix/control/forms/item.py
bockstaller: Interesting, I had a different understanding when talking to Luca. Let me clarify that with him

src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,122 @@ class Meta:
...
+        required=False
+    )
+    subevent = forms.ModelChoiceField(
+        queryset=SubEvent.objects.none(),
+        required=False,
+        empty_label=pgettext_lazy('subevent', 'All dates')
+    )
+    date_range = DateFrameField(
+        required=False,
+        include_future_frames=True
```
wiffbi: Just to keep in mind: If this date_range is applied to ordered-date, then include_future_frames makes no sense.

```suggestion
        include_future_frames=False
```

src/pretix/control/forms/item.py
wiffbi: Okay, it was a misunderstanding on my part then. :)

wiffbi: P.S.: we usually exclude the until/end and the date-helper-function should IIRC accomodate for that, so:

```suggestion
                order__event__subevents__date_from__gte=d_start,
                order__event__subevents__date_from__lt=d_end
```

src/pretix/control/templates/pretixcontrol/items/question.html
bockstaller: Done in c9817e493ae2d1b525a87d5ccaabf76fbf55d9cd

src/pretix/control/forms/item.py
bockstaller: Done in 38feae6744810cf045482f822579758e726a973b

src/pretix/control/forms/item.py
bockstaller: Luca clarified in the ticket that the customer wants to filter by date of the subevent

src/pretix/control/templates/pretixcontrol/items/question.html
bockstaller: I've dropped the inline and added the labels, regarding the field to filter upon see above

src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,126 @@ class Meta:
...
+        required=False,
+        label=_("Order status"),
+    )
+    item = forms.ChoiceField(
+        choices=[],
+        widget=forms.Select(
+            attrs={'class': 'form-control'}
+        ),
+        required=False,
+        label=_("Order item"),
```
wiffbi: Other filterforms use this:
```suggestion
        label=_("Products"),
```

src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,126 @@ class Meta:
...
+    )
+    subevent = forms.ModelChoiceField(
+        queryset=SubEvent.objects.none(),
+        required=False,
+        empty_label=pgettext_lazy('subevent', 'All dates'),
+        label=_("Subevent"),
```
wiffbi: In other places we use:
```suggestion
        label=pgettext_lazy("subevent", "Date"),
```

src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,126 @@ class Meta:
...
+    )
+    date_range = DateFrameField(
+        required=False,
+        include_future_frames=True,
+        label=_('Subevent start date'),
```
wiffbi: I’d say we reuse the label for subevent_date_range in OverviewReport – that way it is already translated.
```suggestion
        label=_('Event date'),
```

src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,126 @@ class Meta:
...
+    STATUS_VARIANTS = [
+        ("", _("All orders")),
+        ("p", _("Paid")),
+        ("pv", _("Paid or confirmed")),
+        ("n", _("Pending")),
+        ("np", _("Pending or paid")),
+        ("o", _("Pending (overdue)")),
+        ("e", _("Expired")),
+        ("ne", _("Pending or expired")),
+        ("c", _("Canceled"))
```
wiffbi: I know it’s taken from the template, but we could optimize it – similar to OrderFilterForm (which of fails using these constants in its filter_qs, but well …)

```suggestion
        (Order.STATUS_PAID, _("Paid")),
        (Order.STATUS_PAID + 'v', _("Paid or confirmed")),
        (Order.STATUS_PENDING, _("Pending")),
        (Order.STATUS_PENDING + Order.STATUS_PAID, _("Pending or paid")),
        ("o", _("Pending (overdue)")),
        (Order.STATUS_EXPIRED, _("Expired")),
        (Order.STATUS_PENDING + Order.STATUS_EXPIRED, _("Pending or expired")),
        (Order.STATUS_CANCELED, _("Canceled"))
```

src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,126 @@ class Meta:
...
+
+    def clean(self):
+        cleaned_data = super().clean()
+        subevent = cleaned_data.get('subevent')
+        date_range = cleaned_data.get('date_range')
+
+        if subevent is not None and date_range is not None:
+            start_d, end_d = resolve_timeframe_to_dates_inclusive(now(), date_range, timezone.utc)
+            if not (start_d >= subevent.date_from.date() >= end_d):
+                self.add_error('subevent', _("Subevent doesn't start in selected date range"))
```
wiffbi: We try to avoid „subevent“ in user-facing messages – it is always „date“. So I suggest:

```suggestion
                self.add_error('subevent', pgettext_lazy('subevent', 'Date doesn't start in selected date range'))
```

src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,126 @@ class Meta:
...
+        else:
+            del self.fields['subevent']
+
+    def clean(self):
+        cleaned_data = super().clean()
+        subevent = cleaned_data.get('subevent')
+        date_range = cleaned_data.get('date_range')
+
+        if subevent is not None and date_range is not None:
+            start_d, end_d = resolve_timeframe_to_dates_inclusive(now(), date_range, timezone.utc)
```
wiffbi: I think we need to use the event’s timezone and looking at other places in the code, we filter subevents based on end-exclusive, so we might use:

```suggestion
            start_d, end_d = resolve_timeframe_to_datetime_start_inclusive_end_exclusive(now(), date_range, self.event.timezone)
```
I have not extensively tested this and I might very well be wrong and comparing to `subevent.date_from` without `astimezone(…)` is actually timezone.utc. So maybe @raphaelm can have a look?

src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,126 @@ class Meta:
         ]
 
 
+class QuestionAnswerFilterForm(forms.Form):
```
wiffbi: Probably should go into `src/pretix/control/forms/filter.py` and extend `FilterForm` from there.

src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,126 @@ class Meta:
...
+        include_future_frames=True,
+        label=_('Subevent start date'),
+    )
+
+    def __init__(self, *args, **kwargs):
+        self.event = kwargs.pop('event')
+        super().__init__(*args, **kwargs)
+        self.initial['status'] = "np"
+        self.fields['item'].choices = [('', _('All products'))] + [(item.id, item.name) for item in
+                                                                   Item.objects.filter(event=self.event)]
```
wiffbi: I know this code is taken from the template, so not inherently wrong, but have a look at `EventOrderFilterForm` how it builds the choices for items – we might want to take items with variations into account. Also note instead of `item.name` we should use `str(item)` as that uses the internal_name if any exists (again, nothing you could have known).

src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,126 @@ class Meta:
...
+    date_range = DateFrameField(
+        required=False,
+        include_future_frames=True,
+        label=_('Subevent start date'),
+    )
+
+    def __init__(self, *args, **kwargs):
+        self.event = kwargs.pop('event')
+        super().__init__(*args, **kwargs)
+        self.initial['status'] = "np"
```
wiffbi:
```suggestion
        self.initial['status'] = Order.STATUS_PENDING + Order.STATUS_PAID
```

src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,126 @@ class Meta:
...
+            del self.fields['subevent']
+
+    def clean(self):
+        cleaned_data = super().clean()
+        subevent = cleaned_data.get('subevent')
+        date_range = cleaned_data.get('date_range')
+
+        if subevent is not None and date_range is not None:
+            start_d, end_d = resolve_timeframe_to_dates_inclusive(now(), date_range, timezone.utc)
+            if not (start_d >= subevent.date_from.date() >= end_d):
```
wiffbi: That does not work for me, I think that check needs to be:
`if not (start_d <= subevent.date_from.date() <= end_d):`

Regarding above comment, if we change to end_exclusive, this needs to be adapted accordingly:

```suggestion
            if not (start_d <= subevent.date_from < end_d):
```

src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,126 @@ class Meta:
...
+        date_range = fdata.get('date_range', None)
+
+        if subevent is not None:
+            opqs = opqs.filter(subevent=subevent)
+
+        if date_range is not None:
+            d_start, d_end = resolve_timeframe_to_dates_inclusive(now(), date_range, timezone.utc)
+            opqs = opqs.filter(
+                subevent__date_from__gt=d_start,
+                subevent__date_from__lt=d_end
```
wiffbi: See comment above regarding timezone, end_exclusive and `__gte`

src/pretix/control/forms/item.py
```diff
@@ -272,6 +279,126 @@ class Meta:
...
+            )
+
+        s = fdata.get("status", "np")
+        if s != "":
+            if s == 'o':
```
wiffbi: I know this has not been used before, but we might change to Order.constants here as well.

wiffbi (changes requested):

src/pretix/control/forms/item.py
raphaelm: Yes, we want the date range to be evaluated in the event timezone, not UTC!

The comparison later is not the issue. Comparisons between datetimes with a timezone are always safe. But `resolve_timeframe_to_datetime_start_inclusive_end_exclusive` is effectively a conversion from a date to a datetime, and that needs to be done in the proper timezone.

src/pretix/control/forms/item.py
bockstaller: Done in 882384a80936f9a104321bd551818cb94097c540

src/pretix/control/forms/item.py
bockstaller: Implemented in 42384e9a90bcb4c2551a920558bf6c6ad2718f83

src/pretix/control/forms/filter.py
```diff
@@ -1219,6 +1223,139 @@ def is_decimal(value):
...
+        fdata = self.cleaned_data
+
+        subevent = fdata.get('subevent', None)
+        date_range = fdata.get('date_range', None)
+
+        if subevent is not None:
+            opqs = opqs.filter(subevent=subevent)
+
+        if date_range is not None:
+            d_start, d_end = resolve_timeframe_to_datetime_start_inclusive_end_exclusive(now(), date_range, timezone.utc)
```
wiffbi:
```suggestion
            d_start, d_end = resolve_timeframe_to_datetime_start_inclusive_end_exclusive(now(), date_range,                 self.event.timezone)
```

src/pretix/control/forms/filter.py
```diff
@@ -1219,6 +1223,139 @@ def is_decimal(value):
...
+        subevent = cleaned_data.get('subevent')
+        date_range = cleaned_data.get('date_range')
+
+        if subevent is not None and date_range is not None:
+            start_d, end_d = resolve_timeframe_to_datetime_start_inclusive_end_exclusive(
+                now(),
+                date_range,
+                self.event.timezone
+            )
+            if not (start_d < subevent.date_from.date() < end_d):
```
wiffbi: I just realized that either or both `start_d` and `end_d` can be `None`. Furthermore start needs to be inclusive. Also fails for me comparing datetime to a date, so remove `.date()`. Note: this suggestion is untested. :)

```suggestion
            if (
                (start_d and not(start_d <= subevent.date_from)) or
                (end_d and not(subevent.date_from < end_d))
            ):
```

I could not find any other example in the code regarding comparing to `date_to`. Not sure how we should handle subevents that start before start_d, but end and after start_d. I guess as a start we only output subevents based on date_from.

src/pretix/control/forms/filter.py
```diff
@@ -32,7 +32,7 @@
 # distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 # License for the specific language governing permissions and limitations under the License.
 
-from datetime import datetime, time, timedelta
+from datetime import datetime, time, timedelta, timezone
```
wiffbi: Removed all references to timezone.utc, so not needed to be imported.
```suggestion
from datetime import datetime, time, timedelta
```

src/pretix/control/forms/filter.py
```diff
@@ -1219,6 +1223,139 @@ def is_decimal(value):
...
+        (Order.STATUS_CANCELED, _("Canceled"))
+    ]
+
+    status = forms.ChoiceField(
+        choices=STATUS_VARIANTS,
+        widget=forms.Select(
+            attrs={
+                'class': 'form-control',
+            }
+        ),
```
wiffbi: Not needed anymore as default `bootstrap_field` in template already sets the CSS-class `form-control`.
```suggestion
```

src/pretix/control/forms/filter.py
```diff
@@ -1219,6 +1223,139 @@ def is_decimal(value):
...
+        required=False,
+        label=_("Order status"),
+    )
+    item = forms.ChoiceField(
+        choices=[],
+        widget=forms.Select(
+            attrs={'class': 'form-control'}
+        ),
```
wiffbi: Same as above, default `bootstrap_field` already sets the needed CSS-class
```suggestion
```

src/pretix/control/forms/filter.py
```diff
@@ -1219,6 +1223,139 @@ def is_decimal(value):
...
+                    choices.append(('%d-%d' % (i.pk, v.pk), '%s – %s' % (str(i), v.value)))
+            else:
+                choices.append((str(i.pk), str(i)))
+        self.fields['item'].choices = choices
+
+        if self.event.has_subevents:
+            self.fields["subevent"].queryset = self.event.subevents.all()
+            self.fields['subevent'].widget = Select2(
+                attrs={
+                    'class': 'form-control simple-subevent-choice',
```
wiffbi: Not needed thanks to default `bootstrap_field`
```suggestion
```

src/pretix/control/views/item.py
```diff
@@ -746,9 +727,12 @@ def get_answer_statistics(self):
 
     def get_context_data(self, **kwargs):
         ctx = super().get_context_data()
-        ctx['items'] = self.object.items.all()
+        ctx['items'] = self.object.items.exists()
+        ctx['has_subevents'] = self.request.event.has_subevents
         stats = self.get_answer_statistics()
-        ctx['stats'], ctx['total'] = stats
+        if stats:
+            ctx['stats'], ctx['total'] = stats
```
wiffbi: `stats` always evaluates to True even if result is empty `([], 0)`
```suggestion
        ctx['stats'], ctx['total'] = stats
```

src/pretix/control/forms/filter.py
bockstaller: Implemented in b46bdfb

wiffbi (approved):
Looks good!

PR #5710, opened 2025-12-10
Fix presale date display in calendar (Z#23216645)

ser8phin:


src/pretix/base/templatetags/html_time.py
```diff
@@ -0,0 +1,58 @@
...
+    Building a <time datetime='{html-datetime}'>{human-readable datetime}</time> html string,
+    where the html-datetime as well as the human-readable datetime can be set
+    to a value from django's formats.py/FORMAT_SETTINGS via comma-separated arguments.
+
+    If only one argument is given, it will be used as the human-readable datetime attribute,
+    and the html datetime attribute will be set to the datetime in iso-format.
+
+    Usage example:
+    {% html_datetime event_start|localtime "SHORT_DATE_FORMAT" attr_fmt="SHORT_DATE_FORMAT" as start_date %}
+    {{start_date}}
```
wiffbi: For the example usage I would use the plain version not assigning the output to a variable. See comments below for skipping `|localtime`
```suggestion
    {% html_datetime event_start "SHORT_DATETIME_FORMAT" %}
    or
    {% html_datetime event_start "TIME_FORMAT" attr_fmt="H:i" %}
```

src/pretix/base/templatetags/html_time.py
```diff
@@ -0,0 +1,58 @@
...
+    """
+    if value in (None, ''):
+        return ''
+    attr_fmt = kwargs["attr_fmt"] if kwargs else None
```
wiffbi, line 51: See daterange.py filter for how to not require |localtime – you still need to 
`from django.utils.timezone import get_current_timezone`
```suggestion
    value = value.astimezone(get_current_timezone())
    attr_fmt = kwargs["attr_fmt"] if kwargs else None
```

src/pretix/base/templatetags/html_datetime.py
```diff
@@ -0,0 +1,58 @@
...
+    attr_fmt = kwargs["attr_fmt"] if kwargs else None
+
+    try:
+        if not attr_fmt:
+            date_html = value.isoformat()
+        else:
+            date_html = date_fast(value, attr_fmt)
+
+        date_human = date_fast(value, dt_format)
+        return f"<time datetime='{date_html}'>{date_human}</time>"
```
wiffbi: See icon.py and use `from django.utils.html import format_html` to return a safe string, that is not html-escaped and does not need to be mapped to a variable and filtered through `|safe`:

```suggestion
        return format_html("<time datetime='{}'>{}</time>", date_html, date_human)
```

src/pretix/presale/templates/pretixpresale/fragment_calendar.html
```diff
@@ -1,3 +1,5 @@
+{% load tz %}
```
wiffbi: Please also apply to other templates.
```suggestion
```

src/pretix/presale/templates/pretixpresale/fragment_calendar.html
```diff
@@ -108,13 +110,12 @@
                                                     <span class="fa fa-ticket" aria-hidden="true"></span> {% trans "Book now" %}
                                                 {% elif event.event.presale_has_ended %}
                                                     {% trans "Sale over" %}
-                                                {% elif event.event.settings.presale_start_show_date and event.event.presale_start %}
+                                                {% elif event.event.settings.presale_start_show_date and event.event.effective_presale_start %}
                                                     <span class="fa fa-ticket" aria-hidden="true"></span>
-                                                    {% with date_human=event.event.presale_start|date_fast:"SHORT_DATE_FORMAT" date_iso=event.event.presale_start|date_fast:"c"  %}
-                                                    {% blocktrans with start_date="<time datetime='"|add:date_iso|add:"'>"|add:date_human|add:"</time>"|safe %}
+                                                    {% html_datetime event.event.effective_presale_start|localtime as start_date %}
```
wiffbi:
```suggestion
                                                    {% html_datetime event.event.effective_presale_start as start_date %}
```

src/pretix/presale/templates/pretixpresale/fragment_calendar.html
```diff
@@ -108,13 +110,12 @@
...
+                                                    {% blocktrans with start_date=start_date|safe %}
```
wiffbi:
```suggestion
                                                    {% blocktrans with start_date=start_date %}
```

src/pretix/base/templatetags/html_datetime.py
```diff
@@ -0,0 +1,58 @@
...
+
+from django import template
+
+from pretix.helpers.templatetags.date_fast import date_fast
+
+register = template.Library()
+
+
+@register.simple_tag
+def html_datetime(value: datetime, dt_format: str = "SHORT_DATE_FORMAT", **kwargs):
```
wiffbi: Hm, thinking about it, maybe we should make the default `SHORT_DATETIME_FORMAT` as the default isformat() for the attribute also includes the time.

```suggestion
def html_datetime(value: datetime, dt_format: str = "SHORT_DATETIME_FORMAT", **kwargs):
```

wiffbi (changes requested):
I have some small improvements making the requirement for |safe and |localtime obsolete. IMHO I would name the tag `html_time` as the HTML-tag is `<time>`.

Could you add the template-tag to the following templates, which use the old string-concat, as well:

- presale/templates/pretixpresale/fragment_event_list_status.html
- presale/templates/pretixpresale/event/fragment_event_info.html
- presale/templates/pretixpresale/event/order.html

src/pretix/base/templatetags/html_time.py
```diff
@@ -0,0 +1,67 @@
...
+    h_fmt = kwargs["h_fmt"] if kwargs else None
+    attr_fmt = kwargs["attr_fmt"] if kwargs else None
+
+    try:
+        if not attr_fmt:
+            date_html = value.isoformat()
+        else:
+            date_html = date_fast(value, attr_fmt)
+
+        if h_fmt and h_fmt == "format_expires":
```
wiffbi: Can’t we just use dt_format and handle the special case instead of adding another attribute/kwarg?
```suggestion
        if dt_format == "format_expires":
```

src/pretix/base/templatetags/html_time.py
```diff
@@ -0,0 +1,67 @@
...
+
+    Usage example:
+    {% html_time event_start "SHORT_DATETIME_FORMAT" %}
+    or
+    {% html_time event_start "TIME_FORMAT" attr_fmt="H:i" %}
+    """
+    if value in (None, ''):
+        return ''
+    value = value.astimezone(get_current_timezone())
+    h_fmt = kwargs["h_fmt"] if kwargs else None
```
wiffbi:
```suggestion
```

src/pretix/presale/templates/pretixpresale/event/order.html
```diff
@@ -92,11 +93,10 @@ <h3 class="panel-title">
                     A payment of {{ total }} is still pending for this order.
                 {% endblocktrans %}</strong>
                 <strong>
-                {% with date_human=order|format_expires|safe date_iso=order.expires|date:"c" %}
-                    {% blocktrans trimmed with date='<time datetime="'|add:date_iso|add:'">'|add:date_human|add:"</time>"|safe %}
+                {% html_time order.expires h_fmt="format_expires" as date %}
```
wiffbi:
```suggestion
                {% html_time order.expires "format_expires" as date %}
```

bockstaller:
Your change looks good from my side

I just got a bit sidetracked by this bit right here in the week calendar
Screenshot 2025-12-17 at 11 02 28

Where `data-time=` is a timestamp in isoformat including the timezone offset `+00:00` placing it in UTC. And then there is the separate "data-timezone=" attribute that adds another offset on top of it.

But as far as I can see, these attributes are only consumed by our own Javascript and not something standardized, so there is probably no need to adjust anything here.

wiffbi:
> But as far as I can see, these attributes are only consumed by our own Javascript and not something standardized, so there is probably no need to adjust anything here.

Yes, it is just there for our internal JS that outputs your local time, if it differs from the event’s timezone.

src/pretix/base/templatetags/html_time.py
```diff
@@ -0,0 +1,66 @@
...
+
+register = template.Library()
+
+
+@register.simple_tag
+def html_time(value: datetime, dt_format: str = "SHORT_DATE_FORMAT", **kwargs):
+    """
+    Building a <time datetime='{html-datetime}'>{human-readable datetime}</time> html string,
+    where the html-datetime as well as the human-readable datetime can be set
+    to a value from django's formats.py/FORMAT_SETTINGS via comma-separated arguments.
```
wiffbi: I think comma-separated is a left-over from when it was a filter.
```suggestion
    to a value from django's FORMAT_SETTINGS or "format_expires".
```

src/pretix/base/templatetags/html_time.py
```diff
@@ -0,0 +1,66 @@
...
+
+    If only one argument is given, it will be used as the human-readable datetime attribute,
+    and the html datetime attribute will be set to the datetime in iso-format.
```
wiffbi:
```suggestion
    If attr_fmt isn’t provided, it will be set to isoformat.
```

src/pretix/base/templatetags/html_time.py
ser8phin: Yes it was :sweat_smile:

src/pretix/base/templatetags/html_time.py
ser8phin: Ah, another leftover and also nice improvement, thanks!

PR #5712, opened 2025-12-10
Waiting list: Add edit view for entry (Z#23215496)

bockstaller:
Allows edits to all fields which aren't controlled by other operations - so effectively to 
- `name`
- `item` and `variation`
- `email` (if enabled and/or required)
- `phone` (if enabled and/or required) 

Adds a simple `icontains` search to the `name`, `phone` and `email` field

### DIFF
--- a/src/pretix/control/forms/waitinglist.py
+++ b/src/pretix/control/forms/waitinglist.py
@@ -29,7 +29,7 @@ from pretix.base.forms import I18nModelForm
 from pretix.base.forms.questions import (
     NamePartsFormField, WrappedPhoneNumberPrefixWidget,
 )
-from pretix.base.models import ItemVariation, WaitingListEntry
+from pretix.base.models import WaitingListEntry
 from pretix.control.forms.widgets import Select2
 
 
@@ -98,7 +98,7 @@ class WaitingListEntryEditForm(I18nModelForm):
                 for variation in item.variations.all():
                     if variation.active:
                         choices.append(
-                            ('{}-{}'.format(item.pk, variation.pk), '{} - {}'.format(str(item),str(variation)))
+                            ('{}-{}'.format(item.pk, variation.pk), '{} - {}'.format(str(item), str(variation)))
                         )
             else:
                 choices.append(('{}'.format(item.pk), str(item)))


src/pretix/control/views/waitinglist.py
```diff
@@ -138,6 +140,16 @@ def get_queryset(self, force_filtered=False):
         elif force_filtered and '__ALL' not in self.request_data:
             qs = qs.none()
 
+        if self.request_data.get("search", "") != "":
+            search_q = Q(email__icontains=self.request_data.get("search", ""))
+
+            if self.request.event.settings.waiting_list_names_asked:
+                search_q = search_q | Q(name_cached__icontains=self.request_data.get("search", ""))
+            if self.request.event.settings.waiting_list_names_asked:
```
wiffbi:
```suggestion
            if self.request.event.settings.waiting_list_phones_asked:
```

src/pretix/control/forms/waitinglist.py
```diff
@@ -54,3 +60,97 @@ class Meta:
...
+
+        super().__init__(*args, **kwargs)
+
+        # Prevent the item field cleaning from complaining that it isn't populated
+        # the value for item is derived and populated during form.clean()
+        self.fields['item'].required = False
+
+        self.fields['name_parts'] = NamePartsFormField(
+            max_length=255,
+            required=True,
```
wiffbi:
```suggestion
            required=self.event.settings.waiting_list_names_required,
```

src/pretix/control/forms/waitinglist.py
```diff
@@ -54,3 +60,97 @@ class Meta:
...
+            scheme=self.event.organizer.settings.name_scheme,
+            titles=self.event.organizer.settings.name_scheme_titles,
+            label=_('Name'),
+        )
+
+        self.fields['name_parts'].required = self.event.settings.waiting_list_names_required
+        self.fields['name_parts'].one_required = self.event.settings.waiting_list_names_required
```
wiffbi: Either it is fully required or not, so not necessary.
```suggestion
```

src/pretix/control/forms/waitinglist.py
```diff
@@ -54,3 +60,97 @@ class Meta:
...
+                    choices.append((
+                        '{}-{}'.format(item.pk, v.pk),
+                        '{} – {}'.format(item, v.value) if item.active else mark_safe(
+                            f'<strike class="text-muted">{escape(item)} – {escape(v.value)}</strike>')
+                    ))
+            else:
+                choices.append(('{}'.format(item.pk), str(item) if item.active else mark_safe(
+                    f'<strike class="text-muted">{escape(item)}</strike>')))
+
+        self.fields['itemvar'].label = _("Item and Variation")
```
wiffbi: In most of the other cases it is simply called „Product“, so I'd reuse that here.
```suggestion
        self.fields['itemvar'].label = _("Product")
```

src/pretix/control/forms/waitinglist.py
```diff
@@ -54,3 +60,97 @@ class Meta:
...
+    def clean(self):
+        cleaned_data = super().clean()
+        itemvar = cleaned_data.get('itemvar')
+        if itemvar is None:
+            self.add_error('itemvar', _('Item and Variation are required'))
+            return cleaned_data
+
+        cleaned_data['item'] = Item.objects.get(pk=itemvar.split('-')[0])
+        if '-' in itemvar:
+            cleaned_data['variation'] = ItemVariation.objects.get(pk=itemvar.split('-')[1])
```
wiffbi: You might want to have a look at pretix.presale.forms.waitinglist.WaitingListForm::clean, the item and variation are set onto the instance directly. Not sure though what is the better way. Maybe @raphaelm can chime in?

src/pretix/control/forms/waitinglist.py
```diff
@@ -54,3 +60,97 @@ class Meta:
...
+        return cleaned_data
+
+    class Meta:
+        model = WaitingListEntry
+        fields = [
+            'email',
+            'name_parts',
+            'phone',
+            'item',
+            'variation',
```
wiffbi: See above regarding itemvar and setting it on the instance directly.

src/pretix/control/forms/waitinglist.py
```diff
@@ -54,3 +60,97 @@ class Meta:
...
+            label=_('Name'),
+        )
+
+        self.fields['name_parts'].required = self.event.settings.waiting_list_names_required
+        self.fields['name_parts'].one_required = self.event.settings.waiting_list_names_required
+        self.fields['phone'].required = self.event.settings.waiting_list_phones_required
+
+        choices = []
+
+        items = self.event.items.prefetch_related('variations')
```
wiffbi: Should we at least check for `allow_waitinglist` on the item or do we allow _everything_ (as it is in the backend)? Should we check quotas, etc.? Or should we simply not allow to change the item/variation?  @raphaelm  what do you think?

src/pretix/control/forms/waitinglist.py
```diff
@@ -54,3 +60,97 @@ class Meta:
...
+                initial['itemvar'] = self.instance.item.pk
+        kwargs['initial'] = initial
+
+        super().__init__(*args, **kwargs)
+
+        # Prevent the item field cleaning from complaining that it isn't populated
+        # the value for item is derived and populated during form.clean()
+        self.fields['item'].required = False
+
+        self.fields['name_parts'] = NamePartsFormField(
```
wiffbi: As we hide name_parts and phone in the template, maybe we should remove mit from form itself? Then we do not need to add e.g. `names_asked` to the template-context but simply check for form.name_parts.

Design question: should we always show name_parts and phone? Because, if at some point in time, the waitinglist asked for those, they cannot be seen. The overview-table hides them, so I’d say we stick to hiding them.

src/pretix/control/forms/waitinglist.py
bockstaller: I think I would limit the selection here to only include products that have `allow_waitinglist` set and leave it at that.

If there isn't quota left - please wait in the line as everybody else
If there is quota left - either pretix is sending a voucher automatically or the user is prompted to do so, either way the customer will be informed that they have to take action to get a ticket

If we would "check quota", meaning only allowing moving the entry to products where there is no quota left, then the situation can become a bit unclear.
Couldn't the customer just buy a ticket for this other product if there is quota left over? And how do we communicate to the customer that they have to take action?

src/pretix/control/forms/waitinglist.py
bockstaller: Done in [8fa492d](https://github.com/pretix/pretix/pull/5712/commits/8fa492d82f8ef1daf392f2c41eec435c12f108af)

src/pretix/control/forms/waitinglist.py
bockstaller: Done in [7c21aed](https://github.com/pretix/pretix/pull/5712/commits/7c21aed1a751c460792c36116f2331c9e6530984)

src/pretix/control/forms/waitinglist.py
bockstaller: Done in [7c21aed](https://github.com/pretix/pretix/pull/5712/commits/7c21aed1a751c460792c36116f2331c9e6530984)

src/pretix/control/forms/waitinglist.py
bockstaller: Done in [f26d260](https://github.com/pretix/pretix/pull/5712/commits/f26d2604713527d758080787eea78901cca4d4bd)

src/pretix/control/forms/waitinglist.py
bockstaller: Implemented in [1deb867](https://github.com/pretix/pretix/pull/5712/commits/1deb86773418628622ac5d3904214b32a7021a96)

I've added an help text for that field as well. Just to avoid support requests that wonder why the customers products don't appear here

bockstaller:
TODO: Clean up testing situation and maybe factor out all of the quota/event/subevent/waitinglist logic

src/pretix/control/forms/waitinglist.py
```diff
@@ -45,12 +61,95 @@ def __init__(self, *args, **kwargs):
...
+
+        if not self.event.settings.waiting_list_names_asked:
+            del self.fields['phone']
+
+        choices = []
+
+        items = self.event.items.prefetch_related('variations').prefetch_related('quotas')
+        for item in items:
+            # don't offer items in the selection if they don't allow waitinglists or aren't on sale at all
+            if not item.allow_waitinglist | item.quotas.exists():
```
wiffbi: Using binary or is rather unusual for our codebase so I would stick to boolean or. But I would actually move that filter to the queryset. `allow_waitinglist` can be a normal `filter()`-call, for `quotas.exists()` you can either combine `annotate(has_quota=Exists(Quota.objects.filter(item=OuterRef('pk')))).filter(has_quota=True)` or you might be able to directly do `.filter(Exists(…))`.

src/pretix/control/forms/waitinglist.py
```diff
@@ -45,12 +61,95 @@ def __init__(self, *args, **kwargs):
...
+                # except if they are currently set in the waitinglist, this will be then caught on submit in the clean step
+                if not self.instance.item.pk == item.pk:
```
wiffbi: You might either add instance.item as an Q to the filter of the queryset or probably easier to add the item manually to items if it is not in there before looping over them.

src/pretix/control/forms/waitinglist.py
```diff
@@ -45,12 +61,95 @@ def __init__(self, *args, **kwargs):
...
+            )
+        else:
+            del self.fields['name_parts']
+
+        if not self.event.settings.waiting_list_names_asked:
+            del self.fields['phone']
+
+        choices = []
+
+        items = self.event.items.prefetch_related('variations').prefetch_related('quotas')
```
wiffbi: You can prefetch_related multiple in one call:

`items = self.event.items.prefetch_related('variations', 'quotas')`

But we might not need to prefetch quotas here when we move the Exists-check into the query.

As the waitinglist is a web-feature, I think we should also limit items available on the web sales_channel. @raphaelm what do you think?

src/pretix/control/forms/waitinglist.py
```diff
@@ -45,12 +61,95 @@ def __init__(self, *args, **kwargs):
...
+                max_length=255,
+                required=self.event.settings.waiting_list_names_required,
+                scheme=self.event.organizer.settings.name_scheme,
+                titles=self.event.organizer.settings.name_scheme_titles,
+                label=_('Name'),
+            )
+        else:
+            del self.fields['name_parts']
+
+        if not self.event.settings.waiting_list_names_asked:
```
wiffbi:
```suggestion
        if not self.event.settings.waiting_list_phones_asked:
```

src/pretix/control/forms/waitinglist.py
```diff
@@ -45,12 +61,95 @@ def __init__(self, *args, **kwargs):
...
+                        '{}-{}'.format(item.pk, v.pk),
+                        '{} – {}'.format(item, v.value) if item.active else mark_safe(
+                            f'<strike class="text-muted">{escape(item)} – {escape(v.value)}</strike>')
+                    ))
+            else:
+                choices.append(('{}'.format(item.pk), str(item) if item.active else mark_safe(
+                    f'<strike class="text-muted">{escape(item)}</strike>')))
+
+        self.fields['itemvar'].label = _("Product")
+        self.fields['itemvar'].help_text = _("Only includes Products with an enabled waiting list.")
```
wiffbi:
```suggestion
        self.fields['itemvar'].help_text = _("Only includes products which have the waiting list enabled.")
```

src/pretix/control/forms/waitinglist.py
```diff
@@ -45,12 +61,95 @@ def __init__(self, *args, **kwargs):
...
+                    'organizer': self.event.organizer.slug,
+                }),
+            },
+            choices=choices
+        )
+        self.fields['itemvar'].choices = choices
+
+    def clean(self):
+        if self.instance.voucher is not None:
+            self.add_error(None, _('A voucher for this waiting list entry was sent out already.'))
```
wiffbi: You might want to raise a ValidationError here as once a voucher has been sent, there is no need to further validate.

src/pretix/control/forms/waitinglist.py
```diff
@@ -45,12 +61,95 @@ def __init__(self, *args, **kwargs):
...
+
+        itemvar = self.data.get('itemvar')
+        if itemvar is None:
+            self.add_error('itemvar', _('Item and Variation are required'))
```
wiffbi: Should already be handled by super().clean()

```suggestion
```

src/pretix/control/forms/waitinglist.py
```diff
@@ -45,12 +61,95 @@ def __init__(self, *args, **kwargs):
...
+                self.instance.variation = ItemVariation.objects.get(pk=itemvar.split('-')[1])
+
+        if not self.instance.item.allow_waitinglist:
+            self.add_error('itemvar', _('The selected product does not allow waiting list entries.'))
+        if self.instance.item and not self.instance.item.quotas.filter(subevent=self.data.get('subevent')).exists():
+            self.add_error('itemvar', _('The selected product is not on sale because there is no quota configured for it.'))
+        if self.instance.variation and not self.instance.variation.quotas.filter(subevent=self.data.get('subevent')).exists():
+            self.add_error('itemvar', _('The selected product is not on sale because there is no quota configured for it.'))
+
+        data = super().clean()
```
wiffbi: For our code base rather unusual to call super().clean() that late. We usually call it the first thing in the clean-method and use the cleaned_data it returns to further validate. I’d move it to the top of this function, call the variable `cleaned_data` and use that for further validation regarding itemvar and quota-subevent.

src/pretix/control/urls.py
```diff
@@ -476,8 +476,10 @@
         re_path(r'^waitinglist/auto_assign$', waitinglist.AutoAssign.as_view(), name='event.orders.waitinglist.auto'),
         re_path(r'^waitinglist/(?P<entry>\d+)/delete$', waitinglist.EntryDelete.as_view(),
                 name='event.orders.waitinglist.delete'),
-        re_path(r'^waitinglist/(?P<entry>\d+)/transfer$', waitinglist.EntryTransfer.as_view(),
-                name='event.orders.waitinglist.transfer'),
+        re_path(r'^waitinglist/(?P<entry>\d+)/edit$', waitinglist.EntryEdit.as_view(),
+                name='event.orders.waitinglist.edit'),
+        re_path(r'^waitinglist/(?P<entry>\d+)/transfer$', waitinglist.EntryEdit.as_view(),  # alias of /edit
+                name='event.orders.waitinglist.edit'),
```
wiffbi: I’d just remove the transfer URL completely, no need for supporting old URLs here.
```suggestion
```

src/pretix/control/views/waitinglist.py
```diff
@@ -238,7 +248,7 @@ class WaitingListView(EventPermissionRequiredMixin, WaitingListQuerySetMixin, Pa
     def get_context_data(self, **kwargs):
         ctx = super().get_context_data(**kwargs)
         ctx['items'] = Item.objects.filter(event=self.request.event)
-        ctx['filtered'] = ("status" in self.request.GET or "item" in self.request.GET)
+        ctx['filtered'] = ("status" in self.request.GET or "item" in self.request.GET or "search" in self.request.GET)
```
wiffbi: When reaching three params, we might compress it to:
```suggestion
        ctx['filtered'] = any(param in self.request.GET for param in ("status", "item", "search"))
```

src/pretix/control/views/waitinglist.py
```diff
@@ -390,25 +400,23 @@ def get_success_url(self) -> str:
...
         return get_object_or_404(WaitingListEntry, pk=self.kwargs['entry'], event=self.request.event, voucher__isnull=True)
 
     @transaction.atomic
     def form_valid(self, form):
-        messages.success(self.request, _('The waitinglist entry has been transferred.'))
         if form.has_changed():
+            if 'subevent' in form.changed_data:
+                messages.success(self.request, _('The waitinglist entry has been transferred.'))
+            else:
+                messages.success(self.request, _('The waitinglist entry has been modified.'))
```
wiffbi: I’d keep it simple. Also, one might transfer and change the item/name/email at the same time, so transfer might not be the only operation.

```suggestion
            messages.success(self.request, _('The waitinglist entry has been modified.'))
```

src/pretix/control/forms/waitinglist.py
```diff
@@ -45,12 +61,95 @@ def __init__(self, *args, **kwargs):
...
+
+    def clean(self):
+        if self.instance.voucher is not None:
+            self.add_error(None, _('A voucher for this waiting list entry was sent out already.'))
+
+        itemvar = self.data.get('itemvar')
+        if itemvar is None:
+            self.add_error('itemvar', _('Item and Variation are required'))
+        else:
+            self.instance.item = Item.objects.get(pk=itemvar.split('-')[0])
```
wiffbi: You might want to use the `self.event.items.get()` to limit choice of items to the actual event and not any item – usually we try to limit selection with our ScopeManager to scope within organizer and event, but if possible we should not rely on it.

src/pretix/control/forms/waitinglist.py
```diff
@@ -45,12 +61,95 @@ def __init__(self, *args, **kwargs):
...
+            if '-' in itemvar:
+                self.instance.variation = ItemVariation.objects.get(pk=itemvar.split('-')[1])
```
wiffbi: See above, use `self.instance.item.variations.get()` to make sure you get a variation that is actually on the item. You also need to `try: … except (ObjectDoesNotExist, ValueError):`  see e.g. presale.forms.WaitingListForm.

src/pretix/control/templates/pretixcontrol/waitinglist/edit.html
```diff
@@ -0,0 +1,36 @@
...
+            Please make the desired changes to the following waiting list entry: <strong>{{ entry }}</strong>
+        {% endblocktrans %}</p>
+
+        {% if form.subevent %}
+            {% bootstrap_field form.subevent layout="control" %}
+        {% endif %}
+
+        {% bootstrap_field form.email layout="control" %}
+
+        {% if form.phone %}
```
wiffbi:
```suggestion
        {% if form.name_parts %}
```

src/pretix/control/views/waitinglist.py
```diff
@@ -138,6 +138,16 @@ def get_queryset(self, force_filtered=False):
         elif force_filtered and '__ALL' not in self.request_data:
             qs = qs.none()
 
+        if self.request_data.get("search", "") != "":
+            search_q = Q(email__icontains=self.request_data.get("search", ""))
+
+            if self.request.event.settings.waiting_list_names_asked:
+                search_q = search_q | Q(name_cached__icontains=self.request_data.get("search", ""))
+            if self.request.event.settings.waiting_list_phones_asked:
+                search_q = search_q | Q(phone__icontains=self.request_data.get("search", ""))
```
wiffbi:
```suggestion
            s = self.request_data.get("search", "")
            search_q = Q(email__icontains=s)

            if self.request.event.settings.waiting_list_names_asked:
                search_q = search_q | Q(name_cached__icontains=s)
            if self.request.event.settings.waiting_list_phones_asked:
                search_q = search_q | Q(phone__icontains=s)
```

src/pretix/control/templates/pretixcontrol/waitinglist/index.html
```diff
@@ -124,6 +124,7 @@ <h1>
                 </option>
             {% endfor %}
         </select>
+        <input name="search" type="text" placeholder={% trans "Search" %} class="form-control" value="{{ request.GET.search }}">
```
wiffbi: At some point we might turn that into a filter-form as well …
```suggestion
        <input name="search" type="text" placeholder="{% trans "Search" %}" class="form-control" value="{{ request.GET.search }}">
```

src/tests/control/test_waitinglist.py
```diff
@@ -191,11 +192,152 @@ def test_delete_bulk(client, env):
...
+
+    response = client.post(
+        '/control/event/dummy/dummy/waitinglist/%s/edit' % wle.id,
+        data={
+            "email": f"1_{wle.email}",
+            "itemvar": item.pk
+        },
+        follow=True
+    )
+    assert response.wsgi_request.path == '/control/event/dummy/dummy/waitinglist/'
```
raphaelm: You could also assert that the edit was actually made and the new value is stored – somewhat more interesting result than just where the redirect goes :)

src/tests/control/test_waitinglist.py
```diff
@@ -191,11 +192,152 @@ def test_delete_bulk(client, env):
...
+    quota = Quota.objects.create(event=event)
+    quota.items.add(item)
+
+    client.login(email='dummy@dummy.dummy', password='dummy')
+    with scopes_disabled():
+        wle = WaitingListEntry.objects.create(
+            event=event, item=item, email='foo@bar.com'
+        )
+
+    client.get('/control/event/dummy/dummy/waitinglist/%s/edit' % wle.id)
```
raphaelm: This doesn't really test anything other than "does not crash", does it? You could test that the correct fields show up (e.g. no phone field).

src/tests/control/test_waitinglist.py
```diff
@@ -191,11 +192,152 @@ def test_delete_bulk(client, env):
...
+
+    response = client.post(
+        '/control/event/dummy/dummy/waitinglist/%s/edit' % wle.id,
+        data={
+            "email": f"1_{wle.email}",
+            "itemvar": itemvar
+        },
+        follow=True
+    )
+    assert response.wsgi_request.path == '/control/event/dummy/dummy/waitinglist/'
```
raphaelm: This also lacks the assertion that the change was actually made.

src/pretix/control/templates/pretixcontrol/waitinglist/index.html
```diff
@@ -267,13 +268,13 @@ <h1>
...
-                                       class="btn btn-default btn-sm" title="{% trans "Transfer to other date" context "subevent" %}"
-                                        data-toggle="tooltip">
-                                        <i class="fa fa-calendar" aria-hidden="true"></i>
-                                    </a>
-                                {% endif %}
+
+                                <a href="{% url "control:event.orders.waitinglist.edit" organizer=request.event.organizer.slug event=request.event.slug entry=e.id %}"
+                                   class="btn btn-default btn-sm" title="{% trans "Edit entry" context "subevent" %}"
+                                    data-toggle="tooltip">
+                                    <i class="fa fa-pencil" aria-hidden="true"></i>
```
raphaelm: I think this should be `fa-edit` for consistency with our other list views :)

src/pretix/control/templates/pretixcontrol/waitinglist/edit.html
```diff
@@ -0,0 +1,36 @@
...
+{% load i18n %}
+{% load bootstrap3 %}
+{% block title %}{% trans "Edit entry" %}{% endblock %}
+{% block content %}
+    <h1>{% trans "Edit entry" %}</h1>
+    <form action="" method="post" class="form-horizontal">
+        {% csrf_token %}
+        <p>{% blocktrans trimmed context "subevent" %}
+            Please make the desired changes to the following waiting list entry: <strong>{{ entry }}</strong>
+        {% endblocktrans %}</p>
```
raphaelm: I think this paragraph can be deleted, since it adds no information that is not otherwise contained on the page.

src/pretix/control/views/waitinglist.py
```diff
@@ -390,25 +401,20 @@ def get_success_url(self) -> str:
...
-        return super().dispatch(request, *args, **kwargs)
-
     def get_object(self, queryset=None) -> WaitingListEntry:
         return get_object_or_404(WaitingListEntry, pk=self.kwargs['entry'], event=self.request.event, voucher__isnull=True)
 
     @transaction.atomic
     def form_valid(self, form):
-        messages.success(self.request, _('The waitinglist entry has been transferred.'))
         if form.has_changed():
+            messages.success(self.request, _('The waitinglist entry has been modified.'))
```
raphaelm: I noticed that this PR uses three verbs for the same thing: "edit", "change" and "modify". It's not like we have a consistent policy for this and usually they are the same (except for orders, where we actually split into change and modify), but going just by the numbers I think we gravitate more towards consistency if we always use "change" at least in the UI.

```
(pretix)raphael ~/work/code/pretix/src py310 $ ack modify pretix/locale/django.pot | wc -l
40
(pretix)raphael ~/work/code/pretix/src py310 $ ack edit pretix/locale/django.pot | wc -l  
258
(pretix)raphael ~/work/code/pretix/src py310 $ ack change pretix/locale/django.pot | wc -l
435
```

src/pretix/control/forms/waitinglist.py
```diff
@@ -45,12 +61,95 @@ def __init__(self, *args, **kwargs):
...
         ]
         field_classes = {
             'subevent': SafeModelChoiceField,
+            'email': EmailField,
+            'phone': PhoneNumberField,
         }
+        exclude = [
+            'voucher',  # handled by the assign operation in the WaitingListActionView
+            'priority'  # handled via thumbs up/down
+        ]
```
raphaelm: I'd argue for a

```
        widgets = {
            'phone': WrappedPhoneNumberPrefixWidget(),
        }
```

to have the consistent widget with country code selection

src/pretix/control/forms/waitinglist.py
```diff
@@ -45,12 +61,95 @@ def __init__(self, *args, **kwargs):
...
+            'name_parts',
+            'phone',
             'subevent',
         ]
         field_classes = {
             'subevent': SafeModelChoiceField,
+            'email': EmailField,
+            'phone': PhoneNumberField,
         }
+        exclude = [
```
raphaelm: Do we need "exclude" if we have "fields"?

src/pretix/control/forms/waitinglist.py
raphaelm: It seems weird to limit the choices here very cleverly, and then just render a select2 field that pulls its options via AJAX and shows all the items and variations. That doesn't really make sense? Either a plain old `<select>` with the options filtered here, or a select2 widget with filtering in clean().

That said, I'm not sure we need to be so strict here with filtering at all. The filtering also raises weird edge cases: What if I have an entry on my waiting list and *then* disable allow_waitinglist for that product? Then I can't edit the name of that waiting list entry any more, which makes some sense, but not a lot.

src/pretix/control/forms/waitinglist.py
raphaelm: Yes, you can do .filter(Exists) or maybe even .filter(quotas__isnull=False).distinct(). But if we filter for item.quotas, we also need to respect variation.quotas, otherwise it doesn't really make sense. I'm not insisting on filtering for that here at all, though.

src/pretix/control/forms/waitinglist.py
raphaelm: Yes, that would be important to check :)

bockstaller:
To sumarize:
I've dropped nearly all kind of filtering regarding which items we want to allow and which not after talking to @raphaelm. The reason for this is that these restrictions can cause confusion and unintuitive edge cases.
The dropped filters are: 

- only products with an assigned quota
- only when the waiting list is enabled
- only items sold via the online sales channel 

IMHO the most problematic aspect is, that these validations turn certain edit operations into one way streets, preventing users from fixing a miss click of theirs without changing the setup of the event. 

The only filter that remains is, if the product is active or not.

This turns this feature into a bit of a sharp knife where you can do (and fix) anything.

src/pretix/control/forms/waitinglist.py
bockstaller: See update below

src/pretix/control/forms/waitinglist.py
bockstaller: Code is removed, see update below

src/pretix/control/forms/waitinglist.py
bockstaller: I've removed the select2 there and dropped the filtering

src/pretix/control/forms/waitinglist.py
bockstaller: Code removed, see update below

src/pretix/control/forms/waitinglist.py
bockstaller: Done

src/pretix/control/forms/waitinglist.py
bockstaller: This is automatically handled by the `field.choices` attribute. Non-existing/conforming Items/ItemVariations aren't listed there and therefore raise a ValidationError even before we get there. In this case `itemvar` is not a part of cleaned data

src/pretix/control/templates/pretixcontrol/waitinglist/index.html
bockstaller: Done

src/pretix/control/templates/pretixcontrol/waitinglist/edit.html
bockstaller: Done

src/pretix/control/views/waitinglist.py
bockstaller: Done

src/pretix/control/forms/waitinglist.py
bockstaller: No, I've dropped it

src/pretix/control/forms/waitinglist.py
bockstaller: Done

src/tests/control/test_waitinglist.py
bockstaller: I've cleaned up the tests quite a bit - done

src/pretix/control/forms/waitinglist.py
```diff
@@ -45,12 +73,76 @@ def __init__(self, *args, **kwargs):
...
+        items = self.event.items.filter(active=True).prefetch_related(
+            Prefetch(
+                'variations',
+                queryset=(ItemVariation.objects.filter(item__event=self.event, active=True)),
+                to_attr='active_variations',
+            )
+        )
+
+        for item in items:
+            if len(item.active_variations) > 0:
```
raphaelm: I think this will be False for an item that *has* variations, but non of them is *active*, leading to this item being wrongly listed as a selectable item without variations.

Easiest fix would be to use an unfiltered prefetch and then just drop the variations with active=False in the python loop, it's not like we expect there to be many

src/pretix/control/templates/pretixcontrol/waitinglist/edit.html
```diff
@@ -0,0 +1,33 @@
...
+        {% if form.phone %}
+            {% bootstrap_field form.phone layout="control" %}
+        {% endif %}
+        {% bootstrap_field form.itemvar layout="control" %}
+        <div class="form-group submit-group">
+            <a href="{% url "control:event.orders.waitinglist" organizer=request.event.organizer.slug event=request.event.slug %}" class="btn btn-default btn-cancel">
+                {% trans "Cancel" %}
+            </a>
+            <button type="submit" class="btn btn-primary btn-save">
+                {% trans "Submit" %}
```
raphaelm:
```suggestion
                {% trans "Save" %}
```
for UI consistency

src/pretix/control/forms/waitinglist.py
```diff
@@ -19,17 +19,45 @@
 # You should have received a copy of the GNU Affero General Public License along with this program.  If not, see
 # <https://www.gnu.org/licenses/>.
 #
+from django.db.models import Prefetch
+from django.forms import ChoiceField, EmailField, ValidationError
```
raphaelm: Style tip: We usually import `from django import forms` and then specify things as `forms.EmailField` etc, because otherwise (a) it gets really long with many imports and (b) when reading the code it is unclear whether `EmailField` is a model field or a form field.

src/pretix/control/templates/pretixcontrol/waitinglist/index.html
```diff
@@ -267,13 +268,13 @@ <h1>
...
-                                {% if request.event.has_subevents %}
-                                    <a href="{% url "control:event.orders.waitinglist.transfer" organizer=request.event.organizer.slug event=request.event.slug entry=e.id %}"
-                                       class="btn btn-default btn-sm" title="{% trans "Transfer to other date" context "subevent" %}"
-                                        data-toggle="tooltip">
-                                        <i class="fa fa-calendar" aria-hidden="true"></i>
-                                    </a>
-                                {% endif %}
+
+                                <a href="{% url "control:event.orders.waitinglist.edit" organizer=request.event.organizer.slug event=request.event.slug entry=e.id %}"
+                                   class="btn btn-default btn-sm" title="{% trans "Edit entry" context "subevent" %}"
```
raphaelm:
```suggestion
                                   class="btn btn-default btn-sm" title="{% trans "Edit entry" %}"
```

PR #5715, opened 2025-12-11
Refactor validation of cart contents, fix purchase of inactive subevent (Z#23217806)

raphaelm:
For the second time this month, a customer reported an issue that boiled down to a missing validation in either the "extend card" or the "perform order" action. These bugs are embarrassing, and is made be wonder if we have an underlying problem, and I think we do:

We need to validate a set of constraints in three processes:

- When adding a product to a cart
- When extending an expired cart
- When creating an order from a cart

It's so easy to forget one check in one of them, and our recent accessibility change pulled the "extending an expired cart" process more into focus. 

This PR refactors these checks into a single function that is called form all three places.

Now, there is a breaking change to this: Previously, there were some cases where we allowed constraints to be ignored as long as the cart is valid. Consider the following scenario:

- SubEvent has been configured to allow the sale of product X until 18:00
- Cart created at 17:50 and valid until 18:20
- Cart submitted at 18:10

Prior to this PR, this was allowed. However, only for the product-subevent-availability combination, not for product availability and subevent availability individually! This made not much sense and was quite inconsistent. Now, it is forbidden.

Whether this is an upside or a downside depends on the situation and there is no one-size-fits-all answer to the validation of non-expired carts. Two scenarios:

- An item is no longer available 2 days before the event, because it reflects an early-bird price and the deadline was reached. If someone created their cart just before the deadline, it would be user-friendly to allow them to continue with their transaction.
- An item is no longer available 2 days before the event, because the organizer needs to prepare something and needs to know the number of participants exactly 48h in advance. It would be quite nasty if some late bookings came in after the deadline unexpectedly for the organizer, just because the carts were created early enough.

We used to think more about the former case, because with a default of 30min, the damage usually isn't too bad. However, with the cart extension feature, our default settings now allow carts to be valid for multiple hours, and this can become quite a big and bad surprise for organizers. So with this PR I propose we move the decision to the more consistent, more clear, but a little less buyer-friendly validation at order time for all validations (other than price changes, which only take effect after the cart expires, as before).

### DIFF
--- a/src/pretix/base/services/cart.py
+++ b/src/pretix/base/services/cart.py
@@ -342,7 +342,7 @@ def _check_position_constraints(
         raise CartPositionError(error_messages['inactive_subevent'])
 
     # Subevent sale not started
-    if subevent and subevent.presale_start and time_machine_now_dt < subevent.presale_start:
+    if subevent and subevent.effective_presale_start and time_machine_now_dt < subevent.effective_presale_start:
         raise CartPositionError(error_messages['not_started'])
 
     # Subevent sale has ended
--- a/src/pretix/base/services/orders.py
+++ b/src/pretix/base/services/orders.py
@@ -769,7 +769,8 @@ def _check_positions(event: Event, now_dt: datetime, time_machine_now_dt: dateti
             delete(cp)
 
     # Create locks
-    if any(cp.expires < now() + timedelta(seconds=LOCK_TRUST_WINDOW) and cp.pk not in deleted_positions for cp in sorted_positions):
+    sorted_positions = [cp for cp in sorted_positions if cp.pk and cp.pk not in deleted_positions]  # eliminate deleted
+    if any(cp.expires < now() + timedelta(seconds=LOCK_TRUST_WINDOW) for cp in sorted_positions):
         # No need to perform any locking if the cart positions still guarantee everything long enough.
         full_lock_required = any(
             getattr(o, 'seat', False) for o in sorted_positions
@@ -788,7 +789,7 @@ def _check_positions(event: Event, now_dt: datetime, time_machine_now_dt: dateti
 
     # Check maximum order size
     limit = min(int(event.settings.max_items_per_order), settings.PRETIX_MAX_ORDER_SIZE)
-    if sum(1 for cp in sorted_positions if not cp.addon_to and cp.pk not in deleted_positions) > limit:
+    if sum(1 for cp in sorted_positions if not cp.addon_to) > limit:
         err = err or (error_messages['max_items'] % limit)
 
     # Check availability
@@ -876,7 +877,7 @@ def _check_positions(event: Event, now_dt: datetime, time_machine_now_dt: dateti
             })
 
     # Check prices
-    sorted_positions = [cp for cp in sorted_positions if cp.pk and cp.pk not in deleted_positions]
+    sorted_positions = [cp for cp in sorted_positions if cp.pk and cp.pk not in deleted_positions]  # eliminate deleted
     old_total = sum(cp.price for cp in sorted_positions)
     for i, cp in enumerate(sorted_positions):
         if cp.listed_price is None:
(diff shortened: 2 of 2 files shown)


src/pretix/base/services/cart.py
```diff
@@ -106,6 +110,9 @@ def __init__(self, *args):
     'unknown_position': gettext_lazy('Unknown cart position.'),
     'subevent_required': pgettext_lazy('subevent', 'No date was specified.'),
     'not_for_sale': gettext_lazy('You selected a product which is not available for sale.'),
+    'positions_removed': gettext_lazy(
+        'Some products have can no longer be purchased and have been removed from your cart for the following reason: %s'
```
wiffbi:
```suggestion
        'Some products can no longer be purchased and have been removed from your cart for the following reason: %s'
```

src/pretix/base/services/cart.py
```diff
@@ -258,6 +265,138 @@ def _get_voucher_availability(event, voucher_use_diff, now_dt, exclude_position_
...
+    # Voucher expired
+    if voucher and voucher.valid_until and voucher.valid_until < time_machine_now_dt:
+        raise CartPositionError(error_messages['voucher_expired'])
+
+    # Subevent has been disabled
+    if subevent and not subevent.active:
+        raise CartPositionError(error_messages['inactive_subevent'])
+
+    # Subevent sale not started
+    if subevent and subevent.presale_start and time_machine_now_dt < subevent.presale_start:
```
wiffbi: As we stumbled upon this in the calendar and widget, shouldn’t this be `effective_presale_start`?

```suggestion
    if subevent and subevent.effective_presale_start and time_machine_now_dt < subevent.effective_presale_start:
```

src/pretix/base/services/cart.py
```diff
@@ -1444,6 +1543,14 @@ def recompute_final_prices_and_taxes(self):
 
         return diff
 
+    def _remove_parents_if_bundles_are_removed(self):
+        removed_positions = {op.position.pk for op in self._operations if isinstance(op, self.RemoveOperation)}
+        for op in self._operations:
+            if isinstance(op, self.RemoveOperation):
+                if op.position.is_bundled and op.position.addon_to_id not in removed_positions:
+                    self._operations.append(self.RemoveOperation(position=op.position.addon_to))
+                    removed_positions.add(op.position.addon_to)
```
wiffbi:
```suggestion
                    removed_positions.add(op.position.addon_to_id)
```

src/pretix/base/services/orders.py
```diff
@@ -130,6 +130,9 @@ def __init__(self, *args):
 
 
 error_messages = {
+    'positions_removed': gettext_lazy(
+        'Some products have can no longer be purchased and have been removed from your cart for the following reason: %s'
```
wiffbi:
```suggestion
        'Some products can no longer be purchased and have been removed from your cart for the following reason: %s'
```

src/pretix/base/services/orders.py
```diff
@@ -744,13 +739,37 @@ def delete(cp):
...
+            # Quota, seat, and voucher availability is checked for below
+            # Prices are checked for below
+            # Memberships are checked in _create_order
+        except cart.CartPositionError as e:
+            err = error_messages['positions_removed'] % str(e)
+            delete(cp)
+
     # Create locks
-    if any(cp.expires < now() + timedelta(seconds=LOCK_TRUST_WINDOW) for cp in sorted_positions):
+    if any(cp.expires < now() + timedelta(seconds=LOCK_TRUST_WINDOW) and cp.pk not in deleted_positions for cp in sorted_positions):
```
wiffbi: Can you look at `lock_objects` in line 782 - it creates locks on deleted_positions as well and I am not sure whether this is needed. IMHO we could already filter deleted_positions from sorted_positions as line 791 filters them, availitbily-check skpis them (line 797) and lateron in line 879 and 910 it is filtered twice (the second time wouldn’t even be needed).

raphaelm: > Can you look at lock_objects in line 782 - it creates locks on deleted_positions as well and I am not sure whether this is needed.

Agreed, is not needed.

> IMHO we could already filter deleted_positions from sorted_positions as line 791 filters them,
>  availitbily-check skpis them (line 797) and lateron in line 879 and 910 it is filtered twice (the second time wouldn’t even be needed).

Yes, but the other checks are needed:

- line 797 is needed in case the position is deleted in a previous iteration of the same loop (eg parent product deleted)
- line 879 is needed in case the position is deleted during the availability checks
- line 910 is needed in case the position is deleted in line 907

wiffbi: Argh, I haven’t hat `delete()` in mind, which changes `deleted_positions`. Thanks, makes total sense!

PR #5719, opened 2025-12-12
Send security notification when recovery code is used or created by admin

raphaelm:
"Where to store recovery codes" is one of these problems there is no right answer to, so many people store them in a less-than-optimal place. If that's the reality we live in, this PR adds at least a little security so one notices when they get used :)


raphaelm:
The email has general

> If this change was not performed by you, please contact us immediately.

verbiage  already but I'll extend it a little bit

PR #5722, opened 2025-12-13
Event cancellation: Fix error message creatoin (fix #5721)

### LINKED ISSUE #5721: When cancelling an event, the error message shows a lot of internal data
### Problem and impact

When cancelling an event, and some errors occur, the error message prints out the whole `value` instead of the number of failures. This results in an error message like the following which is confusing: 

> Die Bestellungen wurden storniert. Bei {'dry_run': False, 'id': '258d26df-45ff-42a4-8f13-09db8e80ee10', 'failed': 23,  ...

This happend in our shop on Nov 09 earlier this year. 

### Expected behaviour

Only the number of failed cancellations should be printed. 

### Steps to reproduce

1. Cancel an event where the cancellation of some orders fails. 

In our case the cancellations failed as the connected PayPal account didn't have the required funds. 

### Screenshots

Problematic code (two occurrences in the file:

https://github.com/pretix/pretix/blob/510c4850a58e4306e657f93a5ec865ea6aaa1be7/src/pretix/control/views/orders.py#L3018-L3019

https://github.com/pretix/pretix/blob/510c4850a58e4306e657f93a5ec865ea6aaa1be7/src/pretix/control/views/orders.py#L3099-L3100

### Link

_No response_

_No response_

_No response_

### Version

_No response_

ahus1:
I'm not a Python programmer, still I hope this change fixes my observation. 

If a test is needed, I would gladly accept some help.

Closes #5721

### DIFF
--- a/src/pretix/control/views/orders.py
+++ b/src/pretix/control/views/orders.py
@@ -3016,7 +3016,7 @@ class EventCancel(EventPermissionRequiredMixin, AsyncAction, FormView):
             return _('All orders have been canceled.')
         else:
             return _('The orders have been canceled. An error occurred with {count} orders, please '
-                     'check all uncanceled orders.').format(count=value)
+                     'check all uncanceled orders.').format(count=value["failed"])
 
     def get_success_url(self, value):
         if settings.HAS_CELERY:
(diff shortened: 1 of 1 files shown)


ahus1:
I've submitted a CLA, see ticket number 23218228

raphaelm (approved):

raphaelm:
@cla-bot[bot] check

PR #5728, opened 2025-12-17
Pluggable permissions

### LINKED ISSUE #39: Better integration of permissions

raphaelm:
Fixes #39, after only 10 years!

- [x] Data model
  - [x] Refactor to groups
- [x] Database migration
- [x] Backend UI
- [x] Teams API
  - [x] Tests
  - [x] Docs
- [x] Device permissions
- [x] Change all filter() and team creations
- [x] Change all permission requirements for views etc
- [x] Review permission tests
- [x] Runtime checks for correct permission usage
- [x] Event copy handling
- [x] Split new permissions
  - [x] Gift cards
    - [x] API
    - [x] UI
  - [x] Customers
    - [x] API
    - [x] UI
  - [x] Reusable Media
    - [x] API
    - [x] UI
  - [x] Devices
    - [x] API
    - [x] UI
  - [x] Settings
    - [x] API
      - [x] Tests
    - [x] UI
  - [x] Subevents
    - [x] API
    - [x] UI
  - [x] Cancel event
    - [x] UI
- [x] Handle exports that require other permissions
  - [x] UI
  - [x] API
  - [x] Tests
  - [x] Docs
- [x] Eliminate all usages of old permissions
- [x] Verify fallbackability of plugins
  - [x] Tests for compatibility layer on Teams
- [x] Adjust plugins
  - [x] pretix-campaigns
    - [x] test fixtures
  - [x] pretix-digital
    - [x] test fixtures
  - [x] pretix-exhibitors
    - [x] test fixtures
    - [x] custom permission
      - [x] refactor to groups
    - [x] exports
  - [x] pretix-offlinesales
    - [x] test fixtures
  - [x] pretix-posbackend
    - [x] test fixtures
    - [x] custom permission
      - [x] refactor to groups
    - [x] device API permissions
    - [x] exports
  - [x] pretix-reports
    - [x] checks
  - [x] pretix-resellers
    - [x] checks
  - [x] pretix-secrets-imports
    - [x] test fixtures
  - [x] pretix-shipping
    - [x] test fixtures
    - [x] checks
  - [x] pretix-superbundles
    - [x] checks
  - [x] pretix-scheduling
    - [x] checks
    - [x] custom permission
      - [x] refactor to groups
  - [x] vrt
    - [x] test fixtures
    - [x] team creation
    - [x] export permission
  - [x] pretixeu
    - [x] test fixtures
    - [x] team creation
    - [x] filters
  - [x] pretix-peppol
    - [x] respect new permission
  - [x] pretix-zugferd
    - [x] respect new permission
  - [x] pretix-invopop
    - [x] respect new permission
  - [x] pretix-datev
    - [x] export permission
  - [ ] All others post-merge
- [ ] Future: OAuth permissions

Useful search regexes for unadjusted places:
```
can_([a-z]+)_([a-z]+)=
\.can_([a-z]+)_([a-z]+)
(['"])can_view_orders['"]
can_([a-z]+)_([a-z]+)
(can_create_events|can_change_teams|can_change_organizer_settings|can_manage_customers|can_manage_reusable_media|can_manage_gift_cards|can_change_event_settings|can_change_items|can_view_orders|can_change_orders|can_view_vouchers|can_change_vouchers|can_checkin_orders|can_change_settings)
```

### DIFF
--- a/src/pretix/base/migrations/0298_pluggable_permissions.py
+++ b/src/pretix/base/migrations/0298_pluggable_permissions.py
@@ -17,9 +17,14 @@ def migrate_teams_forward(apps, schema_editor):
             for k, v in OLD_TO_NEW_EVENT_MIGRATION.items():
                 if getattr(team, k):
                     team.limit_event_permissions.update({kk: True for kk in v})
+
+            # Prevent combinations that were possible previously but no longer make sense
             if team.limit_event_permissions.get("event.orders:checkin") and team.limit_event_permissions.get("event.orders:write"):
-                # Redundant and not allowed
                 team.limit_event_permissions.pop("event.orders:checkin")
+            if team.limit_event_permissions.get("event.orders:write") and not team.limit_event_permissions.get("event.orders:read"):
+                team.limit_event_permissions.pop("event.orders:write")
+            if team.limit_event_permissions.get("event.vouchers:write") and not team.limit_event_permissions.get("event.vouchers:read"):
+                team.limit_event_permissions.pop("event.vouchers:write")
 
         if all(getattr(team, k) for k in OLD_TO_NEW_ORGANIZER_MIGRATION.keys()):
             team.all_organizer_permissions = True


doc/api/resources/devices.rst
```diff
@@ -30,7 +30,7 @@ software_brand                        string                     Device software
 software_version                      string                     Device software version (read-only)
 created                               datetime                   Creation time
 initialized                           datetime                   Time of initialization (or ``null``)
-initialization_token                  string                     Token for initialization
+initialization_token                  string                     Token for initialization (field invisible without without write permission)
```
luelista:
```suggestion
initialization_token                  string                     Token for initialization (field invisible without write permission)
```

doc/api/resources/events.rst
```diff
@@ -615,6 +601,8 @@ organizer level.
 
    Updates event settings. Note that ``PUT`` is not allowed here, only ``PATCH``.
 
+   Permission "Can change event settings" is always required. Some keys requrie additional permissions.
```
luelista:
```suggestion
   Permission "Can change event settings" is always required. Some keys require additional permissions.
```

src/pretix/api/serializers/organizer.py
```diff
@@ -306,21 +313,97 @@ def get_queryset(self):
...
         full_data.update(data)
+
+        if new_data_set:
+            if full_data.get('limit_event_permissions') and full_data.get('all_event_permissions'):
+                raise ValidationError('Do not set both limit_event_permissions and all_event_permissions.')
+            if full_data.get('limit_organizer_permissions') and full_data.get('all_organizer_permissions'):
+                raise ValidationError('Do not set both limit_organizer_permissions and all_organizer_permissions.')
+
+        if old_data_set:
+            # Migrate with same logic as in migration 0297_plugable_permissions
```
luelista:
```suggestion
            # Migrate with same logic as in migration 0297_pluggable_permissions
```

src/pretix/base/signals.py
```diff
@@ -1098,6 +1110,9 @@ class and point to a valid JPEG or PNG file. If no image is available, ``evaluat
 This signal is sent out to collect serializable settings fields for the API. You are expected to
 return a dictionary mapping names of attributes in the settings store to DRF serializer field instances.
 
+These are readable for all users with access to the events, therefore secrets made in the settings store
```
luelista: "secrets stored in the settings store" ?

luelista:
i think we should put some consideration into "permission dependencies" -- at least in the UI, maybe even in the data model.

e.g. a user should not be able to select "event.vouchers:write" without selecting "event.vouchers:read", when the write permission would be ineffective without the read permission. 
as "event.orders:write" automatically grants "event.orders:checkin", maybe we should select the latter automatically in the UI.
not sure if we need to protect just UI users, or API users as well, from such mistakes.

also, we should make it more clear what happens if a "Change" permission is not selected (maybe just update the wording) - e.g. "Change settings" is actually "View and change settings" (if it's off, settings can't be viewed), but "Change seating plans" is just that (if it's off, seating plans are visible readonly)

maybe we should change the UI from check boxes:

```
[x] Change settings
[x] Change seating plans
...
[x] View orders
[x] Change orders
[x] Check-in orders
[x] View vouchers
[x] Change vouchers
```

to dropdowns, radio buttons or similar:

```
Settings       [ No access | View and change ]
Seating plans  [ View | View and change ]
...
Orders         [ No access | View | View and check-in | View, check-in and change ]
Vouchers       [ No access | View | View and change ]
```

"simple" permissions like "Cancel event" could be kept as checkboxes, or also changed.

alternatively, we could work with dependencies between checkboxes, or automatically select/deselect relevant checkboxes.

**edit**: I just noticed that "event.orders:checkin" is a bad example. I thought it always included "event.orders:read", but it seems it doesn't.

----

for the api we might be able to replace `{ "event.vouchers:read": (bool), "event.vouchers:write": (bool) }` with eg. `{ "event.vouchers": "read" / "readwrite" }` (or just "read" / "write"). 
i'm not sure though what something without ascending levels of permission, like "event:cancel" should look like, so maybe we should keep the api as is, and just add checks for ineffective permissions there (or just document the behaviour and change nothing at all).

luelista:
i'm wondering whether we should switch to constants for referring to permissions, now that we need to touch every reference anyways (to give us autocomplete, IDE linting and earlier runtime errors for incorrect permission names)?

either just constants containing strings `EVENT_SETTINGS_GENERAL_WRITE = "event.settings.general:write"` or even the Permission namedtuples themselves
`EVENT_SETTINGS_GENERAL_WRITE = Permission("event.settings.general:write", ...)`

in the latter case, we could use classes instead of `namedtuples` and have
```
class OrganizerPermission(Permission):
	...
class EventPermission(Permission):
	...
```

and even use type hints and/or runtime checks to avoid using the wrong permission class with the wrong method.

doc/development/implementation/permissions.rst
```diff
@@ -195,3 +195,53 @@ staff mode is active. You can check if a user is in staff mode using their sessi
 Staff mode has a hard time limit and during staff mode, a middleware will log all requests made by that user. Later,
 the user is able to also save a message to comment on what they did in their administrative session. This feature is
 intended to help compliance with data protection rules as imposed e.g. by GDPR.
+
+Adding permissions
+------------------
+
+Plugins can add permissions through the ``register_event_permission_groups`` and ``register_organizer_permission_groups``.
+We recommend to use this only for very significant permissions, as the system will become less usable with too many
+permission levels, also because the team page will show all permission options, even those of disabled plugins.
```
luelista, line 207: if meant in the sense of "auch weil" / "insbesondere weil":
```suggestion
permission levels, especially since the team page will show all permission options, even those of disabled plugins.
```
or
```suggestion
permission levels, partly because the team page will show all permission options, even those of disabled plugins.
```

doc/development/implementation/permissions.rst
```diff
@@ -195,3 +195,53 @@ staff mode is active. You can check if a user is in staff mode using their sessi
...
+
+To register your permissions, you need to register a **permission group** (often representing an area of functionality
+or a key model). Below that group, there are **actions**, which represent the actual permissions. Permissions will be
+generated as ``<group_name>:<action>``. Then, you need to define **options** which are the valid combinations of the
+actions that should be possible to select for a team. This two-step mechanism exists to provide a better user experience
+and avoid useless combinations like "write but not read".
```
luelista, line 213:
```suggestion
and avoid impossible combinations like "write but not read".
```
it might not be useless in all cases but more like the combinations won't work?

raphaelm:
> i'm wondering whether we should switch to constants for referring to permissions, now that we need to touch every reference anyways (to give us autocomplete, IDE linting and earlier runtime errors for incorrect permission names)?
…
> and even use type hints and/or runtime checks to avoid using the wrong permission class with the wrong method.

As discussed, we settled for runtime checks for now

doc/development/api/exporter.rst
```diff
@@ -80,8 +80,24 @@ The exporter class
...
 
       This is an abstract method, you **must** override this!
+
+   .. automethod:: available_for_user
+
+   .. automethod:: get_required_event_permission
+
+On organizer level, by default exporters are expected to handle on a *set of events* and the system will automatically
+add a form field that allows the selection of events, limited to events the user has correct permissions for. If this
+does not fit your organizer, because it is not related to events, you should **also** inherit from the following class:
```
luelista, line 99:
```suggestion
On organizer level, by default exporters are expected to handle a *set of events* and the system will automatically
add a form field that allows the selection of events, limited to events the user has correct permissions for. If this
does not fit your exporter, because it is not related to events, you should **also** inherit from the following class:
```

src/pretix/base/permissions.py
```diff
@@ -0,0 +1,332 @@
...
+from pretix.base.signals import (
+    register_event_permission_groups, register_organizer_permission_groups,
+)
+
+logger = logging.getLogger(__name__)
+_ALL_EVENT_PERMISSION_GROUPS = None
+_ALL_ORGANIZER_PERMISSION_GROUPS = None
+_ALL_EVENT_PERMISSIONS = None
+_ALL_ORGANIZER_PERMISSIONS = None
+_CACHE_TIME_APPS_READY = None  # hack: we need to clear the cache after plugins are loaded during startup
```
luelista: this introduces a problem where if multiple results are cached with ready=False, we only re-generate one of them (the one which is retrieved first), which updates `_CACHE_TIME_APPS_READY` but the other results will stay on their old value.

i'd recommend to add a decorator like this:

```suggestion


def cache_until_change(input_value):
    def decorator(func):
        old_input_value = None
        cached_result = None

        @functools.wraps(func)
        def wrapper():
            nonlocal cached_result, old_input_value
            if not cached_result or old_input_value != input_value():
                cached_result = func()
                old_input_value = input_value()
            return cached_result
        return wrapper
    return decorator
```

and decorate the four functions below with `@cache_until_change(input_value=lambda: apps.ready)`

src/pretix/base/permissions.py
```diff
@@ -0,0 +1,332 @@
...
+        return
+    if isinstance(permission, (list, tuple)) and allow_tuple:
+        for p in permission:
+            assert_valid_event_permission(p)
+        return
+    if not allow_legacy and ':' not in permission:
+        raise ValueError(f"Not allowed to use legacy permission '{permission}'")
+    all_permissions = get_all_event_permissions()
+    if permission not in all_permissions:
+        # Warning *and* exception because warning is silently caught when used in if statements in Django templates
```
luelista, line 133: i'd assume the exception is silently caught, not the warning? 🤔

src/pretix/base/services/export.py
```diff
@@ -133,45 +169,138 @@ def set_progress(val):
...
 
-        for recv, response in responses:
-            if not response:
+    responses = register_data_exporters.send(event)
+    for r, response in responses:
+        if not response:
+            continue
+
+        if issubclass(response, OrganizerLevelExportMixin):
+            raise TypeError("Cannot user organizer-level exporter on event level")
```
luelista, line 208:
```suggestion
            raise TypeError("Cannot use organizer-level exporter on event level")
```

src/pretix/base/services/export.py
```diff
@@ -133,45 +169,138 @@ def set_progress(val):
...
+def init_organizer_exporter(identifier, **kwargs):
+    for ex in init_organizer_exporters(**kwargs):
+        if ex.identifier == identifier:
+            return ex
+    return None
+
+
+def init_organizer_exporters(
+    organizer, user=None, token=None, device=None, request=None, progress_callback=None, staff_session=False, event_qs=None
+):
```
luelista, line 231:
```
[2026-02-19 12:13:59,692: ERROR/ForkPoolWorker-15] Scheduled export failed.
  File "///pretix/src/pretix/base/services/export.py", line 354, in _run_scheduled_export
    d = exporter.render(schedule.export_form_data)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "///pretix/src/pretix/base/exporter.py", line 439, in render
    return self._render_xlsx(form_data, output_file=output_file)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "///pretix/src/pretix/base/exporter.py", line 426, in _render_xlsx
    self.progress_callback(counter / total * 100 / n_sheets + 100 / n_sheets * i_sheet)
TypeError: 'NoneType' object is not callable
```


this should fix it (same for init_event_exporters):
```suggestion
def init_organizer_exporters(
    organizer, user=None, token=None, device=None, request=None, progress_callback=lambda v: None, staff_session=False, event_qs=None
):
```

but maybe it's better to keep passing None, and add to `BaseExporter.__init__` something like `if not progress_callback: progress_callback = lambda v: None`

src/pretix/control/views/organizer.py
```diff
@@ -2034,7 +2012,7 @@ def get_context_data(self, **kwargs):
         return ctx
 
     def get_scheduled_queryset(self):
-        if not self.request.user.has_organizer_permission(self.request.organizer, 'can_change_organizer_settings',
+        if not self.request.user.has_organizer_permission(self.request.organizer, 'organizer.settings.general:write',
```
luelista, line 2040: this interacts weirdly with scheduled exports where the owner has more permissions / permission to more events than the current user:
- if you edit an export where some events were selected by the owner that you aren't allowed to see, then saving it quietly removes those events
- if "all events" is selected, you can still edit and run the export, e.g. adding your own email as recipient, allowing you to get event data you normally couldn't
    - you can also change another users' export from "some events" to "all events", with the same problem

quick-fix would be to switch to `organizer.teams:write` here, because that's the highest permission anyway

```suggestion
        if not self.request.user.has_organizer_permission(self.request.organizer, 'organizer.teams:write',
```

src/pretix/control/permissions.py
```diff
@@ -92,9 +97,11 @@ def organizer_permission_required(permission):
     This view decorator rejects all requests with a 403 response which are not from
     users having the given permission for the event the request is associated with.
     """
-    if permission == 'can_change_settings':
+    if permission in ('event.settings.general:write', 'can_change_settings', 'can_change_event_settings'):
         # Legacy support
-        permission = 'can_change_organizer_settings'
+        permission = 'organizer.settings.general:write'
```
luelista, line 102: why do we accept `event.settings.general:write` here?

src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -340,11 +340,12 @@ var form_handlers = function (el) {
     }
 
     el.find("input[data-checkbox-dependency]").each(function () {
+        var initially_disabled = $(this).prop("disabled");
```
luelista, line 343: might be easier to just skip disabled controls completely?

```suggestion
el.find("input[data-checkbox-dependency]:not([disabled])").each(function () {
```

src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -340,11 +340,12 @@ var form_handlers = function (el) {
...
         var dependent = $(this),
             dependency = findDependency($(this).attr("data-checkbox-dependency"), this),
             update = function () {
                 var enabled = dependency.prop('checked');
-                dependent.prop('disabled', !enabled).closest('.form-group, .form-field-boundary').toggleClass('disabled', !enabled);
+                dependent.prop('disabled', !enabled || initially_disabled).closest('.form-group, .form-field-boundary').toggleClass('disabled', !enabled);
```
luelista, line 348:
```suggestion
                dependent.prop('disabled', !enabled).closest('.form-group, .form-field-boundary').toggleClass('disabled', !enabled);
```

src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -366,11 +367,12 @@ var form_handlers = function (el) {
     });
 
     el.find("div[data-display-dependency], textarea[data-display-dependency], input[data-display-dependency], select[data-display-dependency], button[data-display-dependency]").each(function () {
+        var initially_disabled = $(this).prop("disabled");
```
luelista, line 370: same here?

```suggestion
el.find("div[data-display-dependency]:not([disabled]), textarea[data-display-dependency]:not([disabled]), input[data-display-dependency]:not([disabled]), select[data-display-dependency]:not([disabled]), button[data-display-dependency]:not([disabled])").each(function () {
```
or


```suggestion
el.find("div[data-display-dependency], textarea[data-display-dependency], input[data-display-dependency], select[data-display-dependency], button[data-display-dependency]").each(function () {
        if ($(this).prop("disabled")) return;
```

src/pretix/static/pretixbase/js/addressform.js
```diff
@@ -33,6 +33,13 @@ $(function () {
             dependents[cleanName($(this).attr("name"))] = $(this)
         })
 
+        const dependentsDisabled = [];
```
luelista, line 36: i'm wondering if we can skip the whole update logic for disabled forms...?

but i think it's okay either way

src/pretix/control/templates/pretixcontrol/organizers/team_edit.html
```diff
@@ -22,25 +23,24 @@ <h1>{% trans "Create a new team" %}</h1>
...
-            {% bootstrap_field form.can_manage_customers layout="control" %}
-            {% bootstrap_field form.can_manage_reusable_media layout="control" %}
-            {% bootstrap_field form.can_change_teams layout="control" %}
-            {% bootstrap_field form.can_change_organizer_settings layout="control" %}
+            {% bootstrap_field form.all_organizer_permissions layout="control" %}
+            <div class="team-permission-groups col-md-9 col-md-offset-3" data-display-dependency="#id_all_organizer_permissions" data-inverse>
+                {% for f in form.organizer_field_names %}
+                    {% bootstrap_field form|getitem:f layout="control" %}
+                {% endfor %}
+            </div>
```
luelista, line 31: to have it align properly:
```suggestion
            <div class="form-group"><div class="col-md-9 col-md-offset-3">
                <div class="team-permission-groups" data-display-dependency="#id_all_organizer_permissions" data-inverse>
                    {% for f in form.organizer_field_names %}
                        {% bootstrap_field form|getitem:f layout="control" %}
                    {% endfor %}
                </div>
            </div></div>
```

src/pretix/control/templates/pretixcontrol/organizers/team_edit.html
```diff
@@ -22,25 +23,24 @@ <h1>{% trans "Create a new team" %}</h1>
...
-            {% bootstrap_field form.can_change_orders layout="control" %}
-            {% bootstrap_field form.can_checkin_orders layout="control" %}
-            {% bootstrap_field form.can_view_vouchers layout="control" %}
-            {% bootstrap_field form.can_change_vouchers layout="control" %}
+            {% bootstrap_field form.all_event_permissions layout="control" %}
+            <div class="team-permission-groups col-md-9 col-md-offset-3" data-display-dependency="#id_all_event_permissions" data-inverse>
+                {% for f in form.event_field_names %}
+                    {% bootstrap_field form|getitem:f layout="control" %}
+                {% endfor %}
+            </div>
```
luelista, line 43:
```suggestion
            <div class="form-group"><div class="col-md-9 col-md-offset-3">
                <div class="team-permission-groups" data-display-dependency="#id_all_event_permissions" data-inverse>
                    {% for f in form.event_field_names %}
                        {% bootstrap_field form|getitem:f layout="control" %}
                    {% endfor %}
                </div>
            </div></div>
```

src/pretix/base/permissions.py
raphaelm:
```suggestion
        # Warning *and* exception because exception is silently caught when used in if statements in Django templates
```

src/pretix/control/permissions.py
raphaelm: Because historically, can_change_settings was used both on event and organizer level, and can_change_settings is now translated to event.settings.general:write by the compatibility layer, which could break plugins that still use can_change_settings on organizer level

src/pretix/base/services/export.py
raphaelm: I just pass it as **kwargs through init to avoid repeating defaults, might also be safer for the future

doc/api/resources/teams.rst
```diff
@@ -24,21 +24,57 @@ all_events                            boolean                    Whether this te
...
+    organizer.teams:write
+    organizer.seatingplans:write
+    organizer.giftcards:read
+    organizer.giftcards:write
+    organizer.customers:read
+    organizer.customers:write
+    organizer.reusablemedia:read
+    organizer.reusablemedia:write
+    organizer.devices:read
+    organizer.devices:write
```
wiffbi, line 61:
```suggestion
    organizer.devices:write
    organizer.outgoingmails:read
```

src/pretix/base/permissions.py
```diff
@@ -0,0 +1,329 @@
...
+@receiver(register_organizer_permission_groups, dispatch_uid="base_register_default_organizer_permissions")
+def register_default_organizer_permissions(sender, **kwargs):
+    return [
+        PermissionGroup(
+            name="organizer.events",
+            label=_("Events"),
+            actions=["create"],
+            options=[
+                PermissionOption(actions=tuple(), label=pgettext_lazy("permission_level", "Only existing events")),
+                PermissionOption(actions=("create",), label=pgettext_lazy("permission_level", "Create new events")),
```
wiffbi: Having these options under the label „Events“ IMHO seems a bit confusing, especially the "Only existing events", so maybe do:
```suggestion
                PermissionOption(actions=tuple(), label=pgettext_lazy("permission_level", "Access existing events")),
                PermissionOption(actions=("create",), label=pgettext_lazy("permission_level", "Access existing and create new events")),
```

src/pretix/control/forms/event.py
```diff
@@ -294,9 +299,14 @@ def copy_from_queryset(user, session):
...
-                all_events=True, can_change_event_settings=True, can_change_items=True
+                # TODO: review these!
+                # Restrict cross-organizer copying further than same-organizer copying?
+                TeamQuerySet.event_permission_q("event.settings.general:write"),
+                TeamQuerySet.event_permission_q("event.items:write"),
+                all_events=True,
             ).values_list('organizer', flat=True)) | Q(id__in=user.teams.filter(
-                can_change_event_settings=True, can_change_items=True
+                TeamQuerySet.event_permission_q("event.settings.general:write"),
+                TeamQuerySet.event_permission_q("event.items:write"),
```
wiffbi: I don’t understand the copy-wizard with regards to permissions. Why do I need write permission on the event I want to copy? Shouldn’t read be enough? Also copying an event I was able to create items although my permission allowed to create events, but only view items/quotas. Probably an edge-case for strange permission-setting (create events, but do not change items) and already was like this before.

src/pretix/control/templates/pretixcontrol/organizers/reusable_medium.html
```diff
@@ -22,60 +22,69 @@ <h3 class="panel-title">
...
-                                {% else %}
-                                    {% trans "active" %}
-                                {% endif %}
-                            </dd>
-                            <dt>{% trans "Connections" context "reusable_media" %}</dt>
-                            <dd>
-                                {% if medium.customer %}
-                                    <span class="helper-display-block">
-                                    <span class="fa fa-user fa-fw"></span>
+                    {% csrf_token %}
```
wiffbi: No form, so no need for csrf_token here.
```suggestion
```

src/pretix/control/templates/pretixcontrol/organizers/reusable_medium.html
```diff
@@ -22,60 +22,69 @@ <h3 class="panel-title">
...
-                                {% if medium.linked_giftcard %}
-                                    <span class="helper-display-block">
-                                    <span class="fa fa-credit-card fa-fw"></span>
+                            </span>
+                            {% endif %}
+                            {% if medium.linked_orderposition %}
+                                <span class="helper-display-block">
+                                <span class="fa fa-ticket fa-fw"></span>
+                                <a href="{% url "control:event.order" event=medium.linked_orderposition.order.event.slug organizer=request.organizer.slug code=medium.linked_orderposition.order.code %}">
+                                    {{ medium.linked_orderposition.order.code }}</a>-{{ medium.linked_orderposition.positionid }}
```
wiffbi, line 58: This is probably a tough one to check whether the user has the rights to access this order before linking to it. So I’d leave like it is. Just to make sure, that this can link to an order I do not have access to – which correctly shows me an error.

raphaelm: I think that's okay, yes.

raphaelm:
Migration in SQL:

```
-- Some time before code deployment
BEGIN;
ALTER TABLE "pretixbase_team"
    ADD COLUMN "all_event_permissions" boolean DEFAULT false NOT NULL;
ALTER TABLE "pretixbase_team"
    ADD COLUMN "all_organizer_permissions" boolean DEFAULT false NOT NULL;
ALTER TABLE "pretixbase_team"
    ADD COLUMN "limit_event_permissions" jsonb DEFAULT '{}' NOT NULL;
ALTER TABLE "pretixbase_team"
    ADD COLUMN "limit_organizer_permissions" jsonb DEFAULT '{}' NOT NULL;
COMMIT;


-- Right before code deployment
BEGIN;

UPDATE pretixbase_team
SET all_event_permissions       = CASE
                                      WHEN can_change_event_settings = true AND can_change_items = true AND
                                           can_view_orders = true AND can_change_orders = true AND
                                           can_view_vouchers = true AND can_change_vouchers = true
                                          THEN true
                                      ELSE false END,
    limit_event_permissions     = CASE
                                      WHEN can_change_event_settings = true AND can_change_items = true AND
                                           can_view_orders = true AND can_change_orders = true AND
                                           can_view_vouchers = true AND can_change_vouchers = true
                                          THEN '{}'::jsonb
                                      ELSE (
                                          (CASE
                                               WHEN can_change_event_settings = true
                                                   THEN '{"event.settings.general:write": true, "event.settings.payment:write": true, "event.settings.tax:write": true, "event.settings.invoicing:write": true, "event.subevents:write": true}'
                                               ELSE '{}' END)::jsonb ||
                                          (CASE
                                               WHEN can_change_items = true THEN '{"event.items:write": true}'
                                               ELSE '{}' END)::jsonb ||
                                          (CASE
                                               WHEN can_view_orders = true THEN '{"event.orders:read": true}'
                                               ELSE '{}' END)::jsonb ||
                                          (CASE
                                               WHEN can_change_orders = true AND can_view_orders = true
                                                   THEN '{"event.orders:write": true, "event:cancel": true}'
                                               ELSE '{}' END)::jsonb ||
                                          (CASE
                                               WHEN can_checkin_orders = true AND NOT can_change_orders = true
                                                   THEN '{"event.orders:checkin": true}'
                                               ELSE '{}' END)::jsonb ||
                                          (CASE
                                               WHEN can_view_vouchers = true THEN '{"event.vouchers:read": true}'
                                               ELSE '{}' END)::jsonb ||
                                          (CASE
                                               WHEN can_change_vouchers = true AND can_view_vouchers = true THEN '{"event.vouchers:write": true}'
                                               ELSE '{}' END)::jsonb
                                          ) END,
    all_organizer_permissions   = CASE
                                      WHEN can_create_events = true AND can_change_organizer_settings = true AND
                                           can_change_teams = true AND can_manage_gift_cards = true AND
                                           can_manage_customers = true AND
                                           can_manage_reusable_media = true
                                          THEN true
                                      ELSE false END,
    limit_organizer_permissions = CASE
                                      WHEN can_create_events = true AND can_change_organizer_settings = true AND
                                           can_change_teams = true AND can_manage_gift_cards = true AND
                                           can_manage_customers = true AND
                                           can_manage_reusable_media = true
                                          THEN '{}'::jsonb
                                      ELSE (
                                          (CASE
                                               WHEN can_create_events = true
                                                   THEN '{"organizer.events:create": true}'
                                               ELSE '{}' END)::jsonb ||
                                          (CASE
                                               WHEN can_change_organizer_settings = true
                                                   THEN '{"organizer.settings.general:write": true, "organizer.devices:read": true, "organizer.devices:write": true, "organizer.seatingplans:write": true, "organizer.outgoingmails:read": true}'
                                               ELSE '{}' END)::jsonb ||
                                          (CASE
                                               WHEN can_change_teams = true
                                                   THEN '{"organizer.teams:write": true}'
                                               ELSE '{}' END)::jsonb ||
                                          (CASE
                                               WHEN can_manage_gift_cards = true
                                                   THEN '{"organizer.giftcards:read": true, "organizer.giftcards:write": true}'
                                               ELSE '{}' END)::jsonb ||
                                          (CASE
                                               WHEN can_manage_customers = true
                                                   THEN '{"organizer.customers:read": true, "organizer.customers:write": true}'
                                               ELSE '{}' END)::jsonb ||
                                          (CASE
                                               WHEN can_manage_reusable_media = true
                                                   THEN '{"organizer.reusablemedia:read": true, "organizer.reusablemedia:write": true}'
                                               ELSE '{}' END)::jsonb
                                          ) END
;

INSERT INTO "django_migrations" ("app", "name", "applied")
VALUES ('pretixbase', '0298_pluggable_permissions', NOW());
COMMIT;

-- After code deployment
BEGIN;
ALTER TABLE "pretixbase_team"
    DROP COLUMN "can_change_event_settings" CASCADE;
ALTER TABLE "pretixbase_team"
    DROP COLUMN "can_change_items" CASCADE;
ALTER TABLE "pretixbase_team"
    DROP COLUMN "can_change_orders" CASCADE;
ALTER TABLE "pretixbase_team"
    DROP COLUMN "can_change_organizer_settings" CASCADE;
ALTER TABLE "pretixbase_team"
    DROP COLUMN "can_change_teams" CASCADE;
ALTER TABLE "pretixbase_team"
    DROP COLUMN "can_change_vouchers" CASCADE;
ALTER TABLE "pretixbase_team"
    DROP COLUMN "can_checkin_orders" CASCADE;
ALTER TABLE "pretixbase_team"
    DROP COLUMN "can_create_events" CASCADE;
ALTER TABLE "pretixbase_team"
    DROP COLUMN "can_manage_customers" CASCADE;
ALTER TABLE "pretixbase_team"
    DROP COLUMN "can_manage_gift_cards" CASCADE;
ALTER TABLE "pretixbase_team"
    DROP COLUMN "can_manage_reusable_media" CASCADE;
ALTER TABLE "pretixbase_team"
    DROP COLUMN "can_view_orders" CASCADE;
ALTER TABLE "pretixbase_team"
    DROP COLUMN "can_view_vouchers" CASCADE;
ALTER TABLE "pretixbase_team"
    ALTER COLUMN "limit_organizer_permissions" DROP DEFAULT;
ALTER TABLE "pretixbase_team"
    ALTER COLUMN "limit_event_permissions" DROP DEFAULT;
ALTER TABLE "pretixbase_team"
    ALTER COLUMN "all_event_permissions" DROP DEFAULT;
ALTER TABLE "pretixbase_team"
    ALTER COLUMN "all_organizer_permissions" DROP DEFAULT;
COMMIT;
```

src/pretix/control/forms/event.py
```diff
@@ -238,6 +251,18 @@ def clean(self):
                               'check "{field}" above.').format(field=self.fields["no_taxes"].label)
             })
 
+        if self.clone_from and not self.user.has_active_staff_session(self.session.session_key):
+            if data.get("team"):
+                source_event_perms = self.user.get_event_permission_set(self.organizer, self.clone_from)
+                team_perms = data["team"].event_permission_set(include_legacy=False)
+                print(source_event_perms)
+                print(team_perms)
```
wiffbi:
```suggestion
```

src/pretix/base/models/event.py
```diff
@@ -843,6 +843,33 @@ def payment_term_last(self):
...
+            # Copying in the same organizer is always okay with any read access, we just need to ensure it does not
+            # grant more permissions than I had before, but that is handled by the view logic
+            return auth.has_event_permission(self.organizer, self, None)
+
+        if isinstance(auth, User):
+            # Cross-organizer copying requires almost full permission of source to prevent settings extraction
+            required_permissions = get_all_event_permissions() - {
+                # We do not require these, as this data is not copied
+                "event:orders.read", "event:orders.write", "event:vouchers.read", "event:vouchers.write",
+                "event:subevents.write",
```
wiffbi:
```suggestion
                "event.orders:read", "event.orders:write", "event.vouchers:read", "event.vouchers:write",
                "event.subevents:write",
```

src/pretix/control/forms/event.py
```diff
@@ -239,6 +251,18 @@ def clean(self):
...
+                if any(t not in source_event_perms for t in team_perms if ":" in t):
```
wiffbi: Why the `if ":" in t` – all new permissions have a ":" in them and we got the permissions with `include_legacy=False?`

src/pretix/control/forms/event.py
```diff
@@ -333,6 +357,24 @@ def __init__(self, *args, **kwargs):
...
+        if d.get('copy_from_event') and not self.user.has_active_staff_session(self.session.session_key):
+            if not d['copy_from_event'].allow_copy_data(self.organizer, self.user):
+                raise ValidationError({
+                    "copy_from_event": _("You do not have a sufficient level of access on the event you selected "
+                                         "to copy it to the desired organizer.")
+                })
+            if self.team:
+                source_event_perms = self.user.get_event_permission_set(self.organizer, d['copy_from_event'])
+                team_perms = self.team.event_permission_set(include_legacy=False)
+                if any(t not in source_event_perms for t in team_perms if ":" in t):
```
wiffbi: Same as above, why the ` if ":" in t`?

src/pretix/control/views/checkin.py
```diff
@@ -295,7 +295,7 @@ def get_success_url(self, value):
 class CheckinListList(EventPermissionRequiredMixin, PaginationMixin, ListView):
     model = CheckinList
     context_object_name = 'checkinlists'
-    permission = 'can_view_orders'
+    permission = ('event.orders:read', 'event.settings.general:write')
```
luelista: It often confuses me whether this means "any of these" or "all of these". What do you think about adding a marker class called e.g. `AnyPermissionOf` of `any_permission_of`, to make this more obvious?


```suggestion
    permission = AnyPermissionOf('event.orders:read', 'event.settings.general:write')
```

This would be fully backwards compatible and could be changed over time.

src/pretix/base/permissions.py
```diff
@@ -0,0 +1,329 @@
...
+            assert_valid_organizer_permission(p)
+        return
+    if not allow_legacy and ':' not in permission:
+        raise ValueError(f"Not allowed to use legacy permission '{permission}'")
+    all_permissions = get_all_organizer_permissions()
+    if permission not in all_permissions:
+        # Warning *and* exception because warning is silently caught when used in if statements in Django templates
+        warnings.warn(f"Use of undefined permission '{permission}'")
+        raise Exception(f"Undefined permission '{permission}'")
+
```
luelista, line 155:
```suggestion


class AnyPermissionOf(list):
    def __init__(self, *items):
        super().__init__(items)

```
(see other comment)

src/pretix/base/permissions.py
```diff
@@ -0,0 +1,329 @@
...
+
+def assert_valid_organizer_permission(permission, allow_legacy=True, allow_tuple=True):
+    if not apps.ready:
+        # can't really check yet
+        return
+    if permission is None:
+        return
+    if isinstance(permission, (list, tuple)) and allow_tuple:
```
luelista:
```suggestion
    if isinstance(permission, (list, tuple, AnyPermissionOf)) and allow_tuple:
```
(not required, but clarifies intent)

src/pretix/base/permissions.py
```diff
@@ -0,0 +1,329 @@
...
+
+def assert_valid_event_permission(permission, allow_legacy=True, allow_tuple=True):
+    if not apps.ready:
+        # can't really check yet
+        return
+    if allow_legacy and permission == "can_change_settings":
+        permission = "can_change_event_settings"
+    if permission is None:
+        return
+    if isinstance(permission, (list, tuple)) and allow_tuple:
```
luelista:
```suggestion
    if isinstance(permission, (list, tuple, AnyPermissionOf)) and allow_tuple:
```
(not required, but clarifies intent)

doc/development/implementation/permissions.rst
```diff
@@ -66,9 +66,9 @@ Of course, the same is available on event level:
         # Only users with *any* permission on this event can access this
```
luelista, line 72:
```suggestion
   class MyOtherEventView(EventPermissionRequiredMixin, View):
       permission = None
       # Only users with *any* permission on this event can access this


   class MyThirdEventView(EventPermissionRequiredMixin, View):
       permission = AnyPermissionOf('event.settings.payment:write', 'event.settings.general:write')
       # Only users with at least one of the specified permissions on this event 
       # can access this


```

luelista (commented):
I had an idea to make the "tuple-permissions" a bit more obvious to understand. Can also put this in a separate PR though

src/pretix/base/models/event.py
```diff
@@ -843,6 +843,33 @@ def payment_term_last(self):
...
+            }
+            given_permission = auth.get_event_permission_set(self.organizer, self)
+            return all(p in given_permission for p in required_permissions if ":" not in p)
```
wiffbi: Why does this compare legacy permissions? Both legacy and new permissions are included in required_permissions and given_permission, why not just compare the new ones? I.e.:


```suggestion
            return all(p in given_permission for p in required_permissions if ":" in p)
```

Or even compare all of them as both should contain legacy and new permissions?

src/tests/api/test_exporters.py
```diff
@@ -817,3 +824,232 @@ def test_org_scheduled_export_validate_rrule(user_client, organizer, user):
     )
     assert resp.status_code == 400
     assert resp.data == {"schedule_rrule": ["BYEASTER not supported"]}
+
+
+def _can_see_but_not_edit_org_export(client, scheduled):
```
raphaelm: note to self, this is named wrong and tests the wrong thing when negated

PR #5738, opened 2025-12-19
Mail settings: Correctly declare plaintext email (Z#23218835)

raphaelm:


src/pretix/control/forms/event.py
```diff
@@ -1311,9 +1311,17 @@ class MailSettingsForm(FormPlaceholderMixin, SettingsForm):
...
-        widget=I18nMarkdownTextarea,
-        help_text=_("This will only be used if the invoice is sent to a different email address or at a different time "
-                    "than the order confirmation."),
+        widget=I18nTextarea,  # no Markdown supported
+        help_text=lazy(
+            lambda: str(_(
+                "This will only be used if the invoice is sent to a different email address or at a different time "
+                "than the order confirmation."
+            )) + " " + str(_(
+                "Formatting is not supported, as some accounting departments process mail automatically and to not "
```
luelista:
```suggestion
                "Formatting is not supported, as some accounting departments process mail automatically and do not "
```

src/pretix/control/views/event.py
```diff
@@ -851,6 +851,12 @@ def post(self, request, *args, **kwargs):
                                 msgs[self.supported_locale[idx]] = prefix_subject(self.request.event, format_map(
                                     bleach.clean(v), self.placeholders(preview_item), raise_on_missing=True
                                 ), highlight=True)
+                            elif k in MailSettingsForm.plain_rendering:
```
luelista:
```suggestion
                            elif preview_item in MailSettingsForm.plain_rendering:
```
otherwise this never matches due to the index suffix

src/pretix/control/views/event.py
```diff
@@ -851,6 +851,12 @@ def post(self, request, *args, **kwargs):
...
+                                msgs[self.supported_locale[idx]] = mark_safe(
+                                    conditional_escape(format_map(
+                                        bleach.clean(v), self.placeholders(preview_item), raise_on_missing=True
+                                    ), highlight=True).replace("\n", "<br />")
```
luelista: conditional_escape doesn't have a highlight parameter

luelista: conditional_escape doesn't work here, as it escapes our placeholder-marker `<span>`s. i think it's not necessary either, because we already 1. bleach the template and 2. escape the placeholder contents in `get_sample_context`

raphaelm: Some escaping is necessary, though, otherwise if I put `<b>` in the content, will later be rendered, which makes the preview not previewing

raphaelm: now it works :)

image

PR #5744, opened 2025-12-22
Localize state names

wiffbi:
Fixes #5743


src/pretix/base/models/orders.py
```diff
@@ -3480,7 +3480,7 @@ def is_empty(self):
     def state_name(self):
```
raphaelm, line 3480: Same property exists on AbstractPosition and AttendeeProfile, probably needs the same change?

wiffbi: Good catch, fixed!

PR #5756, opened 2026-01-05
Fix race conditions in OrderChange API serializer

stranck:
As commented in issue #5548, there was a small and not critical race condition in the retrieval of newly created Order Positions. The first commit of this PR fixes it. 

While looking at the code I've noticed a similar pattern which is still prone to races in the creation of Fees. I've also updated the code to fix this as well by simply returning the same OrderFee object we pass to the OCM. Internally the OCM doesn't copy the object around, so the instance we create in the serializer is the one where we effectively call `.save()` in the commit phase of the ocm

I've also looked around in the codebase (by searching for `('-` usages) and it doesn't look like there are similar cases of this happening around


stranck:
Hmmm I currently have no idea how to properly add a test to check the addFee line I have changed. Can I ask for any suggestion?

raphaelm (approved):
I think this is fine without an extra test

PR #5781, opened 2026-01-08
Price rounding: Make sum_by_net_keep_gross even more consumer-friendly (Z#23220106)

raphaelm:
The rounding mode sum_by_net_keep_gross is supposed to be the consumer-friendly algorithm. However, some consumer-friendly prices are still impossible, such as 15.00 incl. 19%.  Previously, the system would round to 15.01, because it's the gross price derived from the closest net price. With this PR, it rounds to 14.99 instead, because consumers like paying less more than they like paying more.

### DIFF
--- a/src/tests/base/test_pricing_rounding.py
+++ b/src/tests/base/test_pricing_rounding.py
@@ -134,21 +134,29 @@ def test_revert_net_keep_gross_rounding_to_single_line(sample_lines):
     assert l.tax_rate == Decimal("19.00")
 
 
-@pytest.mark.django_db
-@pytest.mark.parametrize("rounding_mode", [
-    "sum_by_net",
-    "sum_by_net_keep_gross",
-])
-def test_rounding_of_impossible_gross_price(rounding_mode):
+def test_rounding_of_impossible_gross_price():
     l = OrderPosition(
         price=Decimal("23.00"),
     )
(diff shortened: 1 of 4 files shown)


src/tests/base/test_pricing_rounding.py
```diff
@@ -236,10 +244,8 @@ def test_do_not_touch_free(rounding_mode):
     )
     l2._calculate_tax(tax_rule=TaxRule(rate=Decimal("7.00")), invoice_address=InvoiceAddress())
     apply_rounding(rounding_mode, "EUR", [l1, l2])
-    assert l2.price == Decimal("23.01")
-    assert l2.price_includes_rounding_correction == Decimal("0.01")
-    assert l2.tax_value == Decimal("1.51")
-    assert l2.tax_value_includes_rounding_correction == Decimal("0.01")
+    assert l2.price == Decimal("23.01") or l2.price == Decimal("22.99")
```
luelista, line 247:
```suggestion
    assert l2.price == Decimal("23.01") or l2.price == Decimal("22.99")
    assert l2.tax_value == Decimal("1.51") or l2.tax_value == Decimal("1.50")
```
maybe keep that assertion as well?

luelista (approved):

raphaelm:
Currently no more demand, let's keep it around as an idea in case we need it again, and see what happens on the e-invoicing front

PR #5786, opened 2026-01-08
Exclude cancelled orders from paid orders graph

kiancross:
The 'paid orders' time series on the statistics page currently counts orders that were paid and later cancelled.

Filter the paid-by-day queryset to `Order.STATUS_PAID` with at least one non-cancelled position, leaving the placed orders series unchanged, and update the help text to clarify this behaviour.

Discussion: https://github.com/pretix/pretix/discussions/5774


src/pretix/plugins/statistics/views.py
kiancross, line 99: Whilst I'm working on this, I think checking for `all_positions__canceled=False` is inconsistent with the non-subevent behaviour? Does this check exist for a reason, or do we want to remove it?

raphaelm: Hmm, yes, could probably be removed. This all makes little sense.

PR #5787, opened 2026-01-08
Add more log entry types to the org-level logs page

ghickman:
This changes the Organizer's log page to show any log linked to the current Organizer, except those with a related Event, since they're already served by the Event's log page.  The changes to display_object ensure active plugins are checked against Organizer if there is no Event to use.


ghickman:
@raphaelm – I might have a related commit for this PR, lets hold off on merging it until I can can confirm.

PR #5791, opened 2026-01-09
Include fee-cancelled positions in placed orders by product graph

kiancross:
The 'placed orders by product' graph already includes orders that are pending, expired, or fully cancelled without a fee. However, items cancelled with a fee are omitted. This change ensures all placed orders are included in the graph, including those cancelled with a fee.


raphaelm:
This does make it more consistent, but… shouldn't we rather include everything canceled? Doesn't really make the graph useful including them

kiancross:
Do you mean _exclude_ everything cancelled? If so, I probably agree.

raphaelm:
> Do you mean _exclude_ everything cancelled? If so, I probably agree.

Yes.

Issue #5796, opened 2026-01-10
Wrong province list for Italy


stranck:
### Problem and impact

I've found a missing province (which is, in fact, a bit special) from the list for Italy. This is making a customer "unable" to put his correct data while doing an order

In particular, Aosta could not be found with any possible way

Aosta is a special case in italy, since the region Valle d'Aosta is the only region without a province, and the province role is assigned to the region itself, this is maybe why it's not in the list. However it still has a province code which is `AO`

Random official source from the government, this is a list of Cities related to its Province. In the last pages you will find Aosta https://dait.interno.gov.it/documenti/decreto-fl-24-06-2021-2-all-a.pdf

### Link

https://XXXXXXXX/org/event/checkout/questions/

### Version

2025.10.1

raphaelm:
Oh wow, I thought I caught all of the edge cases :D Will look into it!

raphaelm:
PR created: https://github.com/pretix/pretix/pull/5800

PR #5800, opened 2026-01-12
Address form: Add missing province Aosta for Italy (#5796)

raphaelm:


raphaelm:
We'll need to wait for a pycountry release to do that

PR #5805, opened 2026-01-13
Use transparent padding when resizing images with alpha support

kiancross:
Resized images that required padding were always padded with solid white. This caused visible bars around images with transparent elements when rendered on non-white backgrounds (e.g., [emails in dark mode](https://github.com/user-attachments/assets/5b8eca62-4c1a-453a-a026-412a045e76c6)).

Update the padding logic to:

- Prefer RGBA when an image is not already RGB/RGBA
- Use fully transparent padding instead of white

Images without an alpha channel continue to render with white padding, while images with transparency now preserve it correctly.

### DIFF
--- a/src/pretix/helpers/thumb.py
+++ b/src/pretix/helpers/thumb.py
@@ -152,8 +152,8 @@ def resize_image(image, size):
     if min_width > new_size[0] or min_height > new_size[1]:
         padding = math.ceil(max(min_width - new_size[0], min_height - new_size[1]) / 2)
         if image.mode not in ("RGB", "RGBA"):
-            image = image.convert('RGB')
-        image = ImageOps.expand(image, border=padding, fill="white")
+            image = image.convert('RGBA')
+        image = ImageOps.expand(image, border=padding, fill=(255, 255, 255, 0))
 
         new_width = max(min_width, new_size[0])
         new_height = max(min_height, new_size[1])


raphaelm:
I vaguely remember something along the lines of "Outlook does not support transparent PNGs", but I can't find a lot of info atm

kiancross:
I've tested on two Outlook clients: Outlook for Android and the Web App. I haven't tested on Outlook for Desktop (which I imagine is the most likely candidate for failure) as I don't have it installed, but I can install and do a test there if helpful. There were no issues in the afformentioned two tests.

kiancross:
It's from a long time ago, but perhaps [this](https://stackoverflow.com/a/12231908) is what you are referring to with regards to Outlook.

src/pretix/helpers/thumb.py
kiancross, line 217:
```suggestion
        image_out = image_out.convert('RGBA')
```

I think this might also need changing if changed earlier?

Another possible solution is to leave the default mode as `RGB` in both places and only add transparent padding if the original image had an alpha channel?

raphaelm:
Unfortunately we also lack access to Outlook for Desktop, which makes email testing hard – and this change too risky for now. I also don't find any reliable info out there on the web :( 

Litmus has unfortunately removed any pricing that is not "call us": https://www.litmus.com/pricing

kiancross:
Worth noting that I've tested on Outlook for Mac (Version 16.108.1), and it works without issue.

PR #5807, opened 2026-01-13
Invoices: Allow issuing invoices only to businesses (Z#23220397)

raphaelm:
In situations where every invoice has a significant accounting cost and consumers usually do not need invoices, this can save a lot of money or effort.

### DIFF
--- a/src/pretix/control/views/orders.py
+++ b/src/pretix/control/views/orders.py
@@ -509,9 +509,10 @@ class OrderView(EventPermissionRequiredMixin, DetailView):
 
     def get_context_data(self, **kwargs):
         ctx = super().get_context_data(**kwargs)
-        ctx['can_generate_invoice'] = invoice_qualified(self.order) and (
+        ctx['invoice_qualified'] = invoice_qualified(self.order)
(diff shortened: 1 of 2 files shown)


src/pretix/control/views/orders.py
```diff
@@ -1741,7 +1741,7 @@ def post(self, *args, **kwargs):
                     messages.error(self.request, _('The invoice has been cleaned of personal data.'))
                 else:
                     c = generate_cancellation(inv)
-                    if order.status not in (Order.STATUS_CANCELED, Order.STATUS_EXPIRED):
+                    if invoice_qualified(order):
```
wiffbi, line 1746: There is an edge-case: if I had previously either generated invoices for all customers or the invoice adress was a business and now the invoice-address is personal and I click on „cancel and reiusse“, it just cancels the invoice, but the message says „The invoice has been reissued.“ MIght be technically correct as the cancellation is an invoice too, but maybe we change the message here or maybe event better change the text on the button „cancel and reiusse“ to „cancel“ if the order is not invoice_qualified?

PR #5809, opened 2026-01-13
Invoice address: Add convenient autofill for Pepppol in Belgium (Z#23220397)

raphaelm:


src/pretix/static/pretixbase/js/addressform.js
```diff
@@ -166,6 +166,21 @@ $(function () {
...
+            let autofill_peppol_id = !!!dependents.transmission_peppol_participant_id.val();
+            const fill_peppol_id = function () {
+                const vatId = dependents.vat_id.val();
+                if (vatId && vatId.startsWith("BE") && dependents.transmission_type.val() === "peppol" && autofill_peppol_id) {
+                    dependents.transmission_peppol_participant_id.val("0201:" + vatId.substring(2))
+                }
+            }
+            $(dependents.vat_id).on("change", fill_peppol_id);
+            $(dependents.transmission_type).on("change", fill_peppol_id);
+            $(dependents.transmission_peppol_participant_id).on("change", () => { autofill_peppol_id = false; });
```
wiffbi: Wouldn’t it make more sense to just not bind/unbind the change event? Also, dependents.* are already jquery-objects, so no need for $() here. transmission_peppol_participant_id.change also only needs to be watched once, therefore `.one(…`  instead of `.on(…`

```suggestion
            if (!dependents.transmission_peppol_participant_id.val()) {
                const fill_peppol_id = function () {
                    const vatId = dependents.vat_id.val();
                    if (vatId && vatId.startsWith("BE") && dependents.transmission_type.val() === "peppol" && autofill_peppol_id) {
                        dependents.transmission_peppol_participant_id.val("0201:" + vatId.substring(2))
                    }
                }
                dependents.vat_id.add(dependents.transmission_type).on("change", fill_peppol_id);
                dependents.transmission_peppol_participant_id.one("change", () => {
                    dependents.vat_id.add(dependents.transmission_type).unbind("change", fill_peppol_id)
                });
            }
```

PR #5810, opened 2026-01-13
Tax rounding: Allow to apply only for B2B (Z#23220106)

raphaelm:
Most effective in combination with #5807

### DIFF
--- a/src/pretix/base/settings.py
+++ b/src/pretix/base/settings.py
@@ -81,7 +81,7 @@ from pretix.helpers.countries import CachedCountries
 ROUNDING_MODES = (
     ('line', _('Compute taxes for every line individually')),
     ('sum_by_net', _('Compute taxes based on net total')),
-    ('sum_by_net_only_business', _('Compute taxes based on net total for business customers')),
+    ('sum_by_net_only_business', _('For business customers, compute taxes based on net total. For individuals, use line-based rounding')),
     ('sum_by_net_keep_gross', _('Compute taxes based on net total with stable gross prices')),
     # We could also have sum_by_gross, but we're not aware of any use-cases for it
 )


src/pretix/base/settings.py
```diff
@@ -81,6 +81,7 @@
 ROUNDING_MODES = (
     ('line', _('Compute taxes for every line individually')),
     ('sum_by_net', _('Compute taxes based on net total')),
+    ('sum_by_net_only_business', _('Compute taxes based on net total for business customers')),
```
wiffbi: Hm, it is okay as it is, but somehow I read „net total for business customers“ (whatever that means – sorry, my language-brain might be wired wrong) and not „for business customers compute …“, so maybe change it to the follwing – can even help for a11y as it clearly distinguishes this option from the one above with a distinct wording/start:
```suggestion
    ('sum_by_net_only_business', _('For business customers, compute taxes based on net total. For individuals, use line-based rounding')),
```

src/pretix/presale/views/__init__.py
```diff
@@ -262,6 +262,11 @@ def group_key(pos):  # only used for grouping, sorting is done before already
             'max_expiry_extend': max_expiry_extend,
             'is_ordered': bool(order),
             'itemcount': sum(c.count for c in positions if not c.addon_to),
+            'show_rounding_info': (
+                self.request.event.settings.tax_rounding == "sum_by_net_only_business" and
+                not self.request.event.settings.display_net_prices and
+                sum(c.price_includes_rounding_correction for c in positions) + sum(f.price_includes_rounding_correction for f in fees)
```
wiffbi, line 268: Wouldn’t it be enough to check if any position/fee includes a rounding correction? `Decimal("0.0")` is falsy …
```suggestion
                (any(c.price_includes_rounding_correction for c in positions) or any(f.price_includes_rounding_correction for f in fees))
```

raphaelm: For the current algorithms… yes, would be sufficient, I guess, but in the end the user cares about the change in sum?

PR #5811, opened 2026-01-14
Waitinglist: lock entry while we create the voucher (Z#23220154)

bockstaller:
- moves the "has this waitinglist entry a voucher" check into the transaction 
- locks waitinglist entry for the duration of the voucher creation

### DIFF
--- a/src/pretix/base/models/waitinglist.py
+++ b/src/pretix/base/models/waitinglist.py
@@ -35,6 +35,7 @@ from pretix.base.email import get_email_context
 from pretix.base.i18n import language
 from pretix.base.models import User, Voucher
 from pretix.base.services.mail import SendMailException, mail, render_mail
+from pretix.helpers import OF_SELF
 
 from ...helpers.format import format_map
 from ...helpers.names import build_name


src/pretix/base/models/waitinglist.py
```diff
@@ -185,44 +185,47 @@ def send_voucher(self, quota_cache=None, user=None, auth=None):
...
-        if self.voucher:
-            raise WaitingListException(_('A voucher has already been sent to this person.'))
         if '@' not in self.email:
             raise WaitingListException(_('This entry is anonymized and can no longer be used.'))
 
         with transaction.atomic():
-            e = self.email
-            if self.name:
-                e += ' / ' + self.name
+            locked_wle = WaitingListEntry.objects.select_for_update().get(pk=self.pk)
```
wiffbi: As a safety-measure especially in conjunction with our ScopedManager, so we do not accidentally lock the organizer as well, I’d add a `of=OF_SELF`, just to make sure.

```suggestion
            locked_wle = WaitingListEntry.objects.select_for_update(of=OF_SELF).get(pk=self.pk)
```

wiffbi (changes requested):
Looks good, I would just add the of=OF_SELF to make sure we do not lock something accidentally, that we did not intend to lock.

src/pretix/base/models/waitinglist.py
```diff
@@ -185,44 +185,47 @@ def send_voucher(self, quota_cache=None, user=None, auth=None):
...
-        if self.voucher:
-            raise WaitingListException(_('A voucher has already been sent to this person.'))
         if '@' not in self.email:
             raise WaitingListException(_('This entry is anonymized and can no longer be used.'))
 
         with transaction.atomic():
-            e = self.email
-            if self.name:
-                e += ' / ' + self.name
+            locked_wle = WaitingListEntry.objects.select_for_update(of=OF_SELF).get(pk=self.pk)
```
wiffbi, line 193: Sorry, forgot to mention you need to `from pretix.helpers import OF_SELF`

PR #5814, opened 2026-01-16
Allow to combine language variant with region (fixes #3947, Z#23220951)

### LINKED ISSUE #3947: Configured region not honored with de-informal language
### Problem and impact

When setting my user account’s language to de-informal, the configured region (CH) is no longer honored for formatting money values, e.g. in the orders page.

### Expected behaviour

_No response_

### Steps to reproduce

1. Configure region to CH
2. Configure language to de; orders page lists “CHF 97.56” (with dots)
3. Configure language to de-informal; orders page lists “97,56 CHF” (with commas)

Looking at the code, I think the problem is that the language function treats “informal” as a region: https://github.com/pretix/pretix/blob/6bf23b0fddbac3aa8df6f81db35c91fdcc058aba/src/pretix/base/i18n.py#L146-L154

### Screenshots

_No response_

### Link

_No response_

_No response_

_No response_

### Version

2024.2

raphaelm:
This only affects babel-based formatting (currently: currencies and phone numbers), **not** Django-based formatting (currently: date and time formats).

### DIFF
--- a/src/pretix/base/i18n.py
+++ b/src/pretix/base/i18n.py
@@ -128,8 +128,8 @@ def get_babel_locale():
         if current_region:
             try_locales.append(f"{lng_parts[0]}_{lng_parts[1].title()}_{current_region.upper()}")
             try_locales.append(f"{lng_parts[0]}_{current_region.upper()}")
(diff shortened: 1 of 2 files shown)


raphaelm:
why is this breaking the build oO

PR #5819, opened 2026-01-19
Reduce queries for waitinglist autoassign n+1 (PRETIXEU-BJJ)

bockstaller:
I'm taking a stab at PRETIXEU-BJJ in the hope of reducing the number of queries for the plain event object, which is presumably due to cache misses.

I managed to remove two regressions that I introduced in #5811. 
See commits 89287349c7b85665303ed84e648c7ec7b9ec4302 and 24f5e153b04da8861acd1cd51bf76fababbeb9d8 for the respective changes and improvements.

However, I wasn't able to remove all n+1 queries. 
This is because mail_send_task() expects an event_id as a parameter (see here: https://github.com/pretix/pretix/blob/b5ad372bb29edc30199e7b3544fd05c91d0a68e1/src/pretix/base/services/mail.py#L415C5-L418C40).
Therefore, there will always be `n` queries for the event itself, where `n` is equal to the number of vouchers sent out.

### DIFF
--- a/src/tests/base/test_waitinglist.py
+++ b/src/tests/base/test_waitinglist.py
@@ -21,7 +21,6 @@
 #
 from datetime import timedelta
 
-from django.contrib.contenttypes.models import ContentType
 from django.core import mail as djmail
 from django.test import TestCase
 from django.utils.timezone import now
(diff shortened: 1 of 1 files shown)


src/pretix/base/models/waitinglist.py
```diff
@@ -191,6 +192,7 @@ def send_voucher(self, quota_cache=None, user=None, auth=None):
 
         with transaction.atomic():
             locked_wle = WaitingListEntry.objects.select_for_update(of=OF_SELF).get(pk=self.pk)
+            locked_wle.event = ev
```
raphaelm: `self.event`, not `ev`, I think? `ev` can also be a `SubEvent` in which case this will fail or later crash

Issue #5836, opened 2026-01-22
OIDC login can fail when multiple authentication requests are in flight


kiancross:
### Problem and impact

When multiple OIDC login attempts are initiated in quick succession (for example, by double-clicking the login button), pretix may have multiple authentication requests in flight simultaneously.

These concurrent login attempts can interfere with each other while they are still in progress, leading to shared state being overwritten.

As a result, a login attempt may fail even though the OIDC provider reports a successful authentication.

The issue is timing-dependent: when the bug occurs, users see an error message stating "invalid one-time token".

### Expected behaviour

For OIDC-based login methods, multiple rapid clicks on the login button should not cause the authentication flow to fail.

### Steps to reproduce

1. Configure an OIDC login provider in pretix.
2. Open the login page in a browser.
3. Click the OIDC login button and complete authentication at the OIDC provider.
4. Log out.
5. Return to the login page.
6. Double-click the OIDC login button to start two login attempts in quick succession.
7. If both attempts succeed, log out and repeat the previous step, varying the timing between clicks.

The issue tends to reproduce most reliably with a fast double-click. In practice, it occurs roughly 1 in 5 attempts, depending on timing and network conditions.

### Screenshots

<img width="600" src="https://github.com/user-attachments/assets/75961b48-9267-4d53-9db2-fc67d57abe0a" />

### Link

http://host/org/account/login

### Browser (software, desktop or mobile?) and version

Google Chrome

### Version

2025.10.1

kiancross:
I've investigated this further and believe one cause is a race between concurrent requests writing to the same Django session.

In brief:

1. Login attempt A starts and stores OIDC state (nonce / PKCE) in the session.
2. The callback for A begins and loads a snapshot of the session.
3. Before the callback for A finishes, login attempt B starts and writes new OIDC state to the same session.
4. The callback for A then completes and saves its earlier session snapshot, overwriting the state written by B (last write wins).
5. When the callback for B later returns from the IdP, the required OIDC state is missing, and the login fails.

<details>
<summary>A more detailed step-by-step</summary>

1. Request A (OIDC login attempt 1) starts (`/account/login/...`).
2. Request A writes authentication state (e.g., nonce/PKCE/etc.) and saves the session (302 to the IdP).
3. Request A callback starts (`/account/login/.../return?...state=...`).
4. The callback loads the session (snapshot), and validates the login.
[...]

PR #5842, opened 2026-01-25
Prevent double-clicks on SSO login providers

kiancross:
Disable SSO login provider links after the first click to prevent users from starting multiple concurrent login flows.

This avoids accidental double-clicks or taps that can otherwise result in overlapping authentication requests and subsequent login failures.

Fixes one of the issues in #5836.

### DIFF
--- a/src/pretix/static/pretixpresale/js/ui/main.js
+++ b/src/pretix/static/pretixpresale/js/ui/main.js
@@ -733,6 +733,7 @@ $(function () {
     });
 
     $("#customer-account-login-providers a").click(function () {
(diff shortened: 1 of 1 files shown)


src/pretix/static/pretixpresale/js/ui/main.js
```diff
@@ -731,6 +731,10 @@ $(function () {
             $(countInput).trigger("change");
         });
     });
+
+    $("#customer-account-login-providers a").click(function () {
+        $(this).addClass("disabled");
```
raphaelm, line 737:
```suggestion
        // Prevent double-submit, see also https://github.com/pretix/pretix/issues/5836
        $(this).addClass("disabled");
```

PR #5843, opened 2026-01-25
Disable partitioned cookies for Safari due to WebKit bugs

kiancross:
Safari currently exhibits a bug where Partitioned cookies (CHIPS) are not sent back to the originating site after multi-hop cross-site redirects, breaking SSO login flows in pretix.

Partitioned cookies were initially introduced in Safari 18.4, removed again in 18.5 due to a bug, and reintroduced in Safari 26.2, where the current issue is present.

As a mitigation, disable sending the `Partitioned` attribute for Safari user agents. This is intentionally conservative; once the Safari issue is fixed, this check should be refined to be conditional on the affected versions only.

WebKit issues:

  - https://bugs.webkit.org/show_bug.cgi?id=292975
  - https://bugs.webkit.org/show_bug.cgi?id=306194

Fixes the primary cause of #5836.

### DIFF
--- a/src/pretix/helpers/cookies.py
+++ b/src/pretix/helpers/cookies.py
@@ -34,7 +34,10 @@ def set_cookie_without_samesite(request, response, key, *args, **kwargs):
     if not is_secure:
         # https://www.chromestatus.com/feature/5633521622188032
         return
-    if should_send_same_site_none(request.headers.get('User-Agent', '')):
+
+    useragent = request.headers.get('User-Agent', '')
+
+    if should_send_same_site_none(useragent):
         # Chromium is rolling out SameSite=Lax as a default
         # https://www.chromestatus.com/feature/5088147346030592
         # This however breaks all pretix-in-an-iframe things, such as the pretix Widget.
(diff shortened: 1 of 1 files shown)


kiancross:
(Leaving this as a draft for a few days until we confirm the issue is resolved with our users.)

raphaelm:
Uh, tough choice, I feel the price for this will be "break the widget for thousands of users on hundreds of sites", which hurts similarly bad.

kiancross:
Partitioned cookies are only available in Safari from v26.2 (released 12 Dec 2025). Our own data show that only ~35% of Safari users (across macOS and iOS) are currently on 26.2, which aligns with [other reports](https://9to5mac.com/2026/01/08/most-iphone-users-are-still-running-ios-18-with-ios-26-adoption-slow-report/) of slow iOS 26.2 uptake. For the remaining ~65%, Partitioned cookies have no effect today, so disabling them does not change behaviour for those users.

For those who have upgraded, disabling Partitioned cookies regresses the widget only to its pre-26.2 behaviour (i.e., how Safari behaved a month ago), whereas leaving things as-is breaks login flows.

Given that CHIPS support on Safari has already been introduced, removed, and reintroduced due to bugs -- and that the widget has, I believe, been broken there for a while already -- it seems reasonable to wait a little longer before depending on Partitioned cookies on Safari. That said, I appreciate that this is ultimately a judgment call about how urgently pretix wants to restore widget support for Safari users.

raphaelm:
Yeah, if this is so recent in Safari, I think we can live with merging this PR for now. Would be easier to decide if there was acknowledgement by Apple and some insight into when this will be fixed :D

src/pretix/helpers/cookies.py
```diff
@@ -44,8 +47,14 @@ def set_cookie_without_samesite(request, response, key, *args, **kwargs):
...
         response.cookies[key]['secure'] = is_secure
-        # CHIPS
-        response.cookies[key]['Partitioned'] = True
+
+        if can_send_partitioned_cookie(useragent):
+            # CHIPS
+            response.cookies[key]['Partitioned'] = True
+
+
+def can_send_partitioned_cookie(useragent):
```
raphaelm, line 56: Can you add a comment referencing the Safari bug so future readers of this code have a chance of figuring out when it can be removed?

kiancross: Done.

kiancross:
> (Leaving this as a draft for a few days until we confirm the issue is resolved with our users.)

I can confirm that this fixed the issue for our users (it was the primary cause of #5836).

wiffbi:
Thank you!

raphaelm:
Unfortunately, we needed to revert this on master after deploying it to pretix Hosted production, as it caused a high influx of customer reports that backend login and other things broke.

My gut feeling is that not the change in itself is the problem, but the overlap in cookie expiration while the change is made. Don't have a good idea on how to fix that, though.

PR #5848, opened 2026-01-26
API: add organizer-level orderpositions endpoint

bockstaller:


src/pretix/api/views/order.py
```diff
@@ -1104,7 +1105,22 @@ def get_queryset(self):
...
+                qs = qs.filter(
+                    order__event__organizer=self.request.organizer,
+                    order__event__in=self.request.auth.get_events_with_permission(perm, request=self.request)
+                )
+            elif self.request.user.is_authenticated:
+                qs = qs.filter(
+                    order__event__organizer=self.request.organizer,
+                    order__event__in=self.request.auth.get_events_with_permission(perm, request=self.request)
+                )
+
```
bockstaller: If the client doesn't have the required permissions this results in:
- query against safe methods -> empty result
- query against an unsafe method -> 404

Which does the job but our current docs say this about errors:

> [403 Forbidden](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.4) – The requested organizer/event does not exist or you have no permission to view this resource or downloads are not available for this order position at this time. The response content will contain more details.
> [404 Not Found](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4.5) – The requested order position or download provider does not exist.

Depending on how one might read this. Returning a 403 or 404 might be correct if the querying team doesn't have access to an event.

src/pretix/api/views/order.py
```diff
@@ -1174,6 +1172,40 @@ def _get_output_provider(self, identifier):
                 return prov
         raise NotFound('Unknown output provider.')
 
+
+class OrganizerPositionViewSet(OrderPositionViewSetMixin, viewsets.ReadOnlyModelViewSet):
```
wiffbi:
```suggestion
class OrganizerOrderPositionViewSet(OrderPositionViewSetMixin, viewsets.ReadOnlyModelViewSet):
```

src/pretix/api/urls.py
```diff
@@ -67,6 +67,7 @@
 orga_router.register(r'scheduled_exports', exporters.ScheduledOrganizerExportViewSet)
 orga_router.register(r'exporters', exporters.OrganizerExportersViewSet, basename='exporters')
 orga_router.register(r'transactions', order.OrganizerTransactionViewSet)
+orga_router.register(r'orderpositions', order.OrganizerPositionViewSet, basename='orderpositions')
```
wiffbi:
```suggestion
orga_router.register(r'orderpositions', order.OrganizerOrderPositionViewSet, basename='orderpositions')
```

doc/api/resources/orders.rst
```diff
@@ -1719,6 +1719,51 @@ List of all order positions
...
+
+      {
+        "count": 1,
+        "next": null,
+        "previous": null,
+        "results": [
+          {
+            "id:": 23442
+            "code": "ABC12",
+            "event": "sampleconf",
```
wiffbi: I think this is taken from order-endpoint, but needs to be adapted to orderpositions.
```suggestion
            "order": "ABC12",
            "positionid": 1,
            "canceled": false,
            "item": 1345,
```

src/pretix/api/views/order.py
```diff
@@ -1104,7 +1103,6 @@ def get_queryset(self):
         else:
             qs = OrderPosition.objects
 
-        qs = qs.filter(order__event=self.request.event)
         if self.request.query_params.get('pdf_data', 'false').lower() == 'true':
```
wiffbi: Make sure that pdf_data query_param is only honored, when there is an event:


```suggestion
        if self.request.query_params.get('pdf_data', 'false').lower() == 'true' and getattr(request, 'event', None):
```

src/pretix/api/views/order.py
wiffbi: Not for this PR, but we might evaluate, why `select_related` is used for organizer and event instead of `prefetch_related`. Even order and item might be helpful to prefetch instead of select to reduce the size of the query result/number of columns.

src/pretix/api/views/order.py
```diff
@@ -1174,6 +1172,40 @@ def _get_output_provider(self, identifier):
...
+        if isinstance(self.request.auth, (TeamAPIToken, Device)):
+            qs = qs.filter(
+                order__event__organizer=self.request.organizer,
+                order__event__in=self.request.auth.get_events_with_permission(perm, request=self.request)
+            )
+        elif self.request.user.is_authenticated:
+            qs = qs.filter(
+                order__event__organizer=self.request.organizer,
+                order__event__in=self.request.auth.get_events_with_permission(perm, request=self.request)
+            )
```
wiffbi, line 1188: I do not see difference in both filter-calls. I know `OrganizerOrderViewSet` has the same code, but we might optimize it? The organizer-filter should IMHO go into the `OrderPositionViewSetMixin` – technically it is not needed there, as EventOrderPositionViewSet filters on one event, which is garantueed to be from that organizer, but it’s IMHO nice to have as a double check nonetheless.

```suggestion
        if isinstance(self.request.auth, (TeamAPIToken, Device)) or self.request.user.is_authenticated::
            qs = qs.filter(
                order__event__in=self.request.auth.get_events_with_permission(perm, request=self.request)
            )
```

src/pretix/api/views/order.py
```diff
@@ -1104,7 +1103,6 @@ def get_queryset(self):
         else:
             qs = OrderPosition.objects
```
wiffbi: I would add a filter on the organizer here, just to make sure even if the `EventOrderPositionViewSet` does not need it.


```suggestion
        qs = qs.filter(order__event__organizer=self.request.organizer)
```

src/pretix/api/views/order.py
```diff
@@ -1174,6 +1172,40 @@ def _get_output_provider(self, identifier):
...
+            )
+
+        return qs
+
+
+class EventOrderPositionViewSet(OrderPositionViewSetMixin, viewsets.ModelViewSet):
+    def get_serializer_context(self):
+        ctx = super().get_serializer_context()
+        if hasattr(self.request, 'event'):
+            ctx['event'] = self.request.event
```
wiffbi: Should always have the event, so no need to check:


```suggestion
        ctx['event'] = self.request.event
```

wiffbi (changes requested):
Looks good. Just minor suggestions.

doc/api/resources/orders.rst
raphaelm: should we also add the event slug to the output here as well though?

raphaelm: should we also add the event slug to the output here as well though?

wiffbi: That’s what I thought as well. It would help avoid a roundtrip to the order-endpoint to get the event. Another idea I had would be to include the API-URL for that orderposition – so consumers of the API would not need to know how to create proper URLs for getting info on an oderposition. But I think including the event-slug here is more consistent with the current API. So I’d say yes!

bockstaller: done

src/pretix/api/views/order.py
```diff
@@ -1174,6 +1172,37 @@ def _get_output_provider(self, identifier):
...
+
+    def get_queryset(self):
+        qs = super().get_queryset()
+
+        perm = self.permission if self.request.method in SAFE_METHODS else self.write_permission
+
+        if isinstance(self.request.auth, (TeamAPIToken, Device)) or self.request.user.is_authenticated:
+            qs = qs.filter(
+                order__event__in=self.request.auth.get_events_with_permission(perm, request=self.request)
+            )
```
wiffbi: This is a bit special and you probably did not fall into this during testing – when requesting the resource through a logged-in session in the browser, it fails. Also an authenticed user might have access to more organizers, which could result in more orderpositions than expected. You need to adapt it like so:

```suggestion
        if isinstance(self.request.auth, (TeamAPIToken, Device)):
            auth_obj = self.request.auth
        elif self.request.user.is_authenticated:
            auth_obj = self.request.user

        qs = qs.filter(
            order__event__in=auth_obj.get_events_with_permission(perm, request=self.request).filter(
                organizer=self.request.organizer
            )
        )
```

doc/api/resources/orders.rst
wiffbi: @raphaelm  should we only add the event-slug to the orderposition on organizer-level or should the orderposition on the event level have the event-slug as well? Not needed on event-level, but might make the interface more consistent. Currently the only way to see that event-slug is in the response is in the example for organizer-orderpositions …

raphaelm: I would add it only on the organizer level, the same as for invoices. It's not perfect, but it's easier to add more things later than remove them

doc/api/resources/orders.rst
```diff
@@ -1719,6 +1719,54 @@ List of all order positions
    :statuscode 401: Authentication failure
    :statuscode 403: The requested organizer/event does not exist **or** you have no permission to view this resource.
 
+.. http:get:: /api/v1/organizers/(organizer)/orderpositions/
+
+   Returns a list of all order positions within all events of a given organizer (with sufficient access permissions).
```
raphaelm, line 1724: This should include a note of the additional "event" attribute included in the response

src/pretix/api/views/order.py
raphaelm: event and organizer would be a good fit for prefetch_related, yes, order is guesswork at best

src/pretix/api/views/order.py
```diff
@@ -1174,6 +1172,37 @@ def _get_output_provider(self, identifier):
...
+
+class OrganizerOrderPositionViewSet(OrderPositionViewSetMixin, viewsets.ReadOnlyModelViewSet):
+    serializer_class = OrganizerOrderPositionSerializer
+
+    def get_queryset(self):
+        qs = super().get_queryset()
+
+        perm = self.permission if self.request.method in SAFE_METHODS else self.write_permission
+
+        if isinstance(self.request.auth, (TeamAPIToken, Device)) or self.request.user.is_authenticated:
```
raphaelm: I wonder what the "else" case for this if statement would be (probably never happens?) and if it would be safer to avoid future security problems to just not have the if statement.

src/pretix/api/views/order.py
```diff
@@ -1174,6 +1172,37 @@ def _get_output_provider(self, identifier):
...
+            qs = qs.filter(
+                order__event__in=self.request.auth.get_events_with_permission(perm, request=self.request)
```
raphaelm: That fails if `self.request.auth` is an `OAuthAccessToken`. In this case, the get_events_with_permission needs to run against `self.request.user`, not `self.request.auth`

wiffbi (approved):

PR #5851, opened 2026-01-26
Order changes: Do not allow to double-book add-ons (Z#23220592)

raphaelm:


src/pretix/presale/templates/pretixpresale/event/fragment_addon_choice.html
```diff
@@ -372,6 +372,19 @@ <h4 id="cp-{{ form.pos.pk }}-item-{{ item.pk }}-legend">{{ item.name }}</h4>
                 </article>
             {% endif %}
         {% endfor %}
+
+        {% if c.items_missing %}
+            <div class="product-appendix-row">
+                <p class="text-muted">
+                    {% trans "There are currently selected products in this add-on category that currently cannot be changed because they are not on sale:" %}
```
wiffbi: Maybe remove the doubling of „currently“?

```suggestion
                    {% trans "There are products selected in this add-on category that currently cannot be changed because they are not on sale:" %}
```

PR #5866, opened 2026-01-31
Enable email send task rate limits

nicoknoll:
This PR makes it possible to set task rate limits for the the email send task via env variables.

This is especially relevant to prevent ticket selling peaks from getting flagged by your email provider (e.g. most providers have a hourly or daily limit).

**Sample usage**

```bash
PRETIX_MAIL_TASK_RATE_LIMIT=10/m
```

### DIFF
--- a/src/pretix/base/services/mail.py
+++ b/src/pretix/base/services/mail.py
@@ -371,7 +371,7 @@ class CustomEmail(EmailMultiAlternatives):
         return super()._create_mime_attachment(content, mimetype)
 
 
-@app.task(base=TransactionAwareTask, bind=True, acks_late=True)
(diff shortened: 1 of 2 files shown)


raphaelm:
I'm not sure we'll want to merge this, as celery is not really good at rate limits.

See celery docs:

> Note that this is a per worker instance rate limit, and not a global rate limit. To enforce a global rate limit (e.g., for an API with a maximum number of requests per second), you must restrict to a given queue.

This becomes useless quickly for any larger setup. Also, for any multi-tenant setup the rate limit would have to be by organizer or by SMTP server.

Also, it sounds like the implementation works by buffering the tasks over the rate limit in the worker memory, which might have reliability issues.

nicoknoll:
hmm fair points - but having a rate limit option per smtp server still sounds like not the worst idea to me. And I guess I'll have to continue using my forked version then for now :)

PR #5879, opened 2026-02-09
Pay in installments from within pretix

sweenu:
The non-profit I'm a part of is switching systems for its event shop. After reviewing the possibilities, pretix is faaaar better than anything else I found (for our usecase at least). It's pretty(x) awesome!

However, as a non-profit, trying to reduce costs for the participants to our events, using third-party installments providers is not an option for us. Yet, we would very much like to have installments.

I'm in a bit of a tricky situation, because ideally, I would have implemented it as a plugin, but the feature is a bit too core for a plugin. I can also try and carry the burden of maintenance by myself if you don't want to merge this, but it's obviously a loss for me as I would miss out on whole-repo refactorings, etc. 
I also think that if the feature is useful for us, it might also be for others!
If you do merge it, I'd love to help as much as I can with maintenance (of this but also in general) to make up for it.

As it's a big PR, I took time to reorder the changes in digestible commits that could be reviewed independently.

Here's a demo of the main functionalities:
https://github.com/user-attachments/assets/88786129-91d5-409f-8fb0-714456b3fcb2

## Manual testing
I need to continue manually testing the features, here's where I'm at:
### User functionalities
- [x] Pay in installments
- [x] Get charged automatically when next installment is due
- [x] Get reminders emails before the next installment is due
- [x] Get grace period warning emails
- [x] Get installment failure email with link to recovery page
- [x] Recover a failed installment by changing the payment method
### Control
- [x] Filter by installment plan status
- [x] Show installment status in order page
- [x] Cancel installment plan (optionally cancelling the order itself) from order page
- [x] Retry a failed installment from order page

### API
- [x] Get single installment plan
- [x] Cancel installment plan
- [x] Get list of installment plans
- [x] Retry a failed installment

## Open questions
- What should happen on order reactivation ? Currently, we revoke the token when an order is cancelled. Should we not and only revoke after a certain time so that we can reactivate the plan as well?


raphaelm:
Hi there, nice to meet you!

pretix is a hugely complex system, and we need to make sure that all features work together properly and that the overall complexity remains manageable. Therefore, when wanting to add a major feature, we usually recommend to open a discussion first and talk about the requirements, the data model and the implementation approach, to avoid spending a lot of time for something that later does not fit elsewhere. With merging, we assume full responsibility with maintenance, since even if you are happy to help, if it breaks for our customers, we can't wait for that ;) Therefore, this will unlikely be something we merge quickly, since thinking about it and reviewing it will take time. However, installments are a regularly requested feature, so it is certainly interesting!

I have only scrolled through it very briefly (data model only) and watched your video. I'll need to think about it more, but I think it's a good start, but not quite perfect yet.

One thing about the design that bugs me is the way the installment plan is connected to a payment provider, and shown as a sub-option of the payment option in the UI. Wouldn't it be very useful as a customer to be able to pay the first installment via bank transfer and the second one via credit card because I'm late and otherwise the deadline expires? Or, I'd want to recover after a failed installment with a different payment method. (nb: Should one also be able to pay their future installments early if they want to?)

Then, the design assumes that the payment provider is able to "charge" the installments (pull) which only very few payment providers can do. We'd also want to support installments for push-based payments, like bank transfer. Even if they can, there's no guarantee the payment feedback is received immediately, I think process_single_installment would need to be much more complex to account for 

Then, it is still unclear to me how this interacts with the existing payment deadlines and reminder features of pretix. I assume the Order.expires should be set after the last installment? What happens if installments fail? etc

sweenu:
Thank you for your quick reply!

My expectations for getting it merged were not super high, and getting it merged quickly, even less so. So for me, the fact that it's a requested feature is already very good news.
Sorry for not opening a discussion before. We're likely to start using this fork anyway, so for me, it was simpler to start a discussion with something already tangible to propose. I'm very much ready to change the design totally.

At first, I had implemented the settings at the event level instead of at the payment provider level. And the "Pay in installments" option was, at checkout, outside the payment provider selection. But I changed the design as it was complicating the logic by having to dynamically disable payment providers depending on if they support installments or not.

This PR handles failed payments by giving a way to change payment methods (using the same payment provider though). It does not indeed, in any way, support installments across multiple payment providers.

If you think designing it this way would create problems for future improvements, then it makes sense to completely change the design now. Otherwise, I guess, I thought this would be a first good step for the most common usage of installments (at least IMO): I pay now by card, and get automatically debited on the same card the next few months.

For order expiration, I simply filtered out orders with active installment plans for expiration and expiration warnings. But maybe that's too naïve.

Do you want me to open a discussion, or should we continue here?

sweenu:
Brought the branch up to date with master. Also finished doing manual testing and fixed the two issues that came up.

PR #5884, opened 2026-02-11
Fix address-helper wrong locale (Z#23223920)

wiffbi:
This PR adds a locale-URL-param to the address-helper. This is needed if the event has a different language as default than what the browser suggests in its request-header. If the event is in one language only or the user does not switch languages, then there is no cookie for language. Even if the user has a language-cookie, but in a language not supported by the event, the global address-helper-enpoint responded with the wrong locale applied.

I planned on using `get_supported_language` from middleware, but that falls back to the first allowed language. This PR falls back to the middleware’s language, which could differ. Or should it return 404 if the language is unknown?

### DIFF
--- a/src/pretix/base/views/js_helpers.py
+++ b/src/pretix/base/views/js_helpers.py
@@ -164,16 +164,7 @@ def _address_form(request):
 
 def address_form(request):
     locale = request.GET.get('locale')
-    allowed_languages = dict(settings.LANGUAGES)
-    if locale and locale not in allowed_languages and '-' in locale:
-        locale = locale.split('-')[0]
-        if locale not in allowed_languages:
-            for lang in allowed_languages:
-                if lang.startswith(locale + '-'):
-                    locale = lang
-                    break
-
-    if locale in allowed_languages:
+    if locale in dict(settings.LANGUAGES):
         with language(locale):
             info = _address_form(request)
     else:
--- a/src/pretix/presale/templates/pretixpresale/event/checkout_questions.html
+++ b/src/pretix/presale/templates/pretixpresale/event/checkout_questions.html
@@ -42,7 +42,7 @@
                     {% endif %}
                     <div id="invoice" class="profile-scope"
                          data-profiles-id="addresses_json"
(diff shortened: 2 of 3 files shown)


src/pretix/base/views/js_helpers.py
```diff
@@ -91,7 +93,21 @@ def _info(cc):
...
-    info = _info(cc)
+    locale = request.GET.get('locale')
+    allowed_languages = dict(settings.LANGUAGES)
+    if locale and locale not in allowed_languages and '-' in locale:
+        locale = locale.split('-')[0]
+        if locale not in allowed_languages:
+            for lang in allowed_languages:
+                if lang.startswith(locale + '-'):
+                    locale = lang
+                    break
```
raphaelm: Are you sure we need this and can't just drop it? html_locale always contains something that should be in settings.LANGUAGES, does it not?


```suggestion
```

src/pretix/base/views/js_helpers.py
```diff
@@ -91,7 +93,21 @@ def _info(cc):
...
+
+    if locale in allowed_languages:
+        with language(locale):
```
raphaelm: I think the lower part,

```
    if request.GET.get("invoice") == "true":
```

also needs to be executed within this language block, otherwise we get the same problem with things like

```
                    result = {"name": str(t.public_name), "code": t.identifier}
```
being in the wrong language.

Probably makes sense to refactor it into a function to not duplicate the code?

src/pretix/base/views/js_helpers.py
wiffbi: Interestingly it does not. In my tests (mimicking the linked Z-ticket) I had a html_locale „en-de“, which does not exist.

src/pretix/base/views/js_helpers.py
wiffbi: You are right, I just looked at the state-names, which are translated inside `_info(cc)` with non-lazy versions of gettext. Did not think of the transmission-types. I will look into it.

PR #5891, opened 2026-02-13
refactor: potrait -> image crop validator

Legiarrd:
This PR refactors the current portraits validator, turning it into a more generic image crop validator that supports multiple crop formats.
The motivation behind this change is that for certain types of events, such as furry conventions, it’s more common to use 1:1 ratio images for badges instead of the default 4:3 format. With this update, users can choose which image format they prefer to use.

image


raphaelm:
Hey, thanks a lot! I'll discuss it internally, but for compatibility reasons I'm not sure we'll be able to merge this unless someone builds support for it in all of our photo-taking apps, pretixSCAN Android, pretixSCAN Desktop, and pretixSCAN iOS, which probably takes some effort :/

Legiarrd:
> Hey, thanks a lot! I'll discuss it internally, but for compatibility reasons I'm not sure we'll be able to merge this unless someone builds support for it in all of our photo-taking apps, pretixSCAN Android, pretixSCAN Desktop, and pretixSCAN iOS, which probably takes some effort :/

Yeah, I thought so as I made the changes. I would like to take a look into those apps but might take awhile since I am not familiar with Kotlin.

raphaelm:
At the moment, we will not be able to merge this as the Android changes are lacking and we're not able to make it a priority right now. I'm closing the PR for now, if we want to revisit it in the future, we'll still be able to find it :)

PR #5892, opened 2026-02-13
feat: add transparency option and EXIF rotation support for badge images

Legiarrd:
This PR adds an option to make the background transparent when no image is present, instead of displaying the grey box. The grey box option remains available, and the new behavior can be enabled optionally by checking the corresponding checkbox. Additionally, the image rotation is now respected based on its EXIF data.

image

### DIFF
--- a/src/pretix/base/pdf.py
+++ b/src/pretix/base/pdf.py
@@ -975,17 +975,18 @@ class Renderer:
                 )
                 canvas.restoreState()
         else:
-            canvas.saveState()
-            canvas.setFillColorRGB(.8, .8, .8, alpha=1)
-            canvas.rect(
-                x=float(o['left']) * mm,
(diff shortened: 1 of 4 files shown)


raphaelm:
Thanks! This will also need a change on libpretixprint for android, but that's likely an easy fix

Legiarrd:
> a change on libpretixprint for android, but that's likely an easy fix

to double check, this lib is still in use? I see the linked project for the android app is archived

raphaelm:
It is, thanks, I'll update the README

raphaelm:
At the moment, we will not be able to merge this as the Android changes are lacking and we're not able to make it a priority right now. I'm closing the PR for now, if we want to revisit it in the future, we'll still be able to find it :)

PR #5893, opened 2026-02-13
Fix HTML closing tags in pending.html

KarlKeu00:


wiffbi:
Good catch, thank you!

PR #5895, opened 2026-02-16
Mail: Handle all rendering in mail.py, return values for log

raphaelm:


src/pretix/control/views/orders.py
```diff
@@ -2413,9 +2413,9 @@ def form_valid(self, form):
         with language(order.locale, self.request.event.settings.region):
             email_context = get_email_context(event=order.event, order=order)
         email_template = LazyI18nString(form.cleaned_data['message'])
-        email_subject = format_map(str(form.cleaned_data['subject']), email_context)
-        email_content = render_mail(email_template, email_context)
         if self.request.POST.get('action') == 'preview':
+            email_subject = format_map(str(form.cleaned_data['subject']), email_context)
+            email_content = render_mail(email_template, email_context)
```
pajowu, line 2418: both message and subject are CharFields, i.e. `cleaned_data` already has strings for them. Why are they being converted to a `str`/`LazyI18nString`?

```suggestion
            email_subject = format_map(form.cleaned_data['subject'], email_context)
            email_content = render_mail(email_template, email_context)
```

raphaelm: email_template must be a LazyI18nString so it is not treated as a filename by mail().

Making subject a string is pointless, yes.

PR #5904, opened 2026-02-17
Handle PlainHtmlAlternativeString in placeholder help text

pajowu:
before: Screenshot 2026-02-17 at 17 27 31
after: Screenshot 2026-02-17 at 17 27 50


wiffbi:
Wouldn’t it be easier to just add a `__str__` to `PlainHtmlAlternativeString`?

pajowu:
Yes and no: it would fix this (and other occurrences where we want the plaintext), but might hide places we want the html-version. This makes the decision more explicit, but I don’t see a stronger reason not to add a str-method

wiffbi:
Hm, well, you are right, explicit ist better!

PR #5912, opened 2026-02-20
Do not use partitioned cookies for Safari

wiffbi:
#5843 fixed a bug in multi-hop-cross-site SSO-login steps in Safari, but introduced issues for users with exisiting session-cookies. The problem with the original code was, existing partitioned cookies were not deleted and took precedence over the newly introduced non-partitioned cookies. As long as the partitioned cookies contained a valid session, it works fine. But as soon as the session turned invalid, Safari users could not log in as long as the cookie was still in the browser.

This PR deletes any partitioned cookies before setting new non-partitioned cookies. As django does not allow to set cookies with the same key, this PR is kind of a hack sadly.

To reproduce you need to have a dev-setup with HTTPS and a current version of Safari.

1. Log into pretix-control before checking out this PR/branch. You should see your session-cookie with a Partition-Key. 
2. Checkout this PR/branch
3. To imitate an expired session, change the value in Safaris Session-Cookie to something else
4. Reload pretix-control => Login
5. Login - you now should not have any partitioned session-cookie

### DIFF
--- a/src/pretix/helpers/cookies.py
+++ b/src/pretix/helpers/cookies.py
@@ -47,6 +47,34 @@ def set_cookie_without_samesite(request, response, key, *args, **kwargs):
         # CHIPS
         response.cookies[key]['Partitioned'] = True
 
+        if is_safari(useragent):
+            # Safari currently exhibits a bug where Partitioned cookies (CHIPS) are not
+            # sent back to the originating site after multi-hop cross-site redirects,
+            # breaking SSO login flows in pretix.
+            #
+            # Partitioned cookies were initially introduced in Safari 18.4, removed
+            # again in 18.5 due to a bug, and reintroduced in Safari 26.2, where the
+            # current issue is present.
+            #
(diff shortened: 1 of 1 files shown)


raphaelm:
:rotating_light: Merging this will make #6212  more complicated and needs extreme caution.

I wonder if we need it at all. https://bugs.webkit.org/show_bug.cgi?id=292975 is apparently fixed in Safari 26.2 and we have very little customer complaints.

https://bugs.webkit.org/show_bug.cgi?id=306194 ist still open though, @kiancross can you double-check if (a) this is still an issue with Safari 26.2+ and (b) if this PR here even fixes it?

kiancross:
This is still a problem on Safari 26.4. I'll try and test this PR in the next few days.

raphaelm:
We have spent *so* many hours cleaning up consequences from the first attempt, I'm really not sure if "wait until safari fixes this and keep SSO broken until then" isn't better than "keep breaking regular flows"

PR #5989, opened 2026-03-13
Migrate vue2 control components and widget to vue3 and vite

rashfael:
Migrated all vue2 code to vue3 and build them with vite:
- `pretixcontrol/checkin`
- `pretixplugins/webcheckin`
- `pretixpresale/widget`

All vue components have been changed to be single file components using pug, `script setup` and typescript in non-strict mode.
For the widget, a simple shared store has been introduced, replacing the global state that was previously in the root component.

# Usage

The default `runserver.py` manage command has been overwritten, which now starts a vite dev server when `VITE_DEV_MODE` is set. This dev server only build and watches the control components, not the widget.

To develop the widget, run `npm run dev:widget` and navigate to the localhost url that is printed.

```sh
PRETIX_WIDGET_LANG=de npm run dev:widget
npm run dev:control

npm run build
PRETIX_WIDGET_VITE=1 python manage.py runserver # uses widget built by vite (run npm run build first)

PRETIX_WIDGET_VITE=1 pytest […] # builds widget at test setup
PRETIX_WIDGET_VITE_DEV=1 pytest […] # use running dev server in tests
```

# Rollout

Widget can be gradually rolled out if `PRETIX_WIDGET_VITE` is not set by adding http origins to the `widget_vite_origins` global setting

# Todos

- [x] development docs (where do they go?)
- [x] node_modules folder path
- [ ] output paths

## Control

- [x] attribution for code from pretalx ?
- [x] test production / docker build
- [x] remove old VueCompiler and node_prefix folder and code

## Widget

- [ ] [non-english i18n in dev mode](https://github.com/rashfael/pretix/blob/8945e55900ad0b200f86724b13100c2db2e801b9/src/pretix/static/pretixpresale/widget/src/i18n.ts#L17)
- [x] [investigate wrapping html code](https://github.com/rashfael/pretix/blob/8945e55900ad0b200f86724b13100c2db2e801b9/src/pretix/static/pretixpresale/widget/src/button.ts#L57)
- [x] [investigate why these detached elements are created](https://github.com/rashfael/pretix/blob/8945e55900ad0b200f86724b13100c2db2e801b9/src/pretix/static/pretixpresale/widget/src/main.ts#L51)
- [x] revisit siteglobals / debugging

Plus a bunch of TODOs in code just for code organization or modernization, which can also be done later.


Dockerfile
```diff
@@ -20,11 +20,11 @@ RUN apt-get update && \
...
             libmaxminddb0 \
             libmaxminddb-dev \
-            zlib1g-dev \
-            nodejs  \
-            npm && \
+            zlib1g-dev && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/* && \
+    curl -fsSL https://deb.nodesource.com/setup_22.x | sudo bash - && \
+    apt-get install -y nodejs && \
```
luelista: i think the new calls should go before `rm -rf /var/lib/apt/lists` (seems to be size optimization of the resulting image, and calling apt afterwards recreates those caches)

i think the ideal way would be like this (not tested though!)

```
RUN curl -fsSL https://deb.nodesource.com/setup_22.x | sudo bash - && \
    apt-get update && \
    apt-get install -y --no-install-recommends \
            build-essential \
            gettext \
            git \
            libffi-dev \
            libjpeg-dev \
            libmemcached-dev \
            libpq-dev \
            libssl-dev \
            libxml2-dev \
            libxslt1-dev \
            locales \
            nginx \
            python3-virtualenv \
            python3-dev \
            sudo \
            supervisor \
            libmaxminddb0 \
            libmaxminddb-dev \
            zlib1g-dev \
            nodejs && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/* && \
```

(github refuses to add this as a proper 'suggestion', so just a code block...)

.gitignore
```diff
@@ -24,5 +24,6 @@ local/
 .project
 .pydevproject
 .DS_Store
+node_modules/
```
luelista, line 27: for me, vite created a .vite folder in the project root 

```suggestion
node_modules/
.vite/
```

src/pretix/static/pretixpresale/widget/TODOS.md
```diff
@@ -0,0 +1,2 @@
+- modernize the sometimes native form submitting?
+- destructure props?
```
luelista: this file should probably be removed pre-merge
```suggestion
```

.gitignore
rashfael: Hm, interesting, by default it should go in `node_modules/.vite` https://vite.dev/config/shared-options.html#cachedir

Dockerfile
rashfael: Good catch, I'll patch that.

PR #6000, opened 2026-03-17
Email rate limits using sliding windows

nicoknoll:
After we figured out that my initial approach (https://github.com/pretix/pretix/pull/5866) was a little naive and only worked reliably using one worker, this time I worked on a more sophisticated solution making rate limiting emails work via fixed windows and making it configurable per smtp client

**Why:** Our email provider limits sending to a fixed number of emails per time window.                                                                                                           
                                                                                                                                                                                         
**How it works:** A shared Redis counter tracks how many emails have been sent through each SMTP server within a configurable time window. When the limit is reached, emails are deferred and  automatically retried after the window resets.
                                                                                                                                                                                         
Configuration (in order of priority):
- Per-event or per-organizer: in the SMTP setup under "Advanced settings"
- System-wide via pretix.cfg: [mail] → rate_limit_count / rate_limit_window                                                                                                                
- Global admin UI: Settings → Email rate limit                             
                                                                                                                                                                                         
Set count to 0 (default) to disable. If no Redis is available, rate limiting is silently skipped. 

<img width="2314" height="1568" alt="Bildschirmfoto 2026-03-17 um 10 50 08" src="https://github.com/user-attachments/assets/9c9c833b-3bf1-423d-92b4-b1b10aa59bcb" />

Btw: I wasn't sure how translations are handled - like should I provide them in the MR or are they automatically generated using e.g. crowdin or so ?

### DIFF
--- a/src/pretix/base/services/mail.py
+++ b/src/pretix/base/services/mail.py
@@ -91,6 +91,7 @@ from pretix.presale.ical import get_private_icals
 
 logger = logging.getLogger('pretix.base.mail')
 INVALID_ADDRESS = 'invalid-pretix-mail-address'
+MIN_RATE_LIMIT_RETRY_AFTER = 5  # Minimum seconds before retrying a rate-limited email
 
 
 class TolerantDict(dict):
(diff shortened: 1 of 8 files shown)


nicoknoll:
Number #6000 🔥 😄

src/pretix/base/services/mail.py
```diff
@@ -435,6 +436,22 @@ def mail_send_task(self, **kwargs) -> bool:
         assert outgoing_mail.orderposition.order_id == outgoing_mail.order_id
         outgoing_mail.orderposition.order = outgoing_mail.order
 
+    # Rate limit check
+    backend = outgoing_mail.get_mail_backend()
+    is_limited, retry_after = _check_rate_limit(backend, outgoing_mail)
+    if is_limited:
+        logger.info(f"Rate limited email {outgoing_mail.guid}, retrying in {retry_after}s")
+        outgoing_mail.status = OutgoingMail.STATUS_AWAITING_RETRY
+        outgoing_mail.retry_after = now() + timedelta(seconds=retry_after)
```
raphaelm, line 446: I think this should set an error message like "rate limited" so one knows why it is awaiting retry

raphaelm (commented):
This appears to be a sensible solution, I still need to find some time to play with it and think about whether I see any scaling issues.

One thought: This implements the assumption that rate limiting is done by `username@host` combination, which is true for some SMTP hosts, but some might also rate limit on IP address. However, we have no way of knowing which is which, so this is probably the best we can do.

src/pretix/base/services/mail.py
```diff
@@ -1025,6 +1041,56 @@ def _wrap_plain_body(content_plain, signature, event, order, position, no_order_
...
+    # Resolve rate limit settings via hierarkey (event → organizer → global)
+    event = getattr(outgoing_mail, 'event', None)
+    if event:
+        count = event.settings.get('smtp_rate_limit_count', as_type=int)
+        window = event.settings.get('smtp_rate_limit_window', as_type=int)
+    elif outgoing_mail.organizer:
+        count = outgoing_mail.organizer.settings.get('smtp_rate_limit_count', as_type=int)
+        window = outgoing_mail.organizer.settings.get('smtp_rate_limit_window', as_type=int)
+    else:
+        count, window = 0, 600
```
pajowu: Use explicit `None` istead of relying on false-y `0`

```suggestion
        count, window = None, None
```

src/pretix/base/services/mail.py
```diff
@@ -1025,6 +1041,56 @@ def _wrap_plain_body(content_plain, signature, event, order, position, no_order_
...
+
+    # Fallback to pretix.cfg
+    if not count and settings.EMAIL_RATE_LIMIT_COUNT:
```
pajowu: IMHO set pretix.cfg-Limit should always be used when the system email server is used and never if a custom server is used. Users configuring a custom server should be able to send as much as they want. But users using the system server should never be allowed to change the system limits.

src/pretix/control/forms/mailsetup.py
```diff
@@ -76,6 +76,19 @@ class SMTPMailForm(SettingsForm):
         help_text=_("Commonly enabled on port 465."),
         required=False
     )
+    smtp_rate_limit_count = forms.IntegerField(
```
pajowu, line 79: The new fields should also be cleared in the `MailSettingsSetupView.post`-Method

src/pretix/base/services/mail.py
```diff
@@ -1025,6 +1041,56 @@ def _wrap_plain_body(content_plain, signature, event, order, position, no_order_
     return body_plain
 
 
+def _check_rate_limit(backend, outgoing_mail):
+    """
+    Fixed-window Redis counter keyed by SMTP identity + time bucket.
+    Returns (is_limited: bool, retry_after_seconds: int).
+    """
+    if not settings.HAS_REDIS:
```
pajowu, line 1053: Why redis and not the django cache? (+ maybe checking for `settings.REAL_CACHE_USED`)

Is there something that isn't possible e.g. with `memcached`?

pajowu (commented):
I added a few inline comments for things I noticed when having a first look.

However I also have a few more general questions:
* This implements fixed windows. Do you know if fixed windows are actually common? In my personal experience I often saw rate-limits implemented as sliding windows or token buckets, which this couldn't properly address.
* If fixed windows are common, what are the usual intervals? I'm a bit worried about daily limits etc which might be influenced by timezones
* Why redis and not the django cache?

The calculation also currently has the "quirk" that limits with different windows are additive, but limits with the same window are shared. For example if I configure a limit of 5 mails / 30s (count=5, window=30) in one event and one of 5 mails / minute (count=10, window=60) in another event, pretix would send out up to 15 mails per minute. However if I configure both to be 10 mails / minute (count=10, window=60), pretix would send out 10 mails/minute in total.

nicoknoll:
Thanks for the comments - I'll have a deeper look on the code comments and the additive issue after the weekend :)

Regarding your questions:

I checked out different ways on how to do it and didn't want top make it overly complex for the first version (e.g. something like token buckets seemed a little over the top). I did try with sliding windows as well as fixed windows have the boundry issues when the smtp server uses fixed windows or similar w.g. when you have an interval of 30 min and a peak at the end of it and the start of the next you easily get over the limit. but then I came to the conclusion if I make the fixed windows small enough this won't be an issue e.g. instead of 30min and 300 mails -> 10min and 100 mails. 

I think intervals < 1h are quite common e.g. my host has a limit of 300 mails per 30 min. 
I guess sliding windows would be a very valid option as well though even though its slightly more complex

I used redis because I saw in the code there was another rat limtier that was also using redis (the reset password rate limit).

nicoknoll:
@pajowu Regarding the additive issue your describing: I think this is rather an edge case that you define different limits for the same smtp server - like when would I do that? Anyway it's a valid concern though as this could lead to potential blocks from the smtp provider. IMO there are some options here on how to approach that: 

- The easiest option I see would be showing a warning in the event level overwrite scope that defining custom limits can lead to issues. It's not a really nice solution though.
- having a second counter on a global / org level which would basically work like the ticket quotas. E.g. when one is exceeded, no emails will be send until both have space again.
- switch to sliding window with logged timestamps. its a litltle more complex and takes definetly more memory as each send email is logged as a timestamp and then the sliding window just counts and then decides based on that. This seems fairly clean so I'll have a look into that approach now

src/pretix/base/services/mail.py
nicoknoll: would that be populated to admin then as well? I'm just worried when e.g. hundreds of payment reminders go out at the same time this could spam the whole view ?

nicoknoll: I will add a error for the final status failed though

nicoknoll:
I updated it now to sliding windows - this gives the advantages of getting rid of the mentioned additive issue.
Additionally I fixed:
- I changed it for system SMTP to only use the config settings but events don't fallback to the system config
- the settings get cleaned in MailSettingsSetupView.post
- using None inested of 0 for no rate limiting (0 now means now emails are send)

I did this by using a readis pipeline to save redis calls. This also means we cannot easily switch to django cahche anymore though. 

I added a new config variable EMAIL_RATE_LIMIT_MAX_WINDOW that defines the max window size but also defines how long we keep the tracked email timestamps. By default I set it to 24h which probably is more then needed. In high volume production systems this could be reduced to the maximum used window.

nicoknoll:
Let me know if I can do anything else to push this further ahead :)

KarlKeu00:
Have there been any updates on this yet? This feature would be very helpful for me.

raphaelm:
At the moment we're a little swamped with to-dos and PRs but we'll look into it again in due time!

KarlKeu00:
Any updates on this? 🫣😅

nicoknoll:
@KarlKeu00 from my side everything is ready since March 🙈 but probably not the highest prio atm

dchmayer:
We also would really need this feature in October this year...

PR #6019, opened 2026-03-24
Fix delete_cookie for partitioned legacy CSRF and session cookie

wiffbi:
Removing legacy CSRF-cookies failed due to partitioned cookies being used. Not really bad, as everything kept working, but I got warnings in Firefox console and the old cookie wasn’t deleted and the datele_cookie response was sent with every response. Also, the first delete_cookie-call did not do anything as it was overwritten by the second as django keeps cookies in a dict keyed by cookie_name.

### DIFF
--- a/src/pretix/helpers/cookies.py
+++ b/src/pretix/helpers/cookies.py
@@ -22,7 +22,7 @@
 import itertools
 import re
 from http.cookies import Morsel
-import logging
+
 from django.conf import settings
 
 
--- a/src/pretix/helpers/cookies.py
+++ b/src/pretix/helpers/cookies.py
@@ -19,8 +19,10 @@
 # You should have received a copy of the GNU Affero General Public License along with this program.  If not, see
 # <https://www.gnu.org/licenses/>.
 #
+import itertools
 import re
-
+from http.cookies import Morsel
+import logging
 from django.conf import settings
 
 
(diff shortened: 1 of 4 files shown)


pajowu (approved):

src/pretix/helpers/cookies.py
```diff
@@ -48,6 +54,68 @@ def set_cookie_without_samesite(request, response, key, *args, **kwargs):
...
+    # Due to a Safari bug, in some browser, two cookies with the name same but different
+    # values can exist: one unpartitioned, one partitioned. This function generates an
+    # additional Set-Cookie header to get rid of the unpartitioned one.
+
+    if request.is_secure() and cookie_name in request.COOKIES:
+        values = get_all_values_of_cookie(request.headers.get('Cookie'), cookie_name)
+        if len(values) > 1:
+            logger.info('Trying to remove duplicated %s cookies: %r', cookie_name, values)
+
+            # Make sure the set_cookie_without_samesite below will add a new item in the dictionary, placing
```
raphaelm: "the set_cookie_without_samesite below" no longer applies or at least I'm unclear what it is refering to

src/pretix/helpers/cookies.py
```diff
@@ -48,6 +54,68 @@ def set_cookie_without_samesite(request, response, key, *args, **kwargs):
...
+            # Add the deletion Set-Cookie header to the cookie dict under a wrong name, so it doesn't get
+            # overwritten by the set_cookie_without_samesite call below.  This works because the code in
+            # django.core.handlers.wsgi/asgi, that generates the actual Set-Cookie headers, only iterates
+            # over cookie.values(), ignoring the keys.
+            response.cookies['___DELETECOOKIE___' + cookie_name] = make_delete_morsel(cookie_name)
+
+
+def thoroughly_delete_cookie(response, cookie_name, **kwargs):
+    """ Deletes different possible versions of a cookie (SameSite, Partitioned) """
+    properties = {"SameSite": ["", 'None'], "Partitioned": ["", True]}
```
raphaelm, line 55: Not something we currently have in the codebase, but if we'd ever have a cookie with, say, SameSite=Lax, this function would still only delete it with SameSite="" and SameSite=None. Should we for safety of future use check if kwargs['SameSite'] is something other then "" or "None" and then add it to the list?

pajowu: I'm not sure we actually need that, as that should not influence the storage key in the browsers cookie jar. `delete_cookie`s code sounds like it should not be the case? https://github.com/django/django/blob/stable/6.0.x/django/http/response.py#L293-L296

raphaelm: Oh, TIL, thanks!

luelista (approved):
🫣

PR #6020, opened 2026-03-24
wip: self service cancellation

bockstaller:


src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+        "ItemVariation",
+        verbose_name=_("Item variations"),
+    )
+
+    allowed_if_in_order_status=models.CharField(
+        max_length=4,
+        choices=Order.STATUS_CHOICE,
+        verbose_name=_("Cancellation possible if order is in status"),
+        validators=[validate_status_chars],
+        default="".join(Order.ALLOWED_STATUS_CHARS),
```
raphaelm: I don't think this approach for two reasons:

- It is more of an accident than a happy choice that order status is encoded as one-letter codes. while it is unlikely that we add a fifth status, I do not really want to hard-code a dependency on every status being one letter in other places, as it is an unexpected and surprising depenceny.
- It does not really make sense to have a cancellation rule for status "canceled", does it? Neither is "expired" really necessary. In reality, we'd rather need something like a choice field with choices "always", "paid" and "unpaid". Or even just a boolean field with "allow_when_paid", since it's hard to see a situation where cancellation for paid orders is more liberal than for unpaid ones.

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+        decimal_places=2,
+        validators=[
+            MaxValueValidator(
+                limit_value=Decimal("100.00"),
+            ),
+            MinValueValidator(
+                limit_value=Decimal("0.00"),
+            ),
+        ],
+        verbose_name=_("Fee Percentage per Item"),
```
raphaelm: A percentage is not really "per item", is it? 😉

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+        max_digits=10,
+        decimal_places=2,
+        validators=[
+            MaxValueValidator(
+                limit_value=Decimal("100.00"),
+            ),
+            MinValueValidator(
+                limit_value=Decimal("0.00"),
+            ),
+        ],
```
raphaelm:
```suggestion
        max_value=Decimal("100.00"),
        min_value=Decimal("0.00"),
```

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+    )
+    event=models.ForeignKey(
+        Event,
+        verbose_name=_("Event"),
+        related_name="orders",
+        on_delete=models.CASCADE
+    )
+    items = models.ManyToManyField(
+        "Item",
+        verbose_name=_("Items"),
```
raphaelm: We never call items "items" in the UI, the model is named this way for historical reasons, always name them "products" in translated/visible text :)

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+class CancellationRule(models.Model):
+    """
+
+
+    """
+    organizer=models.ForeignKey(
+        "Organizer",
+        related_name="orders",
```
raphaelm: 👀

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+        on_delete=models.CASCADE,
+        null=True,
+        blank=True,
+    )
+    event=models.ForeignKey(
+        Event,
+        verbose_name=_("Event"),
+        related_name="orders",
```
raphaelm: 👀

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+        choices=Order.STATUS_CHOICE,
+        verbose_name=_("Cancellation possible if order is in status"),
+        validators=[validate_status_chars],
+        default="".join(Order.ALLOWED_STATUS_CHARS),
+    )
+    allowed_until=ModelRelativeDateTimeField(null=True, blank=True)
+    except_after=ModelRelativeDateTimeField(null=True, blank=True)
+
+    fee_percentage_per_item=models.DecimalField(
+        max_digits=10,
```
raphaelm:
```suggestion
        max_digits=5,
```
should be enough for 0.00…100.00

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+    if len(value) != len(set(value)):
+        raise ValidationError("Duplicate characters are not allowed.")
+
+
+class CancellationRule(models.Model):
+    """
+
+
+    """
+    organizer=models.ForeignKey(
```
raphaelm: We don't really need a link to the organizer, do we? Some models like OrderPosition have it to allow certain UNIQUE keys, but usually the link via the event model is sufficient.

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+
+    allowed_if_in_order_status=models.CharField(
+        max_length=4,
+        choices=Order.STATUS_CHOICE,
+        verbose_name=_("Cancellation possible if order is in status"),
+        validators=[validate_status_chars],
+        default="".join(Order.ALLOWED_STATUS_CHARS),
+    )
+    allowed_until=ModelRelativeDateTimeField(null=True, blank=True)
+    except_after=ModelRelativeDateTimeField(null=True, blank=True)
```
raphaelm: This will require some good labeling and help texts to make clear what one should use this for

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+        null=True,
+        blank=True,
+    )
+    event=models.ForeignKey(
+        Event,
+        verbose_name=_("Event"),
+        related_name="orders",
+        on_delete=models.CASCADE
+    )
+    items = models.ManyToManyField(
```
raphaelm: I would implement this as a `all_products` and `limit_products`/`limit_variations`  combination similar to e.g. `AutoCheckinRule`. Otherwise, one needs to adjust their cancellation rule every time a product is added, which people will surely forget all the time. Also, we couldn't really have a default rule in place.

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+        # Signal eingesammelt und an CancellationRule.__init__ übergeben werden
+        # Ermöglicht dann: "Shipping modul kann storno geshippter Items verhindern"
+        self.checks: List[CheckFn]=[self._check_order_status, self._check_time_window,
+                                    self._system_check_not_checked_in, self._system_check_not_discounted]
+
+
+    # TODO weitere System Checks
+    # OrderPositions mit Item.min_per_order dürfen nur storniert werden, wenn genug übrig bleiben oder alle des gleichen Items storniert werden
+    # OrderPositions mit addon_to != None dürfen nur über den bestehenden Add-On-Flow storniert werden
+    # OrderPositions mit is_bundled dürfen nur mit der Parent-Position zusammen storniert werden
```
raphaelm: Ich frag mich, ob die Dinge, die hier als @staticmethod passieren oder mehrere items umfassen wie das mit dem is_bundled überhaupt richtig hier aufgehoben sind, oder dann in services/orders.py umziehen sollten. Auch ein Plugin signal sollte wie das für shipping sollte vmtl eher einmal aufgerufen werden für die bestellung, statt `anzahl(positionen) * anzahl(regeln)` mal, oder?

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+
+    @staticmethod
+    def _system_check_not_checked_in(diff: OrderDiff, order_position: OrderPosition) -> CheckResult:
+        check_id = "SYSTEM_TICKET_NOT_USED"
+
+        if order_position.checkins.filter(list__consider_tickets_used=True).exists():
```
raphaelm: Siehe oben, ich glaub es wär cooler, wenn diese Checks auf einem Set von OrderPositions (bzw der ganzen Bestellung) arbeiten können und irgendwie zurückgeben, was stornierbar ist oder nicht.

Sonst brauchen wir bei einer Bestellung mit 500 Tickets erstmal 500 SQL Queries, um zu wissen, ob wir den Storno-Button überhaupt anzeigen oder nicht. Das skaliert nicht gut.

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+            return {check_id: CheckRes(
+                cancellation_possible=False,
+                reason=f"Order position was used",
```
raphaelm: Should these be shown to users? If yes, they need to be translatable. If they are only for debugging, it's fine.

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+        AbstractPositions is introduced, allowing us to be more grannular
+
+        :param diff:
+        :param order_position:
+        :return CheckResults:
+        """
+        check_id = "SYSTEM_TICKET_NOT_DISCOUNTED"
+
+        if order_position in diff.cancellations():
+            if order_position.discount is None:
```
raphaelm: Trick:

```suggestion
            if order_position.discount_id is None:
```

If `select_related` is not used, this would otherwise cause a `SELECT` SQL query. This way, it doesn't.

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+        else:
+            return {check_id: CheckRes(
+                cancellation_possible=False,
+                reason=_("Order position not canceled - check not applicable"),
+            )}
+
+    def _check_time_window(self, diff: OrderDiff, order_position: OrderPosition) -> CheckResult:
+        check_id = "TIME_WINDOW"
+
+        if not self.allowed_until and not self.allowed_until:
```
raphaelm: Seems a bit redundant ;)

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+                cancellation_possible=True,
+                reason=f"No time window specified",
+            )}
+
+        relevant_event=order_position.subevent or order_position.event
+        in_allowed_until=time_machine_now() < self.allowed_until.datetime(
+                relevant_event) if self.allowed_until else False
+        in_exemption=time_machine_now() > self.except_after.datetime(
+                relevant_event) if self.except_after else False
+
```
raphaelm: `ModelRelativeDateTimeField`/`RelativeDateWrapper` still need changes to allow "x days after order", right?

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+                reason=f"Order in status '{diff.order.status}' cannot be canceled",
+            )}
+
+
+    def apply(self, diff: OrderDiff, order_position: OrderPosition) -> Ruling:
+        check_results=reduce(merge_check_results,
+                       [rule(diff, order_position) for rule in self.checks])
+
+        if self.fee_percentage_per_item and self.fee_absolute_per_item:
+            raise NotImplementedError("Should never be reached")
```
raphaelm: We could even support this ("5€ + 5%"), couldn't we?

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+@dataclass(frozen=True)
+class RelativeFee:
+    reference_price: Decimal
+    percentage: Decimal
+
+    @property
+    def amount(self):
+        return self.reference_price * (self.percentage/100)
```
raphaelm: I think we should use round_decimal to round with the proper currency. For example, it makes no sense to charge 3.45 Japanese Yen, since units smaller than 1.00 do not exist.

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+Fee=Union[AbsoluteFee, RelativeFee]
+
+@dataclass(frozen=True)
+class OrderDiff:
+    order: Order
+    prev: Set[OrderPosition]
+    next: Set[OrderPosition]
```
raphaelm: From the current state of the PR, I was unable to guess what these mean and why we need them. What are `prev` and `next` previous and next to here?

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+
+
+CheckFn=Callable[[OrderDiff, OrderPosition], CheckResult]
+
+
+def merge_check_results(a: CheckResult, b: CheckResult) -> CheckResult:
+    result = dict(a)
+    for key, b_inner in b.items():
+        if key in result:
+            result[key] = result[key] | b_inner
```
raphaelm: What's the result of a binary OR between two CheckRes? I don't really get it from reading the code

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+        if not isinstance(other, Ruling):
+            return NotImplemented
+
+        if self.cancellation_possible == other.cancellation_possible:
+            return self.total_fee < other.total_fee
+        else:
+            return self.cancellation_possible and not other.cancellation_possible
+
+
+class CancellationRuleQuerySet(models.QuerySet):
```
raphaelm: This all looks a little bit too clever to be read quickly… and looks like it is doing a large number of queryies later if we're not really careful.

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+    for key, b_inner in b.items():
+        if key in result:
+            result[key] = result[key] | b_inner
+        else:
+            result[key] = b_inner
+    return result
+
+
+@dataclass(frozen=True)
+class Ruling:
```
raphaelm: I don't really get if this is a ruling on an order or on an order position.

If it's on an order, I'm missing the part where the position_fees for multiple positions are summed up.

If it's on an order position, I'm not yet sure how it will later work if different rules match on different positions and have different order fees.

src/pretix/base/models/cancellation.py
```diff
@@ -0,0 +1,326 @@
...
+    cancellation_possible: bool
+    reason: str
+
+CheckResult=Dict[str, CheckRes]
+
+
+CheckFn=Callable[[OrderDiff, OrderPosition], CheckResult]
+
+
+def merge_check_results(a: CheckResult, b: CheckResult) -> CheckResult:
```
raphaelm: I think if you keep to these low-level data types like CheckResult and define operations between them, this is a prime opportunity to create unit tests that test the low-level primitives without needing database objects set up etc.

PR #6026, opened 2026-03-24
Delete unused code

raphaelm:


src/pretix/presale/forms/checkout.py
```diff
@@ -168,46 +165,6 @@ def __init__(self, *args, **kwargs):
             )
 
 
-class AddOnRadioSelect(forms.RadioSelect):
-    option_template_name = 'pretixpresale/forms/addon_choice_option.html'
```
pajowu, line 172: I think this template is unused not as well?

PR #6051, opened 2026-04-01
Centralise order filter logic, properly handle cancelled orders in sendmail (Z#23230149)

pajowu:


src/pretix/base/models/orders.py
```diff
@@ -204,6 +322,68 @@ class Order(LockModel, LoggedModel):
...
+            ('pa', _('Approval pending')),
+        )),
+        (_('Follow-up date'), (
+            ('custom_followup_at', _('Follow-up configured')),
+            ('custom_followup_due', _('Follow-up due')),
+        )),
+        ('testmode', _('Test mode')),
+    )
+
+    STATUS_FILTER_OPTIONS = (
```
raphaelm: I don't really like having this list twice, with the possibility to introduce differences in the set of choices or their labels. In this PR, the order is already inconsistent. Should we prefer to compute STATUS_FILTER_OPTIONS from STATUS_FILTERS?

src/pretix/base/models/orders.py
```diff
@@ -134,6 +134,124 @@ def get_with_secret_check(self, code, received_secret, tag, secret_length=64):
...
+
+        if 'o' in status:
+            filter |= Q(status=Order.STATUS_PENDING, expires__lt=now().replace(hour=0, minute=0, second=0))
+        if 'np' in status:
+            filter |= Q(status__in=[Order.STATUS_PENDING, Order.STATUS_PAID])
+        if 'ne' in status:
+            filter |= Q(status__in=[Order.STATUS_PENDING, Order.STATUS_EXPIRED])
+        if 'pv' in status:
+            filter |= Q(status=Order.STATUS_PAID) | Q(status=Order.STATUS_PENDING, valid_if_pending=True)
+        for s in ('p', 'n', 'e', 'c', 'r'):
```
raphaelm:
```suggestion
        for s in ('p', 'n', 'e', 'c'):
```

r was dropped many pretix versions ago, apparently it still lives on

src/pretix/base/models/orders.py
```diff
@@ -134,6 +134,124 @@ def get_with_secret_check(self, code, received_secret, tag, secret_length=64):
             raise Order.DoesNotExist
         return order
 
+    def filter_by_status(self, qs, status: list[str]):
+        from .invoices import Invoice
+
+        if not isinstance(status, list):
+            raise ValueError("`status` needs to be a list of strings")
```
raphaelm:
```suggestion
            raise TypeError("`status` needs to be a list of strings")
```

src/pretix/plugins/sendmail/views.py
```diff
@@ -262,11 +261,8 @@ def show_history_meta_data(cls, logentry, _cache_store):
                 i.pk: str(i) for i in logentry.event.checkin_lists.all()
             }
         if 'status' not in _cache_store:
-            status = dict(Order.STATUS_CHOICE)
-            status['overdue'] = _('pending with payment overdue')
-            status['valid_if_pending'] = _('payment pending but already confirmed')
-            status['na'] = _('payment pending (except unapproved or already confirmed)')
```
raphaelm, line 268: This was not exactly the same as the "na" from the order filter… One of them also included `valid_if_pending=False`. This was useful behaviour for "send to everyone who would lose their ticket if they don't pay now", which is now no longer possible. Any Ideas on how we could keep that?

src/pretix/plugins/sendmail/forms.py
```diff
@@ -391,6 +385,7 @@ def __init__(self, *args, **kwargs):
             choices.append(
                 ('n__pending_overdue', _('pending with payment overdue'))
             )
+        choices.append(('c', _('Canceled (fully)')))
```
raphaelm, line 388: What's the reason behind this change? Reordering or relabeling?

src/pretix/plugins/sendmail/views.py
```diff
@@ -408,9 +395,10 @@ def get_object_queryset(self, form):
         if form.cleaned_data.get('created_to'):
             opq = opq.filter(order__datetime__lt=form.cleaned_data.get('created_to'))
 
+        orders_without_positions = Order.objects.filter(~Exists(OrderPosition.objects.filter(canceled=False, order_id=OuterRef('pk'))))
         # pk__in turns out to be faster than Exists(subquery) in many cases since we often filter on a large subset
         # of orderpositions
-        return orders.filter(pk__in=opq.values_list('order_id'))
+        return orders.filter(Q(pk__in=opq.values_list('order_id')) | Q(pk__in=orders_without_positions))
```
raphaelm: Can you explain the reason behind this change? It's not obvious to me

src/pretix/base/models/orders.py
pajowu: Sounds like a good Idea. I originally decided against it when STATUS_FILTER still contained the "All orders" option, but since this was moved to the view, we can now do that.

src/pretix/plugins/sendmail/forms.py
pajowu: Relabeling. `'c'` does not include cancelled orders with a fee, as "Canceled" would imply

src/pretix/plugins/sendmail/views.py
pajowu: `opq` only looks at non-cancelled order positions. This means order that were cancelled with a fee (i.e. all positions are cancelled, a fee exists and the order status is `PAID`) would not appear in `opq.values_list('order_id')`

raphaelm: Understood. This will however also make weird effects when combined with recipients=attendees, where then the order matches, but none of the attendees

src/pretix/plugins/sendmail/views.py
pajowu: I changed `na` back to this variant and replaced the other `na` with `na_all`

src/pretix/plugins/sendmail/views.py
```diff
@@ -389,28 +381,25 @@ def get_object_queryset(self, form):
...
-        if form.cleaned_data.get('created_from'):
-            opq = opq.filter(order__datetime__gte=form.cleaned_data.get('created_from'))
-        if form.cleaned_data.get('created_to'):
-            opq = opq.filter(order__datetime__lt=form.cleaned_data.get('created_to'))
 
-        # pk__in turns out to be faster than Exists(subquery) in many cases since we often filter on a large subset
-        # of orderpositions
-        return orders.filter(pk__in=opq.values_list('order_id'))
+        if form.cleaned_data.get('recipients') in ['orders', 'both']:
+            orders_without_positions = Order.objects.filter(~Exists(OrderPosition.objects.filter(canceled=False, order_id=OuterRef('pk'))))
```
pajowu, line 395: move to 398

pajowu:
Too complex, no real win. Replaced with minimal solution in https://github.com/pretix/pretix/pull/6406

PR #6059, opened 2026-04-03
Add `voucher_filter_qs` signal for plugin-based voucher filtering

kiancross:
Allow plugins to filter the voucher queryset on the list page, tag overview page, and tag typeahead. Similar to how waiting list vouchers are excluded from these views, plugins can now hide vouchers they manage from the standard voucher UI.

We have plugins that automatically generate large numbers of vouchers and need the flexibility to hide these, just as waiting list vouchers are already hidden as a special case. This signal generalises that pattern.

Depends on #6058 being merged first. That commit is included here but will disappear from the diff once merged.

### DIFF
--- a/doc/development/api/general.rst
+++ b/doc/development/api/general.rst
@@ -76,7 +76,7 @@ Vouchers
 
 .. automodule:: pretix.control.signals
    :no-index:
-   :members: item_forms, voucher_form_class, voucher_form_html, voucher_form_validation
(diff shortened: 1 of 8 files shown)


raphaelm:
Unfortunately, I don't think this is a feature we want to have in core. Hiding waiting list vouchers in the voucher list is already very "magical" and feels confusing to almost as many users as it confused before we did it. I don't want this behavior to spread across plugins such that we can never again get rid of it. We discussed "voucher types" in the team before, but for now, I think that's also a bit more complex than it needs to be.

PR #6060, opened 2026-04-03
Add `voucher_list_top_buttons` signal for voucher list page

kiancross:
Add a new `EventPluginSignal` that allows plugins to inject additional buttons at the top of the voucher list page, next to the existing "Create a new voucher" and "Download list" buttons.

The signal is rendered outside the write-permission check so that plugins can use it for read-only actions too. Receivers are responsible for checking permissions via the `request` object.

An `empty` boolean is passed to indicate whether the voucher list has results, so plugins can match the larger button styling pretix uses in the empty state.

### DIFF
--- a/doc/development/api/general.rst
+++ b/doc/development/api/general.rst
@@ -76,7 +76,7 @@ Vouchers
 
 .. automodule:: pretix.control.signals
    :no-index:
-   :members: item_forms, voucher_form_class, voucher_form_html, voucher_form_validation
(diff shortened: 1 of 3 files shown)


raphaelm:
I think this is too usecase-specific for a plugin signal, or we'll start adding signals like this to every page in the system, which will be very hard to maintain. If that's a recurring need, we'll need a more generalized mechanism, I think.

PR #6071, opened 2026-04-09
Prevent nullbytes in input data globally

raphaelm:


src/pretix/base/middleware.py
```diff
@@ -346,6 +347,15 @@ def nested_dict_values(d):
...
 
+    def process_request(self, request):
+        # Nullbytes in GET/POST parameters are mostly harmless, as they will later fail on database insertion, but it
+        # keeps spamming our error logs whenever someone tries to run a vulnerability scanner.
+        if "\x00" in request.META['QUERY_STRING'] or "%00" in request.META['QUERY_STRING']:
+            raise BadRequest("Invalid characters in input.")
+        if request.method in ('POST', 'PUT', 'PATCH') and request.content_type == "application/x-www-form-urlencoded":
+            if any("\x00" in value for key, value_list in request.POST.lists() for value in value_list):
+                raise BadRequest("Invalid characters in input.")
+
```
luelista, line 358: SecurityMiddleware now does two unrelated things. I'd recommend either putting the actual logic into methods named according to what they do, or maybe even split into two middlewares.

e.g.
```suggestion
    def process_request(self, request):
        self.raise_if_null_bytes_in_request(request)

    def raise_if_null_bytes_in_request(self, request):
        # Nullbytes in GET/POST parameters are mostly harmless, as they will later fail on database insertion, but it
        # keeps spamming our error logs whenever someone tries to run a vulnerability scanner.
        if "\x00" in request.META['QUERY_STRING'] or "%00" in request.META['QUERY_STRING']:
            raise BadRequest("Invalid characters in input.")
        if request.method in ('POST', 'PUT', 'PATCH') and request.content_type == "application/x-www-form-urlencoded":
            if any("\x00" in value for key, value_list in request.POST.lists() for value in value_list):
                raise BadRequest("Invalid characters in input.")

```
and process_response => add_csp_header

luelista (approved):
👍 , one code style recommendation

PR #6079, opened 2026-04-13
Subevents: Allow to skip conflicting dates in bulk-creation (Z#23217384)

raphaelm:


src/pretix/control/templates/pretixcontrol/subevents/bulk.html
```diff
@@ -379,6 +379,7 @@ <h4 class="panel-title">
                         <i class="fa fa-calendar"></i> {% trans "Add many time slots" %}</button>
                 </p>
             </div>
+            {% bootstrap_field form.skip_if_overlap layout="control" horizontal_label_class='sr-only' horizontal_field_class='col-md-12' %}
```
wiffbi, line 383: IMHO a hr helps here for distinction to the times setting as this setting is more of a generator-flag and not only for times.
```suggestion
            <hr />
            {% bootstrap_field form.skip_if_overlap layout="control" horizontal_label_class='sr-only' horizontal_field_class='col-md-12' %}
```

wiffbi (changes requested):
~~Works as intended.~~ (UPDATE: I think, it should still be possible to create subevents following directly after another) I would add a hr as a separator in the form.

src/pretix/control/views/subevents.py
```diff
@@ -917,6 +917,27 @@ def set_progress(percent):
...
+            ).filter(
+                date_from__lte=date_max,
+                date_fromto__gte=date_min,
+            ).values('date_from', 'date_fromto'))
+            subevents = [
+                se for se in subevents if not any(
+                    (se.date_from <= other['date_from'] <= (se.date_to or se.date_from))
+                    or (se.date_from <= other['date_fromto'] <= (se.date_to or se.date_from))
+                    or (other['date_from'] <= se.date_from <= other['date_fromto'])
+                    or (other['date_from'] <= (se.date_to or se.date_from) <= other['date_fromto'])
```
wiffbi: With this, I cannot create subevents that start immediately after an existing event. Ie. when one subevent goes from 10 to 11 am and I want to add an event starting at 11, it gets skipped.

wiffbi (changes requested):

pajowu (approved):

PR #6080, opened 2026-04-13
Quotas: Add bulk-edit, bulk-delete and filter form

raphaelm:


src/pretix/control/forms/filter.py
```diff
@@ -1528,6 +1528,130 @@ def meta_properties(self):
...
+            ), get_current_timezone())
+            qs = qs.filter(subevent__date_from__gte=date_start)
+
+        if fdata.get('time_until'):
+            qs = qs.filter(subevent__date_from__time__lte=fdata.get('time_until'))
+        if fdata.get('time_from'):
+            qs = qs.filter(subevent__date_from__time__gte=fdata.get('time_from'))
+
+        if fdata.get('ordering'):
+            qs = qs.order_by(*get_deterministic_ordering(Quota, self.get_order_by()))
```
wiffbi, line 1651: When ordering by size, it feels a bit weird to me that unlimited size is first when order is ascending and vice-versa. Not a big deal, but could we – not necessarily in this PR – add a nulls_last=True somehow to ordering? On a first glance no idea how that would easily be possible though. Maybe with some special handling if fdata.get('ordering') is size/-size?

src/pretix/control/templates/pretixcontrol/items/quota_bulk_edit.html
```diff
@@ -0,0 +1,49 @@
...
+{% block content %}
+    <h1>
+        {% trans "Change multiple quotas" %}
+        <small>
+            {% blocktrans trimmed with number=quotas.count %}
+                {{ number }} selected
+            {% endblocktrans %}
+        </small>
+    </h1>
+    <form class="form-horizontal" action="" method="post">
```
wiffbi, line 13: Would it make sense to keep the filter/ordering of the quota-page  e.g. in the GET-params when starting the bulk-edit, so one can show the same filtered/orders quotas after the bulk-edit?

src/pretix/control/templates/pretixcontrol/items/quota_bulk_edit.html
```diff
@@ -0,0 +1,49 @@
...
+            </p>
+            {% bootstrap_field form.itemvars layout="bulkedit" %}
+        </fieldset>
+        <fieldset>
+            <legend>{% trans "Advanced options" %}</legend>
+            {% bootstrap_field form.close_when_sold_out layout="bulkedit" %}
+            {% bootstrap_field form.release_after_exit layout="bulkedit" %}
+            {% bootstrap_field form.ignore_for_event_availability layout="bulkedit" %}
+        </fieldset>
+        <div class="form-group submit-group">
```
wiffbi, line 43: How about a cancel-button? Of course that would lead to the filtered/ordered quota-page.

src/pretix/control/templates/pretixcontrol/items/quota_delete_bulk.html
```diff
@@ -0,0 +1,30 @@
...
+                {% for q in allowed %}
+                    <li>
+                        {{ q }} {% if q.subevent %}({{ q.subevent }}){% endif %}
+                        <input type="hidden" name="quota" value="{{ q.pk }}">
+                    </li>
+                {% endfor %}
+            </ul>
+        {% endif %}
+        <div class="form-group submit-group">
+            <a href="{% url "control:event.items.quotas" organizer=request.event.organizer.slug event=request.event.slug %}"
```
wiffbi, line 25: Would be great to add the filter-* params from POST to GET in this URL so filtering/ordering does not get lost.

src/pretix/control/forms/filter.py
```diff
@@ -1528,6 +1528,130 @@ def meta_properties(self):
...
+            self.fields['subevent'].widget.choices = self.fields['subevent'].choices
+        else:
+            del self.fields['subevent']
+            del self.fields['date_from']
+            del self.fields['date_until']
+            del self.fields['time_from']
+            del self.fields['time_until']
+            del self.fields['weekday']
+
+    def filter_qs(self, qs):
```
wiffbi, line 1616: I think subevent-filter is missing here. Selecting a subevent does not do anything for me.

src/pretix/control/forms/filter.py
```diff
@@ -1528,6 +1528,130 @@ def meta_properties(self):
...
+class QuotaFilterForm(FilterForm):
+    orders = {
+        '-date': ('-subevent__date_from', 'name', 'pk'),
+        'date': ('subevent__date_from', '-name', '-pk'),
+        'size': ('size', 'name', 'pk'),
+        '-size': ('-size', '-name', '-pk'),
+        'name': ('name', 'pk'),
+        '-name': ('-name', '-pk'),
+    }
+    subevent = forms.ModelChoiceField(
```
wiffbi, line 1540: Maybe also add a product-filter, i.e. any quota that contains a product/variant? Just an idea, not for this PR.

src/pretix/control/templates/pretixcontrol/items/quotas.html
wiffbi: Might be removed as well as `request.GET.subevent` doesn’t trigger any filtering anyways anymore.


```suggestion
                {% blocktrans trimmed %}
                    You haven't created any quotas yet.
                {% endblocktrans %}
```

src/pretix/control/templates/pretixcontrol/items/quota_delete_bulk.html
```diff
@@ -0,0 +1,30 @@
+{% extends "pretixcontrol/event/base.html" %}
+{% load i18n %}
+{% load bootstrap3 %}
+{% block title %}{% trans "Delete quotas" %}{% endblock %}
+{% block content %}
+    <h1>{% trans "Delete quotas" %}</h1>
+    <form action="" method="post" class="form-horizontal">
+        {% csrf_token %}
+        {% if allowed %}
+            <p>{% blocktrans %}Are you sure you want to delete the following quotas?{% endblocktrans %}</p>
```
wiffbi: Maybe add the number of quotas `allowed|length` to that message to reassure the user? Would need an additional singular/plural though. Alternatively use an `<ol>` instead of an `<ul>`? I know, other bulk_edits do not do this either, so just an idea.

src/pretix/control/forms/item.py
```diff
@@ -375,6 +376,49 @@ def save(self, *args, **kwargs):
...
+                    selected_items = set(list(self.event.items.filter(id__in=[
+                        i.split('-')[0] for i in self.cleaned_data['itemvars']
+                    ])))
+                    selected_variations = list(ItemVariation.objects.filter(item__event=self.event, id__in=[
+                        i.split('-')[1] for i in self.cleaned_data['itemvars'] if '-' in i
+                    ]))
+                    obj.items.set(selected_items)
+                    obj.variations.set(selected_variations)
+                else:
+                    setattr(obj, k, self.cleaned_data[k])
```
wiffbi: Not sure how this impacts the performance when calling `self.event.items.filter` and `ItemVariation.objects.filter` in the for-loop of objects. Maybe turn it around?
```suggestion
            if k == 'itemvars':
                selected_items = set(list(self.event.items.filter(id__in=[
                    i.split('-')[0] for i in self.cleaned_data['itemvars']
                ])))
                selected_variations = list(ItemVariation.objects.filter(item__event=self.event, id__in=[
                    i.split('-')[1] for i in self.cleaned_data['itemvars'] if '-' in i
                ]))
                for obj in objs:
                    obj.items.set(selected_items)
                    obj.variations.set(selected_variations)
            else:
                for obj in objs:
                    setattr(obj, k, self.cleaned_data[k])
```

wiffbi (commented):
Looks good and works as intended. Just a few comments/ideas. The hint regarding QuotaBulkEditForm.save might be wortha look, the others are really optional IMHO.

src/pretix/control/forms/filter.py
raphaelm: I cannot reproduce the behaviour, for me it is "correct", i.e. null comes first on descending. I'm currently testing on postgres, maybe the behaviour for "unspecified" varies between databases

src/pretix/control/templates/pretixcontrol/items/quota_bulk_edit.html
raphaelm: I tried to keep all the bulk-edits in the system to the same pattern and I'm quite sure there was a reason to have it in POST and having it in moth seems like it could cause consistency issues

src/pretix/control/templates/pretixcontrol/items/quota_bulk_edit.html
raphaelm: I worry about keeping the filter, because then the filter and the explicit selection of checkboxes might interfere

src/pretix/control/templates/pretixcontrol/items/quota_delete_bulk.html
raphaelm: I worry about keeping the filter, because then the filter and the explicit selection of checkboxes might interfere

src/pretix/control/forms/filter.py
raphaelm: Yeah, why not, but not this PR

src/pretix/control/forms/filter.py
wiffbi: Ah, might be sqlite

wiffbi (approved):

src/pretix/control/forms/item.py
```diff
@@ -383,6 +383,16 @@ def __init__(self, *args, **kwargs):
...
         self.fields.pop("subevent", None)  # Would add extra complexity and it's hard to imagine a use case for that
+        self.fields["name"].required = False
+        self.fields["itemvars"].required = False
+
+    def clean(self):
+        d = super().clean()
+        if self.prefix + "name" in self.data.getlist('_bulk') and not d.get("name"):
+            raise ValidationError({"name": _("This field is required.")})
+        if self.prefix + "itemvars" in self.data.getlist('_bulk') and not d.get("itemvars"):
+            raise ValidationError({"itemvars": _("This field is required.")})
```
wiffbi, line 394: Probably a very strange edge case, but this prevents bulk-setting quotas to not contain any products – not really something one really wants, but just to make sure we are aware of this.

PR #6087, opened 2026-04-14
Add locations to program times (Z#23221129)

ser8phin:
- [x] Add location
  - [x] to model/migrations
  - [x] to form/template
  - [x] to ical
  - [x] to pdf-placeholders
  - [x] to serializers
  - [x] to docs
- [x] Add tests
  - [x] for general functionality
  - [x] for event cloning


src/pretix/base/models/items.py
```diff
@@ -2306,10 +2306,17 @@ class ItemProgramTime(models.Model):
...
     :type end: datetime
+    :param location: venue
+    :type location: str
     """
     item = models.ForeignKey('Item', related_name='program_times', on_delete=models.CASCADE)
     start = models.DateTimeField(verbose_name=_("Start"))
     end = models.DateTimeField(verbose_name=_("End"))
+    location = I18nTextField(
+        null=True, blank=True,
+        max_length=200,
```
wiffbi, line 2317: I am not sure I18nTextField has max_length, I think only CharField has. So problably just leave it out – you might want to edit the migration as well.


```suggestion
```

src/pretix/presale/ical.py
```diff
@@ -153,7 +153,7 @@ def get_private_icals(event, positions):
                     # Actual ical organizer field is not useful since it will cause "your invitation was accepted" emails to the organizer
                     descr.append(_('Organizer: {organizer}').format(organizer=event.organizer.name))
                     description = '\n'.join(descr)
-                location = None
+                location = ", ".join(l.strip() for l in str(pt.location).splitlines() if l.strip())
```
wiffbi, line 156: As location can be None in den database, I guess this will create a string „None“. I might be wrong though …

```suggestion
                location = ", ".join(l.strip() for l in str(pt.location).splitlines() if l.strip()) if pt.location else None
```

ser8phin: No, as location will always return a LazyI18nString which will be localized by casting to string which will return an empty string in this case and thus not add a location field (same behaviour if the location is {} which happens if filled from a form with multiple languages in event settings)

We can however still add this condition to be completely sure nothing else might cause mishaps?

src/pretix/base/models/items.py
ser8phin: It does have max_length and we also use it exactly like this for event and subevent locations :wink:

wiffbi: Well, good catch!

src/pretix/presale/ical.py
wiffbi: Ah, right, then it is not necessary! :)

src/pretix/base/pdf.py
```diff
@@ -498,9 +499,9 @@
         ) if op.valid_until else ""
     }),
     ("program_times", {
-        "label": _("Program times: date and time"),
+        "label": _("Program times"),
         "editor_sample": _(
-            "2017-05-31 10:00 – 12:00\n2017-05-31 14:00 – 16:00\n2017-05-31 14:00 – 2017-06-01 14:00"),
+            "2017-05-31 10:00 – 12:00 Room 1\n2017-05-31 14:00 – 16:00 Room 2\n2017-05-31 14:00 – 2017-06-01 14:00 Building A"),
```
pajowu:
```suggestion
            "2017-05-31 10:00 – 12:00, Room 1\n2017-05-31 14:00 – 16:00, Room 2\n2017-05-31 14:00 – 2017-06-01 14:00, Building A"),
```

src/pretix/control/forms/item.py
```diff
@@ -1354,13 +1354,18 @@ class ItemProgramTimeForm(I18nModelForm):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.fields['end'].widget.attrs['data-date-after'] = '#id_{prefix}-start_0'.format(prefix=self.prefix)
+        self.fields['location'].widget.attrs['rows'] = '1'
```
pajowu: Is there a reason to use a `textarea` with `rows=1`?

I found this surprising UX-wise. A `textarea` with `rows=1` looks almost identical to in `input`-element, but behaves differently, for example pressing <kbd>Enter</kbd> creates a newline instead of submitting the form. Also (at least for me) it indicates that the location should be a single line. My suggestion would be to either use at least `rows=2` or to switch to an `I18nTextField`.

ser8phin: I don't remember and do not see a reason either. Changed to 3 rows.

src/pretix/base/pdf.py
```diff
@@ -748,13 +748,17 @@ def get_seat(op: OrderPosition):
...
+    for pt in op.item.program_times.all():
+        ptstr.append([
+            datetimerange(
+                pt.start.astimezone(ev.timezone),
+                pt.end.astimezone(ev.timezone),
+                as_html=False
+            ),
+            ', '.join(l.strip() for l in str(pt.location).splitlines() if l.strip())
+        ])
+    return '\n'.join(''.join(l) for l in ptstr)
```
raphaelm, line 761: This is missing a space in output, e.g.:

> 30.04.2026 00:00 – 14:00Test A

raphaelm (dismissed):

pajowu (approved):

PR #6096, opened 2026-04-17
Vouchers: Allow to bulk-update

raphaelm:
- [x] Tests for validatoin logic


src/pretix/control/forms/vouchers.py
```diff
@@ -245,26 +253,293 @@ def clean(self):
...
+                        _('If you want this voucher to block quota, you need to select a specific date.'))
+
+                if after_change["block_quota"] and not after_change["item"] and not after_change["quota"]:
+                    raise ValidationError(
+                        _('You need to select a specific product or quota if this voucher should reserve '
+                          'tickets.')
+                    )
+
+                if after_change["allow_ignore_quota"]:
+                    # todo: is this the most useful way to do this?
```
luelista, line 381: Not sure... This will allow a transition from:
- allow_ignore_quota = True
- block_quota = True

to 
- allow_ignore_quota = False
- block_quota = True

without adding a quota_diff.  However, the same is possible in the single edit form, so at least its consistent.

To change that, we could add a 'not allow_ignore_quota' to these conditions instead, and remove this `if` here.

```
                if was_valid and current["block_quota"] and not current["allow_ignore_quota"] and current["max_usages"] > current["redeemed"]:
```

```
                if will_be_valid and after_change["block_quota"] and not after_change["allow_ignore_quota"] and after_change["max_usages"] > current["redeemed"]:
```





(I'm not quite sure though why it's even allowed to have both allow_ignore_quota and block_quota selected at the same time?)

luelista (changes requested):
I made a separate PR for my recommended changes:

https://github.com/pretix/pretix/pull/6353/changes

PR #6106, opened 2026-04-21
Enforce uniqueness of EventMetaProperty names (Z#23231419)

luelista:
Todo:

- [x] Test
- [x] Add unique constraint
- [x] Migrations


src/pretix/base/migrations/0299_fixup_eventmetaproperties.py
```diff
@@ -0,0 +1,81 @@
...
+
+def fix_cross_organizer_eventmetavalues(apps, schema_editor):
+    EventMetaProperty = apps.get_model("pretixbase", "EventMetaProperty")
+    EventMetaValue = apps.get_model("pretixbase", "EventMetaValue")
+
+    cross_org_values = EventMetaValue.objects.filter(event__organizer__pk__ne=F('property__organizer__pk'))
+    for emv in cross_org_values:
+        logger.info(f"Fixup cross-organizer value {emv.event.organizer.slug}/{emv.event.slug}\n  before: {emv.property.name}({emv.property.id}@{emv.property.organizer.slug}) = {emv.value}")
+        try:
+            emv.property = emv.event.organizer.meta_properties.filter(name=emv.property.name).first()
```
raphaelm: I don't get this line, it calls first() (which returns None on not-found, which can't be saved) and then catches DoesNotExist (which isn't raised)?

src/pretix/base/migrations/0299_fixup_eventmetaproperties.py
```diff
@@ -0,0 +1,81 @@
...
+                continue
+        except EventMetaProperty.DoesNotExist:
+            meta_prop = emv.property
+            meta_prop.pk = None
+            meta_prop.organizer = emv.event.organizer
+            meta_prop.save(force_insert=True)
+            logger.info(f"  created new EventMetaProperty")
+            emv.property = meta_prop
+        logger.info(f"  after: {emv.property.name}({emv.property.id}@{emv.property.organizer.slug}) = {emv.value}")
+        emv.save()
```
raphaelm:
```suggestion
        emv.save(update_fields=["property"])
```

Micro-optimization but also increases (thread-)safety of the migration

src/pretix/base/migrations/0299_fixup_eventmetaproperties.py
```diff
@@ -0,0 +1,81 @@
...
+def make_eventmetaproperties_unique(apps, schema_editor):
+    EventMetaProperty = apps.get_model("pretixbase", "EventMetaProperty")
+    EventMetaValue = apps.get_model("pretixbase", "EventMetaValue")
+
+    duplicates = EventMetaProperty.objects.values('organizer', 'organizer__slug', 'name').annotate(count=Count('id')).filter(count__gt=1)
+    for dup in duplicates:
+        logger.info(f"Fixup duplicate property {dup['organizer__slug']} {dup['name']}")
+        props = list(EventMetaProperty.objects.filter(organizer=dup['organizer'], name=dup['name']))
+
+        # TODO: any better idea than picking the property to keep more or less randomly (first in database order)?
```
raphaelm: I don't think it matters? If they actually both have values, the order in which they are used probably depends on the database order of the value, not the property, or is inconsistent between code paths

src/pretix/base/migrations/0302_fixup_eventmetaproperties.py
```diff
@@ -0,0 +1,81 @@
...
+        try:
+            with transaction.atomic():
+                affected = EventMetaValue.objects.filter(
+                    event__organizer=dup['organizer'], property__in=invalid
+                ).update(
+                    property=target
+                )
+                logger.info(f"  Switching {affected} value(s) over to {target.name}({target.id}@{target.organizer.slug})")
+
+        except IntegrityError as e:
```
raphaelm, line 66: This is a bit "all or nothing", i.e. if some of the meta values could be switched, this wouldn't switch them. How many customers are affected by this duplication? If its a small number I think it's acceptable

src/pretix/base/migrations/0299_fixup_eventmetaproperties.py
luelista: Me neither, I think this should have been a .get() all along (I was pretty sure I tested that case, but that couldn't have worked...)

src/pretix/base/migrations/0302_fixup_eventmetaproperties.py
luelista: In our production database, no customers are affected. I think it's okay.

raphaelm (approved):

PR #6107, opened 2026-04-21
Navigation: suggest event or organizer by domain (Z#23231404)

raphaelm:


wiffbi (approved):
LGTM, maybe instead of `iexact` ust `istartswith`? But then again, one could also argue for icontains. So maybe leave it as iexact.

PR #6109, opened 2026-04-24
Backend: Support are-you-sure for dynamically added form parts (Z#23232506)

raphaelm:


wiffbi (approved):
I’d change the name to „… dynamically added form-parts“ as it actually does not re-scan for newly added forms but only newly added inputs AFAICT.

PR #6110, opened 2026-04-24
Events/subevents: Fix missing logging for meta changes (Z#23232443)

raphaelm:


src/pretix/control/views/event.py
```diff
@@ -234,13 +234,25 @@ def form_valid(self, form):
...
+                else form.cleaned_data.get(k))
+            for k in form.changed_data
+        }
+        meta_changed = {}
+        for f in self.meta_forms:
+            meta_changed.update({
+                k: (f.cleaned_data.get(k).name
+                    if isinstance(f.cleaned_data.get(k), File)
+                    else f.cleaned_data.get(k))
+                for k in f.changed_data
```
wiffbi: In my tests, f.changed_data for meta_forms is always just one field with key „value“. Each mata-value is represented by one form, which has a field named „value“. I think, what we want here, is `f.property.name` and not looping over f.changed_data – usually should only have one value in f.changed_data, but maybe do something like:

```
if f.changed_data:
    meta_changed[f.property.name] = f.cleaned_data.values()
```

src/pretix/control/views/subevents.py
```diff
@@ -540,20 +540,35 @@ def form_valid(self, form):
...
+            for k in form.changed_data
+        }
+        meta_changed = {}
+        for f in self.meta_forms:
+            meta_changed.update({
+                k: (f.cleaned_data.get(k).name
+                    if isinstance(f.cleaned_data.get(k), File)
+                    else f.cleaned_data.get(k))
+                for k in f.changed_data
+            })
```
wiffbi: See above regarding (Sub)EventMetaValueForm

src/pretix/control/views/subevents.py
```diff
@@ -628,6 +643,18 @@ def form_valid(self, form):
...
+        meta_changed = {}
+        for f in self.meta_forms:
+            meta_changed.update({
+                k: (f.cleaned_data.get(k).name
+                    if isinstance(f.cleaned_data.get(k), File)
+                    else f.cleaned_data.get(k))
+                for k in f.cleaned_data
+            })
+        if meta_changed:
+            data['meta_data'] = meta_changed
```
wiffbi: See above

wiffbi (changes requested):
Logging the changes does not work for me, only one change with the key „value“ is logged. This is due to each meta-value being its own form with just one field named „value“.

PR #6115, opened 2026-04-27
Check-in API: Add reusable media exchange

pc-coholic:


src/pretix/base/services/checkin.py
```diff
@@ -1101,6 +1112,33 @@ def perform_checkin(op: OrderPosition, clist: CheckinList, given_answers: dict,
...
+            if not reusable_media and not linked_media.exists() and media_exchange_supported:
+                raise RequiredMediaExchangeError(
+                    _('You need to exchange your ticket to complete this check-in.'),
+                    'exchange',
+                    required_media_policy,
+                    required_media_type
+                )
+            elif not reusable_media and linked_media.exists():
+                raise CheckInError(
+                    _('This ticket has already been exchanged - use the reusable media instead.'),
```
robbi5:
```suggestion
                    _('This ticket has already been exchanged - use the reusable medium instead.'),
```

src/pretix/base/settings.py
```diff
@@ -217,6 +217,19 @@ def unserialize(cls, s):
...
         )
     },
+    'reusable_media_usage_enforced': {
+        'default': 'False',
+        'type': bool,
+        'form_class': forms.BooleanField,
+        'serializer_class': serializers.BooleanField,
+        'form_kwargs': dict(
+            label=_("Enforce the usage of issued re-usable media for check-in"),
+            help_text=_("If enabled, a ticket barcode will not be accepted anymore, if a re-usable media has been "
```
robbi5:
```suggestion
            help_text=_("If enabled, a ticket barcode will not be accepted anymore, if a re-usable medium has been "
```

src/pretix/control/templates/pretixcontrol/checkin/simulator.html
```diff
@@ -78,6 +81,14 @@ <h3 class="nomargin-top">{% trans "Additional information required" %}</h3>
                             </li>
                         {% endfor %}
                     </ul>
+                {% elif result.status == "exchange" %}
+                    <h3 class="nomargin-top">{% trans "Media exchange required" %}</h3>
+                    <p>
+                        {% blocktrans trimmed with media_policy=media_policies|getitem:result.media_policy media_type=media_types|getitem:result.media_type %}
+                            This ticket needs to be exchanged into a <strong>{{ media_type }}</strong> re-usable media.
```
robbi5:
```suggestion
                            This ticket needs to be exchanged into a <strong>{{ media_type }}</strong> re-usable medium.
```

src/pretix/control/templates/pretixcontrol/checkin/simulator.html
```diff
@@ -78,6 +81,14 @@ <h3 class="nomargin-top">{% trans "Additional information required" %}</h3>
                             </li>
                         {% endfor %}
                     </ul>
+                {% elif result.status == "exchange" %}
+                    <h3 class="nomargin-top">{% trans "Media exchange required" %}</h3>
+                    <p>
+                        {% blocktrans trimmed with media_policy=media_policies|getitem:result.media_policy media_type=media_types|getitem:result.media_type %}
+                            This ticket needs to be exchanged into a <strong>{{ media_type }}</strong> re-usable medium.
```
raphaelm: We seem to be inconsistent in calling them "re-usable" and "reusable", is that possible?

src/pretix/base/settings.py
```diff
@@ -217,6 +217,19 @@ def unserialize(cls, s):
...
         )
     },
+    'reusable_media_usage_enforced': {
+        'default': 'False',
+        'type': bool,
+        'form_class': forms.BooleanField,
+        'serializer_class': serializers.BooleanField,
+        'form_kwargs': dict(
+            label=_("Enforce the usage of issued re-usable media for check-in"),
+            help_text=_("If enabled, a ticket barcode will not be accepted anymore, if a re-usable medium has been "
```
raphaelm:
```suggestion
            help_text=_("If enabled, a ticket barcode will not be accepted anymore once a re-usable medium has been "
```

src/pretix/base/settings.py
```diff
@@ -217,6 +217,19 @@ def unserialize(cls, s):
...
+                        "created and linked to a ticket. Keeping this option turned off will treat the re-usable "
+                        "medium and ticket as equals."),
```
raphaelm, line 229:
```suggestion
                        "medium and ticket as equally valid during check-in."),
```

src/pretix/api/views/checkin.py
```diff
@@ -764,6 +769,16 @@ def _make_context(context, event):
                 ],
                 'list': MiniCheckinListSerializer(list_by_event[op.order.event_id]).data,
             }, status=400)
+        except RequiredMediaExchangeError as e:
+            return Response({
+                'status': 'exchange',
```
raphaelm, line 865: This is missing a docs change

src/pretix/api/views/checkin.py
```diff
@@ -764,6 +769,16 @@ def _make_context(context, event):
                 ],
                 'list': MiniCheckinListSerializer(list_by_event[op.order.event_id]).data,
             }, status=400)
+        except RequiredMediaExchangeError as e:
```
raphaelm, line 863: Just for completeness: This PR needs a full set of tests for every scenario of check-in logic

src/pretix/base/services/checkin.py
```diff
@@ -951,6 +960,8 @@ def perform_checkin(op: OrderPosition, clist: CheckinList, given_answers: dict,
         questions are not filled out.
     :param ignore_unpaid: When set to True, this will succeed even when the order is unpaid.
     :param questions_supported: When set to False, questions are ignored
+    :param media_exchange_supported: When set to False, media exchanges are ignored and access with un-exchanged media
```
raphaelm: We could honestly discuss this. "Check-in in this scenario is not possible at a device that doesn't support media exchange" would also be a valid option, wouldn't it?

src/pretix/base/services/checkin.py
```diff
@@ -1101,6 +1112,23 @@ def perform_checkin(op: OrderPosition, clist: CheckinList, given_answers: dict,
                 require_answers
             )
 
+        required_media_policy = op.item.media_policy
+        required_media_type = op.item.media_type
+        linked_media = op.linked_media
+        if not reusable_media and required_media_policy and required_media_type and not force:
+            if not linked_media.exists() and media_exchange_supported:
+                raise RequiredMediaExchangeError(
+                    _('You need to exchange your ticket to complete this check-in.'),
```
raphaelm:
```suggestion
                    _('Ticket needs to be exchanged to a suitable medium.'),
```

Unlike questions, where "You" could address both check-in staff and a kiosk user, "your ticket" is weird when addressing check-in staff.

src/pretix/base/services/checkin.py
```diff
@@ -1101,6 +1112,23 @@ def perform_checkin(op: OrderPosition, clist: CheckinList, given_answers: dict,
...
+                    'exchange',
+                    required_media_policy,
+                    required_media_type
+                )
+            elif op.organizer.settings.reusable_media_usage_enforced:
+                raise CheckInError(
+                    _('This ticket has already been exchanged - use the reusable medium instead.'),
```
raphaelm:
```suggestion
                    _('This ticket has already been exchanged for a reusable medium that now needs to be used instead.'),
```

src/pretix/base/services/checkin.py
```diff
@@ -1101,6 +1112,23 @@ def perform_checkin(op: OrderPosition, clist: CheckinList, given_answers: dict,
...
+        linked_media = op.linked_media
+        if not reusable_media and required_media_policy and required_media_type and not force:
+            if not linked_media.exists() and media_exchange_supported:
+                raise RequiredMediaExchangeError(
+                    _('You need to exchange your ticket to complete this check-in.'),
+                    'exchange',
+                    required_media_policy,
+                    required_media_type
+                )
+            elif op.organizer.settings.reusable_media_usage_enforced:
```
raphaelm, line 1126: I think this branch will also trigger if a linked medium does not exist, but exchange is also not supported, which is wrong?

src/pretix/control/templates/pretixcontrol/checkin/simulator.html
pc-coholic: Reusable without the hyphen seems to be the correct way of spelling it; with the hyphen it just makes it more obvious to non-native speakers...

I'll adjust to non-hyphenated for now.

pc-coholic:
As of right now, we have a slight discrepancy in the URLs for reusable media: `/api/v1/organizers/(organizer)/reusablemedia/` vs. `/control/organizer/bigevents/reusable_media`.

I feel like `reusable_media` would be more correct (as we are usually also using snake case). While changing the control-URL would be easy, changing the API would require a redirect.

Should we change the API-URL and add a (temporary) redirect? Or just leave it as-is?

src/pretix/base/services/checkin.py
pc-coholic: Very much a fan of this - just didn't think you'd go for it :-)

Adding this now

src/pretix/base/services/checkin.py
pc-coholic: True. Should be fixed with the new if/else branches.

src/pretix/api/serializers/checkin.py
```diff
@@ -88,6 +88,13 @@ class CheckinRPCRedeemInputSerializer(serializers.Serializer):
     nonce = serializers.CharField(required=False, allow_null=True)
     datetime = serializers.DateTimeField(required=False, allow_null=True)
     answers = serializers.JSONField(required=False, allow_null=True)
+    media_type = serializers.ChoiceField(required=False, choices=MEDIA_TYPES)
+    media_identifier = serializers.CharField(required=False)
+    media_policy = serializers.ChoiceField(required=False, choices=Item.MEDIA_POLICIES)
```
robbi5: The `media_policy` shouldn't be provided by the checkinrpc-client, this is something we should pull ourselves from the item of the ticket, or?

src/pretix/api/serializers/checkin.py
```diff
@@ -88,6 +88,13 @@ class CheckinRPCRedeemInputSerializer(serializers.Serializer):
     nonce = serializers.CharField(required=False, allow_null=True)
     datetime = serializers.DateTimeField(required=False, allow_null=True)
     answers = serializers.JSONField(required=False, allow_null=True)
+    media_type = serializers.ChoiceField(required=False, choices=MEDIA_TYPES)
+    media_identifier = serializers.CharField(required=False)
```
robbi5: Shouldn't we prefix the `media_` inputs with `exchange_` or something?
Unsuspecting api users could get the idea, that if they are normally scanning a medium (without exchange) that these fields should be used instead of `secret` and `source_type` 🤔

src/pretix/api/serializers/checkin.py
```diff
@@ -88,6 +88,12 @@ class CheckinRPCRedeemInputSerializer(serializers.Serializer):
...
```
raphaelm: Nitpick: medium is singular, media is plural. Since you act on only one, it should be medium, not media

src/pretix/api/views/checkin.py
```diff
@@ -801,25 +805,58 @@ def _make_context(context, event):
...
-                user=user,
-                auth=auth,
-                type=checkin_type,
-                raw_barcode=raw_barcode_for_checkin,
-                raw_source_type=source_type,
-                from_revoked_secret=from_revoked_secret,
-                simulate=simulate,
-                gate=gate,
-            )
+            if all(k is not None for k in [media_type, media_identifier, media_action]) and not media:
```
raphaelm: If one of them is None, the others will be ignored. This might be unexpected by API users, should this trigger a validation error?

src/pretix/base/services/media.py
```diff
@@ -70,3 +70,37 @@ def get_keysets_for_organizer(organizer):
...
+
+    if media_policy in [Item.MEDIA_POLICY_REUSE, Item.MEDIA_POLICY_REUSE_OR_NEW]:
+        try:
+            medium = ReusableMedium.objects.get(
+                type=media_type,
+                identifier=media_identifier,
+                organizer=organizer,
+            )
+        except ReusableMedium.DoesNotExist:
+            pass
```
raphaelm: Shouldn't this be an error if the policy is "reuse"?

src/pretix/base/services/media.py
```diff
@@ -70,3 +70,37 @@ def get_keysets_for_organizer(organizer):
...
+                type=media_type,
+                identifier=media_identifier,
+                organizer=organizer,
+            )
+        except IntegrityError:
+            raise ReusableMedium.DuplicateEntry()
+
+    if medium:
+        if media_action == 'append':
+            medium.linked_orderpositions.add(*[op])
```
raphaelm:
```suggestion
            medium.linked_orderpositions.add(op)
```

src/pretix/base/services/media.py
```diff
@@ -70,3 +73,107 @@ def get_keysets_for_organizer(organizer):
...
+            )
+
+    else:
+        raise CheckInError(
+            _('Product does not support medium exchange.'),
+            Checkin.REASON_PRODUCT,
+            reason=_('Product does not support medium exchange.'),
+        )
+
+    if link_action == 'append':
```
robbi5, line 200: Should we check that the medium is `active` and not already `expired`?

raphaelm:
> I feel like reusable_media would be more correct (as we are usually also using snake case). While changing the control-URL would be easy, changing the API would require a redirect.

In the API case, we use spelling like reusablemedia more than snake case (orderpositions, membershiptypes, revokedsecrets, …). In fact item_meta_properties and program_times seem to be the only cases of snake-case in the API

raphaelm (approved):

PR #6126, opened 2026-04-28
Enable dragging answer options

sweenu:
The arrows become very painful to use as the number of answers grow. I made an example with a dozen answers, but I have a real case with more than 30 answers.
It felt like it would take less time to make that PR than to actually bring that bottom answer to the top.

A before and after comparison video: 
[answer_options.webm](https://github.com/user-attachments/assets/b0130694-cb24-40a3-a0b1-168eaf581328)

Generated with AI (Claude)

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/items/question_edit.html
+++ b/src/pretix/control/templates/pretixcontrol/items/question_edit.html
@@ -76,13 +76,11 @@
                                             {% bootstrap_form_errors form %}
                                             {% bootstrap_field form.answer layout='inline' form_group_class="" %}
                                         </div>
-                                        <div class="col-xs-2 text-right flip">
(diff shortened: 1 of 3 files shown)


raphaelm:
I think it's probably a bad idea to have a Sortable implementation and the django-formset-js implementation working on the same list of formsets and potentially have them interfere with each other, so we should rather add something like opt-in dragging to django-formset-js (which we [forked anyways](https://pypi.org/project/django-formset-js-improved)). @wiffbi @luelista You also worked on sorting things, do you agree?

raphaelm:
Closed, no time to cleanup AI PR with incorrect approach, sorry

sweenu:
No worries. I'll create a feature request discussion to keep track, as I believe it is a real shortcoming.

PR #6131, opened 2026-04-30
Allow to update -> Allow updating

sweenu:


vlt:
Do all the `.po` files need to be updated manually?

sweenu:
Maybe not? I just used sed on the whole repo. If not, where would they be generated? In the CI?

PR #6145, opened 2026-05-04
Fix Dockerfile syntax for chmod command

caveman99:


raphaelm:
We already merged #6133, but this is also correct, let's have both and clean up another time

caveman99:
thanks, this was actually a correction/fix on top of #6133

PR #6148, opened 2026-05-04
Order details: Link to subevent details (Z#23227664)

raphaelm:


src/pretix/control/templates/pretixcontrol/order/index.html
```diff
@@ -471,7 +471,9 @@ <h3 class="panel-title">
                                 {% endif %}
                                 {% if line.subevent %}
                                     <br/>
-                                    <span class="fa fa-calendar fa-fw"></span> {{ line.subevent.name }} &middot; {{ line.subevent.get_date_range_display_with_times }}
+                                    <span class="fa fa-calendar fa-fw"></span>
+                                    <a href="{% url "control:event.subevent" organizer=request.event.organizer.slug event=request.event.slug subevent=line.subevent_id %}">{{ line.subevent.name }}</a>
```
wiffbi, line 475: Hm, feels a bit „dangerous“ to link to the subevent edit-page. Maybe some inexperienced users think, they can change the date the user wants to attend here ie mistake that link with the „change products“ button. What about linking to the dates-listview with the GET-param `?subevent=line.subevent_id`. This filters that list to the subevent. Not much helpful regarding info for that date, but helps orientation especially for inexperienced pretix-users?

raphaelm: see #6151 instead :)

PR #6160, opened 2026-05-07
Allow RelativeDate in relation to a moment of order

bockstaller:
Adds support for expressing dates in relationship to an order.

- Opt-in, requires setting `relative_to_order` on the field
- Dates in relation to an order can only be expressed as 'after the order has been placed'


src/pretix/base/templates/pretixbase/forms/widgets/reldate.html
```diff
@@ -1,4 +1,6 @@
 {% load i18n %}
+{% load static %}
+<script src="{% static 'pretixbase/js/reldate.js' %}" defer></script>
```
raphaelm: I don't think we should include scripts in widget templates. This will possibly include the script many times on the same page. I'd just add reldate.js to the stack of JS we load everywhere across the backend

src/pretix/base/templates/pretixbase/forms/widgets/reldatetime.html
```diff
@@ -1,4 +1,6 @@
 {% load i18n %}
+{% load static %}
+<script src="{% static 'pretixbase/js/reldate.js' %}" defer></script>
```
raphaelm: Same here

src/pretix/base/reldate.py
```diff
@@ -34,14 +34,20 @@
...
     ('date_from', _('Event start')),
     ('date_to', _('Event end')),
     ('date_admission', _('Event admission')),
     ('presale_start', _('Presale start')),
     ('presale_end', _('Presale end')),
 )
 
+# extend NO_BEFORE_VALUES in reldate.js if changed
+ORDER_CHOICES = (
+    ('datetime', _('Moment of order')),
```
raphaelm:
```suggestion
    ('datetime', _('Order creation')),
```

for consistent labeling and maybe even better

```suggestion
    ('order_datetime', _('Order creation')),
```

In case we ever want another set of CHOICES that also has something called "datetime"

src/pretix/base/reldate.py
```diff
@@ -34,14 +34,20 @@
...
+)
+ORDER_CHOICES_KEYS = [choice[0] for choice in ORDER_CHOICES]
```
raphaelm:
```suggestion
ORDER_CHOICES_KEYS = {choice[0] for choice in ORDER_CHOICES}
```

src/pretix/base/reldate.py
```diff
@@ -51,17 +57,43 @@ class RelativeDateWrapper:
...
+            event = reference
+            base_date = getattr(reference, self.data.base_date_name) or event.date_from
+        else:
+            raise ValueError("Only event, subevent or order objects are supported")
+
+        tz = ZoneInfo(event.settings.timezone)
+
+        return base_date, tz
 
+    def date(self, reference) -> datetime.date:
```
raphaelm: Should we annotate reference with the supported types?

src/pretix/base/reldate.py
```diff
@@ -172,7 +186,9 @@ def from_string(cls, input: str):
                         minutes=None,
                         is_after=len(parts) > 4 and parts[4] == "after",
                     )
-            if data.base_date_name not in [k[0] for k in BASE_CHOICES]:
+            if data.base_date_name in [k[0] for k in ORDER_CHOICES] and parts[4] != "after":
+                raise ValueError('ORDER_CHOICE: {} cannot be combined with "before"'.format(data.base_date_name))
```
raphaelm: This makes sense for order.datetime, but if we'd ever want to support Order.expired, we'll need to change this. Fine with me for now, though

src/pretix/base/reldate.py
```diff
@@ -172,7 +186,9 @@ def from_string(cls, input: str):
                         minutes=None,
                         is_after=len(parts) > 4 and parts[4] == "after",
                     )
-            if data.base_date_name not in [k[0] for k in BASE_CHOICES]:
+            if data.base_date_name in [k[0] for k in ORDER_CHOICES] and parts[4] != "after":
```
raphaelm:
```suggestion
            if data.base_date_name in ORDER_CHOICES_KEYS and parts[4] != "after":
```

src/pretix/base/reldate.py
```diff
@@ -51,17 +57,43 @@ class RelativeDateWrapper:
...
+
+        :param reference:
+        :return:
+        """
+        from .models import Event, Order, SubEvent
+
+        if self.data.base_date_name in ORDER_CHOICES_KEYS:
+            if not isinstance(reference, Order):
+                raise ValueError('A order-based relative datetime choice must be used with an order object')
+            event = reference.event
```
raphaelm: This is not good enough – and will unfortunately make the PR more complex.

If I set a date relative to `date_from` in an event series, I'd expect this to be in reference to the SubEvent, not the Event. Which, for an order, unfortunately could be multiple ones, and it depends on the context if we prefer a min or max of them :( See e.g. Order.user_cancel_deadline or Order.ticket_download_date for current examples

src/pretix/base/reldate.py
```diff
@@ -51,17 +57,43 @@ class RelativeDateWrapper:
...
+        elif isinstance(reference, SubEvent):
+            event = reference.event
+            base_date = (getattr(reference, self.data.base_date_name) or
+                         getattr(reference.event, self.data.base_date_name) or
+                         reference.date_from)
+        elif isinstance(reference, Event):
+            event = reference
+            base_date = getattr(reference, self.data.base_date_name) or event.date_from
+        else:
+            raise ValueError("Only event, subevent or order objects are supported")
```
raphaelm:
```suggestion
            raise TypeError("Only event, subevent or order objects are supported")
```

src/pretix/base/reldate.py
```diff
@@ -20,160 +20,262 @@
...
-    ('date_to', _('Event end')),
-    ('date_admission', _('Event admission')),
-    ('presale_start', _('Presale start')),
-    ('presale_end', _('Presale end')),
-)
+if TYPE_CHECKING:
+    from .models import Event, Order, SubEvent
+
+
+class BaseChoice:
```
raphaelm, line 51: Could this be a `@dataclass(frozen=True)` for even less code but most importantly no accidental mutability? key could be a `@property`

src/pretix/base/reldate.py
```diff
@@ -20,160 +20,262 @@
...
+EVENT_BASE_CHOICES = [
+    x for x in BASE_CHOICES if x.base == 'event'
+]
+
+ORDER_BASE_CHOICES = [
+    x for x in BASE_CHOICES if x.base == 'order'
+]
+
+
+class RelativeDate:
```
raphaelm, line 97: Same here, would a dataclass be a good idea to store the properties immutably? The namedtuple was also immutable

src/pretix/base/reldate.py
```diff
@@ -20,160 +20,262 @@
...
+
+    def __init__(self, days: int = 0, minutes: int = None, time: datetime.time = None, is_after: bool = False,
+                 base_date_name: str = 'event__date_from') -> None:
+        choice = BaseChoice.find(BASE_CHOICES, base_date_name)
+        self.base = choice.base
+        self.attribute = choice.attribute
+
+        if is_after and not choice.supports_after:
+            raise ValueError(
+                "The selected base date and attribute combination does not support relative dates placed after the base date")
```
raphaelm: This is an unusual place for the closing ), we usually do
```suggestion
                "The selected base date and attribute combination does not support relative dates placed after the base date"
            )
```

src/pretix/base/reldate.py
```diff
@@ -309,13 +410,32 @@ def __init__(self, *args, **kwargs):
...
+        self.relative_to_order = kwargs.pop('relative_to_order', False)
+
+        all_choices = EVENT_BASE_CHOICES
+        if self.relative_to_order:
+            all_choices.extend(ORDER_BASE_CHOICES)
+
         if kwargs.get('limit_choices'):
             limit = kwargs.pop('limit_choices')
-            choices = [(k, v) for k, v in BASE_CHOICES if k in limit]
+            if any(["__" not in l for l in limit]):
```
raphaelm, line 474:
```suggestion
            if any("__" not in l for l in limit):
```

src/pretix/base/reldate.py
```diff
@@ -20,160 +20,262 @@
...
+
+
+BASE_CHOICES: List[BaseChoice] = [
+    BaseChoice('event', 'date_from', _('Event start'), True, True),
+    BaseChoice('event', 'date_to', _('Event end'), True, True),
+    BaseChoice('event', 'date_admission', _('Event admission'), True, True),
+    BaseChoice('event', 'presale_start', _('Presale start'), True, True),
+    BaseChoice('event', 'presale_end', _('Presale end'), True, True),
+    BaseChoice('order', 'datetime', _('Order creation'), False, True),
+    BaseChoice('order', 'expires', _('Order expiry'), True, False),
```
raphaelm: I mean, why not

```suggestion
    BaseChoice('order', 'expires', _('Order expiry'), True, True),
```

?

src/pretix/base/reldate.py
```diff
@@ -404,6 +525,18 @@ def clean(self, value):
             raise ValidationError(self.error_messages['incomplete'])
         elif data.status == 'relative_minutes' and (data.rel_mins_number is None or not data.rel_mins_relationto):
             raise ValidationError(self.error_messages['incomplete'])
+        elif data.status == 'relative':
+            choice = BaseChoice.find(BASE_CHOICES, data.rel_days_relationto)
+            if data.rel_days_relation == "before" and not choice.supports_before:
+                raise ValidationError(_("A relative date cannot be expressed as 'before' for '{}'".format(choice.text)))
```
raphaelm: This is really nitpicky, but can you swap ' and " for these, so " is used in user-visible text for consistency?

src/pretix/static/pretixbase/js/reldate.js
```diff
@@ -0,0 +1,44 @@
+if (!window.__reldateInitialized) {
+    window.__reldateInitialized = true;
+    document.addEventListener('DOMContentLoaded', () => {
+         const NO_BEFORE_VALUES = ['datetime'];
```
raphaelm: - I think this needs to be order__datetime now?
- Do we also need NO_AFTER_VALUES now that the backend supports that?
- If we maintain this separately, both this list and the list in the Python code need a comment that reminds to update the other list on changes. Otherwise we could also try overriding the form widget template in a way that passes the information in an `<option data-no-before>` attribute or the like

src/tests/base/test_reldate.py
```diff
@@ -140,10 +141,32 @@ def test_unserialize():
     assert rdw.data == d
 
     rdw = RelativeDateWrapper.from_string('RELDATE/1/-/date_from/')
+    x = RelativeDate(days=1, time=None, base_date_name='date_from', minutes=None)
+    pprint(rdw.data)
+    pprint(x)
```
raphaelm:
```suggestion
```

src/tests/base/test_reldate.py
```diff
@@ -140,10 +141,32 @@ def test_unserialize():
     assert rdw.data == d
 
     rdw = RelativeDateWrapper.from_string('RELDATE/1/-/date_from/')
+    x = RelativeDate(days=1, time=None, base_date_name='date_from', minutes=None)
```
raphaelm: Can you add a comment noting that this is testing the old format on purpose to ensure backwards compat? Otherwise someone might want to clean it up

raphaelm (approved):

src/pretix/base/reldate.py
bockstaller: That is left over from testing

bockstaller:
@raphaelm I've reworked some things.
Subevent relationships are now supported and I've implemented the `<option data-...` suggestion as well. Would you mind taking another look?

src/pretix/base/reldate.py
```diff
@@ -19,161 +19,305 @@
...
+    BaseChoice('event', 'presale_start', "", _('Presale start'), True, True),
+    BaseChoice('event', 'presale_end', "", _('Presale end'), True, True),
+    BaseChoice('order', 'datetime', "", _('Order creation'), False, True),
+    BaseChoice('order', 'expires', "", _('Order expiry'), True, True),
+    BaseChoice('order.subevents', 'date_from', "first", _('Subevent start (first subevent in order)'), True, True),
+    BaseChoice('order.subevents', 'date_from', "last", _('Subevent start (last subevent in order)'), True, True),
+    BaseChoice('order.subevents', 'date_to', "first", _('Subevent end (first subevent in order)'), True, True),
+    BaseChoice('order.subevents', 'date_to', "last", _('Subevent end (last subevent in order)'), True, True),
+    BaseChoice('order.subevents', 'date_admission', "first", _('Subevent admission (first subevent in order)'), True, True),
+    BaseChoice('order.subevents', 'date_admission', "last", _('Subevent admission (last subevent in order)'), True, True),
```
raphaelm: 1. We **never** use the term subevent in user-facing text, always date *with* a translation context key to make sure it's "Termin" and not "Datum" in German
2. I don't think this is a good idea to have order.subevents here as an explicit option because it breaks config when copying from series to non-series or vice versa.

Let's quickly talk this through after your vacation?

src/pretix/base/reldate.py
```diff
@@ -19,161 +19,305 @@
...
+        choice = self._choice
+
+        if choice.base == "order" and isinstance(base, Order):
+            event = base.event
+            base_date = getattr(base, choice.attribute)
+        elif choice.base == "order.subevents" and isinstance(base, Order):
+            if not base.event.has_subevents:
+                raise ValueError("The order is for an event without subevents")
+            if choice.modifier == "first":
+                op = base.all_positions.order_by(f"subevent__{choice.attribute}").first()
```
raphaelm: This does not account for the fact that subevent__date_to and subevent__date_admission are nullable and need a fallback to subevent__date_from

raphaelm (changes requested):
I think 23e37ada59c8eff7f03c913b8f3673d44e981192 is an overshoot in refactoring. I think it was correct to have only "date_from" in the choice and have the *system* figure out if Event or SubEvent is meant, not the user. Happy to quickly talk this through once you're back.

src/pretix/base/templates/pretixbase/forms/widgets/reldate.html
```diff
@@ -1,4 +1,5 @@
 {% load i18n %}
+{% load static %}
```
raphaelm:
```suggestion
```

src/pretix/base/templates/pretixbase/forms/widgets/reldatetime.html
```diff
@@ -1,4 +1,5 @@
 {% load i18n %}
+{% load static %}
```
raphaelm:
```suggestion
```

src/pretix/base/reldate.py
```diff
@@ -19,161 +19,316 @@
...
+                    second=self.time.second
+                )
+            new_date = new_date.astimezone(tz)
+            return new_date
+
+    def to_string(self) -> str:
+        """
+
+        :return:
+        """
```
raphaelm: Can you remove the docstrings that do not add value?

src/pretix/base/reldate.py
```diff
@@ -220,21 +383,29 @@ def placeholder_time_format():
             tf = get_format('TIME_INPUT_FORMATS')[0]
             return datetime.time(8, 30, 0).strftime(tf)
 
-        widgets = reldatetimeparts(
+        widgets = cast(dict[str, Widget | type[Widget]], cast(object, reldatetimeparts(
```
raphaelm, line 369: Personal opinion: If you need a full line to express the typing, you're probably at the level of dynamity where the cost of typing outweighs the returns – especially if it's a local variable and not some external API

src/pretix/base/reldate.py
```diff
@@ -309,17 +480,39 @@ def __init__(self, *args, **kwargs):
...
+
         if kwargs.get('limit_choices'):
             limit = kwargs.pop('limit_choices')
-            choices = [(k, v) for k, v in BASE_CHOICES if k in limit]
-        else:
-            choices = BASE_CHOICES
+            if any(["__" not in l for l in limit]):
+                _warn_skips = (str(os.path.dirname(__file__)),)
+                warnings.warn(
+                    message="Please prefix limit_choices with the base the attributes refer to, for example event__date_from",
```
raphaelm, line 478: Should this be of type DeprecationWarning?

src/pretix/static/pretixbase/js/reldate.js
```diff
@@ -0,0 +1,45 @@
+if (!window.__reldateInitialized) {
+	window.__reldateInitialized = true
+	document.addEventListener('DOMContentLoaded', () => {
```
raphaelm: This will break in multiple ways if a form field is ever dynamically added to the DOM after page load, such as in a formset. I'm not sure if we currently have any relativedatetimefields in formsets yet, but it's not really an unlikely thing to happen.

Even though it's nice if we use less jQuery in the long run, you could just listen to `$.on("pretix:bind-forms")`, this should just work.

raphaelm (commented):
I have read the code and have only very very tiny comments :) I'd like to play around with it briefly which I'll try to do in the next days.

raphaelm (approved):

PR #6164, opened 2026-05-11
use datetime.fromisoformat instead of dateutil.parser (Z#23234093)

bockstaller:
`dateutil.parser.parse()` doesn't accept the complete isoformat grammar, for example seconds in the timezone offset are not supported.

`datetime.fromisoformat()` is able to do so [starting with Python 3.11](https://docs.python.org/3/library/datetime.html) 

Dates like this `0569-01-01T00:00:00-04:56:02` can now be displayed

### DIFF
--- a/src/pretix/control/logdisplay.py
+++ b/src/pretix/control/logdisplay.py
@@ -34,6 +34,7 @@
 # License for the specific language governing permissions and limitations under the License.
 
 from collections import defaultdict
+from datetime import datetime
 from decimal import Decimal
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


pajowu (commented):
`dateutil.parser.parse` supports many more formats than `datetime.fromisoformat`. Do all of these fields only ever contain iso-date-strings?

raphaelm (approved):

raphaelm:
> `dateutil.parser.parse` supports many more formats than `datetime.fromisoformat`. Do all of these fields only ever contain iso-date-strings?

Should be safe, this is only internal

PR #6189, opened 2026-05-18
banktransfer import: Fix prefix confusion if shorter event name contains dash (Z#23234167)

pajowu:
There are still possible collisions here as event slugs are not guaranteed to be prefix-free (or even just collision-free after we apply the fuzzyness regarding dashes to the names). I added an `xfail` testcase that shows at least one possible collision that still exists (`test_event_name_prefix_multiple_dashes`)

### DIFF
--- a/src/tests/plugins/banktransfer/test_import.py
+++ b/src/tests/plugins/banktransfer/test_import.py
@@ -863,6 +863,9 @@ def test_event_name_prefix_contains_dash(env, orga_job):
 @pytest.mark.xfail
 @pytest.mark.django_db
 def test_event_name_prefix_multiple_dashes(env, orga_job):
+    # We decided to ignore the case of multiple events where the event slugs only differ in dashes
+    # for now. If we change that, switch this test case from xfail to regular test.
+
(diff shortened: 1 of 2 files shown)


luelista (approved):

src/tests/plugins/banktransfer/test_import.py
```diff
@@ -834,3 +834,53 @@ def test_ambigious_date_with_region(env, job):
...
+    with scopes_disabled():
+        job = BankImportJob.objects.last()
+        t = job.transactions.last()
+        assert t.state == BankTransaction.STATE_VALID
+
+
+@pytest.mark.xfail
+@pytest.mark.django_db
+def test_event_name_prefix_multiple_dashes(env, orga_job):
+    event, user, o1, o2 = env
```
luelista, line 869:
```suggestion
    # We decided to ignore the case of multiple events where the event slugs only differ in dashes
    # for now. If we change that, switch this test case from xfail to regular test.

    event, user, o1, o2 = env
```

PR #6211, opened 2026-05-22
Search experiments

raphaelm:
- [ ] Test accent handling
- [ ] Test payment provider integration
- [ ] Check if it actually helps
  - [ ] Estimate index size
  - [ ] Estimate cost of writing the index
- [ ] Implement into search form
- [ ] Implement into APIs
- [ ] Implement to shredder process
- [ ] Tokenization
  - [ ] Check tokenization of invoice numbers
  - [ ] Tokenize Words
- [ ] Check for other .update() calls on the affected models
- [ ] Plugin integration (e.g. shipping address)
- [ ] SQLite fallback


src/pretix/base/management/commands/reindex_orders.py
```diff
@@ -0,0 +1,69 @@
...
+            last_pk = 0
+            with tqdm(total=qs.count()) as pbar:
+                while True:
+                    batch = list(qs.filter(pk__gt=last_pk)[:5000])
+                    if not batch:
+                        break
+
+                    for o in batch:
+                        o._update_search_index()
+                        time.sleep(0)
```
luelista, line 67:
```suggestion
                        time.sleep(args.interval / 1000.0)
```

PR #6212, opened 2026-05-22
Remove duplicated csrftoken cookies

luelista:
Due to a Safari bug, in some browser, two csrftoken cookies with different values exist: one unpartitioned, one partitioned ("CHIPS"). This function generates an additional Set-Cookie header to get rid of the unpartitioned one.

As Django usually only allows one Set-Cookie header per cookie name, we need to manually create a cookie 'Morsel' for the deletion and store it in the HttpResponse's cookie dictionary under a different name, so it is not overwritten by the actual, correct Set-Cookie header. This works because the code in django.core.handlers.wsgi/asgi, that generates the actual Set-Cookie headers, only iterates over cookie.values(), ignoring the keys.

### DIFF
--- a/src/pretix/multidomain/middlewares.py
+++ b/src/pretix/multidomain/middlewares.py
@@ -282,7 +282,7 @@ def handle_duplicated_csrftoken(request, response):
 
     cookie_name = '__Host-' + settings.CSRF_COOKIE_NAME
 
-    if request.scheme == 'https' and cookie_name:
+    if request.scheme == 'https' and cookie_name in request.COOKIES:
         values = get_all_values_of_cookie(request.headers.get('Cookie'), cookie_name)
         if len(values) > 1:
             logger.info('Trying to remove duplicated %s cookies: %r', cookie_name, values)
--- a/src/pretix/multidomain/middlewares.py
+++ b/src/pretix/multidomain/middlewares.py
@@ -316,11 +316,6 @@ def get_all_values_of_cookie(cookie_header, cookie_name):
     return values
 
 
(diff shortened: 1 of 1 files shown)


src/pretix/multidomain/middlewares.py
```diff
@@ -265,3 +273,60 @@ def _set_csrf_cookie(self, request, response):
...
+            key, val = "", chunk
+        key, val = key.strip(), val.strip()
+        if key == cookie_name:
+            values.append(val)
+    return values
+
+
+def has_duplicated_csrftoken(request):
+    values = get_all_values_of_cookie(request.headers.get('Cookie'), '__Host-' + settings.CSRF_COOKIE_NAME)
+    return len(values) > 1
```
raphaelm:
```suggestion
```

I think this is unused?

src/pretix/multidomain/middlewares.py
```diff
@@ -265,3 +273,60 @@ def _set_csrf_cookie(self, request, response):
...
+
+
+def handle_duplicated_csrftoken(request, response):
+    # Due to a Safari bug, in some browser, two csrftoken cookies with different values
+    # exist: one unpartitioned, one partitioned. This function generates an additional
+    # Set-Cookie header to get rid of the unpartitioned one.
+
+    cookie_name = '__Host-' + settings.CSRF_COOKIE_NAME
+
+    if request.scheme == 'https' and cookie_name:
```
raphaelm:
```suggestion
    if request.scheme == 'https' and cookie_name in request.COOKIES:
```

raphaelm (approved):

PR #6215, opened 2026-05-24
Private IP webhook protection incompatibility with urllib3.future

TheEnbyperor:
If a plugin has installed urllib3.future (or any package that depends on it, like niquests), then recent versions of Pretix fail to start as the monkeypatching for blocking access to private IP addresses relies too heavily on the internals of urllib3; specifically, it fails to import `_TYPE_SOCKET_OPTIONS`, amongst others.

Since urllib3.future is "a drop-in replacement, fully compatible with its predecessor", then it would make sense to me to support that as the way forward.

Another advantage of using urllib3.future is its resolver framework. This allows the code to filter out private addresses by registering a custom resolver, without having to reimplement and keep up to date all the surrounding code about connection management.

This PR makes pretix always use urllib3.future, and rewrites the private IP protection code to a custom resolver extending the default resolver.


raphaelm:
Interesting, I did not know about urllib3.future. A custom resolver would indeed make it a little cleaner. Could you look into why the tests are failing?

(See also #6220 for potential merge conflicts)

TheEnbyperor:
I cannot reproduce these test failures on my machine. It looks as if GitHub is messing with the network of the test runner, or example.com/example.org was broken at the time the tests ran?

raphaelm:
Pretty sure the old tests didn't haven actual network dependency, though :thinking:

TheEnbyperor:
You're right they didn't previously have a network dependency. Updating to urllib3-future broke the tests in a really annoying way to notice, but I've got it now. 

Specifically, urrlib3-future doesn't use `socket.socket` (which `mock.patch()` was overriding), rather it uses a custom subclass thereof by the name `_with_attr_sock`. This subclass exists, as far as I can tell, to attach Encrypted Client Hello metadata to a socket. 

Updating the test mock to use the subclassed socket makes everything work now.

raphaelm:
After some thoughts we don't think we'll merge it and rather consider pretix incompatible with urllib3.future for now. I like the approach of a subclassable resolver, it is obviously much cleaner code than the monkeypatching we need to do for urllib3. But urllib3.future looks like a single-maintainer project with unclear history and future and we'd like to avoid rely on it for something so central (and potentially prone to security issues).

PR #6217, opened 2026-05-25
Align email footer attribution with site footer

kiancross:
Replace the hardcoded "powered by pretix" with the configured attribution shown on the site.

Adds `get_email_powered_by()`, which links only the name and "pretix" (a full-width link looks heavy in email). The settings lookup is shared with `get_powered_by()` via a new `_get_powered_by_data()` helper.


raphaelm:
Honestly… I think we just want pretix to be promoted there ;) AGPL notice stuff doesn't apply here and I don't see why it's in our interest to make this even easier. Overriding email_footer.html with a local substitute is already plenty easy ;)

PR #6219, opened 2026-05-26
User email verifications: Add rate limiting (Z#23232834)

raphaelm:
Also, refactor rate limiting into a central utility


src/pretix/helpers/ratelimit.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+def rate_limit(key: str, *parameters, include_ip_from_request: HttpRequest=None, max_num: int, expire_time: int):
+    """
+    This is a shared utility to implement simple rate limiting in operations like
+    password resets. This is by far no perfect implementation of rate limiting, as
+    it the window is prolonge
+
+    :param key: The key refering to the feature like "pwreset"
+    :param parameters: Any number of things to be hashed as the bucket key
```
luelista, line 59:
```suggestion
    :param parameters: Any number of things to be hashed as the bucket key
    :param include_ip_from_request: Add IP address from request to the bucket key
```

src/pretix/helpers/ratelimit.py
```diff
@@ -0,0 +1,110 @@
...
+        # This is the private IP of the server, web server not set up correctly
+        return None
+    return str(client_ip)
+
+
+def rate_limit(key: str, *parameters, include_ip_from_request: HttpRequest=None, max_num: int, expire_time: int):
+    """
+    This is a shared utility to implement simple rate limiting in operations like
+    password resets. This is by far no perfect implementation of rate limiting, as
+    it the window is prolonge
```
luelista: some word(s) missing here?

src/pretix/base/forms/user.py
luelista, line 193: We could additionally rate limit the speed at which one can check whether email addresses are in use or not:
```suggestion
        if rate_limit("emailchangeattempt", include_ip_from_request=self.request=True, max_num=5, expire_time=300):
            raise forms.ValidationError(
                self.error_messages['rate_limit'],
                code='rate_limit',
            )
        if User.objects.filter(Q(email__iexact=email) & ~Q(pk=self.user.pk)).exists():
```

src/pretix/base/forms/auth.py
```diff
@@ -85,45 +82,22 @@ def __init__(self, backend, request=None, *args, **kwargs):
...
-                    rc.incr(self.ratelimit_key)
-                    rc.expire(self.ratelimit_key, 300)
                 logger.info("Backend login invalid.")
                 pretix_failed_logins.inc(1, reason="invalid")
                 raise forms.ValidationError(
                     self.error_messages['invalid_login'],
                     code='invalid_login'
                 )
             else:
+                rate_limit_reset("login", include_ip_from_request=self.request)
```
luelista: At least for hosted, this renders the rate limit somewhat useless, as one could simply make every tenth request with a known-good user account (which anyone can easily create). I think it would be useful to have a separate rate_limit_check method that only checks without increasing the rate limit, so we can keep the existing logic.

luelista (commented):
i really like that refactoring 👍 

not tested yet, added some comments from reading

src/pretix/base/forms/auth.py
luelista: This should still be removed, right?
```suggestion
```

src/pretix/base/forms/user.py
```diff
@@ -175,19 +170,35 @@ class UserEmailChangeForm(forms.Form):
...
+                code='rate_limit',
+            )
+
         if User.objects.filter(Q(email__iexact=email) & ~Q(pk=self.user.pk)).exists():
             raise forms.ValidationError(
                 self.error_messages['duplicate_identifier'],
                 code='duplicate_identifier',
             )
+
+        if rate_limit("emailchange", self.user.pk, max_num=1, expire_time=300):
```
luelista:
```suggestion
        if rate_limit("emailchange", self.user.pk, max_num=2, expire_time=300):
```
maybe set to 2, so you can have one typo without having to wait 5 min directly?

src/pretix/presale/forms/customer.py
luelista, line 342: what do you think about adding a rate limit here, to make mass checks of email validity more difficult?
```suggestion
        if rate_limit("customer_pwreset_check", max_num=10, expire_time=600):
            raise forms.ValidationError(
                self.error_messages['rate_limit'],
                code='rate_limit',
            )
        try:
```

src/pretix/control/views/auth.py
```diff
@@ -412,11 +407,7 @@ def post(self, request, *args, **kwargs):
             messages.success(request, _('You can now login using your new password.'))
             user.log_action('pretix.control.auth.user.forgot_password.recovered')
 
-            has_redis = settings.HAS_REDIS
-            if has_redis:
-                from django_redis import get_redis_connection
-                rc = get_redis_connection("redis")
-                rc.delete('pretix_pwreset_%s' % user.id)
+            rate_limit_reset("pwreset", user.pk)
```
luelista, line 410: maybe we should put a `rate_limit_reset("pwreset", user.pk)` in `UserPasswordChangeView.form_valid` too? otherwise, if I request a reset link, then not use it because I remember my password, and then change it in the backend, my link will be invalidated because the hash changed, but I can't get another reset link before 24 hours pass.

luelista (approved):

PR #6222, opened 2026-05-27
Enable sending logs to Sentry

sweenu:
Sentry has a feature to see logs in its UI. My changes allow configuring that.
I've tested the changes on my instance. It works well.

I put the log level to `NOTSET`, but I could make it configurable if you want.


src/pretix/settings.py
```diff
@@ -733,14 +734,16 @@ def traces_sampler(sampling_context):
             CeleryIntegration(),
             LoggingIntegration(
                 level=logging.INFO,
-                event_level=logging.CRITICAL
+                event_level=logging.CRITICAL,
+                sentry_logs_level=logging.NOTSET if SENTRY_ENABLE_LOGS else None,
```
raphaelm, line 738: I think this can just be INFO, which is Sentry's default

PR #6235, opened 2026-06-01
Allow mails to be sent to addon-attendees (Z#23213551)

ser8phin:
As add-on-products can also be used for attendees (for group tickets i.e.), this removes the restriction where pretix did not send mails to those as well as the restriction where pretix did not show product/ticket-pages for add-ons only. This may lead to more mails being sent to the same mail-addresses, if used for multiple products in one order, however, considering we cannot guess all product-add-on-configurations or whether someone uses their mail-address as a placeholder only for example, we found that we cannot restrict these further.

Still to do:
- [x] rethink spamlike multi-mails
- [x] mail-inheritance behavior for product -> add-ons
- [x] check mail-rules


src/pretix/plugins/sendmail/tasks.py
```diff
@@ -84,11 +86,14 @@ def _send_to_order(o):
                         list_id__in=checkin_lists or []
                     )
                 ),
-            ).prefetch_related('addons', 'subevent'):
-                if p.addon_to_id is not None:
-                    continue
+            ).order_by('pk').prefetch_related('addons', 'subevent'):
```
wiffbi: Could add a comment as to why this needs to be ordered by PK? I assume this needs to be oredered by PK to make sure we always have seen the parent-OP before the addon-OP? If so, I think we should oreder by addon-to with NULL-values being first, as maybe through some weird way of database tinkering, the PK-structure for OP and addon-OPs is not as expected.

src/pretix/plugins/sendmail/models.py
```diff
@@ -166,7 +166,8 @@ def send(self):
 
         for o in orders:
             with language(o.locale, e.settings.region):
-                positions = list(o.positions.all())
+                positions = list(o.positions.all().order_by(F("addon_to_id").asc(
+                    nulls_first=True)))  # ordered to make sure we always handle parent order positions before add-ons
```
wiffbi: Hm, this breaks our assumption, that `o.positions.all()` sorts an addon directly after its parent_op. I think this approach with ordering by `addon_to_id` with `nulls_first` came from when we had a parent_ops dictionary at some point? I can’t remember, sorry! But currently the algorithm expects positions to be ordered in a way that addons directly follow after their parent, which is I think by default the case (sorted by positionid?).


```suggestion
                positions = list(o.positions.all())
```

src/pretix/plugins/sendmail/tasks.py
```diff
@@ -84,11 +86,15 @@ def _send_to_order(o):
                         list_id__in=checkin_lists or []
                     )
                 ),
-            ).prefetch_related('addons', 'subevent'):
-                if p.addon_to_id is not None:
-                    continue
+            ).order_by(F("addon_to_id").asc(nulls_first=True)).prefetch_related('addons', 'subevent'):
+                # ordered QuerySet to make sure we always handle parent order positions before add-ons
```
wiffbi: See comment above regarding ordering. I know this was a suggestion by me to make sure, we always have seen the parent op before the addon, but I can’t remember why this ever was the case? Maybe the OrderPosition-queryset, which is used filter orders, lead me to believe we needed that. Sorry!


```suggestion
            ).prefetch_related('addons', 'subevent'):
```

src/pretix/plugins/sendmail/models.py
```diff
@@ -187,13 +188,33 @@ def send(self):
...
+                        if p.addon_to_id is None:
+                            parent_op = p
+                        if not self.rule.all_products and p.item_id not in position_ids:
+                            continue
+
                         if p.id in position_ids:
+                            mail_to_parent = False
+
+                            # position is an add-on
+                            if p.addon_to_id is not None:
```
wiffbi: Although we should be able to rely on correct sorting of positions, we might we add a double-check that parent_op is actually really the parent of this addon, so we do not accidentally send emails that are not meant to be sent.

```suggestion
                            if p.addon_to_id:
                                if not parent_op or parent_op.id != p.addon_to_id:
                                    # something mixed up with this order as addons should always come after their parent-position
                                    continue
```

src/pretix/plugins/sendmail/models.py
```diff
@@ -187,13 +188,33 @@ def send(self):
...
+                                # without attendee-email
```
wiffbi: I think this comment is to verbose, the if statement is IMHO easy to read.
```suggestion
```

src/pretix/plugins/sendmail/models.py
```diff
@@ -187,13 +188,33 @@ def send(self):
...
+
+                        if p.addon_to_id is None:
+                            parent_op = p
+                        if not self.rule.all_products and p.item_id not in position_ids:
+                            continue
+
                         if p.id in position_ids:
+                            mail_to_parent = False
+
+                            # position is an add-on
```
wiffbi: I think this comment is to verbose, the if statement is IMHO easy to read.
```suggestion
```

src/pretix/plugins/sendmail/models.py
```diff
@@ -187,13 +188,33 @@ def send(self):
...
+                            if p.addon_to_id is not None:
+                                # without attendee-email
+                                if not p.attendee_email:
+                                    if p.addon_to_id in sent_to_positions:
```
wiffbi, line 208: Wouldn’t it make sense to just assign `p = parent_op` and check if that is in sent_to_positions? That way we would not need `mail_to_parent` and that second elif-part in line 230?

wiffbi (changes requested):

src/pretix/plugins/sendmail/models.py
ser8phin: true, at this point that works. changed it :)

src/pretix/plugins/sendmail/models.py
```diff
@@ -187,13 +187,32 @@ def send(self):
...
+                    parent_op = None
+                    sent_to_positions = set()
                     for p in positions:
+
+                        if p.addon_to_id is None:
+                            parent_op = p
+                        if not self.rule.all_products and p.id not in position_ids:
+                            continue
+
                         if p.id in position_ids:
```
pajowu: shouldn't this now also be
```suggestion
                        if self.rule.all_products or p.id in position_ids:
```

src/pretix/plugins/sendmail/models.py
```diff
@@ -187,13 +187,32 @@ def send(self):
...
+                            if p.addon_to_id:
+                                if not parent_op or parent_op.id != p.addon_to_id:
+                                    # something got mixed up with this order as addons should always come after their parent-position
```
pajowu, line 205: maybe add a `logger.warning` at least so we see if this happens. just silently ignoring this even though it implies a bigger problem somewhere else seems dangerous to me

src/pretix/plugins/sendmail/models.py
```diff
@@ -187,13 +187,32 @@ def send(self):
...
                         positions = [p for p in positions if p.subevent_id == self.subevent_id]
 
+                    parent_op = None
+                    sent_to_positions = set()
                     for p in positions:
+
+                        if p.addon_to_id is None:
+                            parent_op = p
+                        if not self.rule.all_products and p.id not in position_ids:
+                            continue
```
pajowu: isn't this identical to the `if p.id in position_ids` check in the next line?

src/pretix/plugins/sendmail/models.py
```diff
@@ -187,13 +187,32 @@ def send(self):
                     o_sent = True
 
                 if send_to_attendees:
-                    if not self.rule.all_products:
-                        positions = [p for p in positions if p.item_id in limit_products]
```
pajowu, line 191: not sure this can just be deleted. i think this checks need to be re-added to the inner of the for loop

src/pretix/plugins/sendmail/models.py
pajowu, line 200: this is check seems wrong? `p.id in position_ids` is also checked in the next line.

src/tests/plugins/sendmail/test_rules.py
```diff
@@ -197,6 +197,82 @@ def test_sendmail_rule_send_order_vs_pos(send_to, amount_mails, recipients, orde
     assert set(recipients) == set(_recipients)
 
 
+@pytest.mark.django_db
```
pajowu, line 200: could you also add a test for rules with `all_products=True`?

src/pretix/plugins/sendmail/models.py
```diff
@@ -187,13 +187,32 @@ def send(self):
...
                     if self.subevent_id:
                         positions = [p for p in positions if p.subevent_id == self.subevent_id]
 
+                    parent_op = None
+                    sent_to_positions = set()
                     for p in positions:
+
+                        if p.addon_to_id is None:
```
pajowu, line 197: this loop now has a lot of complexity that makes it hard to understand. can this be simplified / split up to make it more readable?

src/pretix/plugins/sendmail/models.py
wiffbi: It is in the inner for-loop, but it looks a bit different: line 199 checks for `p.id not in position_ids`, where position_ids (or its underlying op_qs) has a filter on limit_products (among others). That `if` is a bit misleading IMHO though as it would not need the `not self.rule.all_products`.

Removing the limit_products here is just there to also get parent_ops in the for-loop in case we need access to them for e-mail-address.

src/pretix/plugins/sendmail/models.py
wiffbi: This check basically is the same as the check in line 190 regarding limit_products. But you are right, we could/should simplify this to basically `if p.id not in position_ids` to skip any ops that are not meant to be in the loop (ie parent_op with wrong item). That would even remove the need of the next `if p.id in position_ids:`

src/pretix/plugins/sendmail/models.py
wiffbi: I will look into this

PR #6237, opened 2026-06-01
Fix #6232 - Remove pagination on categories and discounts

### LINKED ISSUE #6232: Categories cannot be reordered when not all are displayed
### Problem and impact

When creating more than 25 product categories, using the reordering feature throws an error. 

Displaying 50 or 100 categories per page fixes the issue, as this shows all existing categories in my use case. Thus, it is only a major problem when having more than 100 product categories at which the UI cannot be used anymore to reorder sections, even with this workaround. 

### Expected behaviour

When reordering categories, the request should succeed no matter how many categories exist or are currently displayed. 

### Steps to reproduce

1. Create more than 25 product categories in a single event
2. Move a product category up or down
3. Observe error message "Not all objects have been selected."

### Screenshots

More than 25 categories are paginated in the UI: 

Image

Error message shown when reordering: 

Image

The API request in fact only contains 25 sections being reordered, despite there being more in the event. 
Image

### Link

/control/event/\<organizer\>/\<event\>/categories/

_No response_

_No response_

### Version

most recent pretix hosted as well as 2026.4

ser8phin:


luelista (commented):
we could also remove `{% include "pretixcontrol/pagination.html" %}` from the templates (but it's not necessary as the included file contains an `if is_paginated` check)

in pretix/src/pretix/control/templates/pretixcontrol/items/discounts.html:171 and pretix/src/pretix/control/templates/pretixcontrol/items/categories.html:78

luelista (approved):

PR #6254, opened 2026-06-08
add widget's display-event-info property to pretix-button

Qreepex:
When using the Widget, the `<pretix-button>` currently doesnt support adding the `display-event-info` flag.
The button uses the same underlying logic as the `<pretix-widget>`, yet still doesn't support that flag - so i added it:)

If accepted, adjusting the available props documented on https://docs.pretix.eu/guides/widget/#pretix-button would follow

### DIFF
--- a/src/pretix/static/pretixpresale/widget/src/button.ts
+++ b/src/pretix/static/pretixpresale/widget/src/button.ts
@@ -10,6 +10,11 @@ export function createButtonInstance (element: Element, htmlId?: string): App {
 		targetUrl += '/'
 	}
 
+	const displayEventInfoAttr = element.attributes['display-event-info']?.value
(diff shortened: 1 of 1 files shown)


wiffbi:
Hey, thanks for your contribution! What do you want to achieve with adding this attribute to `<pretix-button>`? The button only opens the iframe-overlay and does not show any event/subevent related info itself. The attribute does not affect how the info inside the iframe-overlay will be rendered, so it would not have any effect to add this attribute.

Qreepex:
Thanks for the swift response! I was hoping that the widget thats being opened by the button will then have that property set and show the additional event info.
Looked like they both use the same underlying logic at first. So I didnt check enough of the code and opened the PR too fast, shame on me :)
Thank you again for your time!

PR #6259, opened 2026-06-09
Bank transfer: More restrictive approach to permissions (Z#23235727)

raphaelm:
- [x] "You can only see org-level data if you can see **all** events" instead of "You can only see org-level data if you can see **one** events"
- [x] Graceful handling of read-only permissions
- [ ] Tests


raphaelm:
After internal discussion, we'll keep it as-is until we see real-world use cases that cause us to think otherwise

PR #6268, opened 2026-06-10
Prune wheel and setuptools-rust from build-system

mweinelt:
For wheel the setuptools documentation notes:

> Historically this documentation has unnecessarily listed wheel in
> the requires list, and many projects still do that. This is not
> recommended, as the backend no longer requires the wheel package,
> and listing it explicitly causes it to be unnecessarily required for
> source distribution builds.

https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use

For setuptools-rust I could not find any Rust extension that need to be built. The introduction goes back to c132ccd14, where css-inline, a rust component, was added as a dependency.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,8 +57,7 @@ COPY vite.config.ts /pretix/vite.config.ts
 
 RUN pip3 install -U \
         pip \
-        setuptools \
-        wheel && \
+        setuptools && \
     cd /pretix && \
     PRETIX_DOCKER_BUILD=TRUE pip3 install \
         -e ".[memcached]" \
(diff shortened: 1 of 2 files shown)


raphaelm:
Well, let's see if something breaks, I guess? We still depend on css-inline, I could imagine setuptools-rust still being required when using a distro/architecture with no binary wheels available, but then again, python build tooling progressed over the last 5 years that it could now be isolated enough to work anyways

mweinelt:
Pip should have proper build isolation for pep517/518 builds, where build time dependency constraints could otherwise run into conflicts. Also, the css-inline package uses maturin to build these days, so setuptools-rust was most certainly just there doing nothing.

PR #6274, opened 2026-06-12
Test order deletion: Improve bulk performance (Z#23237160)

raphaelm:


src/pretix/base/models/orders.py
```diff
@@ -354,38 +354,59 @@ def __init__(self, *args, **kwargs):
...
+                save=False,
+            ))
+        LogEntry.bulk_create_and_postprocess(logs_create)
+
+        vouchers = OrderPosition.objects.filter(
+            order__in=orders,
+            voucher__isnull=False
+        ).exclude(order__status=Order.STATUS_CANCELED).values_list("voucher_id", flat=True)
+        for v_id in vouchers:
+            Voucher.objects.filter(pk=v_id).update(redeemed=Greatest(0, F('redeemed') - 1))
```
pajowu: option to reduce nr of db queries if vouchers are used multiple times:

```suggestion
        voucher_ids = OrderPosition.objects.filter(
            order__in=orders,
            voucher__isnull=False
        ).exclude(order__status=Order.STATUS_CANCELED).values_list("voucher_id", flat=True)
        voucher_usages = Counter(voucher_ids)
        for v_id, usage_count in voucher_usages.items():
            Voucher.objects.filter(pk=v_id).update(redeemed=Greatest(0, F('redeemed') - usage_count))

```

pajowu (approved):
My only concern is that `Order.gracefully_delete` could be called outside an transaction, but I couldn't find any cases of that, not even a plugin

Issue #6248, opened 2026-06-05
"Invalidate Secrets" overwrites gift card secrets stored in ticket secrets


dasGoogle:
### Problem and impact

When using the "This product is a gift card" functionality on an item to sell gift cards, the gift card secret is also placed in the ticket secret of the sold ticket. This allows printing out gift cards, even with QR codes that can be scanned by PretixPOS which we use extensively. 

When regenerating secrets, either for the whole order or just one position, this gift card secret is overwritten with a normal ticket secret. The tickets thus do not contain the gift card secret anymore and are thus unusable for customers. 

### Expected behaviour

Regenerating secrets should not have an impact on the secrets stored in the accompanying tickets. 

### Steps to reproduce

1. Create an event with an item with the "This is a gift card" switch set
2. Place an order in the shop
3. Regenerate secrets (e.g. by changing the email address on the order and checking the corresponding box)
4. See that the ticket secret has been changed

### Screenshots

Before regenerating secrets, the gift card secret is also the ticket secret: 
<img width="1054" height="670" alt="Image" src="https://github.com/user-attachments/assets/fa5647b0-bf45-440c-8e6f-27efd8cb230b" />

After regenerating secrets, the gift card secret is replaced with a new random ticket secret: 
<img width="1062" height="690" alt="Image" src="https://github.com/user-attachments/assets/ac3493ed-5b56-4324-9265-bc7e4f32a6e6" />

### Version

Latest pretix hosted and 2026.04

luelista:
I wonder whether it would make more sense to skip these order positions when regenerating secrets, or to actually regenerate the gift card secret, and apply the new gift card secret as the ticket secret.

dasGoogle:
> regenerate the gift card secret, and apply the new gift card secret as the ticket secret.

I see where you're coming from, but I presume that would make things a lot more complex since AFAIK gift card secrets are considered immutable. So regenerating gift card secrets either means reissuing a gift card or changing that assumption. 

My/our use case had the issue that we recreated secrets and thus broke the gift cards in the way described here. Combined with the fact that the API provides no direct way to get the gift card associated with a ticket, this was quite painful to fix. 

So, for us, just ignoring them should be fine. The use case of "I change this order's email and I want to revoke the old user's access" is, however, not covered by that, that is true.

luelista:
In which way exactly did you change the secrets? per-order API, per-ticket API, 'change email' dialog, or 'change products' dialog?

[...]

PR #6281, opened 2026-06-12
Prevent changing ticket secret of gift-card-issuing order positions

luelista:
Fixes issue #6248

image

image

### DIFF
--- a/doc/api/resources/orders.rst
+++ b/doc/api/resources/orders.rst
@@ -864,6 +864,9 @@ Generating new secrets
 
    Triggers generation of new ``secret`` and ``web_secret`` attributes for both the order and all order positions.
 
+   Ticket secrets of order positions that have been used to issue a gift card can not
+   be changed. Only the link (``web_secret``) will be changed in this case.
+
    **Example request**:
 
    .. sourcecode:: http
@@ -894,6 +897,9 @@ Generating new secrets
 .. http:post:: /api/v1/organizers/(organizer)/events/(event)/orderpositions/(id)/regenerate_secrets/
 
    Triggers generation of a new ``secret`` and ``web_secret`` attribute for a single order position.
+
+   Ticket secrets of order positions that have been used to issue a gift card can not
+   be changed. Only the link (``web_secret``) will be changed in this case.
 
    **Example request**:
 
--- a/src/pretix/base/secrets.py
+++ b/src/pretix/base/secrets.py
@@ -245,6 +245,9 @@ def recv_classic(sender, **kwargs):
 
 
 def assign_ticket_secret(event, position, force_invalidate_if_revokation_list_used=False, force_invalidate=False, save=True):
+    if position.issued_gift_cards.exists():
+        return
+
     gen = event.ticket_secret_generator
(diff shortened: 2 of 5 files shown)


src/pretix/control/views/orders.py
```diff
@@ -2307,7 +2313,7 @@ def post(self, *args, **kwargs):
             if self.form.cleaned_data['regenerate_secrets']:
                 changed = True
                 self.order.secret = generate_secret()
-                for op in self.order.all_positions.all():
+                for op in self.order.all_positions.filter(~Exists(GiftCard.objects.filter(issued_in=OuterRef('pk')))):
```
pajowu: Shouldn't `op.web_secret` still be changed in this case?

luelista: oops, yes, i thought i had removed this change again

```suggestion
                for op in self.order.all_positions.all():
```

src/pretix/base/secrets.py
```diff
@@ -245,6 +245,9 @@ def recv_classic(sender, **kwargs):
 
 
 def assign_ticket_secret(event, position, force_invalidate_if_revokation_list_used=False, force_invalidate=False, save=True):
+    if position.issued_gift_cards.exists():
```
luelista:
```suggestion
    if position.pk and position.issued_gift_cards.exists():
```

src/pretix/control/templates/pretixcontrol/order/change.html
```diff
@@ -284,6 +284,14 @@ <h3 class="panel-title">
                             </div>
                             <div class="col-sm-4">
                                 {% bootstrap_field position.form.operation_secret layout='inline' %}
+                                {% if position.issued_gift_cards.exists %}
```
raphaelm, line 287: Will this create N queries for N order positions? Since we allow up to 500 positions per order, can we prefetch this?

doc/api/resources/orders.rst
```diff
@@ -895,6 +898,9 @@ Generating new secrets
 
    Triggers generation of a new ``secret`` and ``web_secret`` attribute for a single order position.
 
+   Ticket secrets of order positions that have been used to issue a gift card can not
+   be changed. Only the link (``web_secret``) will be changed in this case.
```
raphaelm, line 902: Is that what happens or will it fail completely?

luelista: yes, that should be what happens. the new check where an `OrderError` is raised is only triggered during `change_ticket_secret`, but not in `regenerate_secret` which is used here.

src/pretix/control/templates/pretixcontrol/order/change.html
luelista: good point, [this](https://github.com/pretix/pretix/pull/6281/changes/abfec02fea7e62267f2580638427caf530f2f334) should do (that will load some unnecessary data though. but i'm not sure whether something like prefetch_related('issued_gift_cards__exists') is possible... maybe using annotate and a subquery?)

raphaelm (approved):

PR #6284, opened 2026-06-15
include errors.js by default and make it coop with async_task_replace_page (Z#23236752)

bockstaller:
Errors.js is small enough to include by default and as a result sidestep the security implications of [replacing innerHTML](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#security_considerations).

Listening to the new `pretix:async_task_replace_page` event also allows us to change the backwards-link behaviour to a reload of the page whenever the page contend is replaced

### DIFF
--- a/src/pretix/base/templates/error.html
+++ b/src/pretix/base/templates/error.html
@@ -11,6 +11,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <meta charset="utf-8">
     <link rel="icon" href="{% static "pretixbase/img/favicon.ico" %}">
+    <script type="text/javascript" src="{% static "pretixbase/js/errors.js" %}"></script>
     {% block custom_header %}{% endblock %}
     {% if css_theme %}
         <link rel="stylesheet" type="text/css" href="{{ css_theme }}" />
--- a/src/pretix/base/templates/error.html
+++ b/src/pretix/base/templates/error.html
@@ -20,6 +20,5 @@
 <div class="container">
     {% block content %}{% endblock %}
 </div>
-<script src="{% static "pretixbase/js/errors.js" %}"></script>
 </body>
 </html>
--- a/src/pretix/static/pretixbase/js/asynctask.js
+++ b/src/pretix/static/pretixbase/js/asynctask.js
@@ -182,7 +182,7 @@ function async_task_error(jqXHR, textStatus, errorThrown) {
                     jqXHR.responseText.indexOf("<body"),
                     jqXHR.responseText.indexOf("</body")
                 ));
-                document.dispatchEvent(new Event("pretix:async_task_replace_page:on_error"))
+                document.dispatchEvent(new Event("pretix:async-task-error"))
 
             }
 
--- a/src/pretix/static/pretixbase/js/asynctask.js
+++ b/src/pretix/static/pretixbase/js/asynctask.js
@@ -107,6 +107,7 @@ function async_task_replace_page(target, new_html) {
     setup_collapsible_details($(target));
     window.setTimeout(function () { $(window).scrollTop(0) }, 200)
     $(document).trigger("pretix:bind-forms");
+		$(document).trigger("pretix:async_task_replace_page");
 }
 
 function async_task_check_error(jqXHR, textStatus, errorThrown) {
--- a/src/pretix/static/pretixbase/js/asynctask.js
+++ b/src/pretix/static/pretixbase/js/asynctask.js
@@ -107,7 +107,6 @@ function async_task_replace_page(target, new_html) {
     setup_collapsible_details($(target));
     window.setTimeout(function () { $(window).scrollTop(0) }, 200)
     $(document).trigger("pretix:bind-forms");
-		$(document).trigger("pretix:async_task_replace_page");
 }
 
 function async_task_check_error(jqXHR, textStatus, errorThrown) {
@@ -175,15 +174,15 @@ function async_task_error(jqXHR, textStatus, errorThrown) {
         var respdom = $(jqXHR.responseText);
         var c = respdom.filter('.container');
         if (respdom.filter('form') && (respdom.filter('.has-error') || respdom.filter('.alert-danger'))) {
-            // This is a failed form validation, let's just use it
-
             if (respdom.filter('#page-wrapper') && $('#page-wrapper').length) {
+								// This is a failed form validation, let's just use it
                 async_task_replace_page("#page-wrapper", respdom.find("#page-wrapper").html());
             } else {
                 async_task_replace_page("body", jqXHR.responseText.substring(
                     jqXHR.responseText.indexOf("<body"),
                     jqXHR.responseText.indexOf("</body")
                 ));
+								$(document).trigger("pretix:async_task_replace_page:on_error");
             }
 
         } else if (c.length > 0) {
--- a/src/pretix/static/pretixbase/js/asynctask.js
+++ b/src/pretix/static/pretixbase/js/asynctask.js
@@ -182,7 +182,7 @@ function async_task_error(jqXHR, textStatus, errorThrown) {
                     jqXHR.responseText.indexOf("<body"),
                     jqXHR.responseText.indexOf("</body")
                 ));
-								$(document).trigger("pretix:async_task_replace_page:on_error");
+                $(document).trigger("pretix:async_task_replace_page:on_error");
             }
 
         } else if (c.length > 0) {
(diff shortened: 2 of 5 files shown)


src/pretix/static/pretixbase/js/asynctask.js
```diff
@@ -174,15 +174,15 @@ function async_task_error(jqXHR, textStatus, errorThrown) {
...
-
             if (respdom.filter('#page-wrapper') && $('#page-wrapper').length) {
+								// This is a failed form validation, let's just use it
                 async_task_replace_page("#page-wrapper", respdom.find("#page-wrapper").html());
             } else {
                 async_task_replace_page("body", jqXHR.responseText.substring(
                     jqXHR.responseText.indexOf("<body"),
                     jqXHR.responseText.indexOf("</body")
                 ));
+								$(document).trigger("pretix:async_task_replace_page:on_error");
```
pajowu:
```suggestion
                $(document).trigger("pretix:async_task_replace_page:on_error");
```

src/pretix/static/pretixbase/js/asynctask.js
```diff
@@ -174,15 +174,15 @@ function async_task_error(jqXHR, textStatus, errorThrown) {
         var respdom = $(jqXHR.responseText);
         var c = respdom.filter('.container');
         if (respdom.filter('form') && (respdom.filter('.has-error') || respdom.filter('.alert-danger'))) {
-            // This is a failed form validation, let's just use it
-
             if (respdom.filter('#page-wrapper') && $('#page-wrapper').length) {
+								// This is a failed form validation, let's just use it
```
pajowu:
```suggestion
                // This is a failed form validation, let's just use it
```

src/pretix/base/templates/error.html
```diff
@@ -20,6 +20,5 @@
 <div class="container">
     {% block content %}{% endblock %}
 </div>
-<script src="{% static "pretixbase/js/errors.js" %}"></script>
```
pajowu, line 23: `error.html` does not include `fragment_js.html`, breaking the `goback`/`reload`-links on non-ajax error pages. We either need to keep it here or load `fragment_js.html` as well

src/pretix/static/pretixbase/js/errors.js
```diff
@@ -1,5 +1,17 @@
...
 
-document.getElementById('reload').onclick =
- function() {window.location.reload(true)};
+  const reload = document.getElementById('reload');
+  if (reload) {
+    reload.onclick = () => window.location.reload(true);
+  }
+};
+
+registerErrorLinkHandlers();
```
pajowu: since the script is now loaded in the `head`, the function needs to wait until `DOMContentLoaded`
```suggestion
document.addEventListener("DOMContentLoaded", () => registerErrorLinkHandlers());
```

src/pretix/static/pretixbase/js/asynctask.js
```diff
@@ -174,15 +174,15 @@ function async_task_error(jqXHR, textStatus, errorThrown) {
...
                 async_task_replace_page("#page-wrapper", respdom.find("#page-wrapper").html());
             } else {
                 async_task_replace_page("body", jqXHR.responseText.substring(
                     jqXHR.responseText.indexOf("<body"),
                     jqXHR.responseText.indexOf("</body")
                 ));
+                $(document).trigger("pretix:async_task_replace_page:on_error");
```
pajowu: in case we want to ditch jquery in `errors.js`, this would need to be
```suggestion
                document.dispatchEvent(new Event("pretix:async_task_replace_page:on_error"))

```

src/pretix/static/pretixbase/js/errors.js
```diff
@@ -1,5 +1,17 @@
...
+
+$(document).on("pretix:async_task_replace_page:on_error", () => registerErrorLinkHandlers(true));
```
pajowu: depending on how this script is included in `error.html`, this might need to work without jquery (jquery is only loaded in `fragment_js.html`), i.e. the change suggested in asynctask.js and
```suggestion
document.addEventListener("pretix:async_task_replace_page:on_error", () => registerErrorLinkHandlers(true));
```

pajowu (changes requested):
This currently breaks the links on non-ajax pages. I added suggestions on how this could be fixed

src/pretix/static/pretixbase/js/asynctask.js
```diff
@@ -174,15 +174,16 @@ function async_task_error(jqXHR, textStatus, errorThrown) {
...
-
             if (respdom.filter('#page-wrapper') && $('#page-wrapper').length) {
+                // This is a failed form validation, let's just use it
                 async_task_replace_page("#page-wrapper", respdom.find("#page-wrapper").html());
             } else {
                 async_task_replace_page("body", jqXHR.responseText.substring(
                     jqXHR.responseText.indexOf("<body"),
                     jqXHR.responseText.indexOf("</body")
                 ));
+                document.dispatchEvent(new Event("pretix:async_task_replace_page:on_error"))
```
wiffbi: Naming convention for these custom events is to use dashes instead of underscores. I would even simplify the event’s name as I guess when it is async, we can assume it to be HTML-inlined.


```suggestion
                document.dispatchEvent(new Event("pretix:async-task-error"))
```

src/pretix/static/pretixbase/js/errors.js
```diff
@@ -1,5 +1,17 @@
-document.getElementById('goback').onclick =
-   function() {window.history.back()};
+const registerErrorLinkHandlers = (reloadAll = false) => {
```
wiffbi: Die ganze Datei liest sich für mich irgendwie komplizierter als es sein müsste, durch diese Indirektion, wann jetzt reloadAll gemacht wird. Kompakter und für beide Events und optional auch den reload-Button und IMHO klarer, wenn DOMContentLoaded && #goback, dann history.back, sonst reload. Und den optionalen Parameter true beim Reload kennt ja nur Firefox, muss der sein?

Ich kann hier im Review nicht die ganze Datei vorschlagen, aber so hätte ich das eingebaut – aber klar, Geschmackssache, was man lesbarer findet. :) Ansonsten hätte ich das aber mind. in ne IIFE gepackt, um window nicht mit registerErrorLinkHandlers zuzumüllen.

```suggestion
['DOMContentLoaded', 'pretix:async-task-error'].forEach(function (ev) {
  document.addEventListener(ev, function () {
    this.querySelectorAll('#goback, #reload').forEach(function(){
      this.addEventListener("click", ev == 'DOMContentLoaded' && this.id == 'goback' ? window.history.back : window.history.reload);
    })
  });
});
```

wiffbi (commented):
Looks good, just a minor naming suggestion and maybe some more condensed, but IMHO easier to grasp when a reload or history.back happens. But YMMV regarding code-style, so feel free.

src/pretix/static/pretixbase/js/errors.js
bockstaller: Hab die `this` Verwendungen durch Parameter bzw. `document` direkt ersetzt, sowie die eigentlichen Aktionsaufrufe in arrow-Funktionen gepackt.

src/pretix/static/pretixbase/js/errors.js
```diff
@@ -1,5 +1,11 @@
-document.getElementById('goback').onclick =
-   function() {window.history.back()};
-
-document.getElementById('reload').onclick =
- function() {window.location.reload(true)};
+['DOMContentLoaded', 'pretix:async-task-error'].forEach(function (ev) {
+  document.addEventListener(ev, function () {
+    document.querySelectorAll('#goback, #reload').forEach(function (element) {
+      const regularLoad = ev === 'DOMContentLoaded' && element.id === 'goback';
+      element.addEventListener('click', regularLoad
```
wiffbi, line 5: I'd call the boolean the other way around, what to do, not what happened. So:


```suggestion
      const goBack = ev === 'DOMContentLoaded' && element.id === 'goback';
      element.addEventListener('click', goBack
```

src/pretix/static/pretixbase/js/errors.js
```diff
@@ -1,5 +1,11 @@
...
+        ? () => window.history.back()
+        : () => window.location.reload()
```
wiffbi, line 7: Any reason why to not bind the functions directly?

```suggestion
        ? window.history.back
        : window.location.reload
```

wiffbi (approved):
Looks good and works for me! Maybe adapt the naming, other than that, ready to go!

src/pretix/static/pretixbase/js/errors.js
bockstaller: When I am binding the functions directly then it is only binding the plain `back` or `reload` functions. The reference to `window` as the `this`-context inside of them isn't passed on. 
Instead the `element` (I think, might be the eventListener) is used as the reference resulting in -> `Uncaught TypeError: 'reload' called on an object that does not implement interface Location.`

wiffbi: Ah, good to know!

PR #6296, opened 2026-06-17
EventMetaProperties: Restrict changing "allowed values" if already in use (Z#23235265)

luelista:


src/pretix/control/templates/pretixcontrol/organizers/property_edit.html
```diff
@@ -72,6 +72,9 @@ <h1>{% trans "Create a new property" %}</h1>
                                     </div>
                                     <div class="col-sm-6 col-md-4 col-lg-5">
                                         {% bootstrap_field form.key layout='inline' form_group_class="" %}
+                                        {% if form.key.help_text %}
+                                            <span class="help-block">{{ form.key.help_text|safe }}</span>
```
raphaelm, line 76: Wouldn't you rather use mark_save in python code? 😇

luelista:
```suggestion
                                            <span class="help-block">{{ form.key.help_text }}</span>
```

you bet i would 😳 

it's not needed at all though, not sure what i was doing here :/

PR #6297, opened 2026-06-19
use cookieretry only on presale event pages (Z#23236752)

bockstaller:
We are seeing a few reports of missing csrf cookies on event pages, the customers stumble upon them when adding a product to their cart.

Yesterday I was able to reproduce the behaviour by:

- opening a private safari tab
- navigating to a third party website
- entering the url to a pretix shop manually and omitting the last slash

this would lead to a request to `/organizer/shop` which we respond to with a `301` redirect to `/organizer/shop/`.
The request to `/organizer/shop/` would result in a `200` with all the right cookies in the headers, including the csrf cookie.
BUT that cookie wasn't stored in the cookiejar. 

This behaviour matches cases like these here: 
`django.security.csrf log Forbidden (CSRF cookie not set.): /{{event}}/cart/add` of which we are seeing quite a few.

~~This reproduction worked well enough on 18.06. to demo it in the dev call but a day later it doesn't work anymore.~~
I can repro again

The proposed solution here is a bit crude (but did work in testing [^1]) but loading an event index page without having a csrf token will break as soon as the user is adding something to their cart. 
So the tradeoff here is that adding the white flash of the reload is the less of an disruption than seeing the csrf error page after trying to add a product.

[^1]: on the 18.06. now I cannot get the system in the state anymore.

### DIFF
--- a/src/pretix/static/pretixpresale/js/csrfcookieretry.js
+++ b/src/pretix/static/pretixpresale/js/csrfcookieretry.js
@@ -1,0 +1,13 @@
+document.addEventListener("DOMContentLoaded", () => {
+  const COOKIE_NAME = "__Host-pretix_csrftoken";
+  const RELOAD_FLAG = "csrfReloadPerformed";
+
+  const hasCookie = document.cookie
+    .split("; ")
+    .some((c) => c.startsWith(COOKIE_NAME + "="));
+
+  if (!hasCookie && !sessionStorage.getItem(RELOAD_FLAG)) {
+    sessionStorage.setItem(RELOAD_FLAG, "1");
+    location.reload();
+  }
+});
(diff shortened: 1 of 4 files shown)


bockstaller:
For completeness: @pc-coholic also found this [comment](https://issues.chromium.org/issues/40508226#comment59) 
I wasn't able to get this solution to work locally

bockstaller:
The same can happen on order pages as well where we use a POST for the ticket downloads

src/pretix/static/pretixpresale/js/csrfcookieretry.js
```diff
@@ -0,0 +1,13 @@
...
+  const RELOAD_FLAG = "csrfReloadPerformed";
+
+  const hasCookie = document.cookie
+    .split("; ")
+    .some((c) => c.startsWith(COOKIE_NAME + "="));
+
+  if (!hasCookie && !sessionStorage.getItem(RELOAD_FLAG)) {
+    sessionStorage.setItem(RELOAD_FLAG, "1");
+    location.reload();
+  }
```
wiffbi, line 14: Should we remove the item from sessionStorage if hasCookie? We probably lose the session at the same time as the cookie, but maybe more robust to add something like 
```suggestion
  } else if (hasCookie && sessionStorage.getItem(RELOAD_FLAG)) {
      sessionStorage.removeItem(RELOAD_FLAG);
  }
```

wiffbi (approved):

PR #6305, opened 2026-06-23
Allow to convert customers to OIDC on login (Z#23237685)

wiffbi:


src/pretix/base/settings.py
```diff
@@ -181,6 +181,19 @@ def unserialize(cls, s):
             widget=forms.CheckboxInput(attrs={'data-display-dependency': '#id_settings-customer_accounts'}),
         )
     },
+    'customer_accounts_to_oidc': {
```
raphaelm: I think I'd prefer a flag on CustomerSSOProvider

src/pretix/presale/views/customer.py
```diff
@@ -864,11 +864,35 @@ def get(self, request, *args, **kwargs):
...
+                    if self.request.organizer.settings.customer_accounts_to_oidc:
+                        try:
+                            customer = self.request.organizer.customers.get(
+                                email=profile['email'],
+                            )
+                            customer.set_unusable_password()
+                            customer.provider = self.provider
+                            customer.external_identifier = str(profile['uid'])
+                            customer.identifier = identifier
+                            customer.is_active = True
```
raphaelm: This is dangerous, this re-enables a purposefully disabled account, I think this is shouldn't happen

src/pretix/presale/views/customer.py
```diff
@@ -864,11 +864,35 @@ def get(self, request, *args, **kwargs):
...
+                    # no race-condition, try to convert to oidc?
+                    if self.request.organizer.settings.customer_accounts_to_oidc:
+                        try:
+                            customer = self.request.organizer.customers.get(
+                                email=profile['email'],
+                            )
+                            customer.set_unusable_password()
+                            customer.provider = self.provider
+                            customer.external_identifier = str(profile['uid'])
+                            customer.identifier = identifier
```
raphaelm: The identifier of a customer account should be immutable in the best case, we need to discuss this

raphaelm (changes requested):

src/pretix/base/settings.py
wiffbi: Good idea, I changed to a boolean setting on CustomerSSOProvider

src/pretix/presale/views/customer.py
wiffbi: Yes, but currently login via SSO is done based on provider and identifer:

```
            customer = self.request.organizer.customers.get(
                provider=self.provider,
                identifier=identifier,
            )
```

Should we change this to use the external_identifier? Would basically be the same for current SSO-customers as identifier is a combination of `profile['uid']` (or external_identifier) and `self.provider.pk`.

src/pretix/presale/views/customer.py
wiffbi: Okay, removed it.

PR #6306, opened 2026-06-23
Fix event meta property handling when cloning across organizers (Z#23231419)

luelista:


raphaelm (approved):

luelista:
Related: https://github.com/pretix/pretix/pull/6106

PR #6326, opened 2026-07-01
Organizer calendar: Respect event_calendar_future_only (Z#23238776)

raphaelm:
We initially didn't do this for two reasons:

- Performance implications of calling the settings store for every event that shows up in the calendar. As of d43e85da, we need that anyways.
- Performance implications of filtering in Python except SQL but... it can't really be worse than not filtering at all.
- We don't easily know if it's valid for all events so we can't stop rendering the unused calendar rows. That's an acceptable issue for now, still better than nothing. We can always optimize later.

So we might as well implement it.

### DIFF
--- a/src/pretix/presale/views/organizer.py
+++ b/src/pretix/presale/views/organizer.py
@@ -650,6 +650,10 @@ def add_subevents_for_days(qs, before, after, ebd, timezones, sales_channel, eve
             if hide:
                 continue
 
+        if s.event_calendar_future_only:
+            if (se.date_to or se.date_from) < time_machine_now():
+                continue
+
         timezones.add(s.timezone)
         tz = ZoneInfo(s.timezone)
         datetime_from = se.date_from.astimezone(tz)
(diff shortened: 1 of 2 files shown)


src/pretix/presale/views/organizer.py
```diff
@@ -650,6 +650,10 @@ def add_subevents_for_days(qs, before, after, ebd, timezones, sales_channel, eve
             if hide:
                 continue
 
+        if s.event_calendar_future_only:
+            if (se.date_to or se.date_from) < time_machine_now():
```
wiffbi, line 654: Does this need more clever handling for „whole-day“ events? Maybe if no `se.date_to` is given, do a day/date-based instead of datetime-based comparison? Currently, if I have a subevent scheduled for today and using the default time of 00:00, the subevent gets filtered on the day it happens. Or base comparison on `show_times` setting whether to use time- or date-base comparison?

wiffbi (changes requested):

PR #6330, opened 2026-07-01
Drop line numbers from gettext .po files

raphaelm:
Knowing what file a string comes from is useful, but the line number is less useful and changes a lot, causing very unreadable diffs of translation files. I propose we drop them and only include the file names

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -6,8 +6,8 @@ localecompile:
 	./manage.py compilemessages
 
 localegen:
-	./manage.py makemessages --keep-pot --ignore "pretix/static/npm_dir/*" $(LNGS)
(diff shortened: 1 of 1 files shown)


wiffbi (approved):
Love the idea, had that situation recently in a plugin. Should we add this to https://github.com/pretix/pretix-plugin-cookiecutter/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/Makefile and to existing plugins as well?

raphaelm:
> Love the idea, had that situation recently in a plugin. Should we add this to [pretix/pretix-plugin-cookiecutter@`master`/%7B%7Bcookiecutter.repo_name%7D%7D/Makefile](https://github.com/pretix/pretix-plugin-cookiecutter/blob/master/%7B%7Bcookiecutter.repo_name%7D%7D/Makefile) and to existing plugins as well?

Sure!

PR #6332, opened 2026-07-02
Install django-querytagger

raphaelm:
https://github.com/pretix/django-querytagger


wiffbi:
See https://github.com/pretix/django-querytagger/pull/1 to fix failing tests

PR #6337, opened 2026-07-03
Drop all csp_ignore statements from core (except stripe, for now)

raphaelm:


src/pretix/base/middleware.py
```diff
@@ -266,6 +266,22 @@ def _merge_csp(a, b):
...
 
 class SecurityMiddleware(MiddlewareMixin):
+    SAFE_TYPES = (
+        # CSP policies are only used for:
+        # - HTML and SVG in top-level contexts
+        # - SVG or JS Workers delivered in embedded contexts
+        # See: https://www.w3.org/TR/CSP2/#which-policy-applies
+        # Therefore, we can save bandwidth on not including our (sometimes huge) policy
+        # on API responses or CSS. We do however include it with other types as a precaution
+        # (whitelist instead of blacklist) and we also do not whitelist JavaScript in
```
luelista, line 276:
```suggestion
        # (allowlist instead of blocklist). We also do not allowlist JavaScript in case
```

src/pretix/base/middleware.py
```diff
@@ -277,6 +293,11 @@ def process_response(self, request, resp):
         # https://github.com/pretix/pretix/issues/765
         resp['P3P'] = 'CP=\"ALL DSP COR CUR ADM TAI OUR IND COM NAV INT\"'
 
+        if "Content-Type" in resp and resp["Content-Type"].split(";")[0] in self.SAFE_TYPES:
+            if 'Content-Security-Policy' in resp:
+                del resp['Content-Security-Policy']
+            return resp
+
         if not getattr(resp, '_csp_ignore', False):
```
luelista, line 301:
```suggestion
        if self._needs_csp(resp):
```

src/pretix/base/middleware.py
luelista, line 307:
```suggestion
    def _needs_csp(self, resp):
        if "Content-Type" in resp and resp["Content-Type"].split(";")[0] in self.SAFE_TYPES:
            return False
        if getattr(resp, '_csp_ignore', False):
            return False
        return True

```

src/pretix/control/views/pdf.py
```diff
@@ -70,7 +70,6 @@ def get(self, request, *args, **kwargs):
         if 'placeholders' in request.GET:
             return self.get_placeholders_help(request)
         resp = super().get(request, *args, **kwargs)
-        resp._csp_ignore = True
```
luelista, line 73: so this wasn't necessary at all?

luelista (approved):

src/pretix/control/views/pdf.py
raphaelm: let me better double-check it…

raphaelm: It works in my browser except for one feature.

Research says:

- fabricjs [no longer needs unsafe-eval](https://github.com/fabricjs/fabric.js/pull/6296) as of 4.0, we are on 5.x
- pdfjs [no longer needs unsafe-inline style if we upgrade](https://github.com/mozilla/pdf.js/pull/21416) (but its working fine for me already), but will need [wasm-unsafe-eval if we do](https://github.com/mozilla/pdf.js/issues/18457)
- ajv [requires unsafe-eval or extra steps](https://ajv.js.org/security.html). will look into that

raphaelm: Okay, pdfjs would be a major chance since its only available as a module. But it seems to work just fine. Pushed the ajv change

src/pretix/base/middleware.py
luelista: did you maybe accidentally overwrite these when force pushing?

PR #6341, opened 2026-07-03
Improve handling of logout during asynchronous tasks (Z#23238978)

raphaelm:


src/pretix/static/pretixbase/js/asynctask.js
```diff
@@ -113,6 +113,12 @@ function async_task_check_error(jqXHR, textStatus, errorThrown) {
     "use strict";
     var respdom = $(jqXHR.responseText);
     var c = respdom.filter('.container');
+    if (jqXHR.status === 401 && jqXHR.getResponseHeader("X-Login-Url")) {
+      // Append location.hash outside the next parameter so it is not unexpectedly sent to the server
+      // The browser will keep it in the redirect.
```
luelista: but then the hash will apply on the login page (where it most likely won't do anything useful), but be lost when submitting the login form? as long as we don't rely on the hash being a 'secret' from the server, i think it makes more sense to keep it in the `next` parameter (if we wanted to rely on that, we'd still need extra logic on the login form to 'forward' the hash into the form action).

(same comment applies to the two identical snippets below)

luelista (changes requested):

luelista (approved):
apart from that, lgtm

src/pretix/static/pretixbase/js/asynctask.js
raphaelm: You're right. I wanted to prepare for hashes containing something secret some day, but that makes sense.

PR #6347, opened 2026-07-06
CheckInListPDF export: remove double html escaping with PlainTextParagraph (Z#23239571)

pajowu:


src/pretix/plugins/checkinlists/exporters.py
```diff
@@ -409,30 +411,33 @@ def get_story(self, doc, form_data):
             company = op.company or (op.addon_to.company if op.addon_to else '') or iac
             if company:
                 if name:
-                    name += "<br/>"
+                    name += "\n"
```
raphaelm, line 414: I think this will no longer work – without PlainTextParagraph, reportlab is not rendering the newline, so we need to manually convert it back to <br/> later, won't we?

pajowu: oops, i tested this but with a too long company name so it seemed like it worked. fixed now

raphaelm (approved):

PR #6356, opened 2026-07-07
Log RequestId in Celery tasks

luelista:


src/pretix/celery_app.py
```diff
@@ -28,6 +28,10 @@
 os.environ.setdefault("DJANGO_SETTINGS_MODULE", "pretix.settings")
 logger = logging.getLogger(__name__)
 
+from pretix.settings import LOGGING
+
+LOGGING['formatters']['default']['format'] = '[%(asctime)s: %(levelname)s/%(processName)s] RequestId=%(request_id)s %(name)s %(module)s %(message)s'
```
raphaelm, line 33: Hmm, if we go for shared handlers, we should also use  shared format and integrate processName into our regular format, shouldn't we?

PR #6357, opened 2026-07-07
Order change: Fix crash if addon has hidden_if_item_available (Z#23236380)

luelista:


raphaelm (approved):
Please remember that pretix-seating needs an import change

PR #6359, opened 2026-07-07
Remove leftovers from contact url in quick setup

sweenu:
Those should be deleted if we don't include this option in the quick setup.


raphaelm:
Thanks!

PR #6363, opened 2026-07-08
Remove more usages of |safe

luelista:
- html_info from payment providers
   The change of `payment_control_render` and `refund_control_render` seems safe to me, as all our payment provider plugins either return a `template.render(...)` result, which is a SafeString, or a plain string which shouldn't contain any HTML.
- `data-replace-with-qr` QR codes
  I think we should switch from `<script type='text/plain'>` to `<script type='application/json'>`, because for text/plain there is no well-defined escaping mechanism
- some other `<script>{{ some_json_string|safe }}</script>` occurences

### DIFF
--- a/src/pretix/control/templates/pretixcontrol/order/cancel.html
+++ b/src/pretix/control/templates/pretixcontrol/order/cancel.html
@@ -26,7 +26,7 @@
         {% if form.cancellation_fee %}
             {% if fee %}
                 {% with fee|money:request.event.currency as f %}
-                    <p>{% blocktrans trimmed with fee="<strong>"|add:f|add:"</strong>"|safe %}
+                    <p>{% blocktrans trimmed with fee=f|wrap_in:"strong" %}
                         The configured cancellation fee for a self-service cancellation would be {{ fee }} for this
                         order, but for a cancellation performed by you, you need to set the cancellation fee here:
                     {% endblocktrans %}</p>
--- a/src/pretix/control/templates/pretixcontrol/order/cancel.html
+++ b/src/pretix/control/templates/pretixcontrol/order/cancel.html
@@ -2,6 +2,7 @@
 {% load i18n %}
 {% load bootstrap3 %}
 {% load money %}
+{% load wrap_in %}
 {% block title %}
     {% trans "Cancel order" %}
 {% endblock %}
--- a/src/pretix/control/templates/pretixcontrol/order/index.html
+++ b/src/pretix/control/templates/pretixcontrol/order/index.html
@@ -903,7 +903,7 @@
                                             <tr>
                                                 <td colspan="1"></td>
                                                 <td colspan="5">
-                                                    {{ p.html_info|safe }}
+                                                    {{ p.html_info }}
                                                     {% if staff_session %}
                                                         <p>
                                                             <a href="" class="btn btn-default btn-xs admin-only" data-expandpayment data-id="{{ p.pk }}">
@@ -1018,7 +1018,7 @@
                                                             </dl>
                                                         {% endif %}
                                                         {% if r.html_info %}
-                                                            {{ r.html_info|safe }}
+                                                            {{ r.html_info }}
                                                         {% endif %}
                                                         {% if staff_session %}
                                                             <p>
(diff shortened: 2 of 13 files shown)


src/pretix/control/templates/pretixcontrol/order/cancel.html
```diff
@@ -26,7 +26,7 @@ <h1>
         {% if form.cancellation_fee %}
             {% if fee %}
                 {% with fee|money:request.event.currency as f %}
-                    <p>{% blocktrans trimmed with fee="<strong>"|add:f|add:"</strong>"|safe %}
+                    <p>{% blocktrans trimmed with fee=f|wrap_in:"strong" %}
```
raphaelm, line 30: missing {% load %}?

src/pretix/control/templates/pretixcontrol/order/index.html
```diff
@@ -903,7 +903,7 @@ <h3 class="panel-title">
                                             <tr>
                                                 <td colspan="1"></td>
                                                 <td colspan="5">
-                                                    {{ p.html_info|safe }}
+                                                    {{ p.html_info }}
```
raphaelm, line 906: Makes sense, but does this require changes to *all* payment providers? Or are we safe if they all use render()?

src/pretix/plugins/stripe/static/pretixplugins/stripe/pretix-stripe.js
```diff
@@ -323,7 +323,7 @@ $(function () {
             }
         }
     } else if ($("#stripe_payment_intent_next_action_redirect_url").length) {
-        let payment_intent_next_action_redirect_url = $.trim($("#stripe_payment_intent_next_action_redirect_url").html());
+        let payment_intent_next_action_redirect_url = JSON.parse($("#stripe_payment_intent_next_action_redirect_url").html());
```
raphaelm, line 326: Is this tested? Just because we had some "how many decodes does this need" confusions already :)

src/pretix/static/pretixcontrol/js/ui/main.js
```diff
@@ -667,12 +666,11 @@ var form_handlers = function (el) {
 
     el.find("script[data-replace-with-qr]").each(function () {
         var $div = $("<div>");
-        var qrText = this.getAttribute("type") === "application/json" && this.textContent.startsWith('"') ?
-            JSON.parse(this.textContent) : $(this).html();
         $div.insertBefore($(this));
+				var text = (this.getAttribute("type") || "").indexOf("json") !== -1 ? JSON.parse($(this).html()) : $(this).html();
```
raphaelm: I think I did the same thing differently on what's currently on main.

What I don't like is that this is a really bug to search if we ever want to QR-code an actual JSON object in string representation and not a JSON-wrapped string, because then one would probably set the same content type for correctness…

raphaelm (changes requested):

src/pretix/plugins/stripe/static/pretixplugins/stripe/pretix-stripe.js
luelista: yes, tested and it worked :)

src/pretix/control/templates/pretixcontrol/order/cancel.html
luelista: oops, yes. fixed.

src/pretix/control/templates/pretixcontrol/order/index.html
luelista: This seems safe to me, as all our payment provider plugins either return a template.render(...) result, which is a SafeString, or a plain string which shouldn't contain any HTML (only found one actual case of that, `OffsettingProvider` in core (luckily, order codes can't contain special characters 😇), and a few cases of empty string).

src/pretix/static/pretixcontrol/js/ui/main.js
luelista: ah, that's a good workaround, i'll rebase the pr then

raphaelm (approved):

PR #6375, opened 2026-07-13
Fix #6117 -- fr_CA locale renders wrong event times on PDF tickets

### LINKED ISSUE #6117: fr-ca locale renders wrong event times on PDF tickets
### Problem and impact

When generating PDF tickets with the `fr-ca` locale, event times are rendered incorrectly.

In my case, the PDF placeholders `event_begin` and `event_begin_time` show values like:

- `2026-08-23 00 12 20`
- `00 12 20`

instead of the expected Canadian French format, for example:

- `2026-08-23 00 h 20`
- `00 h 20`

I see this both on my development instance and on production, and it seems to have started after updating pretix one or two times recently.

This affects customer-facing ticket PDFs, so attendees receive tickets with obviously wrong date/time formatting.

This looks related to the Django 5.2 upgrade path.

In a local reproduction on April 27, 2026, I could reproduce this on pretix commit `4c0c775baa7ed2688e0983bca58dd1f4bc55c93f` with Django `5.2.12`.

I did not find a pretix issue specifically for this bug, only the broader Django 5.2 tracking issue:
https://github.com/pretix/pretix/issues/5425

### Expected behaviour

With locale `fr-ca`, event times on PDF tickets should render using the normal Canadian French time format, with a literal `h` separator.

For example, an event starting at 00:20 should be rendered as:

- `event_begin_time`: `00 h 20`
- `event_begin`: `2026-08-23 00 h 20`

and not as `00 12 20` / `2026-08-23 00 12 20`.

### Steps to reproduce

1. Use locale `fr-ca`.
2. Create or edit an event with a start time such as `00:20`.
3. Add `event_begin` or `event_begin_time` to a PDF ticket layout.
4. Generate the PDF ticket.
5. Observe that the time is rendered as `00 12 20` instead of `00 h 20`.

### Screenshots

_No response_

### Link

_No response_

_No response_

_No response_

### Version

_No response_

TowyTowy:
Fixes #6117.

**Problem.** With the `fr-ca` locale (French + region Canada), ticket-PDF placeholders like `event_begin` / `event_begin_time` render times wrongly — a begin time of `00 h 20` comes out as `00 12 20` on customer-facing PDFs.

**Cause.** pretix pins `Django==5.2.*`. Django's `fr_CA` locale defines its time formats with an unescaped `h` (`TIME_FORMAT = "H h i"`, etc.). In Django's date mini-language `h` is the 12-hour-clock hour, which is `12` at midnight — so `date_format(dt, "TIME_FORMAT")` yields `00 12 20`. pretix's PDF placeholders (`src/pretix/base/pdf.py`) go through exactly this path. Django 6.0 (commit `b67a36ec`) fixed it by escaping the `h`, but it was not backported to `stable/5.2.x`, so we're affected until we move to Django 6 — matching the "override the file in the meantime" direction discussed in the issue.

**Fix.** Add a `pretix.helpers.formats.fr_CA` override (same mechanism already used for `en_CA`, `en_NZ`, …) escaping the literal `h` in `TIME_FORMAT`, `DATETIME_FORMAT`, `SHORT_DATETIME_FORMAT`. Removable once we require Django >= 6.0.

**Test.** Regression test in `src/tests/helpers/test_i18n.py`, fail-before/pass-after (verified against Django 5.2: default fr_CA renders `00 12 20`, the override renders `00 h 20`).

**AI disclosure (per the project's AI policy).** Prepared with Claude Code (Anthropic). AI was used to locate the affected code path, cross-check Django's upstream fix, and draft the override + regression test; the change and its reasoning were reviewed and understood by a human before submission.

### DIFF
--- a/src/pretix/helpers/formats/fr_CA/__init__.py
+++ b/src/pretix/helpers/formats/fr_CA/__init__.py
@@ -1,0 +1,21 @@
+#
+# This file is part of pretix (Community Edition).
+#
+# Copyright (C) 2014-2020  Raphael Michel and contributors
+# Copyright (C) 2020-today pretix GmbH and contributors
+#
+# This program is free software: you can redistribute it and/or modify it under the terms of the GNU Affero General
+# Public License as published by the Free Software Foundation in version 3 of the License.
+#
+# ADDITIONAL TERMS APPLY: Pursuant to Section 7 of the GNU Affero General Public License, additional terms are
+# applicable granting you additional permissions and placing additional restrictions on your usage of this software.
+# Please refer to the pretix LICENSE file to obtain the full terms applicable to this work. If you did not receive
+# this file, see <https://pretix.eu/about/en/license>.
+#
+# This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied
+# warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU Affero General Public License for more
+# details.
+#
+# You should have received a copy of the GNU Affero General Public License along with this program.  If not, see
+# <https://www.gnu.org/licenses/>.
+#
--- a/src/pretix/helpers/formats/fr_CA/formats.py
+++ b/src/pretix/helpers/formats/fr_CA/formats.py
@@ -1,0 +1,34 @@
+#
+# This file is part of pretix (Community Edition).
+#
+# Copyright (C) 2014-2020  Raphael Michel and contributors
+# Copyright (C) 2020-today pretix GmbH and contributors
+#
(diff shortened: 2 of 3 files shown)


TowyTowy:
Withdrawing this for now — I'm not able to complete the CLA at the moment, so I don't want to leave it sitting unmergeable. Leaving the fix here for the team in case it's useful: the `fr_CA` time formats need the literal `h` escaped (`H\\h\ i`) via a `pretix.helpers.formats.fr_CA` override, matching Django 6.0's fix (b67a36ec) which wasn't backported to 5.2 — that's what makes `00 h 20` render as `00 12 20` on PDFs (#6117). Thanks!

PR #6387, opened 2026-07-16
 Improve CSP handling in vite integration; refactor CSP handling into helper functions (Z#23240534) 

luelista:


src/pretix/base/templatetags/vite.py
```diff
@@ -234,10 +234,16 @@ def vite_importmap(context):
...
+    json_string = json.dumps({"imports": imports})
+
+    # Calculate hash and add it to the CSP info in the request, so that will be merged into the
+    # response header later on by the middleware
     request = context.get('request')
     if request:
-        request.csp_nonce = nonce
+        csp_hash = calculate_csp_hash(json_string)
+        add_to_response_csp_via_request(request, {
+            'style-src': [csp_hash],
```
luelista, line 247:
```suggestion
```

I took this from the original implementation, but I think this actually isn't necessary? At least when I tested it, it works fine without, and I don't see how it would do anything, as the importmap is a script and not a style.

src/pretix/base/middleware.py
```diff
@@ -337,13 +367,6 @@ def _build_csp(self, request, resp):
             h['style-src'] += ["'unsafe-inline'"]
             h['connect-src'] += ["http://localhost:5173", "ws://localhost:5173"]
 
-        if hasattr(request, 'csp_nonce'):
-            nonce = f"'nonce-{request.csp_nonce}'"
-            h['script-src'].append(nonce)
-            if not settings.VITE_DEV_MODE:
-                # can't have 'unsafe-inline' and nonce at the same time
-                h['style-src'].append(nonce)
```
raphaelm, line 395: Do we maybe even want a test for "response at /control/ has no nonce" with a comment to not re-introduce the issue?

src/pretix/base/middleware.py
```diff
@@ -282,6 +305,9 @@ class SecurityMiddleware(MiddlewareMixin):
         # but at the moment it does not seem to be an issue to just send it.
     )
 
+    def process_request(self, request):
+        request.csp_to_merge = {}
```
raphaelm: If external components should interact with this through add_to_response_csp_via_request, maybe prefix it with _?

raphaelm (approved):

src/pretix/base/middleware.py
luelista: I'll add that separately as we should discuss whether we want to extend that test to most frontend pages too

PR #6390, opened 2026-07-16
Orders: Fix N+1 query issues on order detail and change page

raphaelm:
There is one I couldn't fix: Loading the list of tax rules for every select box on the OrderChange page. Unfortunately, Django has a cache-breaking .all() in ModelChoiceField and that would need nasty patching that didn't feel worth it

### DIFF
--- a/src/pretix/control/views/orders.py
+++ b/src/pretix/control/views/orders.py
@@ -81,7 +81,7 @@ from pretix.base.i18n import language
 from pretix.base.models import (
     CachedFile, CachedTicket, Checkin, GiftCard, Invoice, InvoiceAddress, Item,
     ItemVariation, LogEntry, Order, QuestionAnswer, Quota,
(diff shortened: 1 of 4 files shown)


src/pretix/control/views/orders.py
```diff
@@ -1981,24 +1983,45 @@ def add_fee_formset(self):
     def items(self):
         return self.request.event.items.prefetch_related('variations', 'tax_rule').all()
 
+    @cached_property
+    def tax_rules(self):
+        return self.request.event.tax_rules.all()
+
```
pajowu: What does this do?

pajowu (approved):

src/pretix/control/views/orders.py
raphaelm: previous attempt to fix a n+1 I couldn't fix probably…

```suggestion
```

PR #6393, opened 2026-07-17
Fix performance issues in download reminder

raphaelm:


src/pretix/base/services/orders.py
```diff
@@ -1488,83 +1487,96 @@ def send_expiry_warnings(sender, **kwargs):
...
-            o = Order.objects.select_for_update(of=OF_SELF).get(pk=o.pk)
-            if o.download_reminder_sent:
-                # Race condition
-                continue
-            positions = list(o.positions_with_tickets)
-            if not positions:
+        qs = event.orders.annotate(
+            first_date=Min('all_positions__subevent__date_from')
+        ).filter(
+            Q(first_date__isnull=True) | Q(first_date__gte=today),
```
wiffbi: Would it make sense to annotate and filter that qs only if event.has_subevents? Feels needlessly complex for plain events without subevents.

src/pretix/base/services/orders.py
```diff
@@ -48,12 +48,12 @@
 from django.conf import settings
 from django.core.cache import cache
 from django.core.exceptions import ValidationError
-from django.db import transaction
+from django.db import models, transaction
```
wiffbi, line 51: Change not needed as only IntegerField is used from models and that is imported already.

```suggestion
from django.db import transaction
```

src/pretix/base/services/orders.py
```diff
@@ -1488,83 +1487,96 @@ def send_expiry_warnings(sender, **kwargs):
...
+    ).annotate(
+        reminder_days=Subquery(
+            Event_SettingsStore.objects.filter(
+                object=OuterRef('id'),
+                key='mail_days_download_reminder'
+            ).annotate(
+                val=Case(
+                    When(value="None", then=None),
+                    default=Cast(F("value"), output_field=models.IntegerField()),
+                )
```
wiffbi: For me the query (and actual SQL) looks more complex than needed with `Case`. Excluding results with value `None` creates a null-entry in the subquery as well so that filtering on is_null still works. 
```suggestion
            ).exclude(
                value="None"
            ).annotate(
                val=Cast(F("value"), output_field=IntegerField())
```

When using a subquery in annotate, django orm cannot use the column-name in filtering, but adds a second subquery in the WHERE-clause for is_null-filtering. Might not make a difference thanks to the query-optimizer of the DB, but an alternative might be to just check for exists with something like:

```
.filter(Exists(Event_SettingsStore.objects.filter(
            object=OuterRef('id'),
            key='mail_days_download_reminder'
        ).exclude(value="None")
))
```

This of course would result in additional queries lateron to the event’s settings-store to retrieve the actual value of mail_days_download_reminder, but these being simple index-based lookups should be fast.

wiffbi (approved):
Just minor comments, looks good!

PR #6402, opened 2026-07-20
Add central helper for ssrf address protection

pajowu:


wiffbi (dismissed):
lgtm, just need to run isort to fix order of imports.

PR #6408, opened 2026-07-22
Banktransfer: Notify incomplete payments after manual assignment (Z#23241009)

pajowu:


raphaelm (approved):

raphaelm:
read, not tested, but looks good

PR #6410, opened 2026-07-22
Improve performance of overpayment checks (Z#23241035)

raphaelm:
Needs https://github.com/raphaelm/django-scopes/pull/38 for testing


wiffbi (approved):
Works as intended. Putting these 4 queries in a `with scopes_disabled():` gives the same result, but with the new helper, IMHO it is a bit nicer to read and helps to not unintentionally add other queries to the with-block with scopes disabled.

PR #6412, opened 2026-07-23
Log view: New filter form

raphaelm:
The view for event logs gets very slow (or unusable) for large events.

Turns out, the slowest query on the site is often not actually getting the logs, but getting the list of team mates to filter at the top. :exploding_head: 

So let's drop that and while we're at it, do a less weird filter form.

I replaced this select with an email input field, which is a drop in functionality but I see no proper replacement. Getting a list from the teams does not really make sense, since logs also might include past team members. Doing a good typeahead support is hard since we don't usually leak email addresses to all backend users if their team mates have natural names. So email input it is for now.

### DIFF
--- a/src/pretix/control/forms/filter.py
+++ b/src/pretix/control/forms/filter.py
@@ -39,6 +39,7 @@ from urllib.parse import urlencode
 from django import forms
 from django.apps import apps
 from django.conf import settings
+from django.contrib.contenttypes.models import ContentType
 from django.db.models import (
     Count, Exists, F, Max, Model, OrderBy, OuterRef, Q, QuerySet,
 )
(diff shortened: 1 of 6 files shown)


src/pretix/control/forms/filter.py
```diff
@@ -3007,3 +3008,91 @@ def filter_qs(self, qs):
...
+        elif fdata.get('source') == 'device':
+            qs = qs.filter(device__isnull=False)
+        elif fdata.get('source') == 'customer':
+            qs = qs.filter(user__isnull=True, device__isnull=True)
+
+        if fdata.get('device'):
+            qs = qs.filter(device_id=fdata['device'].pk)
+
+        if fdata.get('action_type'):
+            qs = qs.filter(action_type=fdata['action_type'])
```
luelista: might be useful to do some type of list field / comma separated field, and then use `action_type__in` here, so we can build URLs searching for multiple action types at once (alternatively, `action_type__startswith`?)

luelista (commented):
>Doing a good typeahead support is hard since we don't usually leak email addresses to all backend users if their team mates have natural names

but previously we did leak email addresses (only those who did anything that caused a log action) in this dropdown list? 

we could add a typeahead accessible only to users with access to team config? but maybe not a priority right now

src/pretix/control/forms/filter.py
luelista: e.g.
```suggestion
            qs = qs.filter(action_type__in=fdata['action_type'].split(','))
```

luelista (approved):

PR #6419, opened 2026-07-24
Check-in history: Avoid slow JOINs in query

raphaelm:
On a test dataset, this was a speedup from 18s to 900ms

### DIFF
--- a/src/pretix/control/views/checkin.py
+++ b/src/pretix/control/views/checkin.py
@@ -484,10 +484,17 @@ class CheckinListView(EventPermissionRequiredMixin, PaginationMixin, ListView):
 
     def get_queryset(self):
         qs = Checkin.all.filter(
-            list__event=self.request.event,
-        ).select_related(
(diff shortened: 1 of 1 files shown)


src/pretix/control/views/checkin.py
```diff
@@ -484,10 +484,17 @@ class CheckinListView(EventPermissionRequiredMixin, PaginationMixin, ListView):
 
     def get_queryset(self):
         qs = Checkin.all.filter(
-            list__event=self.request.event,
-        ).select_related(
-            'position', 'position__order', 'position__item', 'position__variation', 'position__subevent'
+            list_in=self.request.event.checkin_lists.values_list('id', flat=True),
```
pajowu:
```suggestion
            list_id__in=self.request.event.checkin_lists.values_list('id', flat=True),
```

pajowu (changes requested):

pajowu (approved):

PR #6421, opened 2026-07-24
Event-dashboard: Lazy-load warnings

raphaelm:
I don't like this, it's not pretty, but we have no quick way of fixing the overpaid orders check otherwise and a slow query on the event dashboard feels really bad, because even if you don't wanna see the dashboard ou are always going through it in navigation and that can get very annoying very quickly.


src/pretix/control/templates/pretixcontrol/event/dashboard_partial_warnings.html
```diff
@@ -0,0 +1,48 @@
+{% load i18n %}
```
pajowu, line 1: the overpaid section is missing here

```suggestion
{% load i18n %}
{% if has_overpaid_orders %}
    <div class="alert alert-warning">
        {% blocktrans trimmed %}
            This event contains <strong>overpaid orders</strong>, for example due to duplicate payment attempts.
            You should review the cases and consider refunding the overpaid amount to the user.
        {% endblocktrans %}
        <a href="{% url "control:event.orders" event=request.event.slug organizer=request.event.organizer.slug %}?status=overpaid"
                class="btn btn-primary">{% trans "Show overpaid orders" %}</a>
    </div>
{% endif %}
```

pajowu (changes requested):

PR #6426, opened 2026-07-27
Cart Manager: Round custom price to currency when adding

pajowu:


src/pretix/base/services/cart.py
```diff
@@ -916,6 +917,8 @@ def add_new_items(self, items: List[dict]):
                 if custom_price > 99_999_999_999:
                     raise CartError(error_messages['price_too_high'])
 
+                custom_price = round_decimal(custom_price, currency=self.event.currency)
```
raphaelm, line 920: This should be fairly easy to add a test for 🥺

raphaelm (changes requested):

Issue #1033, opened 2018-09-24
Allow specific sites for iframes


cromefire:
Currently I cannot see any option where I can whitelist domains that are allowed to embed `pretix` into an iframe.

To make it that possible that domains have to be inserted into the `Content-Security-Policy` (at `frame-src`) and if there are two or more whitelisted pages, `X-Frame-Options` has to be removed.

Currently you can only alter this by a downstream server.

cromefire:
If you point me to the code that sets the headers and tell me, how to read the options, I can implement that.

raphaelm:
pretix is not supposed to be put into iframes for security reasons. If you want to embed a ticket shop into a website, please use our [widget](https://docs.pretix.eu/en/latest/user/events/widget.html), which goes to great lengths to make sure there's no compromise in security.

cromefire:
But the widget doesn't seem to support organizers or collections and we need to host a number of events that change quite often

cromefire:
> pretix is not supposed to be put into iframes for security reasons.

That's why I speak of a whitelist, the embedding site is secured.

raphaelm:
Yes, unfortunately this is not yet implemented (see also #878 which is similar)

raphaelm:
> That's why I speak of a whitelist, the embedding site is secured.

I understand that, however, it's not that simple.

> To make it that possible that domains have to be inserted into the Content-Security-Policy (at frame-src) and if there are two or more whitelisted pages, X-Frame-Options has to be removed.

``frame-src`` is not involved here, you mean ``frame-ancestors``. ``frame-ancestors`` is not set, so it currently allows everything. The only thing currently constraining embedding is the ``X-Frame-Options`` header.

To allow a whitelist of two or more, we'd need to drop ``X-Frame-Options`` as you said, and could enforce the whitelist only via ``frame-ancestors``, which is supported by fewer browsers, so it is a compromise in security. Even with a whitelist of only one, ``X-Frame-Options: ALLOW-FROM`` doesn't work in Chrome or Safari (according to [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/X-Frame-Options)), so it's not a really suitable option either.

The other way to implement a whitelist would be to do something hacky with referers and iframes, which only properly works on the first request and probably breaks for people with strong privacy settings.

[...]

PR #6438, opened 2026-07-31
Allow `frame-ancestors` in custom CSP header

fronbasal:
This PR permits the specification of the [`frame-ancestors`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Content-Security-Policy/frame-ancestors) CSP directive in the respective configuration field (`pretix.csp_additional_header`).

Current behavior: raises ValueError and thus HTTP 500 after update.

### DIFF
--- a/src/pretix/base/middleware.py
+++ b/src/pretix/base/middleware.py
@@ -280,8 +280,8 @@ def _parse_csp(header):
 
 
 VALID_CSP_DIRECTIVES = [
-    "child-src", "connect-src", "default-src", "fenced-frame-src", "font-src", "form-action", "frame-src", "img-src",
-    "manifest-src", "media-src", "object-src", "prefetch-src", "report-uri", "script-src", "script-src-elem",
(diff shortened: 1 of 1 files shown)


raphaelm:
Fair fix, but, why would you want to embed the entire system incl the backend in a framey that sounds rather unsafe 👀

(I mean, the whole global csp setting is probably turning out to be a bad idea)

fronbasal:
That's totally understandable and a very valid point! 

My motivation was that I found this as a solution in #1033 a while ago, and a recent update introduced the allowlisting in the CSP header, thus breaking that functionality for us. I'd be happy to assist with a scoped rewrite but I think there are too many open questions in the requirements for me to just "go at it" ;) 

After re-reading #1033 it's dawning on me that I might have gotten the CSP configuration for embedding the Pretix Widget from a self-hosted installation totally wrong and I didn't even have to touch the global CSP settings at all ^^

Happy to see a proper patch for this in a future update; thank you for the effort!

PR #6440, opened 2026-08-03
Presale: Frontend enforcement of max item per order

stranck:
This is a followup of my old PR #4662:
I've generalized the max-item-count frontend checks used in the add-on page and made them work also on the first shop page. This makes the checks for "Maximum number of items per order" work client-side as well and not validated only after an user clicks for adding products to their cart.

Contrary to what I have done in the previous PR, now I don't disable the buttons, but, instead, make a tooltip appear like in the add-on selection. I've also fixed a small bug with that tooltip: Previously if, in a category, all products but one had 0 quantity, multiple selection per add-on category was permitted, the quantity for that single selected product already reached `c.max_count` and the user presses "+", no tooltip was displayed. This has been fixed by setting the upper bound javascript side, and this works also for the first shop page.

This PR changed the name of the attributes used for this checks to a more generalized one, and added a new one, `data-count-in-cart`: If present on a fieldset, it makes the check count as well the number of products from that fieldset already in the cart. 
Finally, it also hides the "+" from the cart's selection if we have already reached the maximum items per order

### DIFF
--- a/src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
+++ b/src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
@@ -7,6 +7,25 @@
 {% load getitem %}
 {% load eventsignal %}
 {% load rich_text %}
+<fieldset data-max-count="{{ max_items }}" data-multi-allowed data-count-in-cart>
+<p class="count-desc" id="count-desc">
+    {% if max_items == 1 %}
+        {% blocktrans trimmed %}
+            You can choose only one product.
+        {% endblocktrans %}
+    {% else %}
+        {% blocktrans trimmed with max_count=max_items %}
+            You can choose up to {{ max_count }} products.
+        {% endblocktrans %}
+    {% endif %}
+    {% if cart.itemcount > 0 %}
+        {% blocktrans trimmed count count=cart.itemcount %}
+            You have already made one selection.
+        {% plural %}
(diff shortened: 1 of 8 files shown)


src/pretix/presale/templates/pretixpresale/event/fragment_product_list.html
```diff
@@ -36,12 +55,12 @@
...
-                                   data-title="{{ item.name|force_escape|force_escape }}"
+                                data-title="{{ item.name|force_escape|force_escape }}"
                                         {# Yes, double-escape to prevent XSS in lightbox #}
-                                   data-lightbox="{{ item.id }}"
-                                   aria-label="{% blocktrans trimmed with item=item.name %}Show full-size image of {{ item }}{% endblocktrans %}">
+                                data-lightbox="{{ item.id }}"
+                                aria-label="{% blocktrans trimmed with item=item.name %}Show full-size image of {{ item }}{% endblocktrans %}">
                                     <img src="{{ item.picture|thumb:'60x60^' }}"
-                                         alt="{{ item.name }}"/>
+                                        alt="{{ item.name }}"/>
```
luelista: please avoid doing these whitespace-only changes

stranck:
Good morning
I've just updated my branch so there are no merge conflicts anymore. Any updates on this? Sorry for the bump

PR #6447, opened 2026-08-04
CachedFileField: Bind file access to session (Z#23242301)

raphaelm:


src/pretix/control/forms/__init__.py
```diff
@@ -172,7 +172,9 @@ def value_from_datadict(self, data, files, name):
         from ...base.models import CachedFile
         v = super().value_from_datadict(data, files, name)
         if v is None and data.get(name + '-cachedfile'):  # An explicit "[x] clear" would be False, not None
-            return CachedFile.objects.filter(id=data[name + '-cachedfile']).first()
+            v = CachedFile.objects.filter(id=data[name + '-cachedfile']).first()
+            if self.request and not v.allowed_for_session(self.request):
```
luelista, line 176: We could also forbid using any CachedFile objects which are bound to a session (session_key not None) in case no request is set, right?

```suggestion
            if (self.request and not v.allowed_for_session(self.request)) or (not self.request and v.session_key):
```

or

```suggestion
            if not v.allowed_for_session(self.request):
```
(simpler logic but would raise AttributeError in case `v.session_key and not self.request`)

luelista (approved):
AsyncFormView

PR #6448, opened 2026-08-04
Password reset: Prevent race condition that uses token twice (Z#23242132)

raphaelm:
Thanks to NH Limon

### DIFF
--- a/src/pretix/presale/views/customer.py
+++ b/src/pretix/presale/views/customer.py
@@ -54,6 +54,7 @@ from pretix.base.models import Customer, InvoiceAddress, Order, OrderPosition
 from pretix.base.services.mail import mail
 from pretix.base.settings import PERSON_NAME_SCHEMES
 from pretix.base.signals import customer_created, customer_signed_in
+from pretix.helpers import OF_SELF
 from pretix.helpers.compat import CompatDeleteView
 from pretix.helpers.http import redirect_to_url
 from pretix.multidomain.models import KnownDomain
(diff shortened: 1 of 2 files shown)


src/pretix/presale/views/customer.py
```diff
@@ -280,6 +281,16 @@ def get_success_url(self):
 
     def form_valid(self, form):
         with transaction.atomic():
```
luelista, line 283: could we avoid the recheck by just wrapping the dispatch in `with transaction.atomic():` or adding a decorator there?

raphaelm: We could wrap dispatch, but we'd also need to do the select_for_update, and we probably wouldn't want to do it on a GET request, so… I don't think that's better?

src/pretix/presale/views/customer.py
```diff
@@ -280,6 +281,16 @@ def get_success_url(self):
...
+            # Re-check token in transaction to prevent race condition
+            try:
+                self.customer = Customer.objects.select_for_update(of=OF_SELF).get(pk=self.customer.pk)
+            except Customer.DoesNotExist:
+                messages.error(self.request, _('You clicked an invalid link.'))
+            else:
+                if not TokenGenerator().check_token(self.customer, self.request.GET.get('token', '')):
+                    messages.error(self.request, _('You clicked an invalid link.'))
+                    return HttpResponseRedirect(self.get_success_url())
```
luelista, line 292:
```suggestion
            self.customer = Customer.objects.select_for_update(of=OF_SELF).get(pk=self.customer.pk)
            if not TokenGenerator().check_token(self.customer, self.request.GET.get('token', '')):
                return HttpResponseRedirect(self.get_success_url())
```
- Customer.DoesNotExist seems unlikely enough that we could just let it fail (and it would have failed before as well)
- only legit case where check_token could fail would be 'double click' on submit button, in which we wouldn't need an error message as it would have actually succeeded

luelista (approved):

PR #6450, opened 2026-08-04
Add Stripe per-method fee overrides

fronbasal:
This PR proposes adding support for overriding the base and variable fee for every Stripe method to account for more expensive methods such as Klarna.

Use case: we want incentivize credit card payments, but cannot individually customize fees for methods.

This is deliberately still phrased as a draft, as I'd love to figure out whether this is viable before adding e.g. translations, further tests, etc.

I tried to keep the edits minimal and chose not to abstract the fees logic away into it's own module, if the changes bloat the payment.py too much, I'd be happy to refactor it. 

AI declaration: the test file is entirely AI generated and wasn't checked thoroughly - obviously needs a manual rewrite if viable. I used Cursor with Gemini Pro 3.1 (Thinking) for sketching out the viability and fully reviewed and customized the output. Will further manually rewrite if proven viable, of course. I'm not unfamiliar with the plugin codebase so this should hopefully not smell 😅 

Thank you for taking the time to review the proposal :)

### DIFF
--- a/src/pretix/plugins/stripe/payment.py
+++ b/src/pretix/plugins/stripe/payment.py
@@ -39,7 +39,7 @@ import urllib.parse
 import zoneinfo
 from collections import OrderedDict
 from datetime import datetime
-from decimal import Decimal
+from decimal import ROUND_HALF_UP, Decimal
 from json import JSONDecodeError
 
 import stripe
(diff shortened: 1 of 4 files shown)


raphaelm:
Honestly, I'm mostly not happy about the dozens of settings fields this adds to the already very long list. The "proper" way to refactor the entire system into a proper split between payment providers and payment methods, which I don't see prioritized any time soon.

Given that payment fees for most methods are not even legal in Europe, either by law or by the payment company's terms, payment fees are not really one of our priorities.

Therefore, I'd likely not merge it.

PR #6454, opened 2026-08-05
Money representation in templates: Allow more precision

raphaelm:
When rendering money in templates, we used to have the following logic:

- When the decimal places fit the currency, render with Babel
- When they don't, e.g. we stored 123.67 JPY, even though there are no fractional Yens, render without Babel with a custom format, but render the fractional Yens because we'd rather *show* wrong data and make the bug obvious than hide it.

However, we only did that up to a prevision of two places, we never showed more. This is still sufficient for core pretix, but we have plugins that need to operate in fractional cents. Also, we CAN render everything through babel for consistent formatting.

There is one **risk**: This might cause weird results on SQLite. Since SQLite has no concept of precise decimal math, results of in-SQL computations can sometimes experience floating point errors and show with A LOT of decimal palces. This used to be invisible since the UI performed the rounding. With this PR – not any more. We'll need to see how annoying it is, but it should only affect development mode.

This PR also fixes a bug in tax_rate_format that for some reason did not do what it was supposed to do, even though I tested it back then, weird. Might even be a Python version thing?

### DIFF
--- a/src/pretix/base/templatetags/money.py
+++ b/src/pretix/base/templatetags/money.py
@@ -48,19 +48,12 @@ def money_filter(value: Decimal, arg='', hide_currency=False):
         raise ValueError("No currency passed.")
     arg = arg.upper()
 
-    places = settings.CURRENCY_PLACES.get(arg, 2)
-    rounded = value.quantize(Decimal('1') / 10 ** places, ROUND_HALF_UP)
-    if places < 2 and rounded != value:
-        # We display decimal places even if we shouldn't for this currency if rounding
-        # would make the numbers incorrect. If this branch executes, it's likely a bug in
-        # pretix, but we won't show wrong numbers!
-        if hide_currency:
-            return floatformat(value, "2g")
-        else:
-            return '{} {}'.format(arg, floatformat(value, "2g"))
+    currency_places = settings.CURRENCY_PLACES.get(arg, 2)
+    required_places = -value.normalize().as_tuple().exponent
+    render_places = max(currency_places, required_places)
 
     if hide_currency:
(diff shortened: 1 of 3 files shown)


src/pretix/base/templatetags/money.py
```diff
@@ -48,29 +48,37 @@ def money_filter(value: Decimal, arg='', hide_currency=False):
...
     except UnknownLocaleError:
         locale = "en"
 
     try:
-        return format_currency(value, arg, locale=locale)
+        return format_currency(
+            value,
+            arg,
+            locale=locale,
+            # We only allow Babel to restrict the digits to the digits by the currency if this does not remove any
```
pajowu, line 68:
```suggestion
            # We only allow Babel to restrict the digits to the digits defined by the currency if this does not remove any
```

src/pretix/base/templatetags/money.py
```diff
@@ -91,11 +99,18 @@ def tax_rate_format(number):
     """
     Display a Decimal to its significant decimal places, used for tax rates.
     """
-    assert isinstance(number, Decimal)
+    if isinstance(number, (float, int, str)):
+        number = Decimal(number)
+    if number is None:
+        number = Decimal('0.00')
+    if not isinstance(number, Decimal):
+        if number == '':
```
pajowu, line 107: when can that happen as we already check and convert strings above?

pajowu (approved):

src/pretix/base/templatetags/money.py
pajowu, line 39:
```suggestion
    if isinstance(value, (float, int, str)):
```

pajowu (approved):

PR #6455, opened 2026-08-05
Device list: Show online status

raphaelm:


src/pretix/control/templates/pretixcontrol/organizers/devices.html
```diff
@@ -118,13 +118,26 @@ <h3 class="panel-title">{% trans "Filter" %}</h3>
                                 </td>
                             {% endif %}
                             <td>
+                                {% if d.last_seen and d.last_seen.is_recent %}
+                                    <span class="fa fa-check-circle text-success"
```
pc-coholic: I would remove the check and just to the colored circles - looks cleaner to me.

src/pretix/control/templates/pretixcontrol/organizers/devices.html
```diff
@@ -118,13 +118,26 @@ <h3 class="panel-title">{% trans "Filter" %}</h3>
...
+                                          data-toggle="tooltip"
+                                          title="{% if d.last_seen %}{% blocktrans with time=d.last_seen.last_seen|date:"SHORT_DATETIME_FORMAT" %}Last seen: {{ time }}{% endblocktrans %}{% else %}{% trans "No recent contact" %}{% endif %}"
```
pc-coholic: Duplicate with below...

I think we could just merge both blocks into a single one with a `text-{% if d.last_seen and d.last_seen.is_recent %}success{% else %}danger{% endif %}` within the circle's `span`?

pc-coholic (approved):

src/pretix/control/templates/pretixcontrol/organizers/devices.html
raphaelm: nope, acessbility wins

PR #6456, opened 2026-08-06
Paypal2: handle incomming capture webhooks (Z#23240966)

bockstaller:
We have to handle webhooks differently depending on the incoming resource type.
In this case "capture" has its own id. We need to pry out the checkout-order id in order to look up the order object.

Additionally we store the payment info asap

### DIFF
--- a/src/pretix/plugins/paypal2/views.py
+++ b/src/pretix/plugins/paypal2/views.py
@@ -361,10 +361,13 @@ def webhook(request, *args, **kwargs):
         return HttpResponse("Not interested in this resource type", status=200)
 
     # Retrieve the Charge ID of the refunded payment
-    if event_json['resource_type'] == 'refund':
+    if event_json['resource_type'] == 'checkout-order':
+        payloadid = event_json['resource']['id']
+    elif event_json['resource_type'] == 'refund' or event_json['resource_type'] == 'capture':
         payloadid = get_link(event_json['resource']['links'], 'up')['href'].split('/')[-1]
     else:
-        payloadid = event_json['resource']['id']
+        logger.exception('Unknown webhook ressource received. Event data: %s' % str(event_json))
+        return HttpResponse("Unable to handle ressource", status=200)
 
     refs = [payloadid]
     if event_json['resource'].get('supplementary_data', {}).get('related_ids', {}).get('order_id'):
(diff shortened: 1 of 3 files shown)


src/pretix/plugins/paypal2/views.py
```diff
@@ -361,10 +361,13 @@ def webhook(request, *args, **kwargs):
...
     # Retrieve the Charge ID of the refunded payment
-    if event_json['resource_type'] == 'refund':
+    if event_json['resource_type'] == 'checkout-order':
+        payloadid = event_json['resource']['id']
+    elif event_json['resource_type'] == 'refund' or event_json['resource_type'] == 'capture':
         payloadid = get_link(event_json['resource']['links'], 'up')['href'].split('/')[-1]
     else:
-        payloadid = event_json['resource']['id']
+        logger.exception('Unknown webhook ressource received. Event data: %s' % str(event_json))
+        return HttpResponse("Unable to handle ressource", status=200)
```
ser8phin:
```suggestion
        logger.exception('Unknown webhook resource received. Event data: %s' % str(event_json))
        return HttpResponse("Unable to handle resource", status=200)
```
typos :)

bockstaller: danke dir

ser8phin (approved):

src/pretix/plugins/paypal2/views.py
```diff
@@ -361,10 +361,13 @@ def webhook(request, *args, **kwargs):
...
     # Retrieve the Charge ID of the refunded payment
-    if event_json['resource_type'] == 'refund':
+    if event_json['resource_type'] == 'checkout-order':
+        payloadid = event_json['resource']['id']
+    elif event_json['resource_type'] == 'refund' or event_json['resource_type'] == 'capture':
         payloadid = get_link(event_json['resource']['links'], 'up')['href'].split('/')[-1]
     else:
-        payloadid = event_json['resource']['id']
+        logger.exception('Unknown webhook resource received. Event data: %s' % str(event_json))
+        return HttpResponse("Unable to handle resource", status=200)
```
luelista: is it correct to return status=200 here?

bockstaller: It is correct but was a duplication of logic. I've cleaned that up here now
https://github.com/pretix/pretix/pull/6456/changes/4ae628737d2a65ab2de21c4cabea85e75cb15c03

PR #6458, opened 2026-08-07
Display invoice destination in the (re)transmit button tooltip (Z#23241800)

luelista:
image

### DIFF
--- a/src/pretix/base/invoicing/email.py
+++ b/src/pretix/base/invoicing/email.py
@@ -82,6 +82,14 @@ class EmailTransmissionType(TransmissionType):
             }
         return {}
 
+    def describe_invoice_destination(self, invoice):
+        info = (invoice.invoice_to_transmission_info or {})
+        if info.get("transmission_email_address"):
+            recipient = info["transmission_email_address"]
+        else:
+            recipient = invoice.order.email
+        return _("Via Email to {recipient}").format(recipient=recipient)
+
 
 @transmission_providers.new()
 class EmailTransmissionProvider(TransmissionProvider):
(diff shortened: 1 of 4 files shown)


src/pretix/base/invoicing/email.py
```diff
@@ -82,6 +82,14 @@ def form_data_to_transmission_info(self, form_data: dict) -> dict:
             }
         return {}
 
+    def describe_invoice_destination(self, invoice):
+        info = (invoice.invoice_to_transmission_info or {})
+        if info.get("transmission_email_address"):
+            recipient = info["transmission_email_address"]
+        else:
+            recipient = invoice.order.email
+        return _("Via Email to {recipient}").format(recipient=recipient)
```
raphaelm:
```suggestion
        return _("Via email to {recipient}").format(recipient=recipient)
```

src/pretix/control/templates/pretixcontrol/order/index.html
```diff
@@ -357,7 +357,7 @@ <h3 class="panel-title">
                                               action="{% url "control:event.order.retransmitinvoice" event=request.event.slug organizer=request.event.organizer.slug code=order.code id=i.pk %}">
                                             {% csrf_token %}
                                             <button class="btn btn-default btn-xs" data-toggle="tooltip"
-                                                title="{{ i.transmission_type_instance.verbose_name }}">
+                                                title="{{ i.transmission_type_destination_description }}">
```
raphaelm, line 360: please do a quick test if this is XSS-safe if e.g. someone would be able to insert quotes into e.g. a Peppol ID or email address, tooltips are weird around escaping

raphaelm (approved):

src/pretix/control/templates/pretixcontrol/order/index.html
luelista: should be fine as it's the same as e.g. here: https://github.com/pretix/pretix/blob/7fe31634e62baf62584570fc04097209d7b1e0d4/src/pretix/control/templates/pretixcontrol/items/index.html#L85 (where no xss is possible)

seems to be only a problem if data-toggle="tooltip_html" is used

PR #6459, opened 2026-08-07
Allow to set payment term per sales channel

raphaelm:


src/pretix/base/settings.py
```diff
@@ -1023,7 +1023,7 @@ def unserialize(cls, s):
         'form_class': forms.IntegerField,
         'serializer_class': serializers.IntegerField,
         'write_permission': 'event.settings.payment:write',
-        'form_kwargs': dict(
+        'form_kwargs': lambda suffix="", parent=0: dict(
```
luelista:
```suggestion
        'form_kwargs': lambda suffix="", parent=1: dict(
```

wiffbi (approved):

PR #6471, opened 2026-08-11
Order-level questions

luelista:
To do:
- [x] Test it doesnt break pretixPOS
    - order-level questions are completely ignored (not displayed, cannot be answered, having required ones does not block checkout)
- [x] Test with resellers
    - works as expected, questions can/must be answered (answering questions is optional if the according flag is set)
- [x] Test with event series
- [x] Hide behind feature flag
- [x] Fix tests
- [ ] Add some tests for new features

Separate PRs:
- [ ] API for reading / changing answers to order-level questions
- [ ] (probably with questionnaires: pretixPOS support)

Not to do:
- API for manipulating order-level questions (will be resolved with questionnaires)


src/pretix/base/services/orders.py
```diff
@@ -1113,6 +1114,17 @@ def _create_order(event: Event, *, email: str, positions: List[CartPosition], no
...
+    if cart_id:
+        try:
+            session = CheckoutSession.objects.get(event=event, cart_id=cart_id)
+        except CheckoutSession.DoesNotExist:
+            pass
+        else:
+            for answ in session.answers.all():
+                answ.order = order
+                answ.checkoutsession = None
+                answ.save()
```
wiffbi: Wouldn’t a `update` on the queryset be sufficient as all entries will be updated with the same value? E.g.:


```suggestion
            session.answers.update(order=order, checkoutsession=None)
```

src/pretix/base/forms/questions.py
```diff
@@ -636,13 +637,356 @@ def __init__(self, *args, **kwargs):
...
+)
+
+
+def get_fake_questions(settings):
+    def b(s):
+        return s == 'True'
+    fq = []
+    sqo = settings.system_question_order
+
+    if b(settings.get('attendee_names_asked', 'True')):
```
wiffbi: I think hierarkeys `as_type` and `default` params might be better. That way, you can completely remove that local function `b`. Apply to the other places accordingly:
```suggestion
    if settings.get('attendee_names_asked', as_type=bool, default=False):
```

src/pretix/base/forms/questions.py
```diff
@@ -636,13 +637,356 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
 
+FakeQuestion = namedtuple(
+    'FakeQuestion', 'system_question label position required help_text'
+)
+
+
+def get_fake_questions(settings):
```
wiffbi: Hm, I somehow wonder, whether „fake“ is the right wording. It is not a real question as in one manually created by the user, so it is indeed somewhat fake. But for me, fake feels more like „bogus“ as-in a question that is not really used. What about the wording „virtual“? So it is distinct to „real“ questions, but it is still a question that is meaningful. Do you know what I mean? I also would name that helper-function accordingly.

```suggestion
VirtualQuestion = namedtuple(
    'VirtualQuestion', 'system_question label position required help_text'
)


def get_virtual_attendee_questions(settings):
```

src/pretix/base/forms/questions.py
```diff
@@ -636,13 +637,356 @@ def __init__(self, *args, **kwargs):
...
+        fq.append(FakeQuestion('zipcode', _('ZIP code'), sqo.get('zipcode', 0), b(settings.get('attendee_addresses_required')), ''))
+        fq.append(FakeQuestion('city', _('City'), sqo.get('city', 0), b(settings.get('attendee_addresses_required')), ''))
+        fq.append(FakeQuestion('country', _('Country'), sqo.get('country', 0), b(settings.get('attendee_addresses_required')), ''))
+    return fq
+
+
 class BaseQuestionsForm(forms.Form):
     """
-    This form class is responsible for asking order-related questions. This includes
+    This form class is responsible for asking order- and ticket-related questions.
```
wiffbi: I stumbled upon the „order- and ticket-related“, which technically is right, but I wondered whether using „or“ makes it more clear, that it is a base class used for either order or ticket-related questions.
```suggestion
    This is the base form class responsible for asking order- or ticket-related questions.
```

src/pretix/base/forms/questions.py
```diff
@@ -636,13 +637,356 @@ def __init__(self, *args, **kwargs):
...
+
+        # Strip False answers to required yes/no questions even if all_optional is set, as our data model assumes that
+        # required yes/no questions can only be answered with yes
+        for q in question_cache.values():
+            if q.required and q.type == Question.TYPE_BOOLEAN:
+                if 'question_%d' % q.pk in d and d['question_%d' % q.pk] is False:
+                    d['question_%d' % q.pk] = None
+
+
+class OrderLevelQuestionsForm(BaseQuestionsForm):
```
wiffbi, line 949: I’d use the same naming convention as `TicketQuestionsForm`
```suggestion
class OrderQuestionsForm(BaseQuestionsForm):
```

src/pretix/base/forms/questions.py
```diff
@@ -636,13 +637,356 @@ def __init__(self, *args, **kwargs):
...
+        Takes two additional keyword arguments:
+
+        :param checkoutsession: The checkout session the form should be for
+        :param order: The order the form should be for
+        :param event: The event this belongs to
+        """
+        request = kwargs.pop('request', None)
+        checkoutsession = self.checkoutsession = kwargs.pop('checkoutsession', None)
+        order = self.order = kwargs.pop('order', None)
+        container = checkoutsession or order
```
wiffbi: Do we really need a local var for this. And should we prefer order over checkoutsession? I guess either one of them should be None? If so, would it make sense to even add an `assert not (self.checkoutsession and self.order)`?


```suggestion
        self.checkoutsession = kwargs.pop('checkoutsession', None)
        self.order = kwargs.pop('order', None)
        container = self.checkoutsession or self.order
```

src/pretix/base/forms/widgets.py
wiffbi:
```suggestion
        self.container = kwargs.pop('container')
```

wiffbi:
```suggestion
            return self.FakeFile(value, self.container, self.event, self.answer)
```

wiffbi (changes requested):
First round of review. I stumbled upon an error on WidgetUploadedFile (see comments) and system-level attendee questions on personalized products did not work when I tested. Could you look into this?

src/pretix/base/forms/questions.py
luelista: yes, i copied that function from https://github.com/pretix/pretix/pull/6465/changes#diff-5db2c98abedaedc3a297332b211e16c7872469a6b86432cee5e037451cc5df94R20-R41 and didn't notice that hierarkey behaves differently in migrations and in normal code

src/pretix/base/forms/questions.py
```diff
@@ -636,13 +637,354 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
 
+FakeQuestion = namedtuple(
+    'FakeQuestion', 'system_question label position required help_text'
+)
+
+
+def get_fake_questions(settings):
```
wiffbi: As discussed, we stick to the „fake“ wording, but would it make sense to call this 
```suggestion
def get_fake_attendee_questions(settings):
```

src/pretix/base/forms/questions.py
```diff
@@ -636,13 +637,354 @@ def __init__(self, *args, **kwargs):
...
+        fq.append(FakeQuestion('attendee_email', _('Attendee email'), sqo.get('attendee_email', 0), settings.attendee_emails_required, ''))
+
+    if settings.attendee_company_asked:
+        fq.append(FakeQuestion('company', _('Company'), sqo.get('company', 0), settings.attendee_company_required, ''))
+
+    if settings.attendee_addresses_asked:
+        fq.append(FakeQuestion('street', _('Street'), sqo.get('street', 0), settings.attendee_addresses_required, ''))
+        fq.append(FakeQuestion('zipcode', _('ZIP code'), sqo.get('zipcode', 0), settings.attendee_addresses_required, ''))
+        fq.append(FakeQuestion('city', _('City'), sqo.get('city', 0), settings.attendee_addresses_required, ''))
+        fq.append(FakeQuestion('country', _('Country'), sqo.get('country', 0), settings.attendee_addresses_required, ''))
```
wiffbi: In my tests, this at least needed -1 for all FakeQuestion (as already done in `reorder_questions`) otherwise they were sorted behind the first custom question, which I did not have initially (I actually did not sort the questions at all in my test event). To make sure we always get the default ordering, we might do:

```suggestion
        fq.append(FakeQuestion('attendee_name_parts', _('Attendee name'), sqo.get('attendee_name_parts', -7), settings.attendee_names_required, ''))

    if settings.attendee_emails_asked:
        fq.append(FakeQuestion('attendee_email', _('Attendee email'), sqo.get('attendee_email', -6), settings.attendee_emails_required, ''))

    if settings.attendee_company_asked:
        fq.append(FakeQuestion('company', _('Company'), sqo.get('company', -5), settings.attendee_company_required, ''))

    if settings.attendee_addresses_asked:
        fq.append(FakeQuestion('street', _('Street'), sqo.get('street', -4), settings.attendee_addresses_required, ''))
        fq.append(FakeQuestion('zipcode', _('ZIP code'), sqo.get('zipcode', -3), settings.attendee_addresses_required, ''))
        fq.append(FakeQuestion('city', _('City'), sqo.get('city', -2), settings.attendee_addresses_required, ''))
        fq.append(FakeQuestion('country', _('Country'), sqo.get('country', -1), settings.attendee_addresses_required, ''))
```

src/pretix/base/forms/questions.py
```diff
@@ -664,115 +1006,164 @@ def __init__(self, *args, **kwargs):
...
-                    help_text='' if initial else _('If you keep this empty, the ticket will be valid starting at the time of purchase.'),
-                    required=bool(initial),
-                    initial=pos.requested_valid_from or initial,
-                    widget=DatePickerWidget(attrs),
-                    validators=([MaxDateValidator(max_date.date())] if max_date else []) + [MinDateValidator(min_date.date())]
-                )
+            self.fields['requested_valid_from'] = self.build_requested_valid_from_field(event, pos, item)
+
+        if item.ask_attendee_data:
+            questions = questions + get_fake_questions(event.settings)
```
wiffbi: See comment above. Currently only used here, so we might move into the class?

```suggestion
            questions = questions + get_fake_attendee_questions(event.settings)
```

src/pretix/base/forms/questions.py
luelista: Oh thanks, missed that, should be fixed now by 1c13a16f7935bc102bf30856b0938f8d9f9504d8. 
It seems to be sufficient to just change the order in which the lists are concatenated (system_questions + user_questions instead of user_questions + system_questions) to fix the ordering (that was also the way it's done currently).

src/pretix/base/forms/questions.py
luelista: I'll rather change it the other way round (TicketQuestionsForm -> TicketLevelQuestionsForm), as we also have an OrderQuestionsViewMixin, where the 'Order' prefix means something different (belonging to a submitted order as opposed to a cart; not belonging to the order as opposed to an order position)

src/pretix/base/forms/questions.py
luelista: Even better, we don't need to distinguish here between order and checkoutsession at all, allowing us to get rid of bunch of if's

src/pretix/api/views/item.py
```diff
@@ -475,7 +475,7 @@ class QuestionViewSet(ConditionalListView, viewsets.ModelViewSet):
     write_permission = 'event.items:write'
 
     def get_queryset(self):
-        return self.request.event.questions.prefetch_related('options').all()
+        return self.request.event.questions.filter(container_type=Question.ContainerType.ORDERPOSITION).prefetch_related('options').all()
```
wiffbi: I know, we wanted to limit the API to old-style orderposition-based questions for now, but when adding a question dependency onto an order-level question, this query fails to respond with the question I selected as a dependency. Should we risk to just output order-level questions as well? We hide the actual creation behind a feature-flag and do not allow to update/create order-level questions through the API at the moment. So this should be somewhat safe to output all questions here.

src/pretix/presale/templates/pretixpresale/event/checkout_questions.html
```diff
@@ -68,6 +68,18 @@ <h3 class="panel-title">
...
+                    <summary class="panel-heading">
+                        <h3 class="panel-title">
+                            <strong>{% trans "Additional order information" %}</strong>
+                        </h3>
+                    </summary>
+                    <div class="panel-body">
+                        {% bootstrap_form order_questions_form layout="checkout" %}
+                    </div>
+                </details>
+            {% endif %}
```
wiffbi, line 82: Hm, looks a bit crowded adding another details-box when I maybe just want to add one or two additional questions. I would treat this similar to questions added through the signal `contact_form_fields`, which get listet below the system questions for email and optionally phone. We probably need to think about changing the summary „Contact information“ to something more general like „Order information“ or „Your information“. Of course we would need to move the boxes on the confirm step into the contact info box as well. The actual order-paid-page already has it inside the box „Your information“. :)

src/pretix/presale/templates/pretixpresale/event/checkout_questions.html
wiffbi, line 24: What about adding the order-level questions after here instead?

src/pretix/api/views/item.py
luelista: Fixed by adding an undocumented `container_type` parameter to the API endpoint

src/pretix/presale/templates/pretixpresale/event/checkout_questions.html
luelista: Mh, not sure.. In some views we don't have the contact info box (pretixpresale/event/order_modify.html), and the contact info is a different view actually. If you have a good idea how to do this consistently feel free to add it to the PR, otherwise I'd keep it like this for now. Or would you just move it into the contact info box during checkout, and keep it separate in the modify views?

luelista: When we add the questionnaires, we could even have the possibility to configure per-questionnaire in which place the questions should appear (contact form, separate box, maybe even invoice address box)

wiffbi (approved):

PR #6472, opened 2026-08-11
BasePaymentProvider & PayPal2: allow to cancel pending payments on a per payment basis (Z#23240966)

bockstaller:
**BasePaymentProvider:**
- extends BasePaymentProvider with `abort_pending_payment_allowed(self, payment: OrderPayment)` to allow per payment decisions if it can be canceled or not
 
**PayPal2:**
- moves payments into pending when we receive "APPROVED" by PayPal
- moves payments into pending when we receive "PENDING_REVIEW" by PayPal, should be a NoOp because we should™️ get an "APPROVED" update beforehand
- adds two new options to allow/forbid retrying PayPal payments that are currently pending, defaults match the current behaviour of not moving PayPal payments into `PENDING`
- adjusts the messaging we show to customers, to inform the about the double charge risk if they aren't patient

### DIFF
--- a/src/pretix/base/payment.py
+++ b/src/pretix/base/payment.py
@@ -338,8 +338,9 @@ class BasePaymentProvider:
         """
         return False
 
-    def payment_abort_pending_allowed(self, payment: OrderPayment) -> bool:
+    def _payment_abort_pending_allowed(self, payment: OrderPayment) -> bool:
         """
+        Experimental: This might change during upcomming releases.
         Whether or not a user can abort a payment in pending state to switch to another
         payment method. This returns ``self.abort_pending_allowed`` by default which is
         no guarantee that aborting a pending payment can never happen, it just hides the
@@ -1035,7 +1036,7 @@ class BasePaymentProvider:
         """
 
         if payment.state == OrderPayment.PAYMENT_STATE_PENDING:
-            if not self.payment_abort_pending_allowed(payment):
+            if not self._payment_abort_pending_allowed(payment):
                 raise PaymentException(_(
                     "This payment is already being processed and cannot be canceled any more."
                 ))
--- a/src/pretix/base/payment.py
+++ b/src/pretix/base/payment.py
@@ -1037,7 +1037,7 @@ class BasePaymentProvider:
         if payment.state == OrderPayment.PAYMENT_STATE_PENDING:
             if not self.payment_abort_pending_allowed(payment):
                 raise PaymentException(_(
-                    "This payment is already being processed and can not be canceled any more."
+                    "This payment is already being processed and cannot be canceled any more."
                 ))
 
         payment.state = OrderPayment.PAYMENT_STATE_CANCELED
--- a/src/pretix/base/payment.py
+++ b/src/pretix/base/payment.py
@@ -333,6 +333,15 @@ class BasePaymentProvider:
         """
         return False
 
+    def abort_pending_payment_allowed(self, payment) -> bool:
+        """
+        Whether this specific payment can be aborted in pending state or not.
+        If both ``abort_pending_allowed`` and ``abort_pending_payment_allowed`` are True,
+        then the frontend button will be shown to users.
+        This returns ``True`` by default, which makes implementing this method optional.
+        """
+        return True
+
     @property
     def requires_invoice_immediately(self):
         """
@@ -1019,10 +1028,12 @@ class BasePaymentProvider:
         On success, you should set ``payment.state = OrderPayment.PAYMENT_STATE_CANCELED`` (or call the super method).
         On failure, you should raise a PaymentException.
         """
-        if payment.state == OrderPayment.PAYMENT_STATE_PENDING and not self.abort_pending_allowed:
-            raise PaymentException(_(
-                "This payment is already being processed and can not be canceled any more."
-            ))
+
+        if payment.state == OrderPayment.PAYMENT_STATE_PENDING:
+            if self.abort_pending_allowed and self.abort_pending_payment_allowed(payment):
+                raise PaymentException(_(
+                    "This payment is already being processed and can not be canceled any more."
+                ))
 
         payment.state = OrderPayment.PAYMENT_STATE_CANCELED
         payment.save(update_fields=['state'])
--- a/src/pretix/base/payment.py
+++ b/src/pretix/base/payment.py
@@ -1030,7 +1030,7 @@ class BasePaymentProvider:
         """
 
         if payment.state == OrderPayment.PAYMENT_STATE_PENDING:
-            if self.abort_pending_allowed and self.abort_pending_payment_allowed(payment):
+            if not (self.abort_pending_allowed and self.abort_pending_payment_allowed(payment)):
                 raise PaymentException(_(
                     "This payment is already being processed and can not be canceled any more."
                 ))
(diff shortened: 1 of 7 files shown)


src/pretix/plugins/paypal2/templates/pretixplugins/paypal2/pending.html
```diff
@@ -6,9 +6,12 @@
...
             payment succeeding on a second or third attempt. You can also try other payment methods, if available.
         {% endblocktrans %}</div>
-    {% else %}
+{% elif stuck_in_compliance %}
         <p>{% blocktrans trimmed %}
-            Our attempt to execute your payment via PayPal has failed. Please try again or contact us.
+            Your payment is being processed by PayPal. This takes longer than usual. You can wait until PayPal
+            acknowledges the payment or you can try paying again with this or another payment method.
+            This might result in you being charged twice in case PayPal allows your initial payment attempt.
+            Please contact us, to resolve this case.
```
pajowu:
```suggestion
            Please contact us to resolve this case.
```

src/pretix/plugins/paypal2/payment.py
```diff
@@ -193,6 +193,31 @@ def settings_form_fields(self):
...
                  )
              )),
+            ('allow_retries_during_compliance_hold',
+             forms.BooleanField(
+                 label=_('Allow further payments during compliance hold'),
+                 help_text=_(
+                     'PayPals fraud prevention might block processing of individual payments for a considerable amount '
+                     'of time. The payment is marked as "pending" during this time window. You can allow your customers to '
+                     'start another payment attempts during that window. This might result in overpayment of orders if the'
+                     'original payment is approved.'
```
pajowu: i found "might result in overpayment" unintuitive to understand, maybe something like that instead:

```suggestion
                     'start another payment attempts during that window. This might result in them being charged twice if the'
                     'original payment is approved.'
```

src/pretix/plugins/paypal2/payment.py
```diff
@@ -857,22 +896,34 @@ def log_payment_duration(payment: OrderPayment):
             logger.info('{}: {} - paypal payment processing time'.format(str(payment.global_id), str(duration)))
 
     def payment_pending_render(self, request, payment) -> str:
+        stuck_in_compliance = False
         retry = True
+
+        if payment.state == OrderPayment.PAYMENT_STATE_PENDING:
+            retry = self.abort_pending_payment_allowed(payment) and self.abort_pending_allowed
+
```
pajowu: doesn't payment_pending_render automatically imply state PENDING? also wrong method name

```suggestion
        retry = self.payment_abort_pending_allowed(payment)
```

src/pretix/plugins/paypal2/payment.py
```diff
@@ -857,22 +896,34 @@ def log_payment_duration(payment: OrderPayment):
...
                     payment.info
                     and payment.info_data['purchase_units'][0]['payments']['captures'][0]['status'] == 'PENDING'
             ):
-                retry = False
+                stuck_in_compliance = True
         except (KeyError, IndexError):
             pass
 
+        try:
+            if payment.info and payment.info_data['status'] == "APPROVED":
```
pajowu:
```suggestion
            if payment.info_data.get('status') == "APPROVED":
```

src/pretix/plugins/paypal2/payment.py
```diff
@@ -857,22 +896,34 @@ def log_payment_duration(payment: OrderPayment):
...
+        stuck_in_compliance = False
         retry = True
+
+        if payment.state == OrderPayment.PAYMENT_STATE_PENDING:
+            retry = self.abort_pending_payment_allowed(payment) and self.abort_pending_allowed
+
         try:
             if (
                     payment.info
                     and payment.info_data['purchase_units'][0]['payments']['captures'][0]['status'] == 'PENDING'
```
pajowu: is only checking the first capture here correct? in the webhook view all captures are checked

src/pretix/plugins/paypal2/payment.py
```diff
@@ -517,6 +542,18 @@ def format_price(self, value):
 
     @property
     def abort_pending_allowed(self):
+        return True
+
+    def abort_pending_payment_allowed(self, payment) -> bool:
+        if not self.settings.get('allow_retries_during_compliance_hold', as_type=bool, default=True):
+            return False
+
+        if payment.info_data.get('create_time', False):
```
pajowu, line 547: is this field guaranteed to exist at this point? or could this break aborting a failed/abandoned payment somehow?

src/pretix/base/payment.py
```diff
@@ -330,9 +330,23 @@ def abort_pending_allowed(self) -> bool:
...
         return False
 
+    def payment_abort_pending_allowed(self, payment: OrderPayment) -> bool:
+        """
+        Whether or not a user can abort a payment in pending state to switch to another
+        payment method. This returns ``self.abort_pending_allowed`` by default which is
+        no guarantee that aborting a pending payment can never happen, it just hides the
+        frontend button to avoid users accidentally committing double payments.
+        """
+        return True and self.abort_pending_allowed
```
luelista: whats the `True and` for?

src/pretix/plugins/paypal2/views.py
```diff
@@ -483,16 +483,19 @@ def webhook(request, *args, **kwargs):
...
                 try:
                     payment.info = json.dumps(sale.dict())
                     payment.save(update_fields=['info'])
                     payment.confirm()
                     prov.log_payment_duration(payment)
                 except Quota.QuotaExceededException:
                     pass
+            if any_pending_review:
+                payment.state = OrderPayment.PAYMENT_STATE_PENDING
+                payment.save()
```
pajowu:
```suggestion
            if any_pending_review and payment.state != OrderPayment.PAYMENT_STATE_PENDING:
                payment.state = OrderPayment.PAYMENT_STATE_PENDING
                payment.save(update_fields=['state'])
```

src/pretix/base/payment.py
```diff
@@ -1019,10 +1033,12 @@ def cancel_payment(self, payment: OrderPayment):
...
         """
-        if payment.state == OrderPayment.PAYMENT_STATE_PENDING and not self.abort_pending_allowed:
-            raise PaymentException(_(
-                "This payment is already being processed and can not be canceled any more."
-            ))
+
+        if payment.state == OrderPayment.PAYMENT_STATE_PENDING:
+            if not self.payment_abort_pending_allowed(payment):
+                raise PaymentException(_(
+                    "This payment is already being processed and can not be canceled any more."
```
pajowu:
```suggestion
                    "This payment is already being processed and cannot be canceled any more."
```

src/pretix/plugins/paypal2/payment.py
bockstaller: Yes it does

src/pretix/plugins/paypal2/payment.py
bockstaller: I've upgraded it

src/pretix/plugins/paypal2/payment.py
bockstaller: I think so, we need successful communication with PayPal to get to this point and the OrderDetails endpoint promises a create_time field https://docs.paypal.ai/reference/api/rest/orders/show-order-details#show-order-details

src/pretix/base/payment.py
bockstaller: Oversight from refactoring, removed

PR #6476, opened 2026-08-13
Always base64-encode email attachments

pajowu:
It might sound unneccesary to base64-encode plaintext attachments, but some smtp providers modify them otherwise


src/pretix/base/services/mail.py
```diff
@@ -382,15 +389,25 @@ def mail(email: Union[str, Sequence[str]], subject: Union[str, FormattedString],
 class CustomEmail(EmailMultiAlternatives):
     def _create_mime_attachment(self, content, mimetype):
         """
-        Convert the content, mimetype pair into a MIME attachment object.
-
```
raphaelm, line 386: Maybe add a "why" to this comment for the future?

raphaelm (approved):

PR #6506, opened 2026-08-31
Event Settings: Make contact_url and imprint_url I18nURLField (Z#23238623)

pc-coholic:
Also changing the docs to not use the imprint-URL. We could - but then we would need to either explain the whole "Single String in, Dict with default Event lang out" or "Dict in, Dict out"... Both not crazy concepts; but seemed easier to stay with a simple string.

Tests are adjusted to use "String in, Dict out".

### DIFF
--- a/doc/api/resources/events.rst
+++ b/doc/api/resources/events.rst
@@ -566,7 +566,7 @@ organizer level.
       Content-Type: application/json
 
       {
-        "imprint_url": "https://pretix.eu",
+        "region": "DE",
         …
       }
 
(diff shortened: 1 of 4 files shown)


doc/api/resources/events.rst
```diff
@@ -579,12 +579,14 @@ organizer level.
       Content-Type: application/json
 
       {
-        "imprint_url":
+        "locale":
```
raphaelm:
```suggestion
        "region":
```

PR #6507, opened 2026-08-31
VAT ID validation: Alternative API for German shops

raphaelm:


src/pretix/base/services/tax.py
```diff
@@ -343,6 +343,66 @@ def _validate_vat_id_EU(vat_id, country_code):
...
+                # evatr-0000: Die angefragte Ust-IdNr. ist zum Anfragezeitpunkt gültig.
+                # evatr-2002: Die angefragte USt-IdNr. ist zum Anfragezeitpunkt nicht gültig.
+                #             Sie ist erst gültig ab dem Datum im Feld gueltigAb.
+                return vat_id
+            # evatr-2006: Die angefragte Ust-IdNr. ist zum Anfragezeitpunkt nicht gültig.
+            #             Sie war gültig im Zeitraum, der durch die Werte in den Feldern gueltigAb und gueltigBis beschrieben ist.
+            # evatr-2008: Die angefragte Ust-IdNr. ist zum Anfragezeitpunkt gültig.
+            #             Für die qualifizierte Bestätigungsanfrage liegt einer Besonderheit vor.
+            #             Für Rückfragen wenden Sie sich an das BZSt.
+            raise VATIDFinalError(error_messages['invalid'])
```
ser8phin, line 374: The comments for status 2002 and 2008, while correct, are respectively in the wrong place (because the API-Documentation has them in number order, but here we have them corresponding to their result)

(I would have made an inline-suggestion, but it messed up the code, so I refrained)

src/pretix/base/services/tax.py
```diff
@@ -343,6 +343,66 @@ def _validate_vat_id_EU(vat_id, country_code):
         return vat_id
 
 
+def _validate_vat_id_EU_fallback_germany(vat_id, country_code, requester_id):
```
ser8phin, line 346: We could add a test with the test vat-ids from the docs:
"Mit folgenden Werten können Testaufrufe mit Beispiel USt-IdNr erfolgen: "anfragendeUstid":"DE123456789","angefragteUstid":"ATU12345678"

ser8phin (changes requested):
Just some comments, looks good otherwise

src/pretix/base/services/tax.py
ser8phin: fixed with https://github.com/pretix/pretix/pull/6507/commits/98f41561b1ff10f7254fe0d966f7baca2ef8e387

src/pretix/base/services/tax.py
ser8phin: :x: discussed - won't fix

ser8phin (approved):

PR #6515, opened 2026-09-02
Event calendar: consider events without products "not available" for filtering

raphaelm:
Also, drop the help text of the flag that is no longer accurate.

### DIFF
--- a/src/pretix/presale/views/organizer.py
+++ b/src/pretix/presale/views/organizer.py
@@ -604,7 +604,7 @@ def filter_subevents_with_plugins(subevents, sales_channel=None):
 def should_hide_subevent(settings, subevent, voucher=None):
     hide = False
     if settings.event_list_available_only:
-        hide = hide or (
+        hide = (
             # Presale is over → the subevent is not available → hide
             subevent.presale_has_ended or
             # Not a single product is available on this sales channel → hide
--- a/src/pretix/presale/views/organizer.py
+++ b/src/pretix/presale/views/organizer.py
@@ -601,6 +601,32 @@ def filter_subevents_with_plugins(subevents, sales_channel=None):
     return subevents
 
 
+def should_hide_subevent(settings, subevent, voucher=None):
+    hide = False
+    if settings.event_list_available_only:
+        hide = hide or (
+            # Presale is over → the subevent is not available → hide
+            subevent.presale_has_ended or
+            # Not a single product is available on this sales channel → hide
+            # Note that means there could be products which are ignored for calendar availability (Quota.ignore_for_event_availability)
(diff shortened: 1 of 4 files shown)


src/pretix/presale/views/organizer.py
```diff
@@ -601,6 +601,32 @@ def app_active(event):
     return subevents
 
 
+def should_hide_subevent(settings, subevent, voucher=None):
+    hide = False
+    if settings.event_list_available_only:
+        hide = hide or (
```
wiffbi: This additional `hide` is probably an artefact from a previous for loop? As hide will always be `False`here, we can leave it out.
```suggestion
        hide = (
```

src/pretix/presale/views/organizer.py
```diff
@@ -601,6 +601,32 @@ def app_active(event):
...
+                # Sold out → hide, unless we have a voucher active that can bypass all quotas
+                (not voucher or not voucher.allow_ignore_quota) and
+                subevent.best_availability_state < Quota.AVAILABILITY_RESERVED
+            )
+        )
+
+    if settings.event_calendar_future_only:
+        if (subevent.date_to or subevent.date_from) < time_machine_now():
+            hide = True
+    return hide
```
wiffbi, line 627: For me returning early is easier to read. I guess this originally came from a loop over multiple events? Anyway, what about something like the following – I know, lots of ifs that could be combined with `or`s, but for me this is easier to process in my head than one long expression.

```suggestion
    if not settings.event_list_available_only:
        return True
    if settings.event_calendar_future_only and (subevent.date_to or subevent.date_from) < time_machine_now():
        return True
    if subevent.presale_has_ended:
        # Presale is over → the subevent is not available → hide
        return True
    if subevent.best_availability_state is None:
        # Not a single product is available on this sales channel → hide
        # Note that means there could be products which are ignored for calendar availability (Quota.ignore_for_event_availability)
        # or products only visible with a voucher. However, for customers with these scenarios, the event_list_available_only
        # makes only very little sense as it would never do anything, so the flag can just be removed -- or the products should
        # be made visible so people know why there are no products. In case a voucher is already entered on the calendar view,
        # this is already respected and subevents are shown correctly.
        return True
    if (not voucher or not voucher.allow_ignore_quota) and subevent.best_availability_state < Quota.AVAILABILITY_RESERVED:
        # Sold out → hide, unless we have a voucher active that can bypass all quotas
        return True

    return False
```

wiffbi (approved):
Just a minor code-style suggestion, feel free to ignore.

src/pretix/presale/views/organizer.py
raphaelm: that's not equivalent on the first lines, but i'll see if I rework it before I deploy it

PR #6516, opened 2026-09-02
Payment step: Allow to postpone payment choice on some sales channels

raphaelm:
image

Use case 1: pretix-resellers being used for telephone order or mail order and customer should receive a link without any started payment.

Use case 2: Target audience where usually the person who orders cannot make the payment decision but needs to forward the link internally.


bockstaller (approved):
Only two small, non-blocking edgecases:
https://github.com/pretix/pretix/pull/6526

PR #6527, opened 2026-09-07
product image dialog: fix CSS to always fit image to viewport

fmalcher:
Product images had no CSS max-height. On wide screens, the image would take full width and overflow the viewport at the bottom. This PR adds a maximum height so that images always fit into the viewport, regardless of height and width of screen and image.

## Before

Screenshot 2026-09-07 at 14 50 42

## After

Screenshot 2026-09-07 at 14 51 02


fmalcher:
@cla-bot I signed it

raphaelm:
Funny coincidence, I did not see your PR and did something similar in #6534

raphaelm:
Yours is simpler and seems to work the same

PR #6528, opened 2026-09-07
Organizer API: Add endpoint for event-meta-properties (Z#23220103, Z#23243414)

ser8phin:


doc/api/resources/event_meta_properties.rst
```diff
@@ -0,0 +1,244 @@
+Event Meta Properties
```
raphaelm, line 1: this needs to be listed in index.rst to be accessible

doc/api/resources/event_meta_properties.rst
```diff
@@ -0,0 +1,244 @@
...
+=====================
+
+Resource description
+--------------------
+
+An Event Meta Property is used to to define meta information fields for its events.
```
raphaelm:
```suggestion
An event meta property is used to to define meta information fields for its events.
```

doc/api/resources/event_meta_properties.rst
```diff
@@ -0,0 +1,244 @@
...
+This information can be re-used, for example, in ticket layouts.
+
+The Event Meta Properties resource contains the following public fields:
```
raphaelm:
```suggestion
The event meta property resource contains the following public fields:
```

doc/api/resources/event_meta_properties.rst
```diff
@@ -0,0 +1,244 @@
...
+===================================== ========================== =======================================================
+Field                                 Type                       Description
+===================================== ========================== =======================================================
+id                                    integer                    Unique ID for this property
+name                                  string                     Name of the property
+default                               string                     Value of the default option
+required                              boolean                    If ``true``, an event can only be taken live if the
+                                                                 property is set. In event series, it's always optional
+                                                                 to set a value for individual dates
+protected                             boolean                    If ``true``, this property can only be changed by
```
raphaelm:
```suggestion
protected                             boolean                    If ``true``, the value for an event can only be changed by
```

doc/api/resources/event_meta_properties.rst
```diff
@@ -0,0 +1,244 @@
...
+                                                                 consisting of a dictionary of i18n string translations,
+                                                                 as well as other implementation based key-value-pairs
+===================================== ========================== =======================================================
+
+Endpoints
+---------
+
+.. http:get:: /api/v1/organizers/(organizer)/event_meta_properties/
+
+   Returns a list of all Meta Properties for the organizer.
```
raphaelm:
```suggestion
   Returns a list of all meta properties for the organizer.
```

doc/api/resources/event_meta_properties.rst
```diff
@@ -0,0 +1,244 @@
...
+            "default": "blue",
+            "required": false,
+            "protected": false,
+            "filter_public": false,
+            "public_label": {},
+            "filter_allowed": true,
+            "choices": [
+                {
+                    "key": "blue",
+                    "ORDER": 1,
```
raphaelm, line 79: This is a Django formset artifact and should IMHO not be visible on the API and confuse API users. Ideally, it should not even be in the database.

doc/api/resources/event_meta_properties.rst
```diff
@@ -0,0 +1,244 @@
...
+                    "label": {
+                        "en": "Blue"
+                    },
+                    "DELETE": false
```
raphaelm, line 83: This is a Django formset artifact and should IMHO not be visible on the API and confuse API users. Ideally, it should not even be in the database.

src/pretix/api/serializers/organizer.py
```diff
@@ -640,3 +641,12 @@ def get_new_filename(self, name: str) -> str:
...
         # TODO: make sure pub is always correct
         return 'pub/' + fname
+
+
+class EventMetaPropertiesSerializer(I18nAwareModelSerializer):
+    class Meta:
+        model = EventMetaProperty
+        fields = (
+            'id', 'name', 'default', 'required', 'protected', 'filter_public', 'public_label', 'filter_allowed',
+            'choices'
```
raphaelm, line 651: This needs custom logic to verify that `choices`, when changed by the API client, only contains allowed values.

src/pretix/api/views/organizer.py
```diff
@@ -846,3 +847,50 @@ def perform_destroy(self, instance):
             data={'id': instance.pk}
         )
         instance.delete()
+
+
+class EventMetaPropertiesViewSet(viewsets.ModelViewSet):
+    serializer_class = EventMetaPropertiesSerializer
+    queryset = EventMetaProperty.objects.none()
+    write_permission = 'organizer.settings.general:write'
```
raphaelm, line 855: Please also add to test_permissions.py

src/pretix/api/serializers/organizer.py
```diff
@@ -640,3 +641,12 @@ def get_new_filename(self, name: str) -> str:
...
+        )
```
raphaelm, line 652: The backend UI validates that `default` is valid according to `choices`, should we do the same?

src/pretix/api/views/organizer.py
```diff
@@ -846,3 +847,50 @@ def perform_destroy(self, instance):
...
+            auth=self.request.auth,
+            data={'id': instance.pk}
+        )
+        instance.delete()
+
+    @transaction.atomic()
+    def perform_create(self, serializer):
+        inst = serializer.save(organizer_id=self.request.organizer.pk)
+        serializer.instance.log_action(
+            'pretix.organizer.event_meta_property.added',
```
raphaelm: same here

src/pretix/api/views/organizer.py
```diff
@@ -846,3 +847,50 @@ def perform_destroy(self, instance):
...
+
+    def get_serializer_context(self):
+        ctx = super().get_serializer_context()
+        ctx['organizer'] = self.request.organizer
+        return ctx
+
+    @transaction.atomic()
+    def perform_destroy(self, instance):
+        instance.log_action(
+            'pretix.organizer.event_meta_property.deleted',
```
raphaelm: The backend uses pretix.property.deleted, why inventing a new one?

src/pretix/api/views/organizer.py
```diff
@@ -846,3 +847,50 @@ def perform_destroy(self, instance):
...
+            auth=self.request.auth,
+            data=self.request.data,
+        )
+        return inst
+
+    @transaction.atomic()
+    def perform_update(self, serializer):
+        inst = serializer.save(organizer_id=self.request.organizer.pk)
+        serializer.instance.log_action(
+            'pretix.organizer.event_meta_property.changed',
```
raphaelm: same here

doc/api/resources/event_meta_properties.rst
```diff
@@ -0,0 +1,244 @@
...
+                                                                 to set a value for individual dates
+protected                             boolean                    If ``true``, this property can only be changed by
+                                                                 organizer-level administrators
+filter_public                         boolean                    If ``true``, this property will be shown to filter
+                                                                 events in the public event list and calendar
+public_label                          string                     Public name of the property
+filter_allowed                        boolean                    If ``true``, this property will be shown to filter
+                                                                 events or reports in the backend, and it can also be
+                                                                 used for hidden filter parameters in the frontend
+choices                               list of objects            List of JSON objects representing all permitted values
```
raphaelm, line 31: What happens if I set `"choices": []` as an API consumer? Will it be treated as "nothing is valid" or "everything is valid"? In the latter case (which I think is more likely), we should not confuse API users with `null` and `[]` being the same thing but normalize to either one in the API so we still have the flexibility later (which one, I don't care much).

raphaelm (changes requested):

src/pretix/api/views/organizer.py
ser8phin: Artifacts from copy-paste, I even added the logdisplay entries for the ones already in use in the backend, I just overlooked that I hadn't fixed this afterwards 🤦

doc/api/resources/event_meta_properties.rst
wiffbi: I added a change that will normalize it to None.

doc/api/resources/event_meta_properties.rst
```diff
@@ -0,0 +1,244 @@
...
+                                                                 for this property, or ``null`` for no limitation.
+                                                                 Each choice object has a required internal name named
+                                                                 ``key`` and optional public name named ``label``
+                                                                 consisting of a dictionary of i18n string translations,
+                                                                 as well as other implementation based key-value-pairs
```
wiffbi: What does "as well as other implementation based key-value-pairs" mean? I think we should only allow `key` and `label` all other existing keys are just django-form-related which IMHO should not even be saved?


```suggestion
                                                                 consisting of a dictionary of i18n string translations
```

doc/api/resources/event_meta_properties.rst
```diff
@@ -0,0 +1,244 @@
...
+protected                             boolean                    If ``true``, the value for an event can only be changed
+                                                                 by organizer-level administrators
+filter_public                         boolean                    If ``true``, this property will be shown to filter
+                                                                 events in the public event list and calendar
+public_label                          string                     Public name of the property
+filter_allowed                        boolean                    If ``true``, this property will be shown to filter
+                                                                 events or reports in the backend, and it can also be
+                                                                 used for hidden filter parameters in the frontend
+choices                               list of objects            List of JSON objects representing all permitted values
+                                                                 for this property, or ``null`` for no limitation.
```
wiffbi, line 32:
```suggestion
                                                                 for this property, or ``[]`` or ``null`` for no limitation.
```

doc/api/resources/event_meta_properties.rst
wiffbi: I added a `KeyLabelObjectListField` for `choices` and filtered all unknown keys in its `to_representation`. Or should we fix the django-formset and add a migration to fix all existing EventMetaProperties?

doc/api/resources/event_meta_properties.rst
wiffbi: See above

doc/api/resources/event_meta_properties.rst
```diff
@@ -0,0 +1,244 @@
...
+            "public_label": {},
+            "filter_allowed": true,
+            "choices": [
+                {
+                    "key": "blue",
+                    "ORDER": 1,
+                    "label": {
+                        "en": "Blue"
+                    },
+                    "DELETE": false
```
wiffbi:
```suggestion
```

doc/api/resources/event_meta_properties.rst
```diff
@@ -0,0 +1,244 @@
...
+            "default": "blue",
+            "required": false,
+            "protected": false,
+            "filter_public": false,
+            "public_label": {},
+            "filter_allowed": true,
+            "choices": [
+                {
+                    "key": "blue",
+                    "ORDER": 1,
```
wiffbi:
```suggestion
```

src/pretix/api/serializers/organizer.py
wiffbi: I added even more validation to choices.

doc/api/resources/event_meta_properties.rst
raphaelm: I thought you normalized it?

src/pretix/api/serializers/organizer.py
```diff
@@ -640,3 +641,62 @@ def get_new_filename(self, name: str) -> str:
...
+    def to_internal_value(self, data):
+        if data is None:
+            return data
+        if not isinstance(data, list):
+            raise ValidationError("Choices need to be a list or null.")
+        if not data:
+            # empty list
+            return None
+
+        if any([not isinstance(choice, dict) for choice in data]):
```
raphaelm, line 663:
```suggestion
        if not all(isinstance(choice, dict) for choice in data):
```

src/pretix/api/serializers/organizer.py
```diff
@@ -640,3 +641,62 @@ def get_new_filename(self, name: str) -> str:
         )
         # TODO: make sure pub is always correct
         return 'pub/' + fname
+
+
+class KeyLabelObjectListField(serializers.Field):
+
+    def to_representation(self, value):
```
raphaelm, line 648: Normalize empty list to None here as well?

src/pretix/api/serializers/organizer.py
```diff
@@ -640,3 +641,62 @@ def get_new_filename(self, name: str) -> str:
...
+        if not data:
+            # empty list
+            return None
+
+        if any([not isinstance(choice, dict) for choice in data]):
+            raise ValidationError("Choices need to contain only objects.")
+
+        required_keys = {"key"}
+        allowed_keys = {"key", "label"}
+        if not all([required_keys <= set(choice.keys()) <= allowed_keys for choice in data]):
```
raphaelm, line 668:
```suggestion
        if not all(required_keys <= set(choice.keys()) <= allowed_keys for choice in data):
```

src/tests/api/test_event_meta_properties.py
```diff
@@ -0,0 +1,185 @@
...
+        '/api/v1/organizers/{}/event_meta_properties/'.format(organizer.slug),
+        format='json',
+        data={
+            "name": "Color",
+            "default": "Red",
+            "required": False,
+            "choices": ["Red", "Green", "Blue"]
+        }
+    )
+    assert resp.status_code == 400
```
raphaelm, line 91: I strongly suggest to actually check for the expected error message. Otherwise this test may have written with the intent of testing the type validation in choices,  but if e.g. a new required attribute is added to the serializer, it might suddenly test something else entirely!

doc/api/resources/event_meta_properties.rst
wiffbi: Ah, well, by „normalizing“ I meant if you pass in `[]` it gets normalized to `None`. Should we disallow `[]` instead and respond with an error?

src/pretix/api/serializers/organizer.py
wiffbi: There should never be an empty list here as `to_internal_value` already converts it to `None` and previously there was no way to save an empty list. But I guess as a double safe-guard it is a good idea. I will add it.

doc/api/resources/event_meta_properties.rst
wiffbi: Or just not list `[]` as a possible value (as the API currently will never respond with an `[]`). Sorry, misunderstanding on my part.

src/tests/api/test_event_meta_properties.py
wiffbi: THank, good idea! Added checks for error-messages.

src/tests/api/test_event_meta_properties.py
```diff
@@ -0,0 +1,240 @@
...
+            }
+        ],
+    )
+
+
+TEST_TYPE_RES = {
+    "name": "Color",
+    "default": "Red",
+    "required": False,
+    "choices": [{"key": "Red", "label": "Rot"}],
```
raphaelm: Either the tests are wrong or the docs. The docs say:

> `label` consisting of a dictionary of i18n string translations

raphaelm (changes requested):

src/tests/api/test_event_meta_properties.py
wiffbi: Well, it is basically anything a `LazyI18nString` supports, so not only a basic string or a dict of i18n strings, but even a list of anything! I will add some validation that it is either a string or a dict with strings as keys and values.

raphaelm: No, we should normalize that. The fact that LazyI18nString supports plain `str` is an (unfortunate) implementation detail of ours and should not concern API consumers who do not expect it.

For all other i18nfields in the API, it's normalized by the `I18nField` serializer field class:
```
    def to_representation(self, value):
        if hasattr(value, 'data'):
            if isinstance(value.data, dict):
                return value.data
            elif value.data is None:
                return None
            else:
                return {
                    settings.LANGUAGE_CODE: str(value.data)
                }
        elif value is None:
            return None
        else:
            return {
                settings.LANGUAGE_CODE: str(value)
            }

```
this can never return str

src/tests/api/test_events.py
```diff
@@ -252,10 +252,7 @@ def test_event_create(team, token_client, organizer, event, meta_prop):
     resp = token_client.post(
         '/api/v1/organizers/{}/events/'.format(organizer.slug),
         {
-            "name": {
-                "de": "Demo Konference 2020 Test",
-                "en": "Demo Conference 2020 Test"
-            },
+            "name": ["Demo Konference 2020 Test"],
```
raphaelm: uh, no?

src/tests/api/test_events.py
```diff
@@ -275,6 +272,7 @@ def test_event_create(team, token_client, organizer, event, meta_prop):
         },
         format='json'
     )
+    print(resp.data)
```
raphaelm: 🖨️

src/tests/api/test_events.py
wiffbi: sorry, somehow came through while I was investigating what I18nField does for validation and I somehow faild to not commit that

src/tests/api/test_event_meta_properties.py
wiffbi: This uses the `I18nField` for labels inside `MetaPropertyDictField` now.

src/pretix/api/serializers/organizer.py
```diff
@@ -640,3 +642,88 @@ def get_new_filename(self, name: str) -> str:
...
+            "key": value["key"]
+        }
+        if "label" in value:
+            d["label"] = self.label_child.to_representation(value["label"])
+
+        return super().to_representation(d)
+
+    def to_internal_value(self, data):
+        if not isinstance(data, dict):
+            raise ValidationError("Meta properties must be a dict.")
```
raphaelm:
```suggestion
            raise ValidationError("Meta property value options must be a dict.")
```

src/pretix/api/serializers/organizer.py
```diff
@@ -640,3 +642,88 @@ def get_new_filename(self, name: str) -> str:
...
+
+        if not isinstance(data.get("key"), str):
+            raise ValidationError("Meta properties must have a key of type string.")
```
raphaelm:
```suggestion
            raise ValidationError("Meta property value options must have a key of type string.")
```

src/pretix/api/serializers/organizer.py
```diff
@@ -640,3 +642,88 @@ def get_new_filename(self, name: str) -> str:
...
+
+        if any(k not in {"key", "label"} for k in data.keys()):
+            raise ValidationError("Meta properties may only have a key and optionally a label.")
```
raphaelm:
```suggestion
            raise ValidationError("Meta property value options may only have a key and optionally a label.")
```

src/tests/api/test_event_meta_properties.py
```diff
@@ -0,0 +1,233 @@
...
+    'public_label': None,
+}
+
+
+@pytest.mark.django_db
+def test_meta_property_list(token_client, organizer, event_meta_property):
+    res = dict(TEST_TYPE_RES)
+
+    resp = token_client.get('/api/v1/organizers/{}/event_meta_properties/'
+                            .format(organizer.slug))
```
raphaelm: Where is this incredibly short line length coming from? ;)

src/tests/api/test_event_meta_properties.py
```diff
@@ -0,0 +1,233 @@
...
+                {"label": "Green"},
+                {"key": "g", "label": "Green", "foo": "bar"},
+            ],
+        }
+    )
+    assert resp.status_code == 400
+    assert str(resp.data["choices"][0][0]) == "Meta properties must have a key of type string."
+    assert str(resp.data["choices"][1][0]) == "Meta properties must be a dict."
+    assert str(resp.data["choices"][2][0]) == "Meta properties must have a key of type string."
+    assert str(resp.data["choices"][3][0]) == "Meta properties may only have a key and optionally a label."
```
raphaelm:
```suggestion
    assert str(resp.data["choices"][0][0]) == "Meta property value options must have a key of type string."
    assert str(resp.data["choices"][1][0]) == "Meta property value options must be a dict."
    assert str(resp.data["choices"][2][0]) == "Meta property value options must have a key of type string."
    assert str(resp.data["choices"][3][0]) == "Meta property value options may only have a key and optionally a label."
```

src/pretix/api/serializers/organizer.py
```diff
@@ -640,3 +642,88 @@ def get_new_filename(self, name: str) -> str:
...
+class MetaPropertyListField(serializers.ListField):
+    def __init__(self, *args, **kwargs):
+        kwargs["validators"] = kwargs.pop("validators", [])
+
+        def validate_keys_unique(choices):
+            if not choices:
+                return
+            keys = [c.get("key") for c in choices]
+            if len(set(keys)) < len(keys):
+                raise ValidationError("The key for each meta property must be unique.")
```
raphaelm:
```suggestion
                raise ValidationError("The key for each meta property value option must be unique.")
```

src/tests/api/test_event_meta_properties.py
```diff
@@ -0,0 +1,233 @@
...
+            "default": "r",
+            "required": False,
+            "choices": [
+                {"key": "r", "label": {"en": "Red"}},
+                {"key": "r", "label": {"en": "Razzmatazz"}},
+            ],
+        }
+    )
+    assert resp.status_code == 400
+    assert str(resp.data["choices"][0]) == "The key for each meta property must be unique."
```
raphaelm:
```suggestion
    assert str(resp.data["choices"][0]) == "The key for each meta property value option must be unique."
```

raphaelm (approved):

src/tests/api/test_event_meta_properties.py
wiffbi: Uh, you are right, that is very short and IMHO ugly. Fixed it.

PR #6530, opened 2026-09-08
Delete PayPal integration v1

raphaelm:
Leaving only migrations behind


pc-coholic (approved):
LGTM.

I changed the references to paypal2 in a few more tests and docs - but those are mostly cosmetic.

I was also wondering if we should do another, last pass over the database and forecefully move everything to paypal2 that has not been changed yet - but I think we can also do without...

PR #6539, opened 2026-09-09
Fix OPTIONS request on SeatViewSet (PRETIXEU-FBJ)

luelista:


pajowu (approved):

src/pretix/api/serializers/event.py
```diff
@@ -1104,8 +1104,8 @@ def prefetch_expanded_data(self, items, request, expand_fields):
                 raise PermissionDenied('event.vouchers:read permission required for expand=voucher')
             prefetch_by_id(items, Voucher.objects, 'voucher_id', 'voucher')
 
-    def __init__(self, instance, *args, **kwargs):
-        if not kwargs.get('data'):
+    def __init__(self, instance=None, *args, **kwargs):
+        if instance and not kwargs.get('data'):
```
pajowu, line 1108:
```suggestion
        if instance is not None and not kwargs.get('data'):
```

pajowu (changes requested):

PR #6542, opened 2026-09-10
Event index: Add subevent time and name to <title> (Z#23246174)

raphaelm:


src/pretix/presale/templates/pretixpresale/event/index.html
```diff
@@ -23,7 +23,11 @@
             {% endblocktrans %} ::
         {% endif %}
     {% elif subevent %}
-        {{ subevent.get_date_range_display }} ::
+        {{ subevent.get_date_range_display_with_times }} ::
+        {% if subevent.name|upper != event.name|upper %}
+            {# The |upper is a trick to force LazyI18nString→str conversion before comparison #}
+            {{ subevent.name }} ::
+        {% endif %}
```
wiffbi: I would put the data+time after the subevent. for special time-slot booking, one might have the subevent.name the same as the event.name, which the causes the subevent.name not to be displayed so the title then starts with the date+time.
```suggestion
        {% if subevent.name|upper != event.name|upper %}
            {# The |upper is a trick to force LazyI18nString→str conversion before comparison #}
            {{ subevent.name }} ::
        {% endif %}
        {{ subevent.get_date_range_display_with_times }} ::
```

PR #6545, opened 2026-09-10
Phone number input: Fix workaround for prefixes shared between countries (fixes #6544)

### LINKED ISSUE #6544: Phone prefix in Italian: default is "Vatican City +39" → should be "Italia +39"
### Problem and impact

When the language is Italian, I noticed the default phone prefix is "Città del Vaticano +39" (🤯)

Image

The current situation is technically correct for only 882 inhabitants of the Vatican City country.

### Expected behaviour

Suggest "Italia +39" as default, not "Città del Vaticano +39":

Image

In this way we have a default which is excellent for 58 millions inhabitants of Italy, and still technically correct for the remaining 882 inhabitants of the Città del Vaticano.

### Steps to reproduce

1. Use Pretix in Italian
2. Get a ticket
3. See the phone number selector

### Screenshots

_No response_

### Link

_No response_

Firefox

Linux, macOS, Windows

### Version

latest on https://pretix.eu

raphaelm:


src/pretix/base/forms/questions.py
```diff
@@ -350,16 +350,21 @@ def render(self, name, value, *args, **kwargs):
         return super().render(name, value or self.initial, *args, **kwargs)
 
     def get_context(self, name, value, attrs):
-        if value and self.choices[1][0] != value:
-            matching_choices = len([1 for p, c in self.choices if p == value])
+        choices = list(self.choices)
```
pajowu, line 354:
```suggestion
        # self.choices is lazy evaluated, needs to be realized to be modifiable
        choices = list(self.choices)
```

pajowu (approved):

PR #6553, opened 2026-09-14
PDF: fix canvas-size when artbox/trimbox is present in background-PDF

wiffbi:


wiffbi:
See #6560 for a better solution

PR #6554, opened 2026-09-14
Badges: fix rotated background-PDF

wiffbi:
When exporting multiple badges that have a background-PDF with rotation, handling the rotation with `.transfer_rotation_to_content()` during the merging-process results in wrongly rotated media/assets on the background-PDF-page. mediabox/artbox/etc. are correctly rotated, but for each page, rotation on the assets seem to accumulate for each page.

Transfering the rotation to content during creation of the multipage-bg-pdf for merging fixes the issue.

accumulated rotation:
bg-rotation-accumulated

Test-PDF
[badge-bg-a7-rotated-90.pdf](https://github.com/user-attachments/files/32193327/badge-bg-a7-rotated-90.pdf)


wiffbi:
better solved through #6561

PR #6566, opened 2026-09-18
Add pdf render tests

pajowu:
These currently live in the badges pdf tests (as this is where we encountered the bugs). Once we add more, we should move them to an own file and directly test `pretix.base.pdf.Renderer`


src/tests/plugins/badges/test_pdf.py
```diff
@@ -118,3 +123,53 @@ def test_generate_pdf_multi(env):
...
+
+
+def compare_pdfs(pdf_dir: Path, inp_a: Path | bytes, inp_b: Path | bytes):
+    logging.info(f"Comparing pdfs, writing files to {pdf_dir}")
+
+    pdf_a = pypdfium2.PdfDocument(inp_a)
+    pdf_b = pypdfium2.PdfDocument(inp_b)
+
+    pdf_a.save(pdf_dir / "a.pdf")
+    pdf_a.save(pdf_dir / "b.pdf")
```
wiffbi, line 139:
```suggestion
    pdf_b.save(pdf_dir / "b.pdf")
```

src/tests/plugins/badges/test_pdf.py
```diff
@@ -118,3 +123,53 @@ def test_generate_pdf_multi(env):
...
+
+    assert len(pdf_a) == len(pdf_b)
+
+    for i, (expected_page, output_page) in enumerate(zip(pdf_a, pdf_b)):
```
wiffbi, line 143: Isn’t pdf_b the expected pdf actually? Although in the compare-function nomenclature is not output vs. expected, it is anyways a vs. b, so why not stay with e.g. page_a, page_b?

src/tests/plugins/badges/test_pdf.py
wiffbi: Is saving actually necessary?

wiffbi: Probably helps to examine the pdfs by hand, when an assert fails, so therefore saving the PDFs?

PR #6568, opened 2026-09-18
Fix addon legend overlapping with profile checkbox (Z#23247075)

pajowu:


src/pretix/static/pretixpresale/scss/_checkout.scss
```diff
@@ -157,6 +157,11 @@
 .profile-scope:last-child .profile-save-container {
   margin-bottom: -15px;
 }
+
+fieldset:not(:last-child) .profile-scope:last-child .profile-save-container {
```
wiffbi, line 161: That selector does not work for me when I have multiple addons. The following is ugly, but works. It basically selects fieldsets, that do not have a .profile-add-on inside, but are followd by a fieldset (which currently only happens if there are addon-questions).

```suggestion
/* the rule above removes blank space after any profile-save-container at the end of every fieldset, which causes an overlay if addon-questions are used. So re-add margin for non-addon-fieldsets followed by fieldsets (currently only happens it addon-questions are available) */
fieldset:not(:has(.profile-add-on)):has( + fieldset) .profile-scope:last-child .profile-save-container {
```

We might think about changing the HTML in checkout_questions.html and move the `{% if profiles_data %} profile-add-on{% endif %}` from the div to the fieldset in line 153. This would make the selector here a bit easier although still complicated enough, so I’d leave it for now.

wiffbi (changes requested):

PR #6569, opened 2026-09-18
User details: Show list of 2FA devices and allow to reset drift

raphaelm:


src/pretix/control/views/users.py
```diff
@@ -107,6 +110,21 @@ def get_context_data(self, **kwargs):
...
+            for obj in objs:
+                if dt == TOTPDevice:
+                    obj.devicetype = 'totp'
+                elif dt == U2FDevice:
+                    obj.devicetype = 'u2f'
+                elif dt == WebAuthnDevice:
+                    obj.devicetype = 'webauthn'
+                elif dt == StaticDevice:
+                    obj.devicetype = 'emergency'
+            ctx['devices'] += objs
```
luelista, line 126: This currently displays devices of the logged in user, not the edited user.

Also, I'd rather add a helper somewhere:

```
    def get_2fa_devices(user, include_emergency):
        devices = []
        for dt in [*REAL_DEVICE_TYPES, StaticDevice] if include_emergency else REAL_DEVICE_TYPES:
            objs = list(dt.objects.filter(user=user, confirmed=True))
            for obj in objs:
                if dt == TOTPDevice:
                    obj.devicetype = 'totp'
                elif dt == U2FDevice:
                    obj.devicetype = 'u2f'
                elif dt == WebAuthnDevice:
                    obj.devicetype = 'webauthn'
                elif dt == StaticDevice:
                    obj.devicetype = 'emergency'
            devices += objs
        return devices
```

```suggestion
        ctx['devices'] = get_2fa_devices(self.object, include_emergency=True)
```

and then use that helper in User2FAMainView as well:
```
        ctx['devices'] = get_2fa_devices(self.request.user, include_emergency=False)
```

luelista (changes requested):

luelista (approved):

PR #6577, opened 2026-09-21
InvoiceShredder: Fix shredding of new fields

raphaelm:


src/pretix/base/shredder.py
```diff
@@ -598,18 +598,30 @@ def generate_files(self) -> List[Tuple[str, str, str]]:
     def shred_data(self, progress_callback=None):
         qs_i = self.event.invoices.filter(shredded=False)
         total = qs_i.count()
+        ignore_fields = (
```
wiffbi, line 601: Should we add a comment that date, boolean and int are (currently by default/algorithm) ignored and not shredded?

wiffbi (approved):

src/pretix/base/migrations/0311_fix_unshredded_invoices.py
```diff
@@ -0,0 +1,48 @@
+# Generated by Django 5.2.17 on 2026-09-21 11:30
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+def fix_unshredded_invoices(apps, _):
+    Invoice = apps.get_model("pretixbase", "Invoice")
+    InvoiceLine = apps.get_model("pretixbase", "InvoiceLine")
+    ignore_fields = (
```
raphaelm, line 10:
```suggestion
    ignore_fields = (
        # bool/int fields are not listed and skipped automatically
```

PR #6578, opened 2026-09-21
Event metadata: Fix server-side validation of write protection

raphaelm:


src/pretix/control/views/subevents.py
```diff
@@ -1276,6 +1276,11 @@ def meta_forms(self):
             self._default_meta = self.request.event.meta_data
 
         for p in self.request.organizer.meta_properties.all():
+            if p.protected and not self.request.user.has_organizer_permission(
+                    self.request.organizer, 'organizer.settings.general:write', request=self.request
+            ):
+                continue
```
wiffbi, line 1282: Should protected properties be shown and `disabled` instead of completely hidden? Label for protected reads "Can only be changed by organizer-level administrators", but says nothing about being viewed.

wiffbi (approved):

src/pretix/control/views/subevents.py
raphaelm: Yeah, but this is only for the "bulk edit" screen

PR #6581, opened 2026-09-21
2FA: Show info on throttled request

raphaelm:


src/pretix/control/views/auth.py
```diff
@@ -536,7 +554,23 @@ def post(self, request, *args, **kwargs):
             return redirect('control:index')
         else:
             pretix_failed_logins.inc(1, reason="2fa")
-            messages.error(request, _('Invalid code, please try again.'))
+            msg = _('Invalid code, please try again.')
+            if retry_after:
+                seconds = int((retry_after - now()).total_seconds())
+                minutes = int(seconds // 60)
```
pajowu: nit: imho telling users to retry after 1 minute if they still have to wait 1:59min is bad ux. i'd suggest to `math.ceil(seconds/60)` here and skip the if below (or extend it to show `math.ceil(seconds/60)` minutes with `seconds > 59` and `seconds` otherwise)

pajowu (approved):

